PR #249, opened 2015-10-28
U2F token images are not shown when using in https://www.example.site…

vDorst:
…/somesubdir

This is because of the use of a absolute url instead of a relative url in css

### DIFF
--- a/privacyidea/lib/tokens/u2ftoken.py
+++ b/privacyidea/lib/tokens/u2ftoken.py
@@ -168,8 +168,8 @@ the signatureData and clientData returned by the U2F device in the *u2fResult*:
 
 """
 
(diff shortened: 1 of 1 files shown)


cornelinux:
Thanks a lot for this fix.
But it only seams to work when accessing

    https://example.com/subdir/

not

    https://example.com/subdir

Without the trailing slash it still tries to load the image from /static.
I will look into this. If you have additional ideas, it would be great.

cornelinux:
OK, I figured out, how to add the instanceUrl (subdir) withint the javascript code.
Nevertheless it makes sense in this case to remove the prepended slashes.

vDorst:
great!

PR #250, opened 2015-10-29
Added proxy support

Keithsc:
Hi, As disused previously here's our modified code to add proxy support with a static UserAgent into the owncloud privacyIdea app. This is our first contribution to GitHub so hopefully we've done it right.


cornelinux:
looks cool!
I will merge the pull requests, when the tests are run.
Reading through http://curl.haxx.se/libcurl/c/CURLOPT_PROXY.html, I think we should not set CURLOPT_PROXY when $proxy is empty. Otherwise the environment parameters $http_proxy will be overwritten.

Issue #64, opened 2014-12-25
LDAP Resolver: Windows 2012 DC has binary objectGUID


cornelinux:
A windows 2012 domain controller has binary objectGUID which will result in unicode errors during json dumps in the LDAP resolver.

The binary objectGUID should be hexlified.

PR #325, opened 2016-02-18
Resolve the issue #64 

salvorapi:
This patch resolve the issue #64.

### DIFF
--- a/privacyidea/lib/resolvers/LDAPIdResolver.py
+++ b/privacyidea/lib/resolvers/LDAPIdResolver.py
@@ -251,7 +251,14 @@ class IdResolver (UserIdResolver):
         for ldap_k, ldap_v in attributes.items():
             for map_k, map_v in self.userinfo.items():
                 if ldap_k == map_v:
-                    if type(ldap_v) == list and map_k not in ["mobile"]:
+		    if ldap_k == "objectGUID":
+                       try:
+                          import uuid
+                          uuid_v = uuid.UUID(bytes_le=ldap_v[0])
+                          ret[map_k] = str(uuid_v)
+                       except:
+                          ret[map_k] = ''
+                    elif type(ldap_v) == list and map_k not in ["mobile"]:
                         # All lists (except) mobile return the first value as
                         #  a string. Mobile is returned as a list
                         ret[map_k] = ldap_v[0]


cornelinux:
Hello Salvo,

thanks a lot for the pull request and the hint with the uuid module.
But his is the wrong location. The objectGUID does not need to be part of the users attributes, but it needs to be the users uid. Since this is the location, that needs to be modified:

https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/resolvers/LDAPIdResolver.py#L156

like:

    if  uidtype == "objectGUID":
         rawuid = attributes.get(uidtype)
         uid = str(uuid.UUID(bytes_le=rawuid))

Kind regards
Cornelius

salvorapi:
Hi Cornelius,

I know that objectGUID attribute can be the uid type but in my case DN it's correct for uid and I want objectGUID attribute only on attribute mapping.

However, if you think to add this snip on _get_uid function it's ok.

Do you want that I do the modify and a new pull request ?

Salvo.

cornelinux:
Hi Salvo,

if you like, too, you are very welcome to do the pull request. Otherwise I will do it at the next occasion.

So you need the objectGUID also as an attribute? What for?

Kind regards
Cornelius

cornelinux:
...I will add a test for this!

salvorapi:
I need objectGUID because the product  federated with SimpleSAMLphp (and PrivacyIdea) have a user connection with LDAP source and user_id based on objectGUID. For the moment I want to mantain both authetication methods (local & LDAP & SAML) and this is the reason.

Salvo.

cornelinux:
Hi Salvo,
using the objectGUID within the attributes is fairly simple. (As your patch suggest).
I was looking at using it as the UID. I added a branch "objectGUID". It fails when I try to do the LDAP search with

   objectGUID = *what*?

I have this test case
https://github.com/privacyidea/privacyidea/blob/objectGUID/tests/test_lib_resolver.py#L840
which internally returns an empty result list from the ldap search in
https://github.com/privacyidea/privacyidea/blob/objectGUID/privacyidea/lib/resolvers/LDAPIdResolver.py#L178

I would very much like to get this objectGUID stuff straigt...

Kind regards
Cornelius

salvorapi:
Could be because objectGUID it's an attribute present in Microsoft AD LDAP ?

cornelinux:
I have a Windows 2012 DC. I can query the LDAP and see the objectGUID:

<pre>
# Administrator, Users, mimikatz.intranet
dn: CN=Administrator,CN=Users,DC=mimikatz,DC=intranet
objectClass: top
objectClass: person
objectClass: organizationalPerson
objectClass: user
cn: Administrator
description: Built-in account for administering the computer/domain
distinguishedName: CN=Administrator,CN=Users,DC=mimikatz,DC=intranet
instanceType: 4
whenCreated: 20160110201623.0Z
whenChanged: 20160219173841.0Z
uSNCreated: 8196
memberOf: CN=Group Policy Creator Owners,CN=Users,DC=mimikatz,DC=intranet
memberOf: CN=Domain Admins,CN=Users,DC=mimikatz,DC=intranet
memberOf: CN=Enterprise Admins,CN=Users,DC=mimikatz,DC=intranet
memberOf: CN=Schema Admins,CN=Users,DC=mimikatz,DC=intranet
memberOf: CN=Administrators,CN=Builtin,DC=mimikatz,DC=intranet
uSNChanged: 16406
name: Administrator
objectGUID:: 7zabA8Dn80Kb+cpqbA1Ndw==
userAccountControl: 512
badPwdCount: 0
codePage: 0
countryCode: 0
</pre>

But I can not create an ldap filter for it.
Neither in binary format, nor in hex nor in base64...

Kind regards
Cornelius

salvorapi:
I think that you have to remove 

userId = escape_bytes(userId_b)

and use userId_b for the filter.

Can you try ?

cornelinux:
Great! Sometimes it is only important, that someone is insisting!

We in fact need the escape_bytes. I checked with the 2012DC. But there is something wrong in my LDAP mock why it is not working...

Thanks!

cornelinux:
Hi Salvo,

ok. I fixed #64 so that objectGUID can be the default uid of LDAP resolvers. 

Now I would like to merge your change. As it is ok to add objectGUID to the list of attributes. But I would like to suggest the following changes:

1. please fix the indentation starting in line 254. It is only 3 spaces. It should be 4
2. move the "import uuid" to the top of the file to all other imports
3. I understand the exception is necessary if the LDAP object has not attribute "objectGUID" and therefor the "ldap_v[0]" bail out with something like "index out of range". Or ldap_v might be None. Can you please improve the code and name the exception (Is it IndexError or TypeError or whatever...)

Thanks a lot 
Cornelius

PR #340, opened 2016-03-12
yubikey: identify a token by the prefix sent for /ttype/yubikey

jh23453:
A yubikey is often identified by the serial number in privacyidea, e.g.
when using /validate/check.  This works also for yubikeys in AES mode.
But when we use the yubikey validation protocoll, we don't have the
serial number in the request, so we must use other data.

Yubico cloud service (and pam_yubico) identify the token by the prefix
defined when enrolling which is sent with every request.  We'll do the
same here.

If there is no info in the table tokeninfo stored, we add yubikey.prefix.
When we handle /ttype/yubikey we search alle tokentype=yubikey tokens for
a matching prefix and verify against that token.

This works fine for newly created tokens with --yubikeyrandomprefix 6
[see my patch to privacyideaadm] and after using the token once
[see https://github.com/privacyidea/privacyideaadm/issues/18].
I'm not sure that removing the usage of the serial number completly
is the right solution here. I guess we need some tests with already
enrolled and used tokens.

### DIFF
--- a/privacyidea/lib/tokens/yubikeytoken.py
+++ b/privacyidea/lib/tokens/yubikeytoken.py
@@ -205,8 +205,6 @@ class YubikeyTokenClass(TokenClass):
         # The variable otp val is the last 32 chars
         yubi_otp = anOtpVal[-32:]
 
-        # TODO: We can also check the PREFIX! At the moment, we do not use it!
-
         try:
             otp_bin = modhex_decode(yubi_otp)
         except KeyError:
@@ -251,6 +249,11 @@ class YubikeyTokenClass(TokenClass):
             log.debug("Got no tokenid for %r. Setting to %r." % (serial, uid))
             tokenid = uid
             self.add_tokeninfo("yubikey.tokenid", tokenid)
+
+        prefix = self.get_tokeninfo("yubikey.prefix")
+        if not prefix:
+            log.debug("Got no prefix for %r. Setting to %r." % (serial, yubi_prefix))
+            self.add_tokeninfo("yubikey.prefix", yubi_prefix)
 
         if tokenid != uid:
             # wrong token!
@@ -379,29 +382,23 @@ h={h}
         res = False
 
         token_list = []
+        token_candidate_list = []
 
         # strip the yubico OTP and the PIN
-        modhex_serial = passw[:-32][-16:]
-        try:
-            serialnum = "UBAM" + modhex_decode(modhex_serial)
-        except TypeError as exx:  # pragma: no cover
-            log.error("Failed to convert serialnumber: %r" % exx)
-            return res, opt
-
-        # build list of possible yubikey tokens
-        serials = [serialnum]
-        for i in range(1, 3):
-            serials.append("%s_%s" % (serialnum, i))
+        prefix = passw[:-32][-16:]
 
         from privacyidea.lib.token import get_tokens
         from privacyidea.lib.token import check_token_list
-        for serial in serials:
-            tokenobject_list = get_tokens(serial=serial)
-            token_list.extend(tokenobject_list)
+
+        token_candidate_list = get_tokens(tokentype='yubikey')
+        for tokenobject in token_candidate_list:
+            token_prefix = tokenobject.get_tokeninfo("yubikey.prefix")
+            if prefix == token_prefix:
+                token_list.append(tokenobject)
 
         if not token_list:
-            opt['action_detail'] = ("The serial %s could not be found!" %
-                                    serialnum)
+            opt['action_detail'] = ("The prefix %s could not be found!" %
+                                    prefix)
             return res, opt
 
         (res, opt) = check_token_list(token_list, passw)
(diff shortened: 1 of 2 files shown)


cornelinux:
The TravisCI tests should succeed. The test tests/test_lib_tokens_yubikey.py fails - it might need adaption.

privacyidea/lib/tokens/yubikeytoken.py
```diff
@@ -379,29 +382,23 @@ def check_yubikey_pass(passw):
...
-            serials.append("%s_%s" % (serialnum, i))
+        prefix = passw[:-32][-16:]
 
         from privacyidea.lib.token import get_tokens
         from privacyidea.lib.token import check_token_list
-        for serial in serials:
-            tokenobject_list = get_tokens(serial=serial)
-            token_list.extend(tokenobject_list)
+
+        token_candidate_list = get_tokens(tokentype='yubikey')
```
cornelinux: I am not content with this. This will fetch all yubikey tokens from the database. This could be thousand tokens. But the get_tokens() function is limited. It can not match for a tokeninfo.
But it should be possible to enhance the get_tokens() to match for certain tokeninfos. (but this will be another commit)

cornelinux:
Can this work at all?

You only create the tokeninfo yubikey.prefix during the check_otp function. But this is only called during a normal /validate/check API request. 
But you need the yubikey.prefix in the check_yubikey_pass static method, which is called in /ttype/yubikey.

If you are using the yubikey this way, you probably alway want to call /ttype/yubikey and you will end up in check_yubikey_pass. But as you never hit check_otp (since you never called /validate/check) the yubikeys will have no yubikey.prefix in the tokeninfo!

So the yubikey.prefix already needs to be written to the tokeninfo during the update() function.

...we should discuss this during daylight!

privacyidea/lib/tokens/yubikeytoken.py
jh23453: Cornelius Kölbel <notifications@github.com> writes:

>>  
>>          from privacyidea.lib.token import get_tokens
>>          from privacyidea.lib.token import check_token_list
>> -        for serial in serials:
>> -            tokenobject_list = get_tokens(serial=serial)
>> -            token_list.extend(tokenobject_list)
>> +
>> +        token_candidate_list = get_tokens(tokentype='yubikey')
>
> I am not content with this. This will fetch all yubikey tokens from
> the database. This could be thousand tokens. But the get_tokens()
> function is limited. It can not match for a tokeninfo.
> But it should be possible to enhance the get_tokens() to match for certain tokeninfos. (but this will be another commit)

Yes, that's most likely the best way forward.

Jochen

-- 
The only problem with troubleshooting is that the trouble shoots back.

jh23453:
Cornelius Kölbel <notifications@github.com> writes:

> Can this work at all?
>
> You only create the tokeninfo yubikey.prefix during the check_otp
> function. But this is only called during a normal /validate/check API
> request.
> But you need the yubikey.prefix in the check_yubikey_pass static method, which is called in /ttype/yubikey.
>
> If you are using the yubikey this way, you probably alway want to call
> /ttype/yubikey and you will end up in check_yubikey_pass. But as you
> never hit check_otp (since you never called /validate/check) the
> yubikeys will have no yubikey.prefix in the tokeninfo!
>
> So the yubikey.prefix already needs to be written to the tokeninfo during the update() function.

Yes, I think so, see https://github.com/privacyidea/privacyideaadm/issues/18

I just didn't find the correct place to add it yet.

Jochen

-- 
The only problem with troubleshooting is that the trouble shoots back.

cornelinux:
But the yubikey.prefix is not present during initialization at the moment. 

I have an idea to do this during enrollment. I need to enhance the enrollment function
https://github.com/privacyidea/privacyideaadm/blob/master/scripts/privacyidea#L113
and return the yubikey.prefix.

This needs to be added to the submit_params and the yubikey.update function needs to be enhanced.
But first we should merge your pull request
https://github.com/privacyidea/privacyideaadm/pull/17#discussion-diff-55929116L991
(please with default=0)

THanks a lot 
Cornelius

jh23453:
Cornelius Kölbel <notifications@github.com> writes:

> This needs to be added to the submit_params and the yubikey.update function needs to be enhanced.
> But first we should merge your pull request
> https://github.com/privacyidea/privacyideaadm/pull/17#discussion-diff-55929116L991
> (please with default=0)

I'll check what needs to be done, but it may take some days until I get
to it and can test it again.

Jochen

-- 
The only problem with troubleshooting is that the trouble shoots back.

cornelinux:
Hi Jochen,
if you are just fine with accepting default=0 i can merge the pull request at privacyideaadm and then I will add the yubikey.prefix parameter.

Thanks
Cornelius

jh23453:
Cornelius Kölbel <notifications@github.com> writes:

> Hi Jochen,
> if you are just fine with accepting default=0 i can merge
> the pull request at privacyideaadm and then I will add the
> yubikey.prefix parameter.

Of course I'm fine with it.  Thanks!

Jochen

-- 
The only problem with troubleshooting is that the trouble shoots back.

jh23453:
Cornelius Kölbel <notifications@github.com> writes:

> Merged #340.

Wonderful. Your fixes after that are far better than what I had in mind
after the last discussion.  So, many thanks.  I'll try to test the
latest changes soon.

Jochen

-- 
The only problem with troubleshooting is that the trouble shoots back.

PR #345, opened 2016-03-24
Added userlist right to user scope. Enable users with updateuser abil…

wheldom01:
…ity to edit their account details. Enable users with password_reset ability to reset their password

Hi,

Please could you take a look at this commit.

Best Regards

Martin

### DIFF
--- a/privacyidea/lib/policy.py
+++ b/privacyidea/lib/policy.py
@@ -971,6 +971,9 @@ def get_static_policy_definitions(scope=None):
             ACTION.AUDIT: {
                 'type': 'bool',
                 'desc': _('Allow the user to view his own token history.')},
+            ACTION.USERLIST: {'type': 'bool',
+                                'desc': _("The user is allowed to view his "
+                                          "own user information.")},
             ACTION.UPDATEUSER: {'type': 'bool',
                                 'desc': _("The user is allowed to update his "
                                           "own user information, like changing "
--- a/privacyidea/static/components/user/views/user.html
+++ b/privacyidea/static/components/user/views/user.html
@@ -4,7 +4,7 @@
         <div class="col-lg-3">
             <div class="pa-sidebar panel">
                 <ul class="nav nav-pills nav-stacked"
-                    ng-show="loggedInUser.role === 'user'">
+                    ng-show="loggedInUser.role == 'user' && checkRight('updateuser') && checkRight('password_reset')">
                    <li role="presentation" ng-class=
                            "{active: $state.includes('user.password')}">
                        <a ui-sref="user.password">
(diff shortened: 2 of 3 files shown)


cornelinux:
Hi Martin,
thanks a lot for this pull request.
What is your use case?
Why should a user not being allowed to view his own information.
I.e. why do you need this policy?

Kind regards
Cornelius

cornelinux:
OK, I had to think for it for a while.

The password_reset policy is not used for the user to CHANGE his password.
If the user is to be allowed to change his password, than he needs the update_user action.

password_reset means if the user has forgotten his password and starts the reset process via the login screen:
http://privacyidea.readthedocs.org/en/latest/policies/user.html#password-reset

But you are right about the update_user action:
which is not implemented at the moment in the web UI.
http://privacyidea.readthedocs.org/en/latest/policies/user.html#updateuser

So you should not use password_reset in this commit!

privacyidea/lib/policy.py
```diff
@@ -971,6 +971,9 @@ def get_static_policy_definitions(scope=None):
             ACTION.AUDIT: {
                 'type': 'bool',
                 'desc': _('Allow the user to view his own token history.')},
+            ACTION.USERLIST: {'type': 'bool',
```
cornelinux: In my opinion this makes no sense

privacyidea/static/components/user/views/user.html
```diff
@@ -4,7 +4,7 @@
         <div class="col-lg-3">
             <div class="pa-sidebar panel">
                 <ul class="nav nav-pills nav-stacked"
-                    ng-show="loggedInUser.role === 'user'">
+                    ng-show="loggedInUser.role == 'user' && checkRight('updateuser') && checkRight('password_reset')">
```
cornelinux: updateuser is fine.
But you need to remove password_reset as mentioned in the comment

privacyidea/static/components/user/views/user.html
```diff
@@ -13,7 +13,17 @@
                    </li>
                 </ul>
                 <ul class="nav nav-pills nav-stacked"
-                    ng-show="loggedInUser.role === 'admin'">
+                    ng-show="loggedInUser.role == 'user' && checkRight('updateuser') && !checkRight('password_reset')">
```
cornelinux: I think you can remove this!

wheldom01:
Hi Cornelius,

I'm afraid I missed that in the documentation, I'll remove it as I have 
used it for a
unintended purpose.

Best Regards

Martin

privacyidea/static/components/user/views/user.html
wheldom01: Hi Cornelius,

No problem, I remove this too.

Best Regards

Martin

privacyidea/lib/policy.py
wheldom01: Hi Cornelius,

I added this as the get_users() function in api/user.py has a prepolicy 
decorator that requires
the userlist action.

Maybe I have worked around a problem that should be fixed else where? 
This did occur to me when
I wrote the code..

Best Regards

Martin

cornelinux: Oh, you are totally right.
As soon as we define a user policy, the user will not be able to list his own data anymore.
So this needs to be added!
THanks!

cornelinux:
I think I now know, which parts to merge!
Thx

wheldom01:
Thanks Cornelius

PR #347, opened 2016-03-24
Fix layout in token details info section

wheldom01:
Hi,

Just a tiny patch to fix the layout of the list in the info section of token details.

Best Regards

Martin


cornelinux:
Thanks!

PR #354, opened 2016-03-29
yubico: make the Yubico cloud URL configurable

jh23453:
This patch adds a field to Config -> Tokens -> Yubico which can
contain another validation server than the Yubico servers.

If there is no configuration stored we still use the Yubico
servers as default, but we switch to the https-URL.

The field is marked tranlateable, but I don't have grunt available,
so there is no updated translation.


cornelinux:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/privacyidea/privacyidea/pull/354'>Refresh</a>.<a href='#crh-update-none'></a>

PR #373, opened 2016-04-13
FIX: deb packages break Apache and FreeRADIUS sites.

tuxmartin:
FIX: deb packages break Apache and FreeRADIUS sites.
Deb packages "privacyidea-apache2" and "privacyidea-radius" have
"rm -f /etc/apache2/sites-enabled/*" and "rm /etc/freeradius/sites-enabled/*"
in "postinst" files.
This disable all Apache VirtualHost and FreeRADIUS VirtualServers.
<a href='#crh-start'></a><a href='#crh-data-%7B%7D'></a>
<a href='https://www.codereviewhub.com/'></a>
<a href='https://www.codereviewhub.com/privacyidea/privacyidea/pull/373?approve=1'></a>&nbsp;<a href='https://github.com/privacyidea/privacyidea/pull/373'></a>
<a href='#crh-end'></a>

### DIFF
--- a/deploy/debian-jessie/privacyidea-radius.postinst
+++ b/deploy/debian-jessie/privacyidea-radius.postinst
@@ -35,7 +35,6 @@ HERE
 }
 
 activate_site() {
-	rm /etc/freeradius/sites-enabled/*
 	ln -s /etc/freeradius/sites-available/privacyidea /etc/freeradius/sites-enabled/	
 }
 
--- a/deploy/debian-ubuntu/privacyidea-apache2.postinst
+++ b/deploy/debian-ubuntu/privacyidea-apache2.postinst
@@ -95,7 +95,6 @@ create_database() {
 }
 
 enable_apache() {
-        rm -f /etc/apache2/sites-enabled/*
         a2enmod wsgi
         a2enmod headers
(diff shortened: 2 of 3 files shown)


cornelinux:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/privacyidea/privacyidea/pull/373'>Refresh</a>.<a href='#crh-update-none'></a>

cornelinux:
Hi Martin,
thanks a lot for looking into this and committing the pull request. But I can not merge it this way.

The most common installation scenario is installing this packages on a plain ubuntu system turning this into a privacyIDEA server. This is why we need to disable the default settings/servers.

Especially in the case of the RADIUS server the default sites (default and inner-tunnel) would simply disturb in this way, that the configuration would not work as expected.

The package privacyidea-apache2 is a meta package, that is ment to work this way. If you have a more complex apache or RADIUS configuration, you probably would only want to install the base package (python-privacyidea) and configure your whole system on your own. Then you could also decide about the database used etc. pp.

So unless you come up with a more sophisticated suggestion that also preserves the intended functionality, I can not merge this.

Thanks again 
Cornelius

PR #1021, opened 2018-04-17
Require ldap3 2.5 in setup.py and requirements.txt

### LINKED ISSUE #912: Problematic ldap3 / pyasn1 version combinations
This is an issue to keep track of problematic combinations of ldap3 / pyasn1 versions. See #887, #911.

Functionality I check for each combination:
* STARTTLS
* Complete user info on /validate/check
* Size Limit < number of users

Everything tested on a62fd9b3d0555b43cf31be3c4092a549fd73f724

| | pyasn 0.1.9 | pyasn 0.4.2 |
| --- | --- | --- 
| ldap3 2.1.1| ok | STARTTLS fails (#885)
| ldap3 2.3 | Empty user list (#887) | Empty user list (#887)
| ldap3 2.4 | Empty user list (#887), Empty user info (#911) | Empty user list (#887), Empty user info (#911)
| ldap3 2.4.1 | Empty user list (#887), **Error on /validate/check and resolver error** in token list (#911) | Empty user list (#887), **Error on /validate/check and resolver error** in token list (#911)

Combinations in current 2.21.4 packages:
* pip:
```
ldap3==2.1.1
privacyIDEA==2.21.4
pyasn1==0.4.2
```
* CentOS:
```
ldap3==2.3
privacyIDEA==2.21.4
pyasn1==0.4.2
```
* Ubuntu (Launchpad):
```
ldap3==2.1.1
privacyIDEA==2.21.4
pyasn1==0.1.9
```

fredreichbier:
This:
* fixes ldap3 2.5 and pyasn1 0.4.2 in requirements.txt (for development environments)
* requires ldap3 >= 2.5 and pyasn1>= 0.4.2 in setup.py (for future virtualenv-based packages)

Though I'm not sure if the changes to setup.py will break our Ubuntu-based setups which still run on ldap3 2.1.1+pyasn1 0.1.9?

Fixes #912


cornelinux:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/privacyidea/privacyidea/pull/1021'>Refresh</a>.<a href='#crh-update-none'></a>

fredreichbier:
Alternatively, we could also write a requirement like "ldap3>=2.1.1, but neither ldap3 2.3, nor 2.4, nor 2.4.1" to the setup.py.

cornelinux:
We also need to update ldap3 as a ubuntu package here:
https://github.com/privacyidea/privacyidea/blob/master/contrib/Makefile#L10

And in the same Makefile we would need to add pyasn, since Ubuntu comes with 0.1.9.

fredreichbier:
Cool @ ldap3 2.5!

Therereotically, ldap3 2.5 should also be compatible with pyasn1 0.1.9, in case we don't want to package it.

cornelinux:
Look! pyasn 0.4.2 builds fine on ubuntu 16.04 :-)

p53:
hi, i think for stability of software would be great if all dependencies and subdependencies at least for stable versions would be fixed, we have many bugs also with other python software related to such issues...

splashx:
My ops experience dictates that using `>=` can be rather reckless :)

cornelinux:
@p53 I am sorry, I do not get your point or question?

@splashx: We could do a 

    pyasn1>=0.1.9<0.5
    ldap3>=2.5<2.6

for the time being.

fredreichbier:
I think we need to separate here two locations where dependencies are defined: `setup.py` and `requirements.txt`.

It [seems to be best practice](https://packaging.python.org/discussions/install-requires-vs-requirements/) to do the following:
* Pin specific versions in `requirements.txt`. This is what we do (with two exceptions being `cffi>=1.7.0` and `psycopg2>=2.6`, but we can probably turn these into specific versions too).
* Specify abstract dependencies in `setup.py`. This is also what we do, as we only specifiy minimum version requirements there. We could improve this and add guards for packages that follow semantic versioning as described in the link above.

So I think we more or less follow the best practices here. We can always get a nearly determinstic environment by running ``pip install -r requirements.txt``. But I also understand the desire for more reproducible installations via ``pip install privacyidea``, as it also bothers me sometimes that two pip-based installations may have very different versions of dependencies and may behave very differently.

Maybe we should move this discussion to a new issue.

fredreichbier:
Better late than never: I've opened #1127 for this.

Issue #1057, opened 2018-05-08
TAN list token type


cornelinux:
The TAN list token is a simple, unordered list of TANs.

The TANs can be used in any arbitrary order. If a TAN is used, it is deleted from the token.

The TANs can be imported with a script and the following CSV:

    username, resolver, realm, serial, tokentype, otplen, tan1, tan2....

# Idea of implementation

The TANs are stored in the tokeninfo field of the token as key/value pairs. The key might be something like "token_tan" for all TANs.

The TAN is stored in a salted hash way. The salt is saved in the otpkey of the token.

In the event of authentication, we can salt the given OTP value, hash it and run a select on the tokeninfo table. If a value is found, authentication is successful and the tan is removed.

**Note:** We need to think about the token detail view in the UI. Otherwise we will fill up the view with all the hashed TANs.

berndnicklas:
Hi Cornelius,

as answer to your personal E-Mail: Yes, the required token type is no iTAN, which means no challenge-response and each user can decide the order of the provided TANs.

Best,

Bernd

cornelinux:
Some initial thoughts:

* Tokentype: **tan**
* Token Identifier: **PITN**
* otpkey is used for encrypted **salt**
* Key of the tokeninfo: **tan.tan**, to identifiy, that this is a value of a tan token (compare to "radius.identifier".)
* init parameters: **tantoken_count** (compare paper_token)

Two ways to enroll:

1. Import via CSV as mentioned above
2. Generate a random tan list, as with the paper token (genkey=1).


TANs could be generated based on the HMAC with ``get_multi_otp`` and be returned in "otps".
Then the salt gets generated and stored in otpkey.
Then the values can be hashed and stored in tokeninfo.

The update token method must be capable of accecpting a list of TANs. Maybe we simply use either the parameter otpkey (if possible) or a new parameter **tans** to pass a list of comma separated tans.

cornelinux:
Note: The key in tokeninfo is unique!

cornelinux:
We can add the tokeninfo.type **hide_ui**, and use this to hide all the tans from the UI.

cornelinux:
# TODO List

[x] Basic token functionality
[x] save tans to tokeninfo and check OTP
[ ] make TANs invisible in the UI-Token Details
[ ] import TANs

cornelinux:
The TANs should not be visible in the UI and should not appear in the REST API.
Instead of this, a list of TANs would be great.

[...]

PR #1071, opened 2018-05-17
1057/tan token

cornelinux:
This PR implements the TAN token.
It adds the token type "tan",
the ui enrollment
and a new way to import this token type (via a tan list in the oath csv file).

This is the PR for #1057. It probably does not completely close this issue, since we need to clarify if and how we want to hide the tans from the token detail view.

### DIFF
--- a/doc/configuration/tokens/tan.rst
+++ b/doc/configuration/tokens/tan.rst
@@ -1,0 +1,26 @@
+.. _tan_token:
+
+TAN Token
+---------
+
+.. index:: TAN Token
+
+(added in version 2.23)
+
+The token type *tan* is related the the :ref:`paper_token`.
+
+In contrast to the *paper* token, a user can use the OTP values of a *tan* token in
+any arbitrary order.
+
+A *tan* token can either be initialized with random OTP values. In this case the HOTP mechanism is used.
+Or it can be initialized or imported with a dedicated list of TANs.
+
+Import of TAN token
+~~~~~~~~~~~~~~~~~~~
+
+The import schema for TAN tokens via the OATH CSV file look like this:
+
+    <serial>, <seed>, tan, <white space separated list of tans>
+
+The TANs are located in the 4th column. TANs are separated by blanks or whitespaces.
+The *<seed>* is not used with a TAN token. You can leave this blank or set to any (not used) value.
--- a/doc/workflows_and_tools/import/index.rst
+++ b/doc/workflows_and_tools/import/index.rst
@@ -79,6 +79,22 @@ This is usually 6 or 8.
 
 **ocra suite** is the ocra suite of the OCRA token according to [#ocra]_.
 
+Import format version 2
+~~~~~~~~~~~~~~~~~~~~~~~
+
+A new import format allows to prepend a user, to whom the imported token should be assigned.
+
+The file format needs to start with the first line
+
+   # version: 2
+
+and the first three colums will be the user:
+
+   <username>, <resolver>, <realm>, <serial>, <seed>, <type>, ...
+
+.. note:: The import will bail out, if a specified user does not exist.
+
+
 Yubikey CSV
 -----------
 
--- a/doc/workflows_and_tools/import/index.rst
+++ b/doc/workflows_and_tools/import/index.rst
@@ -79,6 +79,12 @@ This is usually 6 or 8.
 
 **ocra suite** is the ocra suite of the OCRA token according to [#ocra]_.
 
+For TAN tokens it looks like this::
+
+   <serial, <n/a>, TAN, <list of tans>
+
+The list of tans is a whitespace separated list.
+
 Import format version 2
 ~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/privacyidea/lib/token.py
+++ b/privacyidea/lib/token.py
@@ -80,7 +80,7 @@ from privacyidea.lib.config import get_from_config
 from privacyidea.lib.config import (get_token_class, get_token_prefix,
                                     get_token_types,
                                     get_inc_fail_count_on_false_pin)
-from privacyidea.lib.user import get_user_info
+from privacyidea.lib.user import get_user_info, User
 from privacyidea.lib import _
 from privacyidea.lib.realm import realm_is_defined
 from privacyidea.lib.resolver import get_resolver_object
@@ -863,18 +863,25 @@ def import_token(serial, token_dict, default_hashlib=None, tokenrealms=None):
     """
     init_param = {'serial': serial,
                   'type': token_dict['type'],
-                      'description': token_dict.get("description",
-                                                        "imported"),
-                      'otpkey': token_dict['otpkey'],
-                      'otplen': token_dict.get('otplen'),
-                      'timeStep': token_dict.get('timeStep'),
-                      'hashlib': token_dict.get('hashlib')}
+                  'description': token_dict.get("description",
+                                                "imported"),
+                  'otpkey': token_dict['otpkey'],
+                  'otplen': token_dict.get('otplen'),
+                  'timeStep': token_dict.get('timeStep'),
+                  'hashlib': token_dict.get('hashlib')}
+
+    user_obj = None
+    if token_dict.get("user"):
+        user_obj = User(token_dict.get("user").get("username"),
+                        token_dict.get("user").get("realm"),
+                        token_dict.get("user").get("resolver"))
 
     if default_hashlib and default_hashlib != "auto":
         init_param['hashlib'] = default_hashlib
 
     # Imported tokens are usually hardware tokens
-    token = init_token(init_param, tokenrealms=tokenrealms,
+    token = init_token(init_param, user=user_obj,
+                       tokenrealms=tokenrealms,
                        tokenkind=TOKENKIND.HARDWARE)
     if token_dict.get("counter"):
         token.set_otp_count(token_dict.get("counter"))
--- a/privacyidea/lib/tokens/tantoken.py
+++ b/privacyidea/lib/tokens/tantoken.py
@@ -121,10 +121,18 @@ class TanTokenClass(PaperTokenClass):
         return ret
 
     def update(self, param, reset_failcount=True):
-        param["papertoken_count"] = param.get("tantoken_count") or DEFAULT_COUNT
-        PaperTokenClass.update(self, param, reset_failcount=reset_failcount)
-        # After this creation, the init_details contain the complete list of the TANs
-        for tankey, tanvalue in self.init_details.get("otps", {}).iteritems():
+        tan_dict = {}
+        if "tans" in param:
+            # init tokens with tans
+            tans = param.get("tans").split()
+            tan_dict = {k: v for k, v in enumerate(tans)}
+        else:
+            # Init token without tans, so we create tans
+            param["papertoken_count"] = param.get("tantoken_count") or DEFAULT_COUNT
+            PaperTokenClass.update(self, param, reset_failcount=reset_failcount)
+            # After this creation, the init_details contain the complete list of the TANs
+            tan_dict = self.init_details.get("otps", {})
+        for tankey, tanvalue in tan_dict.iteritems():
             # Get a 4 byte salt from the crypto module
             salt = geturandom(SALT_LENGTH, hex=True)
             # Now we add all TANs to the tokeninfo of this token.
@@ -159,3 +167,24 @@ class TanTokenClass(PaperTokenClass):
                     return 1
 
         return res
+
+    @staticmethod
+    def get_import_csv(l):
+        """
+        Read the list from a csv file and return a dictionary, that can be used
+        to do a token_init.
+
+        :param l: The list of the line of a csv file
+        :type l: list
+        :return: A dictionary of init params
+        """
+        params = TokenClass.get_import_csv(l)
+        # Delete the otplen, if it exists. The fourth column is the TANs!
+        if "otplen" in params:
+            del params["otplen"]
+
+        # tans
+        if len(l) >= 4:
+            params["tans"] = l[3]
+
+        return params
--- a/privacyidea/lib/tokens/tantoken.py
+++ b/privacyidea/lib/tokens/tantoken.py
@@ -1,0 +1,151 @@
+# -*- coding: utf-8 -*-
+#
+#  2018-05-09 Cornelius Kölbel <cornelius.koelbel@netknights.it>
+#             TAN token with to be randomly used TAN values
+#
+#  (c) 2018 Cornelius Kölbel - cornelius.koelbel@netknights.it
+#
+# This code is free software; you can redistribute it and/or
+# modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
+# License as published by the Free Software Foundation; either
+# version 3 of the License, or any later version.
+#
+# This code is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+#
+# You should have received a copy of the GNU Affero General Public
+# License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+This file contains the definition of the paper token class
+It depends on the DB model, and the lib.tokenclass.
+"""
+
+import logging
+from privacyidea.lib.log import log_with
+from privacyidea.lib.tokenclass import TokenClass
+from privacyidea.lib.tokens.papertoken import PaperTokenClass
+from privacyidea.lib.policy import SCOPE
+from privacyidea.lib import _
+from privacyidea.lib.policydecorators import libpolicy
+
+log = logging.getLogger(__name__)
+DEFAULT_COUNT = 100
+
+
+class TANACTION(object):
+    TANTOKEN_COUNT = "papertoken_count"
+
+
+class TanTokenClass(PaperTokenClass):
+    """
+    The TAN token allows to print out the next e.g. 100 OTP values.
+    This sheet of paper can be used to authenticate and strike out the used
+    OTP values. In contrast to the paper token, the OTP values of the TAN token
+    can be used randomly.
+    """
+
+    @log_with(log)
+    def __init__(self, db_token):
+        """
+        This creates a new TAN token object from a DB token object.
+
+        :param db_token: instance of the orm db object
+        :type db_token:  orm object
+        """
+        TokenClass.__init__(self, db_token)
+        self.set_type(u"tan")
+        self.hKeyRequired = False
+
+    @staticmethod
+    def get_class_type():
+        """
+        return the token type shortname
+
+        :return: 'paper'
+        :rtype: string
+        """
+        return "tan"
+
+    @staticmethod
+    def get_class_prefix():
+        """
+        Return the prefix, that is used as a prefix for the serial numbers.
+        :return: PITN
+        """
+        return "PITN"
+
+    @staticmethod
+    @log_with(log)
+    def get_class_info(key=None, ret='all'):
+        """
+        returns a subtree of the token definition
+
+        :param key: subsection identifier
+        :type key: string
+        :param ret: default return value, if nothing is found
+        :type ret: user defined
+        :return: subsection if key exists or user defined
+        :rtype: dict or scalar
+        """
+        res = {'type': 'tan',
+               'title': 'TAN Token',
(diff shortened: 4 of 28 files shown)


cornelinux:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/privacyidea/privacyidea/pull/1071'>Refresh</a>.<a href='#crh-update-none'></a>

doc/configuration/tokens/tan.rst
```diff
@@ -0,0 +1,26 @@
+.. _tan_token:
+
+TAN Token
+---------
+
+.. index:: TAN Token
+
+(added in version 2.23)
+
+The token type *tan* is related the the :ref:`paper_token`.
```
fredreichbier: "to the"

doc/workflows_and_tools/import/index.rst
```diff
@@ -79,6 +79,28 @@ This is usually 6 or 8.
 
 **ocra suite** is the ocra suite of the OCRA token according to [#ocra]_.
 
+For TAN tokens it looks like this::
+
+   <serial, <n/a>, TAN, <list of tans>
```
fredreichbier: missing a ``>`` here

privacyidea/lib/tokens/tantoken.py
```diff
@@ -0,0 +1,190 @@
...
+# This code is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+#
+# You should have received a copy of the GNU Affero General Public
+# License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+This file contains the definition of the paper token class
```
fredreichbier: tan token class

privacyidea/lib/tokens/tantoken.py
```diff
@@ -0,0 +1,190 @@
...
+            PaperTokenClass.update(self, param, reset_failcount=reset_failcount)
+            # After this creation, the init_details contain the complete list of the TANs
+            tan_dict = self.init_details.get("otps", {})
+        for tankey, tanvalue in tan_dict.iteritems():
+            # Get a 4 byte salt from the crypto module
+            salt = geturandom(SALT_LENGTH, hex=True)
+            # Now we add all TANs to the tokeninfo of this token.
+            hashed_tan = binascii.hexlify(hash(tanvalue, salt))
+            self.add_tokeninfo("tan.tan{0!s}".format(tankey),
+                               "{0}:{1}".format(binascii.hexlify(salt),
```
fredreichbier: Hmm, shouldn't the salt already be a hexstring because we passed ``hex=True``?

privacyidea/lib/tokens/tantoken.py
```diff
@@ -0,0 +1,190 @@
...
+        :return: the counter state or -1
+        :rtype: int
+        """
+        res = -1
+        tans = self.get_tokeninfo()
+        for tankey, tanvalue in tans.iteritems():
+            if tankey.startswith("tan.tan"):
+                salt, tan = tanvalue.split(":")
+                if tan == binascii.hexlify(hash(anOtpVal,
+                                                binascii.unhexlify(salt))):
```
fredreichbier: we wouldn't need to unhexlify the salt here, if we remove the hexlify step in ``update``

privacyidea/lib/tokens/tantoken.py
```diff
@@ -0,0 +1,190 @@
...
+               'user':  ['enroll'],
+               # This tokentype is enrollable in the UI for...
+               'ui_enroll': ["admin", "user"],
+               'policy': {
+                   SCOPE.ENROLL: {
+                       TANACTION.TANTOKEN_COUNT: {
+                           "type": "int",
+                           "desc": _("The number of OTP values, which are "
+                                     "printed on the paper.")
+                       }
```
fredreichbier: Hmm, does this policy currently have an effect (I think it has the same action name as the corresponding paper token action)

privacyidea/lib/tokens/tantoken.py
```diff
@@ -0,0 +1,190 @@
...
+        tans = self.get_tokeninfo()
+        for tankey, tanvalue in tans.iteritems():
+            if tankey.startswith("tan.tan"):
+                salt, tan = tanvalue.split(":")
+                if tan == binascii.hexlify(hash(anOtpVal,
+                                                binascii.unhexlify(salt))):
+                    self.del_tokeninfo(tankey)
+                    return 1
+
+        return res
```
fredreichbier: as ``res`` is not used, we could also just ``return -1`` here

privacyidea/lib/tokens/tantoken.py
```diff
@@ -0,0 +1,190 @@
...
+            if ret == 'all':
+                ret = res
+        return ret
+
+    def update(self, param, reset_failcount=True):
+        tan_dict = {}
+        if "tans" in param:
+            # init tokens with tans
+            tans = param.get("tans").split()
+            tan_dict = {k: v for k, v in enumerate(tans)}
```
fredreichbier: In case we provide a list of TANs, we do not call the parent class' ``update`` implementation. I guess that's intentional because we do not need a OTP key and such, but this way we also lose the functionality of ``TokenClass.update``, which also sets the token description, PIN and token kind. This could be a problem, I think?

cornelinux:
> A question: The TAN token with predefined TANs does not really play well with a token PIN, does it? Because we cannot be sure that the TANs all have the same length, we cannot chop off the dynamic part reliably. So I guess we could either disable PINs altogether, or check that all TANs have the same length when enrolling?

You are right. I totally missed this! Of course we need to be able to define the length of the TAN :-/

I thought it would be that nice, that we could even import TANs of arbitrary lengths. But then - no pin. I will adapt this!

doc/configuration/tokens/tan.rst
cornelinux: done

doc/workflows_and_tools/import/index.rst
cornelinux: done

privacyidea/lib/tokens/tantoken.py
cornelinux: done

privacyidea/lib/tokens/tantoken.py
cornelinux: The policy actually does not have the same name, since if a token type specific policy does not start with the tokentype, the tokentype is automatically added as prefix. So this policy results in the action name "tan_papertoken_count". I think nevertheless I should rename it.

But you are right. We are missing a decorator for this policy!

privacyidea/lib/tokens/tantoken.py
cornelinux: True. I thought this would be more consistent? (Or I did not think at all)
After all, we *could* return a value>0, like the number of the TAN...

privacyidea/lib/tokens/tantoken.py
cornelinux: done!

privacyidea/lib/tokens/tantoken.py
cornelinux: You are right. Done.

privacyidea/lib/tokens/tantoken.py
cornelinux: Done. I removed the double hashing.

privacyidea/lib/tokens/tantoken.py
cornelinux: hm. You are right. I did not want to add a useless otpkey. But maybe we should...
(The problem with the PIN and the different length is still open)

cornelinux: done. Added the update method and also created an otplen from the tan list.

privacyidea/lib/tokens/tantoken.py
```diff
@@ -126,9 +126,15 @@ def update(self, param, reset_failcount=True):
             # init tokens with tans
             tans = param.get("tans").split()
             tan_dict = {k: v for k, v in enumerate(tans)}
+            # Avoid to generate TANs, since we get the from params
+            param["tantoken_count"] = 0
+            # Determine the otplen from the TANs
+            if len(tans) > 0:
+                param["otplen"] = len(tans[0])
```
fredreichbier: Hm, should we maybe validate the lengths of all supplied TANs? Or should we handle this as an instance of SISO? :)

privacyidea/lib/tokens/tantoken.py
```diff
@@ -126,9 +126,15 @@ def update(self, param, reset_failcount=True):
...
+            PaperTokenClass.update(self, param, reset_failcount=reset_failcount)
         else:
             # Init token without tans, so we create tans
-            param["papertoken_count"] = param.get("tantoken_count") or DEFAULT_COUNT
+            param["tantoken_count"] = param.get("tantoken_count") or DEFAULT_COUNT
```
fredreichbier: I think if we rename the parameter here, the underlying ``PaperTokenClass.update`` will always generate 100 TANs because it gets the number of OTPs to generate from the non-existing ``papertoken_count`` parameter.

privacyidea/lib/tokens/tantoken.py
```diff
@@ -126,9 +126,15 @@ def update(self, param, reset_failcount=True):
             # init tokens with tans
             tans = param.get("tans").split()
             tan_dict = {k: v for k, v in enumerate(tans)}
+            # Avoid to generate TANs, since we get the from params
+            param["tantoken_count"] = 0
```
fredreichbier: As in the other case, the ``PaperTokenClass.update`` function will still generate 100 OTP values here because it reads the ``papertoken_count`` parameter.

privacyidea/lib/tokens/tantoken.py
cornelinux: I like siso a lot.
What should we do, if there were TANs with different length.
These tans are only imported tans. So the import file has to be crafted carefully anyways.

If you come up with a method that verifies tans with the same length, go for it. What should we do then, if we encounter different tan-length (which will neverTM happen) ;-)

privacyidea/lib/tokens/tantoken.py
cornelinux: Oh, you are right. Now I know, why the parameter was used this way...
I will rechange this!

privacyidea/lib/tokens/tantoken.py
cornelinux: Will change this! thx!

privacyidea/lib/tokens/tantoken.py
cornelinux: The first and probably only use case will be the import of a csv file.
To my knowledge the CSV file import does behave nicely if it encounters a ParameterError or any other exception. It will simply break the complete import.

Far from perfect. But this is probably another issue.

fredreichbier:
Is it possible that importing CSVs with preexisting TANs via the WebUI does not work yet?
I tried with
```
ABC123,,tan,123465 798111
```
However, privacyIDEA seems to ignore the given TANs and generates 100 new TANs. I think this is because ``tans`` is never passed to ``TanTokenClass.update``.

privacyidea/lib/tokens/tantoken.py
```diff
@@ -0,0 +1,218 @@
...
+        return ret
+
+    def update(self, param, reset_failcount=True):
+        tan_dict = {}
+        if "tans" in param:
+            # init tokens with tans
+            tans = param.get("tans").split()
+            tan_dict = {k: v for k, v in enumerate(tans)}
+            # Avoid to generate TANs in the superclass PaperToken, since we get the from params
+            param["papertoken_count"] = 0
```
fredreichbier: This does not seem to work, because PaperTokenClass uses
```
papertoken_count = int(param.get("papertoken_count") or DEFAULT_COUNT)
```
and since ``param.get("papertoken_count")`` evaluates to 0, DEFAULT_COUNT is used.

... if we want to be especially hacky, we could use ``param["papertoken_count"] = "0"`` ... :)

privacyidea/lib/tokens/tantoken.py
```diff
@@ -0,0 +1,218 @@
...
+
+        if key:
+            ret = res.get(key, {})
+        else:
+            if ret == 'all':
+                ret = res
+        return ret
+
+    def update(self, param, reset_failcount=True):
+        tan_dict = {}
```
fredreichbier: As ``tan_dict`` is defined in both branches, we can omit this line.

cornelinux:
> Is it possible that importing CSVs with preexisting TANs via the WebUI does not work yet?
I tried with
ABC123,,tan,123465 798111
However, privacyIDEA seems to ignore the given TANs and generates 100 new TANs. I think this is because tans is never passed to TanTokenClass.update.

I can confirm this. :-(

privacyidea/lib/tokens/tantoken.py
cornelinux: You are right. On the other hand the two if branches are very long! If we change them one day, we might end up with an undefined dict.
What do you think should be the best practice here? Maybe you are right...

privacyidea/lib/tokens/tantoken.py
cornelinux: Well I think actually in the line

~~~~python
    papertoken_count = int(param.get("papertoken_count") or DEFAULT_COUNT
~~~~

what we mean is:

~~~~python
    if "papertoken_count" not in param:
       papertoken_count = DEFAULT_COUNT
    else:
       papertoken_count = param.get("papertoken_count")
~~~~

Or

~~~~python
    papertoken_count = DEFAULT_COUNT
    if "papertoken_count" in param:
        papertoken_count = param.get("papertoken_count")
~~~~

But since usually param really contains strings, since they are passed from the API, we could also use the string "0".

Whhat do you think?

fredreichbier: Hm, I think both alternatives you mentioned are equivalent to
```python
papertoken_count = param.get("papertoken_count", DEFAULT_COUNT)
```
(well, we need to use ``int(...)`` as well though)

The difference to the current behavior, however, is that ``papertoken_count`` can be the empty string currently (we would choose DEFAULT_COUNT then). With the changed line, this is not allowed anymore. Do you think we should allow the empty string?

privacyidea/lib/tokens/tantoken.py
fredreichbier: Hmm, I would think that in that case, having a undefined dict would be better than having a dictionary which is unintentionally empty, because the undefined dict would get noticed faster :)
However, this is probably a bit nit-picky. So I would be okay with the above extra line as well.

privacyidea/lib/token.py
```diff
@@ -862,13 +862,11 @@ def import_token(serial, token_dict, default_hashlib=None, tokenrealms=None):
...
                   'description': token_dict.get("description",
-                                                "imported"),
-                  'otpkey': token_dict['otpkey'],
-                  'otplen': token_dict.get('otplen'),
-                  'timeStep': token_dict.get('timeStep'),
-                  'hashlib': token_dict.get('hashlib')}
+                                                "imported")}
+    for p in ['type', 'otpkey', 'otplen', 'timeStep', 'hashlib', 'tans']:
+        if p in token_dict:
+            init_param[p] = token_dict[p]
```
fredreichbier: This slightly changes behavior: Currently, if ``token_dict["hashlib"]`` is None, ``init_param["hashlib"]`` is None. With this PR, ``init_param`` does not contain a key ``hashlib``.
But I think this won't cause problems ...?

Apart from that, I think having the token-specific parameter names in ``import_token`` is not perfect in terms of modularity. But I can imagine changing this would require some major refactoring, so I would keep it as it is for now.

privacyidea/api/lib/prepolicy.py
```diff
@@ -342,6 +342,37 @@ def papertoken_count(request=None, action=None):
     return True
 
 
+def tantoken_count(request=None, action=None):
```
fredreichbier: I think this policy is not actually in effect yet?

privacyidea/lib/tokens/tantoken.py
cornelinux: I think you are right. I do not know if there was a scenario, where we explicitly wanted to have ``DEFAULT_COUNT`` paper tokens, although the user specified ``papertoken_count = 0``

If will got with

    int(param.get("papertoken_count", DEFAULT_COUNT))

privacyidea/api/lib/prepolicy.py
cornelinux: done

privacyidea/lib/tokens/tantoken.py
cornelinux: Done.

privacyidea/lib/tokens/tantoken.py
```diff
@@ -0,0 +1,194 @@
+# -*- coding: utf-8 -*-
+#
+#  2018-05-09 Cornelius Kölbel <cornelius.koelbel@netknights.it>
+#             TAN token with to be randomly used TAN values
+#
```
fredreichbier: :+1:<a href='#crh-update-none'></a>

privacyidea/lib/tokens/tantoken.py
fredreichbier: :+1:<a href='#crh-update-none'></a>

privacyidea/api/lib/prepolicy.py
fredreichbier: :+1:<a href='#crh-update-none'></a>

privacyidea/lib/tokens/tantoken.py
fredreichbier: :+1:<a href='#crh-update-none'></a>

privacyidea/lib/tokens/tantoken.py
```diff
@@ -0,0 +1,194 @@
...
+#  (c) 2018 Cornelius Kölbel - cornelius.koelbel@netknights.it
+#
+# This code is free software; you can redistribute it and/or
+# modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
+# License as published by the Free Software Foundation; either
+# version 3 of the License, or any later version.
+#
+# This code is distributed in the hope that it will be useful,
```
fredreichbier: :+1:<a href='#crh-update-none'></a>

privacyidea/lib/tokens/tantoken.py
fredreichbier: :+1:<a href='#crh-update-none'></a>

privacyidea/lib/tokens/tantoken.py
fredreichbier: :+1:<a href='#crh-update-none'></a>

doc/workflows_and_tools/import/index.rst
fredreichbier: :+1:<a href='#crh-update-none'></a>

privacyidea/lib/tokens/tantoken.py
fredreichbier: :+1:<a href='#crh-update-none'></a>

privacyidea/lib/tokens/tantoken.py
fredreichbier: :+1:<a href='#crh-update-none'></a>

privacyidea/lib/tokens/tantoken.py
```diff
@@ -0,0 +1,194 @@
+# -*- coding: utf-8 -*-
+#
+#  2018-05-09 Cornelius Kölbel <cornelius.koelbel@netknights.it>
+#             TAN token with to be randomly used TAN values
+#
+#  (c) 2018 Cornelius Kölbel - cornelius.koelbel@netknights.it
+#
+# This code is free software; you can redistribute it and/or
```
fredreichbier: :+1:<a href='#crh-update-none'></a>

privacyidea/lib/tokens/tantoken.py
fredreichbier: :+1:<a href='#crh-update-none'></a>

privacyidea/lib/tokens/tantoken.py
fredreichbier: :+1:<a href='#crh-update-none'></a>

doc/configuration/tokens/tan.rst
fredreichbier: :+1:<a href='#crh-update-complete'></a>

fredreichbier:
:shipit:

PR #1137, opened 2018-07-27
Script to make base32 secrets importable

cornelinux:
This script takes base32 encoded secrets and creates
an import file to be used by privacyIDEA.

This is supposed to be a simple script running under supervision.
I know that it will create false import data, if the paramter "-s" would be used without the paramter "-t". 
But the admin should realize the garbage when written to stdout.

### DIFF
--- a/tools/convert-base32.py
+++ b/tools/convert-base32.py
@@ -1,0 +1,57 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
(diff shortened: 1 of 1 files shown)


cornelinux:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/privacyidea/privacyidea/pull/1137'>Refresh</a>.<a href='#crh-update-none'></a>

PR #1234, opened 2018-09-08
Only show limited assign-err message for user

### LINKED ISSUE #1232: User tries to assign an already assigned token
When a user tries to assign an already assigned user, the system responds back with a message like this:

    ERR1103: Token already assigned to user User(...)

This gives out unnecessarily too much information. It should just say "Token already assigned to another user"

The ``assign`` API is the same for *admin* and *user*. We could already test on the server side, if this request is done by an admin or by a user and craft the respond accordingly.

But on the library level (https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/token.py#L1133) we currently do not have the information, if this is an admin or a user. So we would have to pass down this information from here:
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/api/token.py#L425

cornelinux:
If user assigns a token and the token already
belongs to another user, only a limited error
message is displayed. It does not tell to whom
this token is already assigned.

Closes #1232

### DIFF
--- a/privacyidea/api/token.py
+++ b/privacyidea/api/token.py
@@ -422,7 +422,12 @@ def assign_api():
     serial = getParam(request.all_data, "serial", required, allow_empty=False)
     pin = getParam(request.all_data, "pin")
     encrypt_pin = getParam(request.all_data, "encryptpin")
-    res = assign_token(serial, user, pin=pin, encrypt_pin=encrypt_pin)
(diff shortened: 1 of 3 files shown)


cornelinux:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/privacyidea/privacyidea/pull/1234'>Refresh</a>.<a href='#crh-update-none'></a>

Issue #331, opened 2016-02-24
remove flask imports from lib


cornelinux:
There are some flask imports in privacyidea.lib.

We should remove them, to be able to use privacyidea.lib without any reference to flask.

cornelinux:
This probably won't work out.

fredreichbier:
While we probably won't be able to remove all Flask imports from privacyidea.lib, we could try to minimize them: For example, we use `flask.g` for per-request caching. We could define a central function like the following:

<pre>
def get_request_local_context():
    return flask.g
</pre>

and use this function instead of `flask.g` in `privacyidea.lib`.

fredreichbier:
The code cleanup milestone may be a good opportunity to minimize the Flask imports in ``privacyidea.lib``. I think we essentially need an "application context" (shared between threads) and "request-local context" (thread-local) which different responsibilities:
* the application context stores objects like the configuration, the policies, the engine registry, the HSM object, maybe at some point an LDAP server pool?
* the request context stores objects like the request-local configuration snapshot, request-local policy snapshot, the current resolver objects

Like above, I'm not sure if we can completely get rid of Flask imports. But we could avoid to explicitly import Flask names by having two functions
```python
def get_application_local_context():
    return something

def get_request_local_context():
    return flask.g
```
and only using these two functions in ``privacyidea.lib``. We could do this without touching too much existing code.

cornelinux:
The flask dependency was reduced to ``lib/framework.py`` (and one occurrance in the federation handler).
This is as good as we should get. 
In aspect of being able to migrate to a potential other request framework (like bottle ;-) one day, the effort of code changes would be small enough.

PR #1254, opened 2018-09-19
Avoid reload of static configuration

### LINKED ISSUE #1253: Avoid reloading of Token/Resolver classes and types
Caching the Token/Resolver classes and types improves performance.

plettich:
The pretty static configuration of available Token/Resolver classes and
types was reloaded every time which lead to a performance degradation when
called for a lot of tokens.
Caching the configuration on a module level solves this.
In addition this reduces the dependency on flask in the libs (See #331).

Fixes #1253

### DIFF
--- a/privacyidea/lib/config.py
+++ b/privacyidea/lib/config.py
@@ -33,6 +33,7 @@ It provides functions to retrieve (get) and and set configuration.
 The code is tested in tests/test_lib_config
 """
 
+import sys
 import logging
 import inspect
 from flask import current_app, g
@@ -53,6 +54,11 @@ import datetime
 log = logging.getLogger(__name__)
 
 ENCODING = 'utf-8'
+
+# this is a pointer to the module object instance itself.
+this = sys.modules[__name__]
+
+this.config = {}
 
 
 class Singleton(type):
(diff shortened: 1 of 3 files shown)


cornelinux:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/privacyidea/privacyidea/pull/1254'>Refresh</a>.<a href='#crh-update-none'></a>

privacyidea/lib/config.py
```diff
@@ -331,15 +332,12 @@ def get_token_class(tokentype):
...
-    for module, ttype in type_dict.items():
-        if ttype.lower() == tokentype.lower():
-            tokenmodule = module
-            break
-    if tokenmodule:
-        tokenclass = class_dict.get(tokenmodule)
+    tokenclasses = get_token_classes()
+    for tclass in tokenclasses:
+        if tclass.get_class_type().lower() == tokentype.lower():
+            tokenclass = tclass
```
cornelinux: I think we should ``break`` here out of the for loop.

tests/test_lib_config.py
```diff
@@ -79,16 +79,12 @@ def test_01_resolver(self):
...
-        # At the beginning the resolver classes are not cached.
-        self.assertFalse("pi_resolver_classes" in current_app.config,
-                                current_app.config)
-        r = get_resolver_classes()
-        self.assertTrue(PWResolver in r, r)
-        # Now the resolver classes are cached
-        self.assertTrue("pi_resolver_classes" in current_app.config,
-                                current_app.config)
+        # resolver classes are not cached at this point
+        self.assertFalse("pi_resolver_classes" in this.config, this.config)
```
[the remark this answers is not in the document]
plettich: already investigating. It seems we can't predict if the value was already cached from a test before so we would just assume it is 'undefined' here.
But we know, after calling `get_resolver_classes()` the entry must be there.

privacyidea/lib/config.py
cornelinux: :+1:<a href='#crh-update-none'></a>

tests/test_lib_config.py
cornelinux: :+1:<a href='#crh-update-complete'></a>

cornelinux:
:shipit:

PR #1312, opened 2018-11-16
fix dictionary usage for python3 compatibility

plettich:
The usage of dictionaries changed between python2 and python3.
See <https://portingguide.readthedocs.io/en/latest/dicts.html#dictionaries>


cornelinux:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/privacyidea/privacyidea/pull/1312'>Refresh</a>.<a href='#crh-update-none'></a>

plettich:
Regarding the inefficiencies of python3 idioms in python2 i have done same tests:
Python2:
> In [1]: D = dict(zip(range(100), range(100)))
In [4]: %timeit for k,v in D.iteritems(): pass
100000 loops, best of 3: 2.06 µs per loop
In [5]: %timeit for k,v in D.items(): pass
100000 loops, best of 3: 3.17 µs per loop
In [6]: %timeit for k,v in six.iteritems(D): pass
100000 loops, best of 3: 2.17 µs per loop

Python3:
> In [255]: %timeit for k,v in D.items(): pass                                                                                                                                                  
1.93 µs ± 14.3 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
In [256]: %timeit for k,v in six.iteritems(D): pass                                                                                                                                           
2.2 µs ± 43.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

privacyidea/api/policy.py
```diff
@@ -523,9 +523,9 @@ def get_policy_defs(scope=None):
     dynamic_pol = get_dynamic_policy_definitions()
 
     # combine static and dynamic policies
-    keys = static_pol.keys() + dynamic_pol.keys()
-    pol = {k: dict(static_pol.get(k, {}).items()
-                   + dynamic_pol.get(k, {}).items()) for k in keys}
+    keys = list(static_pol.keys()) + list(dynamic_pol.keys())
```
cornelinux: Wouldn't it be

~~~~python
keys = list(static_pol) + list(dynamic_pol)
~~~~

privacyidea/lib/policy.py
```diff
@@ -761,9 +761,9 @@ def ui_get_rights(self, scope, realm, username, client=None):
...
             # possible actions in this scope.
             log.debug("No policies defined, so we set all rights.")
-            static_rights = get_static_policy_definitions(scope).keys()
-            enroll_rights = get_dynamic_policy_definitions(scope).keys()
-            rights = static_rights + enroll_rights
+            tmp_rights = get_static_policy_definitions(scope)
+            tmp_rights.update(get_dynamic_policy_definitions(scope))
+            rights = list(tmp_rights)
         # reduce the list
         rights = list(set(rights))
```
cornelinux: We can drop this line, since the dict ``tmp_rights`` contains unique keys. Using ``list(set())`` was used to make the entris in the list unique.

privacyidea/lib/user.py
```diff
@@ -597,20 +597,12 @@ def get_user_list(param=None, user=None):
     searchDict = {"username": "*"}
     param = param or {}
 
-    # we have to recreate a new searchdict without the realm key
-    # as delete does not work
-    for key in param:
```
cornelinux: You must not drop the ``for``-loop. There could be other search parameters like *email* which you are now missing and which are not contained in the ``searchDict``.

tests/test_api_lib_policy.py
```diff
@@ -1802,7 +1802,7 @@ def test_07_sign_response(self):
...
+        self.assertEqual(jresult.get("signature"), "11355158914966210201410734667484298031497086510917116878993822963793177737963323849914979806826759273431791474575057946263651613906587629736481370983420295626001055840803201448376203681672140726404056349423937599480275853513810616624349811159346536182220806878464577429106150903913526744093300868582898892977164229848617413618851794501457802670374543399415905458325601994002527427083792164898293507308423780001137468154518279116138010266341425663850327379848131113626641510715557748879427991785684858504631545256553961505159377600982900016536629720752767147086708626971940835730555782551222922985302674756190839458609")
```
cornelinux: I understand that the keys of the response are now sorted for signature, so the signature changes but will stay the same?

tests/test_lib_policy.py
```diff
@@ -497,7 +497,7 @@ def test_17_ui_get_rights(self):
         P = PolicyClass()
         rights = P.ui_get_rights(SCOPE.USER, "realm2", "user")
         # there was still another policy...
-        self.assertEqual(rights, ["enable", "disable"])
+        self.assertEqual(set(rights), {"enable", "disable"})
```
cornelinux: is there a reason why this can not be compares as a list anymore?
WOuld this change the program behaviour?

privacyidea/api/policy.py
plettich: Yes, You are right.

privacyidea/lib/policy.py
plettich: correct

privacyidea/lib/token.py
```diff
@@ -2027,7 +2027,7 @@ def check_token_list(tokenobject_list, passw, user=None, options=None):
     # add the user to the options, so that every token, that get passed the
     # options can see the user
     options = options or {}
-    options = dict(options.items() + {'user': user}.items())
+    options.update({'user': user})
```
fredreichbier: Might be a bit nitpicking :-) But if the caller passes ``options``, this change will modify the object in-place, which was not the case before. I think it would be safer to have something like
```python
if options:
    options = options.copy()
else:
    options = {}
options.update(...)
```

cornelinux: oups I was not aware of the copy. I think I changed this also in my PR. But you are right.

privacyidea/lib/policy.py
plettich: What about the block before? Could we get duplicate entries from the
current policies?

privacyidea/lib/user.py
plettich: You are right, i am on it.

tests/test_api_lib_policy.py
plettich: Yes, as long as there are the same keys/values in the dictionary. Before we
dumped the dictionary as a JSON-string without any order (which was fine
since the order did not change, but it wasn't guaranteed).
Now the the keys of the dictionary get sorted before the dump and thus the
dumped JSON-String should be the same regardless of the sorting of the
dictionary.

tests/test_lib_policy.py
plettich: Since ordering is part of the list, we can not (safely) compare lists
generated from dictionaries (since their order can not be guaranteed).
We should check If we compare lists like that somewhere in the code.
Unfortunately lists ans sets arent't the same since lists can contain
multiple occurrences of the same value.

privacyidea/lib/token.py
plettich: No nitpicking, will fix.

tests/test_api_lib_policy.py
cornelinux: :+1:<a href='#crh-update-none'></a>

tests/test_lib_policy.py
cornelinux: :+1:<a href='#crh-update-complete'></a>

privacyidea/lib/policy.py
plettich: So it doesn't really matter if some 'rights' occur multiple times in the list?
I meant the block in line 746:757 where we create the list of 'rights' from the current policies.

tests/test_lib_policy.py
plettich: Correct.
BTW: `{"disable", "enable"}` denotes a `set`-literal, not a dictionary.

privacyidea/lib/policy.py
cornelinux: It *should*TM not.
The rights are read in the [auth API](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/api/auth.py#L273) and [returned by the /auth endpoint](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/api/auth.py#L305).

The the UI uses the [checkRight](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/static/components/login/factories/auth.js#L54)-function to determine, if a user e.g. is allowed to set the Token PIN, so that we can display the "set PIN" button or not.

However, it would also probably not hurt to avoid duplicate entries.

privacyidea/api/policy.py
plettich: Done

privacyidea/lib/policy.py
plettich: changed the code so that we have only unique elements in the `rights` list at the end of the function.

privacyidea/lib/user.py
plettich: Fixed the loop, checks for `username` and `user` (in this order) are done after the loop.

privacyidea/lib/token.py
plettich: Done

privacyidea/lib/policy.py
cornelinux: :+1:<a href='#crh-update-none'></a>

privacyidea/api/policy.py
cornelinux: :+1:<a href='#crh-update-none'></a>

tests/test_lib_policy.py
cornelinux: :+1:<a href='#crh-update-complete'></a>

privacyidea/lib/user.py
```diff
@@ -597,7 +597,22 @@ def get_user_list(param=None, user=None):
...
+    # as delete does not work
+    for key in param:
+        lval = param[key]
+        if key == "realm":
+            continue
+        if key == "resolver":
+            continue
+        if key == "user" or key == "username":
+            # If "user" or "username" in param we skip to handle the user afterwards
+            continue
```
cornelinux: Sorry for nitpicking and probably my original code was even worse.
But should be do:

~~~~python
for key in param:
    lval = param[key]
    if key in ["realm", "resolver", "user", "username"]:
        continue
    searchDict[key] = lval
    log.debug(...)
~~~~

I was not aware of this earlier.

privacyidea/lib/token.py
fredreichbier: :+1:

privacyidea/lib/user.py
cornelinux: :+1:<a href='#crh-update-none'></a>

privacyidea/lib/user.py
```diff
@@ -605,13 +605,20 @@ def get_user_list(param=None, user=None):
...
-            # If "user" is in the param we overwrite the username
-            key = "username"
-
+        if key == "user" or key == "username":
+            # If "user" or "username" in param we skip to handle the user afterwards
+            continue
         searchDict[key] = lval
         log.debug("Parameter key:{0!r}={1!r}".format(key, lval))
 
+    # update searchdict depending on existence of 'user' or 'username' in param
```
cornelinux: :+1:<a href='#crh-update-complete'></a>

cornelinux:
:shipit:

cornelinux:
@fredreichbier This looks good to me. My comments have been covered. If you have nother else we could merge this!

PR #1315, opened 2018-11-21
Remove audit based stats

### LINKED ISSUE #1314: Remove audit based statistics
Since we provide new possibilities with periodic tasks and the event counter to have great statistics, we should remove the simple statistics that were done using audit log, matplotlib and pandas.
They have not been used anyways.
The pandas package blew up the UCS packages to over 80 MBs.

cornelinux:
Completely remove the statistics that are audit based
and are using matplotlib and pandas.

Closes #1314


cornelinux:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/privacyidea/privacyidea/pull/1315'>Refresh</a>.<a href='#crh-update-none'></a>

privacyidea/static/components/audit/controllers/auditControllers.js
```diff
@@ -128,10 +116,4 @@ myApp.controller("auditController", function (AuditFactory, $scope,
         return true;
     };
```
fredreichbier: I'm not really sure, but I think ``openDate`` can be removed too because it is unused now?

cornelinux: I think you are right. Removed it in d2c62ab

PR #1351, opened 2018-12-20
Add check for client_component

### LINKED ISSUE #1347: Bug in Two Step Enrollment Key
Upon Token Enrollment, once the user scans the QR Code,  privacyIDEA Authenticator generates the key.
privaycyIDEA server is supposed to validate the entire key as a whole.

But there is bug, wherein if the key generated in Authenticator is 12 character long, privacyIDEA accepts any character in the end. 

For example if authenticator generates the key as AAA-BBB-CCC-DDD, but if user enters the key in privacyIDEA window as AAA-BBB-CCC-DD**E** its accepted and OTP works.

cornelinux:
Closes #1347


cornelinux:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/privacyidea/privacyidea/pull/1351'>Refresh</a>.<a href='#crh-update-none'></a>

PR #1368, opened 2019-01-10
1288/more users per token

cornelinux:


migrations/versions/48ee74b8a7c8_.py
```diff
@@ -0,0 +1,141 @@
...
+
+            to = TokenOwner(token_id=token.id, user_id=token.user_id,
+                            resolver=token.resolver, resolver_type=token.resolver_type,
+                            realm_id=realm_id)
+            session.add(to)
+        session.commit()
+
+    except Exception as exx:
+        print("Failed to migrate token assignment data!")
+        print (exx)
```
fredreichbier: This should probably also drop the unused columns from ``Token``, if the data has been migrated successfully.

privacyidea/lib/tokenclass.py
```diff
@@ -190,9 +190,9 @@ def set_user(self, user, report=None):
         :return: None
         """
         (uid, resolvertype, resolvername) = user.get_user_identifiers()
-        self.token.resolver = resolvername
-        self.token.resolver_type = resolvertype
-        self.token.user_id = uid
+        r = TokenOwner(token_id=self.token.id,
+                       user_id=uid, resolver=resolvername,
+                       resolver_type=resolvertype, realmname=user.realm).save()
```
fredreichbier: Hm, what happens if the token already has a user? In the old behavior, the user was just overwritten. I think here, the token would be assigned a second user?

privacyidea/models.py
```diff
@@ -204,25 +205,27 @@ def __init__(self, serial, tokentype=u"",
...
             tr = TokenRealm(realm_id=realm_id, token_id=token_id)
             if tr:
                 db.session.add(tr)
+
+            # get type of resolver
+            res_type = Resolver.query.filter_by(name=resolver).first().rtype
+            to = TokenOwner(token_id=token_id, user_id=userid, resolver=resolver, resolver_type=res_type,
+                            realm_id=realm_id)
+            if to:
+                db.session.add(to)
```
fredreichbier: Could it ever be the case that ``to`` is false-y?

privacyidea/models.py
```diff
@@ -482,14 +485,13 @@ def get(self, key=None, fallback=None, save=False):
...
-        else:
-            return fallback
-    
+        td = self.get_vars(save=save)
+        return td.get(key, fallback)
+
     @log_with(log)
     def get_vars(self, save=False):
         log.debug('get_vars()')
+        tokenowner = self.owners.first()
```
fredreichbier: I think ``token.owners.first()`` is used at so many places that it would be warranted to add a property to ``Token`` named ``first_owner``, which could be used instead. What do you think?

privacyidea/models.py
```diff
@@ -1050,9 +1052,69 @@ def __init__(self, resolver_id=None, realm_id=None,
...
+        :param resolver_type: The type of the resolver
+        :param realm_id: The database ID of the realm
+        :param realmname: The alternate name of realm
+        """
+        if realm_id:
+            self.realm_id = realm_id
+        elif realmname:
+            r = Realm.query.filter_by(name=realmname).first()
+            self.realm_id = r.id
+        if token_id:
```
fredreichbier: Are there databases for which the ID of the first generated token would be ``0``? If yes, we should change the default argument value to ``None`` and check ``token_id is not None`` here.

cornelinux:
Sounds fair to me. The *association* alredy seems to exist since version 1.0.x.
I could create  a 2nd pull request with the assosiation and we could check how they compare in regards to code and performance...?

privacyidea/models.py
cornelinux: I am not sure. It might be if some constraints of the tokenowner table are not met. (doublicate entry).
At a first glance it looks like you are problably right.
For certain reasons we use this way of code at several places - especially since ``db.session.add(None)`` would raise an exception.

privacyidea/models.py
cornelinux: You are right.

Will change this.

fredreichbier:
Just for the record, I don't think we can use associations in this case because users are not organized in a table -- so nothing to do in this regard :-)

Regarding performance costs: I did some very quick benchmarks, 1000 authentication requests for varying users in a LDAP resolver. The ``TokenOwner`` table didn't seem to have any noticeable negative impact on performance. On the contrary, privacyIDEA was actually able to handle more requests per second, which is kind of weird. Maybe it's due to my not-so-scientific experimental setup.

```
$ cat without-tokenowner.txt 
Transactions:		        1000 hits
Availability:		      100.00 %
Elapsed time:		      781.34 secs
Data transferred:	        0.88 MB
Response time:		        0.78 secs
Transaction rate:	        1.28 trans/sec
Throughput:		        0.00 MB/sec
Concurrency:		        1.00
Successful transactions:        1000
Failed transactions:	           0
Longest transaction:	        1.93
Shortest transaction:	        0.20
$ cat with-tokenowner.txt 
Transactions:		        1000 hits
Availability:		      100.00 %
Elapsed time:		      738.95 secs
Data transferred:	        0.86 MB
Response time:		        0.74 secs
Transaction rate:	        1.35 trans/sec
Throughput:		        0.00 MB/sec
Concurrency:		        1.00
Successful transactions:        1000
Failed transactions:	           0
Longest transaction:	        1.92
Shortest transaction:	        0.06
```

privacyidea/models.py
```diff
@@ -1050,9 +1052,69 @@ def __init__(self, resolver_id=None, realm_id=None,
...
+class TokenOwner(MethodsMixin, db.Model):
+    """
+    This tables stores the owner of a token.
+    A token can be assigned to several users.
+    """
+    __tablename__ = 'tokenowner'
+    id = db.Column(db.Integer(), Sequence("tokenowner_seq"), primary_key=True,
+                   nullable=True)
+    token_id = db.Column(db.Integer(), db.ForeignKey('token.id'))
+    token = db.relationship('Token', lazy='joined', backref='token_list')
```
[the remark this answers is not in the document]
cornelinux: I understand, will look into this and remove it.

privacyidea/models.py
```diff
@@ -185,7 +185,8 @@ class Token(MethodsMixin, db.Model):
     info = db.relationship('TokenInfo',
                            lazy='dynamic',
                            backref='info')
-        
+    owners = db.relationship('TokenOwner', lazy='dynamic', backref='owners')
```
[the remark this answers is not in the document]
cornelinux: ok, will to and adapt.

privacyidea/models.py
```diff
@@ -1050,9 +1052,69 @@ def __init__(self, resolver_id=None, realm_id=None,
...
+    This tables stores the owner of a token.
+    A token can be assigned to several users.
+    """
+    __tablename__ = 'tokenowner'
+    id = db.Column(db.Integer(), Sequence("tokenowner_seq"), primary_key=True,
+                   nullable=True)
+    token_id = db.Column(db.Integer(), db.ForeignKey('token.id'))
+    token = db.relationship('Token', lazy='joined', backref='token_list')
+    resolver = db.Column(db.Unicode(120), default=u'', index=True)
+    resolver_type = db.Column(db.Unicode(120), default=u'')
```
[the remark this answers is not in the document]
cornelinux: I will look into this. 
Well, it will not break anything ;-)

migrations/versions/48ee74b8a7c8_.py
cornelinux: OK, I addressed this thing in 72f79b2e

privacyidea/models.py
cornelinux: Done in 65cf90e8

privacyidea/models.py
cornelinux: done in 0d1b013f

privacyidea/models.py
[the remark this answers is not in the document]
cornelinux: backrefs naming fixed in b82765e9

privacyidea/models.py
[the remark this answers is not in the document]
cornelinux: backrefs naming fixed in b82765e9

privacyidea/models.py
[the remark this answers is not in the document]
cornelinux: I checked this and I was able to remove the resolver_type.
It was never read from the database...

Done in 0cbffc5a

privacyidea/models.py
```diff
@@ -1050,9 +1052,69 @@ def __init__(self, resolver_id=None, realm_id=None,
...
+    user_id = db.Column(db.Unicode(320), default=u'', index=True)
+    realm_id = db.Column(db.Integer(), db.ForeignKey('realm.id'))
+    realm = db.relationship('Realm', lazy='joined', backref='realm_list')
```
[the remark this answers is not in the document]
cornelinux: I renamted this to "tokenowner". Not quite sure, in which commit.

Well, the only question remains, what we really should do with the *laziness*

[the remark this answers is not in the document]
cornelinux: I think Lazy loading (lazy='select' https://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html) is good if
1. you usually do not need the relation or need it later or
2. if you related object is very big.

The lazy=select issues a 2nd select statement for the relation. It many cases it is probably better to jave *one* joines select statement than 2 select statements.

So we need to look at, where we need the relationship immediately or where we need the relationship later - we have no *big* object (So this reason for lazy loading does not count).

# class Token

*info* and *owner*. I think we *mostly* need the additional tokeninfo and we *always* need the owner of the token, if we are handling with tokens.
So my guess would be to change these two
https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L180
https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L184
actually to 'joined'.

# TokenInfo

oups, we have a double realtion here, 
https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L642
although we only have an 1:m relation (one token, many attributes)
But also in this case, I would prefer a 'joined' laziness.
I will check, if we can drop the info in the Token.

# Resolver / ResolverConfig

https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L921
https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L960token/privacyidea/models.py#L921

Here we also have a probably unnessary double relation.
But this needs to be joined, since a resolver is never used without its config.

# ResolverRealm

is used for the n:m relation to bind any possible resolver to any possible realm.
https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L1019
https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L1024

To my understanding it is always neccessary to do a joined laziness.

# TokenOwner

I am not sure, if we always need the realm, when we want to find, which user belongs to a token.
https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L1068

Usually we get the user and want to find his tokens, so we should take a look at the ``get_tokens(user)``.
But from a data size perspective, this join would not hurt, since the realm is only a name.

# TokenRealm

https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L1126
https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L1130

This connects any token to any realm.
I am not sure, if this needs to be joined. At a first glance it looks like this information is not necessariely needed durch authentication. The realm would already be available via the tokenowner.

# Eventhandler

https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L1656
https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L1660
https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L1770
https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L1818

Eventhandler depend on the configuration and conditions.
So we should fetch all information from the start in a joined laziness.

# Periodic Tasks

https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L2615
https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L2618

You say ;-)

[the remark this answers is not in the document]
cornelinux: We should define a nice test scenario for investigating this.
But we might also put the question of laziness into another issue and PR.

privacyidea/lib/token.py
```diff
@@ -191,23 +191,29 @@ def _create_token_query(tokentype=None, realm=None, assigned=None, user=None,
                                           TokenRealm.token_id ==
                                           Token.id)).distinct()
 
-    if resolver is not None and resolver.strip("*"):
+    stripped_resolver = None if resolver is None else resolver.strip("*")
+    stripped_userid = None if userid is None else userid.strip("*")
+    if stripped_userid or stripped_resolver:
+        # Join the search with the token owner
+        sql_query = sql_query.filter(TokenOwner.token_id == Token.id)
```
fredreichbier: This will result in duplicate comparisons in the WHERE clauses if we, for example, pass a ``resolver`` *and* ``assigned=True``:
```python
>>> get_tokens(resolver='passwd', assigned=True)
```
emits
```sql
SELECT ...
FROM token, tokenowner 
WHERE token.id = tokenowner.token_id AND tokenowner.token_id = token.id AND tokenowner.resolver = ?
```
... but even though it looks weird, it works, so I guess this is OK for now.

privacyidea/lib/token.py
```diff
@@ -732,15 +740,14 @@ def get_all_token_users():
...
-        if tokenobject.token.user_id and len(user_info) == 0:
-            user_info['username'] = u'/:no user info:/'
+            if len(user_info) == 0:
+                user_info['username'] = u'/:no user info:/'
 
-        if user_info:
-            tokens[tokenobject.token.serial] = user_info
+        tokens[tokenobject.token.serial] = user_info
 
     return tokens
```
fredreichbier: I was curious because I didn't encounter ``get_all_tokens_users`` before :-) It seems it is only used in the tests. Likewise, ``get_user_info`` is only used by ``get_all_tokens_users`` and in the tests. What do you think about removing them?

privacyidea/lib/token.py
```diff
@@ -1771,15 +1779,15 @@ def copy_token_user(serial_from, serial_to):
...
     tokenobject_to = get_one_token(serial=serial_to)
-    user_id = tokenobject_from.token.user_id
-    resolver = tokenobject_from.token.resolver
-    resolver_type = tokenobject_from.token.resolver_type
-    tokenobject_to.set_user_identifiers(user_id, resolver,
-                                                resolver_type)
+    TokenOwner(token_id=tokenobject_to.token.id,
+               user_id=tokenobject_from.token.first_owner.user_id,
+               realm_id=tokenobject_from.token.first_owner.realm_id,
+               resolver=tokenobject_from.token.first_owner.resolver).save()
```
fredreichbier: If I understand correctly, if ``tokenobject_to`` had an owner before the copy operation, it will have two owners after the copy operation. I guess we should first unassign ``tokenobject_to`` to rule out that case?

privacyidea/lib/tokenclass.py
```diff
@@ -190,9 +190,9 @@ def set_user(self, user, report=None):
         :return: None
         """
         (uid, resolvertype, resolvername) = user.get_user_identifiers()
-        self.token.resolver = resolvername
-        self.token.resolver_type = resolvertype
-        self.token.user_id = uid
+        r = TokenOwner(token_id=self.token.id,
```
fredreichbier: I'm not so happy about this function because we change its meaning:
Before: ``set_user`` causes ``user`` to be the only owner of the token.
After: ``set_user`` causes ``user`` to be added as an owner of the token.

Even though we have the implicit precondition that we only call ``set_user`` if the token has no user, I would prefer to have one of the following safeguards here:
* We could rename the function to ``add_owner`` to make its meaning clear.
* We could assert that the token is unassigned and throw an error otherwise.
* We could unassign all current owners of the token.

What do you think?

privacyidea/lib/tokenclass.py
```diff
@@ -205,30 +205,25 @@ def user(self):
...
-            username = get_username(self.token.user_id, self.token.resolver)
-            rlist = self.token.realm_list
-            # FIXME: What if the token has more than one realm assigned?
-            if len(rlist) == 1:
-                realmname = rlist[0].realm.name
-            if username and realmname:
-                user_object = User(login=username,
-                                   resolver=self.token.resolver,
-                                   realm=realmname)
+        user_object = User()
```
fredreichbier: Hm, we change the return value of this function in case the token has no owner. Before, we returned ``None``, now we return the empty user object. Is this intentional?

privacyidea/models.py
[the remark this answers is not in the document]
cornelinux: THere are a lot of comment/conversations. Could you please resolver the conversation? Thanks.

privacyidea/models.py
```diff
@@ -204,25 +201,29 @@ def __init__(self, serial, tokentype=u"",
...
+            # We can not create the tokenrealm-connection and owner-connection, yet
             # since we need to token_id.
             token_id = self.save()
             realm_id = Realm.query.filter_by(name=realm).first().id
             tr = TokenRealm(realm_id=realm_id, token_id=token_id)
             if tr:
                 db.session.add(tr)
+
+            # get type of resolver
+            res_type = Resolver.query.filter_by(name=resolver).first().rtype
```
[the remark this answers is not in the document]
cornelinux: Will do

privacyidea/models.py
```diff
@@ -204,25 +201,29 @@ def __init__(self, serial, tokentype=u"",
...
+            to = TokenOwner(token_id=token_id, user_id=userid, resolver=resolver, realm_id=realm_id)
+            if to:
+                db.session.add(to)
+
+            if tr or to:
                 db.session.commit()
+
+    @property
+    def first_owner(self):
+        return self.owners.first()
```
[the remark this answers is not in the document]
cornelinux: No, I don't think so.
We would have to revert this.
After the migration there is no token that can have two owners. I want to add the possibility for multiple owners very quickly, like 3.1.

migrations/versions/48ee74b8a7c8_.py
cornelinux: What happend to this f* PR?
I can not comment on your comment. (now I can) :-(
Well, a file would be nice, but on the other hand, the admin created a backup, right?
To which file should this be dumped?

privacyidea/models.py
```diff
@@ -1050,9 +1052,66 @@ def __init__(self, resolver_id=None, realm_id=None,
...
+    """
+    __tablename__ = 'tokenowner'
+    id = db.Column(db.Integer(), Sequence("tokenowner_seq"), primary_key=True,
+                   nullable=True)
+    token_id = db.Column(db.Integer(), db.ForeignKey('token.id'))
+    resolver = db.Column(db.Unicode(120), default=u'', index=True)
+    user_id = db.Column(db.Unicode(320), default=u'', index=True)
+    realm_id = db.Column(db.Integer(), db.ForeignKey('realm.id'))
+    # This creates an attribute "tokenowner" in the realm objects
+    realm = db.relationship('Realm', lazy='joined', backref='tokenowner')
```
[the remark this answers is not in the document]
cornelinux: will do 2019-02-14 ;-)

privacyidea/models.py
```diff
@@ -1050,9 +1052,66 @@ def __init__(self, resolver_id=None, realm_id=None,
...
 
 
+class TokenOwner(MethodsMixin, db.Model):
+    """
+    This tables stores the owner of a token.
+    A token can be assigned to several users.
+    """
+    __tablename__ = 'tokenowner'
+    id = db.Column(db.Integer(), Sequence("tokenowner_seq"), primary_key=True,
+                   nullable=True)
```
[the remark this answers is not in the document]
cornelinux: will do 2019-02-14 ;-)

privacyidea/lib/token.py
```diff
@@ -177,9 +177,9 @@ def _create_token_query(tokentype=None, realm=None, assigned=None, user=None,
     if assigned is not None:
         # filter if assigned or not
         if assigned is False:
-            sql_query = sql_query.filter(Token.user_id == "")
+            sql_query = sql_query.filter(Token.owners == None)
```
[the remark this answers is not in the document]
cornelinux: I have not seen ``~Token.owners.any()`` before.

privacyidea/lib/tokenclass.py
cornelinux: I think yes. Because it allows calling functions to handle things better, since they always get a user object.

privacyidea/lib/token.py
cornelinux: This was intended also for subscription purposes or statistics purposes, like "how many tokens with users are there"?

It could be that it was used and the use case was changed a bit.
It could however be, that it could be used again...?

So I am a bit hesitant to remove this. (Although it usually makes no sense to have unused code)

privacyidea/lib/token.py
cornelinux: I think this would be half of  the intented behaviour.
I would actually expect to copy *all* users of the old token to the new token and then the new token have the old users and the new users.

Currently this function is only used in *losttoken*, where the destination token does not have any users anyways.

The destination token loosing all the existing users would not be the expected behaviour. After all it is ``copy_token_users`` and not ``replace_token_users``.

privacyidea/lib/tokenclass.py
cornelinux: Again from an expectation level of the naming of the method I agree.

In the long run I think it makes sense to have a function ``add_user`` and ``set_user``.
Guess, what each function would be doing! ;-)
So I will rename this method to ``add_user``.

Will do 2019-02-14

privacyidea/lib/token.py
fredreichbier: I think the copying behavior instead of replacing makes total sense, once we allow a token to have multiple owners. :-)
Right now, however, ``copy_token_users`` it is also used in the ``/token/copyuser`` endpoint. With this PR, if I do
```
$ http POST http://localhost:5000/token/copyuser from=OATH0001E293 to=OATH0003E7E7 PI-Authorization:$TOKEN
```
the token ``OATH0003E7E7`` actually has two owners:
```python
>>> get_one_token(serial="OATH0003E7E7").token.owners.all()
[<TokenOwner 5>, <TokenOwner 6>]
```
so I think it makes more sense to have ``copy_token_users`` *replace* the owner of the destination token.

privacyidea/lib/subscriptions.py
```diff
@@ -82,10 +82,9 @@ def get_users_with_active_tokens():
     :return: Number of users
     :rtype: int
     """
-    from privacyidea.models import Token
-    sql_query = Token.query.with_entities(Token.resolver,
-                                          Token.user_id).filter(Token.active == True,
-                                                                Token.user_id != "").distinct()
+    from privacyidea.models import Token, TokenOwner
+    sql_query = TokenOwner.query.with_entities(TokenOwner.resolver, TokenOwner.user_id)
+    sql_query = sql_query.filter(Token.active == True).distinct()
```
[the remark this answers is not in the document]
cornelinux: Will do 2019-02-14

   add .filter(Token.id == TokenOwner.token_id)

privacyidea/lib/token.py
cornelinux: In subscriptions.py we have get_users_with_active_tokens.

I will remove this! (both)

privacyidea/lib/token.py
cornelinux: OK, will do. (Since /token/assign) also only allows to assign a token without a user.
The old user will be removed (although usually there will never be an old user;)
(Dont forget to add a TODO/FIXME for Version 3.1)

Will do  2019-02-14

privacyidea/lib/tokenclass.py
cornelinux: Will do 2019-02-14

privacyidea/models.py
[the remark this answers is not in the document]
cornelinux: done in d7dabaae

privacyidea/models.py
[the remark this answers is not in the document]
cornelinux: done in d7dabaae

privacyidea/models.py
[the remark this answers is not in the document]
cornelinux: done in d7dabaae

privacyidea/lib/subscriptions.py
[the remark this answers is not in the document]
cornelinux: done in d7dabaae

privacyidea/lib/token.py
cornelinux: done in 2ae64abf

privacyidea/lib/tokenclass.py
cornelinux: done in 49f1ac75

privacyidea/lib/tokenclass.py
cornelinux: done in 49f1ac75

privacyidea/lib/token.py
cornelinux: done in 49f1ac75

migrations/versions/48ee74b8a7c8_.py
```diff
@@ -0,0 +1,154 @@
...
+        op.create_table('tokenowner',
+        sa.Column('id', sa.Integer()),
+        sa.Column('token_id', sa.Integer(), nullable=True),
+        sa.Column('resolver', sa.Unicode(length=120), nullable=True),
+        sa.Column('user_id', sa.Unicode(length=320), nullable=True),
+        sa.Column('realm_id', sa.Integer(), nullable=True),
+        sa.ForeignKeyConstraint(['realm_id'], ['realm.id'], ),
+        sa.ForeignKeyConstraint(['token_id'], ['token.id'], ),
+        sa.PrimaryKeyConstraint('id')
+        )
```
fredreichbier: Sorry, one last thing: Should we create this table with ``mysql_row_format="DYNAMIC"``? Like for the periodic task tables?
https://github.com/privacyidea/privacyidea/blob/466d562db1167b7e4a8ed2dcf5ff420bf9af5d44/migrations/versions/2c9430cfc66b_.py#L30

plettich: if the total number of characters in a row exceeds 768 we need this setting.
As of [mysql 5.7.9](https://dev.mysql.com/doc/refman/5.6/en/glossary.html#glos_dynamic_row_format) and [mariadb 10.1.32/10.2.2](https://mariadb.com/kb/en/library/innodb-system-variables/#innodb_default_row_format) the `DYNAMIC` row format is the default.

fredreichbier: Thanks! Then I guess we don't really need to set it explicitly, because the table is fairly small.

PR #1374, opened 2019-01-14
Remove password hash calculation

### LINKED ISSUE #1372: remove password hash calculation and switch to passlib completely
Currently there are several password hashes supported in the [`SQLIdResolver`](https://github.com/privacyidea/privacyidea/blob/b0cd514c17bea43337a01c33b72da136c27d52be/privacyidea/lib/resolvers/SQLIdResolver.py) which are more or less generated/verified with homegrown code.
We can remove most of the stuff and switch to [`passlib`](https://pypi.org/project/passlib/) (which is already used in some cases).
Some hashes (drupal, SSHA512 and SSHA256) are currently not available in passlib but can be added.

plettich:
To simplify the code the password hash calculation (used in the
SQLIdResolver) is removed and replaced with the passlib module.
Unfortunately passlib does not provide all necessary password hashes, so
these are implemented as passlib handlers.

Fixes #1372


tests/test_lib_resolver.py
```diff
@@ -403,15 +460,15 @@ def test_07_add_user_update_delete_ssha512(self):
...
         # Add a new user
-        uid = y.add_user({"username":"hans",
+        uid = y.add_user({"username": "hans",
                           "email": "hans@world.net",
                           "password": u"foo",
                           "mobile": "12345"})
         self.assertTrue(y.checkPass(uid, u"foo"))
         self.assertFalse(y.checkPass(uid, u"bar"))
-        # check that we actually store SSHA512 now
+        # check that we actually store MD5CRYPT now
```
cornelinux: 256

PR #1415, opened 2019-02-01
Migrate test_lib_apps.py, HOTP and TOTP tests

fredreichbier:
This migrates ``test_lib_tokens_{hotp,totp}.py`` and ``test_lib_apps.py``.


tests/test_lib_tokens_totp.py
```diff
@@ -499,11 +499,14 @@ def test_19_pin_otp_functions(self):
         #self.assertTrue(res == 47251647, res)
         self.assertTrue(res == -1, res)
 
-        # simple get_otp of current time
-        r = token.get_otp()
-        self.assertTrue(r > 47251648, r)
-        r = token.get_otp(current_time=datetime.datetime.now())
-        self.assertTrue(r > 47251648, r)
+        # simple OTPs of current time
+        # TODO: Not sure if that's what should be tested here
```
fredreichbier: This is a bit tricky here. The assertion ``r > 47251648`` apparently always succeeded in Python 2, because ``r`` is a tuple:
```python
>>> (1, "pin", "otp", "pinotp") > 47251648
True
```
But under Python 3, tuples and ints are incomparable:
```python
>>> (1, "pin", "otp", "pinotp") > 47251648
  File "<stdin>", line 1, in <module>
TypeError: '>' not supported between instances of 'tuple' and 'int'
```
so the test above fails.

So I tried to find out what should be tested here: If I understand correctly, the test was intended to check that the TOTP counter value is bigger than 47251648. I tried to fix the test accordingly, but I'm not sure if it makes sense. @cornelinux you got any idea?

privacyidea/lib/apps.py
```diff
@@ -56,19 +57,25 @@
...
+        if isinstance(key, six.text_type):
             key = key.encode('utf-8')
-        if isinstance(value, unicode):
+        if isinstance(value, six.text_type):
             value = value.encode('utf-8')
-        extra_data_list.append('{key}={value}'.format(key=quote(str(key)),
-                                                      value=quote(str(value))))
+        # Thus, ``quote`` returns (ASCII) bytes on Python 2. On Python 3, it returns str.
+        extra_data_list.append('{key}={value}'.format(key=quote(key),
+                                                      value=quote(value)))
```
plettich: i was thinking about this section as well and thought about a helper function like `to_unicode_string(s)` which
 * does nothing if `s` already is a unicode string
 * converts `s` to a unicode string if it is a byte string
 * converts `s` to a (unicode) string if it is something else (int, bool, ...)
Similar to `convert_column_to_unicode()`
What do You think?

privacyidea/lib/apps.py
```diff
@@ -126,7 +133,7 @@ def create_google_authenticator_url(key=None, user=None,
 
     key_bin = binascii.unhexlify(key)
     # also strip the padding =, as it will get problems with the google app.
-    otpkey = base64.b32encode(key_bin).strip('=')
+    otpkey = base64.b32encode(key_bin).decode('utf-8').strip('=')
```
plettich: i would prefer something like `to_unicode(...)` instead of `.decode()`, i find it more 'literal'. What do You think?

cornelinux: hm, I think there are sill a lot of occurrences of "decode".
We should then also change this in the rest of the code.

tests/test_lib_tokens_totp.py
cornelinux: Yes, it is probably the OTP counter. I also stumbled upon this at another python3 PR I issued some days ago.

fredreichbier: OK thanks, then I'll just keep the updated test and remove the TODO comment.

privacyidea/lib/apps.py
fredreichbier: This section is really a bit hacky :-) Maybe a helper function would also improve readability. But I'm not sure if a ``to_unicode_string`` function would be helpful here: ``key`` and ``value`` actually need to be converted to byte strings, because ``quote()`` on Python 2 crashes if we pass a unicode string:
```python
>>> urllib2.quote(u'ö')
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/urllib.py", line 1298, in quote
    return ''.join(map(quoter, s))
KeyError: u'\xf6'
```

privacyidea/lib/apps.py
fredreichbier: Yeah, ``.decode('utf-8')`` isn't that nice. :-/ But as far as I can see, ``to_unicode`` is more of a Holzhammer function, because it does nothing if the input is already a unicode string, and decodes to a unicode string only if the input is a byte string. But in the case above, we *know* that ``b32encode`` returns a byte string, so we know that we always need to decode. But maybe that's nitpicking -- what do you think?

I was also thinking about creating more literal encoding functions that return unicode strings, similarly to the ``hexlify_and_unicode`` we already have:
```python
def b32encode_and_unicode(s):
    return b32encode(s).decode('utf-8')
```

plettich: You are right, if the output is known to be a byte string, a following `.decode()` is probably more straight forward.
I like the idea of a helper function like `b32encode_and_unicode()`, maybe also for `b64encode`.

privacyidea/lib/apps.py
plettich: I ran into this myself. That is why i added the [`to_byte_string()`](https://github.com/privacyidea/privacyidea/blob/911ac0da6baba09e5677cae6790cd8e66f027c84/privacyidea/lib/utils.py#L179) function.

privacyidea/lib/apps.py
```diff
@@ -43,32 +43,31 @@
 """
```
plettich: since You are already working on `lib/apps.py` could You also check if the `test_lib_apps.py` tests run? I think they should work out of the box now.

privacyidea/lib/apps.py
fredreichbier: ``_construct_extra_parameters`` now uses ``to_byte_string`` and it's much shorter and, I think, readable :-)

privacyidea/lib/apps.py
fredreichbier: I've added a ``b32encode_and_unicode()`` function.

Issue #676, opened 2017-04-10
privacyIDEA should be able to run on python 3


cornelinux:
In regards to dependent packages, we might look at this.

https://medium.com/@anthonypjshaw/oh-no-this-package-is-python-2-only-8e6316f9a02

fredreichbier:
Just read that the next RHEL major release [will deprecate Python 2](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/7.5_release_notes/chap-red_hat_enterprise_linux-7.5_release_notes-deprecated_functionality), also, Python 2.7 will go end-of-life in 2020. If we want to start looking into this: the above page links an [interesting porting guide](https://portingguide.readthedocs.io/en/latest/).

plettich:
I started a [Wiki page](https://github.com/privacyidea/privacyidea/wiki/python3)

plettich:
Further steps for making privacyIDEA compatible with Python 2 and Python 3
* [ ] migrate tests to Python 3
  * [ ] migrate database testcases
  * [ ] migrate crypo library testcases
  * [ ] migrate other lib-level testcases
  * [ ] migrate API testcases
* [ ] migrate tools and plugins
* [ ] refactor crypto interfaces

PR #1418, opened 2019-02-01
Make certificate tests run with python 3

plettich:
This also fixes the certificate token tests.

Working on #676


privacyidea/lib/caconnectors/localca.py
```diff
@@ -332,15 +332,14 @@ def _filename_from_x509(x509_name, file_extension="pem"):
...
-        :rtype: basestring
+        :rtype: str
         """
         name_components = x509_name.get_components()
-        filename = ""
-        for (key, value) in name_components:
-            filename += value+"_"
-        filename = filename[:-1] + "." + file_extension
-        return filename
+        filename = "_".join([value.decode('utf8') for (key, value) in name_components])
```
cornelinux: Here you are doing a ``decdode``. When should this be a ``to_unicode``?

privacyidea/lib/caconnectors/localca.py
```diff
@@ -402,15 +401,9 @@ def sign_request(self, csr, options=None):
                 csr_obj.get_subject(), file_extension="pem")
             #csr_extensions = csr_obj.get_extensions()
         csr_filename = csr_filename.replace(" ", "_")
-        if type(csr_filename) == str:
-            csr_filename = csr_filename.decode("utf-8")
-        csr_filename = csr_filename.encode("ascii", "ignore")
```
cornelinux: So why would this suddently be not necessary under python 2.7 anymore?

privacyidea/lib/caconnectors/localca.py
plettich: You are right, very inconsequential of me...

privacyidea/lib/caconnectors/localca.py
plettich: You are right, i misinterpreted this. Since we are using the updated path later on there should be no trouble. I'll fix this.

privacyidea/lib/caconnectors/localca.py
plettich: fixed in 1fb2839

privacyidea/lib/caconnectors/localca.py
plettich: fixed in 1fb2839

cornelinux: I think we also should use ``to_unicode`` instead of decode?

PR #1421, opened 2019-02-04
Changed favicon to png

### LINKED ISSUE #1420: Change favicon to png
https://github.com/privacyidea/privacyidea/pull/1361 added the favicon, but I think it would be better, if we use a png instead of an ico, because it can be transparent.

Mipronimo:
closes #1420


privacyidea/static/templates/header.html
```diff
@@ -6,7 +6,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta name="description" content="">
     <meta name="author" content="Cornelius Kölbel" >
-    <link rel="icon" href="{{ instance }}/static/favicon.ico">
```
plettich: Please delete the old favicon file.

Mipronimo: See https://github.com/privacyidea/privacyidea/commit/50c6c153f5a927299334c4337405b086644e0b6e

PR #1423, opened 2019-02-04
Update daplug token tests to work with Python 3

plettich:
Working on #676


privacyidea/lib/tokens/daplugtoken.py
```diff
@@ -55,29 +58,33 @@
...
+
+
 def _daplug2digit(daplug_otp):
     hex_otp = ""
     for i in daplug_otp:
         digit = MAPPING.get(i)
         hex_otp += digit
-    otp = binascii.unhexlify(hex_otp)
+    # we know the result is a string of digits
+    otp = to_unicode(binascii.unhexlify(hex_otp))
```
cornelinux: We have a function ``hexlify_and_unicode`` --- should we have a similar function ``unhexlify_and_unicode``?
Till now there seems only one appearance of such a construct in the code.

> grep -r unhexlify privacyidea/* | grep unicode

plettich: The problem is, that we can't be sure if it is a printable string that gets returned from unhexlify. It might as well be a cryptographic key in a bytes-representation.
So this would depend very much on the context.

cornelinux: Actually, I understand this ;-)

PR #1424, opened 2019-02-04
Make yubikey tests run with Python 3

plettich:
- create a `b64encode_and_unicode()` function to automatically convert the
  result of a `b64encode()` call to a unicode string
- update crc checksum calculation to work with Python 2 and 3

Working on #676

### DIFF
--- a/privacyidea/lib/utils.py
+++ b/privacyidea/lib/utils.py
@@ -286,9 +286,17 @@ def modhex_decode(m):
 
 
 def checksum(msg):
+    """
+    Calculate CRC-16 (16-bit ISO 13239 1st complement) checksum.
+    (see Yubikey-Manual - Chapter 6: Implementation details)
+
+    :param msg: input byte string for crc calculation
+    :type msg: bytes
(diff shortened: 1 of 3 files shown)


privacyidea/lib/utils.py
```diff
@@ -286,9 +286,17 @@ def modhex_decode(m):
...
 def checksum(msg):
+    """
+    Calculate CRC-16 (16-bit ISO 13239 1st complement) checksum.
+    (see Yubikey-Manual - Chapter 6: Implementation details)
+
+    :param msg: input byte string for crc calculation
+    :type msg: bytes
+    :return: crc16 checksum of msg
+    :rtype: int
+    """
```
cornelinux: I think it might be a good idea to add an additional test case in ``test_lib_utils.py``. 
Well, we should have had a test case for checksum earlier, but now it looks like we should! ;-)

plettich: I was going to update the `test_lib_utils.py` and the `lib/utils.py` anyway.

PR #1429, opened 2019-02-06
Python3 update passwd remote token tests

plettich:
Working on #676


privacyidea/lib/tokens/remotetoken.py
```diff
@@ -218,34 +218,24 @@ def check_otp(self, otpval, counter=None, window=None, options=None):
...
         """
         otp_count = -1
-        otpval = otpval.encode("utf-8")
 
         remoteServer = self.get_tokeninfo("remote.server") or ""
-        remoteServer = remoteServer.encode("utf-8")
 
         # in preparation of the ability to relocate privacyidea urls,
         # we introduce the remote url path
         remotePath = self.get_tokeninfo("remote.path") or ""
```
fredreichbier: With this change, ``removePath`` is not stripped anymore. I'm not sure it's necessary?

plettich: Oh, totally missed that. Will ad a test for remote.path as well.

cornelinux: The background is, that the remote privacyIDEA server could run e.g. at https://remote.example.com/pi2 while the originial privacyIDEA server would run at https://local.exampl.com/

plettich: Fixed in 6141b198

fredreichbier: :+1:

PR #1438, opened 2019-02-08
Update to make radius token tests work with Python 3

plettich:


privacyidea/lib/tokens/radiustoken.py
```diff
@@ -197,25 +196,27 @@ def check_otp(self, otpval, counter=None, window=None, options=None):
...
             radius_secret = get_from_config("radius.secret")
-            # Is returned as unicode, so we convert it to utf-8
-            radius_secret = radius_secret.encode("utf-8")
         else:
             # individual token settings
             radius_server = self.get_tokeninfo("radius.server")
             # Read the secret
             secret = self.token.get_otpkey()
+            # TODO: can we assume that this is a printable string? Otherwise the
+            #       to_unicode() in the debug log below might fail.
```
fredreichbier: I think strictly speaking, a RADIUS secret is a just a sequence of bytes, so it may not be valid UTF-8, which means ``to_unicode()`` may fail. But the question is whether our API even allows to set a byte string as the RADIUS secret. If the API only accepts Unicode strings (which are then encoded to UTF-8), we could also make this restriction explicit in the documentation.

plettich: That was my last Information: "Who would use bytes as a RADIUS shared secret..."
I couldn't find any information on the type of the shared secret, but since it is written in the config-file, it should contain only printable characters (or not?).

fredreichbier: I found the following in the default FreeRADIUS clients.conf:
```
        #  The secret can be any string, up to 8k characters in length.
        #
        #  Control codes can be entered vi octal encoding,
        #       e.g. "\101\102" == "AB"
        #  Quotation marks can be entered by escaping them,
        #       e.g. "foo\"bar"
        #
        #  A note on security:  The security of the RADIUS protocol
        #  depends COMPLETELY on this secret!  We recommend using a
        #  shared secret that is composed of:
        #
        #       upper case letters
        #       lower case letters
        #       numbers
```
so technically one *could* set some random bytes for the shared secret by using the octal escape . But I think it's sensible to only allow UTF-8 bytestrings for secrets, so I think internally treating RADIUS secrets as unicode strings is okay.

PR #1440, opened 2019-02-08
Update motp test case to run with Python 3

plettich:
Working on #676

### DIFF
--- a/privacyidea/lib/tokens/mOTP.py
+++ b/privacyidea/lib/tokens/mOTP.py
@@ -36,12 +36,12 @@ import time
 
 from hashlib import md5
 
+from privacyidea.lib.utils import to_unicode, to_bytes
 from privacyidea.lib.crypto import zerome
 from privacyidea.lib.log import log_with
 
 
 log = logging.getLogger(__name__)
-
 
 
 class mTimeOtp(object):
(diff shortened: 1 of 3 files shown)


privacyidea/lib/tokens/mOTP.py
```diff
@@ -97,18 +98,16 @@ def checkOtp(self, anOtpVal, window=10, options=None):
...
             pin = self.pin
         else:
             key = self.secretObject.getKey()
             pin = self.secPin.getKey()
 
-
         for i in range(otime - window, otime + window):
-            otp = unicode(self.calcOtp(i, key, pin))
-            if unicode(anOtpVal) == otp:
+            otp = self.calcOtp(i, to_unicode(key), to_unicode(pin))
```
fredreichbier: Hm, a question, just to be sure: Do I understand correctly that ``key`` is always a hex-string in the case of mOTP, which means it's valid UTF-8?

plettich: It's a little tricky. If the key is generated (which it usually is if i read the code in [`update()`](https://github.com/privacyidea/privacyidea/blob/42cf5d841465a3aaafc9750b25fd3b71f551b488/privacyidea/lib/tokens/motptoken.py#L152) correctly) it is a hex-string.
I don't know how the `otpkey` would be set otherwise.
But according to http://motp.sourceforge.net/ the keys are hex-strings.

fredreichbier: Sounds good! I just tested this with DroidOTP and 
* generating a hex-string secret on the phone and entering it manually in privacyIDEA during enrollment
* generating a hex-string secret in privacyIDEA and entering it manually on the phone
both seems to work fine :-)

fredreichbier: :+1:

PR #1441, opened 2019-02-08
Python3 lib importotp

cornelinux:


tests/test_lib_importotp.py
```diff
@@ -637,6 +638,6 @@ def test_00_gpg_decrypt(self):
         self.assertTrue("2F25BAF8645350BB" in pubkeys)
 
         r = GPG.decrypt(str(HALLO_PAYLOAD))
```
plettich: I think we can safely remove the `str()` here, the payload is already a string.

privacyidea/lib/importotp.py
```diff
@@ -615,6 +616,7 @@ def export_pskc(tokenobj_list, psk=None):
         issuer = "privacyIDEA"
         try:
             manufacturer = tokenobj.token.description.encode("ascii")
+            manufacturer = to_unicode(manufacturer)
```
fredreichbier: We could also use the ``ignore`` option to just delete non-ASCII characters:
```
>>> 'Hællöhe'.encode('ascii', 'ignore')
b'Hllhe'
```
What do you think?

privacyidea/lib/importotp.py
```diff
@@ -630,7 +632,9 @@ def export_pskc(tokenobj_list, psk=None):
         otpkey = tokenobj.token.get_otpkey().getKey()
         try:
             if tokenobj.type.lower() in ["totp", "hotp"]:
-                encrypted_otpkey = aes_encrypt_b64(psk, binascii.unhexlify(otpkey))
+                encrypted_otpkey = to_unicode(aes_encrypt_b64(psk, binascii.unhexlify(otpkey)))
+            elif tokenobj.type.lower() in ["pw"]:
+                encrypted_otpkey = to_unicode(aes_encrypt_b64(psk, otpkey))
             else:
                 encrypted_otpkey = aes_encrypt_b64(psk, otpkey)
```
fredreichbier: Wouldn't it be sensible to also apply ``to_unicode`` here? Otherwise, I think we'll end up with string representations of base64 bytestrings in the XML, like ``<xenc:CipherValue>b'ABAB...'</xenc:CipherValue>``

privacyidea/lib/importotp.py
```diff
@@ -683,5 +687,8 @@ def export_pskc(tokenobj_list, psk=None):
             number_of_exported_tokens += 1
         except Exception as e:
             log.warning(u"Failed to export the token {0!s}: {1!s}".format(serial, e))
+            tb = traceback.format_exc()
+            log.debug(tb)
+            raise(e)
```
fredreichbier: With this change, PSKC export fails if the export fails for any token. Just want to make sure that's intended? :-)

privacyidea/lib/importotp.py
```diff
@@ -500,7 +499,9 @@ def parsePSKCdata(xml_data,
                 enc_data = enc_data.strip()
                 secret = aes_decrypt_b64(binascii.unhexlify(preshared_key_hex), enc_data)
                 if token["type"].lower() in ["hotp", "totp"]:
-                    token["otpkey"] = binascii.hexlify(secret)
+                    token["otpkey"] = hexlify_and_unicode(secret)
+                elif token["type"].lower() in ["pw"]:
+                    token["otpkey"] = to_unicode(secret)
                 else:
                     token["otpkey"] = secret
```
fredreichbier: Hm, I think ``secret`` is a bytestring in this case. I'm not sure if this is supported by the underlying ``import_token`` API?

... but on a second thought, can we even import any token types besides HOTP, TOTP and PW?

privacyidea/lib/importotp.py
cornelinux: No. Will do.

privacyidea/lib/importotp.py
cornelinux: We can import Spass tokens. Which have no otpkey anyways.

privacyidea/lib/importotp.py
cornelinux: WIll check.

privacyidea/lib/importotp.py
cornelinux: will check.

tests/test_lib_importotp.py
cornelinux: WIll do.

privacyidea/lib/importotp.py
cornelinux: No, the decryption also assumes, that this would be a bytestring.

privacyidea/lib/importotp.py
cornelinux: RIght, it was for testing purposes and remained.

tests/test_lib_importotp.py
cornelinux: done in d367ea32

privacyidea/lib/importotp.py
cornelinux: done in d367ea32

privacyidea/lib/importotp.py
cornelinux: done in d367ea32

privacyidea/lib/importotp.py
cornelinux: done in d367ea32

privacyidea/lib/importotp.py
cornelinux: done in d367ea32

PR #1443, opened 2019-02-11
Fix RADIUS token and passthru policy for non-ASCII usernames

### LINKED ISSUE #1442: RADIUS token and passthru policy fail with non-ASCII usernames
If we assign a RADIUS token to a user whose username contains non-ASCII characters, the user cannot authenticate. The debug log shows the following error:
```
[2019-02-08 16:41:22,926][7670][139677530805632][ERROR][privacyidea.lib.tokens.radiustoken:281] Error contacting radius Server: UnicodeEncodeError('ascii', u'k\xf6lbel', 1, 2, 'ordinal not in range(128)')
```
The problem is that we encode the username to ASCII before passing it to pyrad here:
https://github.com/privacyidea/privacyidea/blob/42cf5d841465a3aaafc9750b25fd3b71f551b488/privacyidea/lib/tokens/radiustoken.py#L249

I guess it would be sensible to just encode the username to UTF-8.

fredreichbier:
Closes #1442

### DIFF
--- a/tests/test_lib_radiusserver.py
+++ b/tests/test_lib_radiusserver.py
@@ -1,3 +1,4 @@
+# coding: utf-8
 """
 This test file tests the lib/radiusserver.py
 """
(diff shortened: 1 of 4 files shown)


tests/test_lib_radiusserver.py
```diff
@@ -99,3 +100,13 @@ def test_06_test_radius(self):
                           identifier="myserver", server="1.2.3.4",
                           user="user", password="password",
                           secret="x" * 96, dictionary=DICT_FILE)
+
```
plettich: should we also test for a non-ascii RADIUS-user/password?

PR #1446, opened 2019-02-13
Update tiqr/ocra token tests to work with python 3

plettich:
Working on #676


plettich:
We should add more tests for the tiqr/ocra tokens...

privacyidea/lib/tokens/ocra.py
```diff
@@ -260,14 +261,14 @@ def create_data_input(self, question, pin=None, pin_hash=None,
             question = '{0:x}'.format(int(question))
             question += '0' * (len(question) % 2)
             question = binascii.unhexlify(question)
-            question += '\0' * (128-len(question))
+            question += b'\0' * (128-len(question))
             data_input += question
         elif self.ocrasuite_obj.challenge_type == "QA":
             question += '\0' * (128-len(question))
```
fredreichbier: Hm, I think it would be cleaner to also use ``b'\x00' * (128-len(question))`` here, analogously to the other cases. What do you think?

fredreichbier: Ah, it would actually be better to convert to bytes before padding anyway, in case ``question`` contains non-ASCII characters :-)

plettich: Fixed in b2026028683fde876391eef855f9739fb96d9dd4

PR #1451, opened 2019-02-14
1435/resolver in event condition

cornelinux:


privacyidea/lib/eventhandler/base.py
```diff
@@ -118,17 +125,29 @@ def conditions(cls):
...
                 "desc": _("The token realm, for which this event should "
                           "apply."),
-                "value": [{"name": r} for r in realms.keys()]
+                "value": [{"name": r} for r in list(realms)]
+            },
+            CONDITION.TOKENRESOLVER: {
+                "type": "multi",
+                "desc": _("The token resolver, for which this event should "
+                          "apply."),
+                "value": [{"name": r} for r in list(resolvers)]
```
plettich: 1. just `for r in resolvers` is enough
2. is it intended that multiple (token) resolvers can be selected here but only one (user) resolver?

cornelinux: yes. For whichever reason.

cornelinux: done in 081876c1

privacyidea/lib/eventhandler/base.py
```diff
@@ -311,12 +335,30 @@ def check_condition(self, options):
...
+            # There is a token involved, so we determine it's resolvers and realms
             token_obj = token_obj_list[0]
             tokenrealms = token_obj.get_realms()
             tokentype = token_obj.get_tokentype()
 
-        if "realm" in conditions:
-            if user.realm != conditions.get("realm"):
+            all_realms = get_realms()
+            for tokenrealm in tokenrealms:
+                resolvers = all_realms.get(tokenrealm, {}).get("resolver")
```
fredreichbier: Is there any situation in which there is no ``"resolver"`` key in the dictionary (e.g. if ``all_realms`` does not contain ``tokenrealm``?)? In this case, the line below would fail with ``AttributeError: 'NoneType' object has no attribute 'get'``

Should we maybe change this to ``.get("resolver", {})``?

privacyidea/lib/eventhandler/base.py
```diff
@@ -356,10 +398,19 @@ def check_condition(self, options):
                 if not ','.join([tok.get_serial() for tok in token_objects]):
                     return False
 
-        if "tokenrealm" in conditions and tokenrealms:
+        if CONDITION.TOKENREALM in conditions and tokenrealms:
             res = False
             for trealm in tokenrealms:
-                if trealm in conditions.get("tokenrealm").split(","):
+                if trealm in conditions.get(CONDITION.TOKENREALM).split(","):
```
fredreichbier: An idea: We could get rid of ``res`` and change this to 
```python
for trealm in tokenrealms:
    if trealm in conditions.get(CONDITION.TOKENREALM).split(","):
        break
else:
    return False
```
because Python executes the ``else`` clause only if the loop exited normally, i.e. it is not executed if a ``break`` was encountered.
But the ``for...else`` syntax is a bit obscure, so I'm okay with keeping this like it is. What do you think?

privacyidea/lib/eventhandler/base.py
```diff
@@ -356,10 +398,19 @@ def check_condition(self, options):
                 if not ','.join([tok.get_serial() for tok in token_objects]):
                     return False
 
-        if "tokenrealm" in conditions and tokenrealms:
+        if CONDITION.TOKENREALM in conditions and tokenrealms:
```
fredreichbier: Hmm, I'm not sure about the logic of ``and tokenrealms`` here: Assume we have an event handler with condition ``realm=defrealm``, and we have a token without any realms. In a request with that token, the handler is executed, because ``tokenrealms`` is empty. I would expect the handler not to be executed, because the token realm is not ``defrealm``. What do you think?

privacyidea/lib/eventhandler/base.py
```diff
@@ -356,10 +398,19 @@ def check_condition(self, options):
...
             res = False
             for trealm in tokenrealms:
-                if trealm in conditions.get("tokenrealm").split(","):
+                if trealm in conditions.get(CONDITION.TOKENREALM).split(","):
+                    res = True
+                    break
+            if not res:
+                return False
+
+        if CONDITION.TOKENRESOLVER in conditions and tokenresolvers:
```
fredreichbier: Same question as above: What if the token has no resolvers?

privacyidea/lib/eventhandler/base.py
plettich: Personally i find this a bit difficult to read/understand. I just don't expect an `else:` here and for me it implies, that the loop did **not** finish correctly.

privacyidea/lib/eventhandler/base.py
cornelinux: I am not sure, if

    .get(tokenrealm, {}) 

returns {}.
But I see where you are aiming at. It will fail in ``for  r in resolvers`` in case of None.

Will do!

privacyidea/lib/eventhandler/base.py
cornelinux: There are several of these conditions: "and serial", "and user".
The originial reason for this was not that the token would have no tokenrealm but rather, that the tokenrealm information is not available in this request/ could not be determined, since the serialnumber of a token was not specified or could not be determined.

If you think we should change this, this is absolutely ok. But I would prefer opening a new issue for that, so that it is easier to track the actual logical change in the behaviour.

privacyidea/lib/eventhandler/base.py
cornelinux: s.o.

privacyidea/lib/eventhandler/base.py
cornelinux: ...so when will the ``else`` branch be called?
I look at this and I do not unterstand it ;-)

privacyidea/lib/eventhandler/base.py
cornelinux: Done in 3fdefe23

fredreichbier: Thanks!

privacyidea/lib/eventhandler/base.py
fredreichbier: The ``else`` branch is executed only if we did *not* break out of the loop -- if you're used to it, it's quite useful, but I agree that it's not very readable. So let's keep ``res`` :-)

privacyidea/lib/eventhandler/base.py
fredreichbier: Ah okay, thanks for the clarification! I opened #1460 to track this issue.

PR #1452, opened 2019-02-15
Allow to concatenate challenge text in a more flexible way

### LINKED ISSUE #1384: Enhance the challenge response text
Allow the challenge text in case of challege response to be combined of:

1. An introduction text
2. a list of the texts of all tokens, while the text of one tokentype will only be added once
3. An end text.

To allow this, we would have to add policies for the introduction text and the end text. (Note: HTML tags can be used)

cornelinux:
In case of multi_challenge, double challenge texts will be removed.
The list of the challenge texts can have an additional footer and
header.
It may contain HTML to format an ordered or unordered list.

Closes #1384

### DIFF
--- a/doc/policies/authentication.rst
+++ b/doc/policies/authentication.rst
@@ -314,7 +314,8 @@ challenge_text, challenge_text_header, challenge_test_footer
 .. index:: Challenge Text Policy
 
 Using these policies the administrator can modify the challenge texts
-of e.g. Email-Token or SMS-Token.
+of e.g. Email-Token or SMS-Token. The action *challenge_text* changes the
+challenge text in general, no matter which challenge response token is used.
 
 If the *challenge_text_header* is set and if there are more matching
 challenge response tokens, then the texts of all tokens are
--- a/doc/policies/authentication.rst
+++ b/doc/policies/authentication.rst
@@ -304,3 +304,27 @@ dependent on the clients IP address and the user agent.
 .. note:: The AuthCache only works for user authentication, not for
    authentication with serials.
 
+
+.. _policy_challenge_text:
+
+
+challenge_text, challenge_text_header, challenge_test_footer
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. index:: Challenge Text Policy
+
+Using these policies the administrator can modify the challenge texts
+of e.g. Email-Token or SMS-Token.
+
+If the *challenge_text_header* is set and if there are more matching
+challenge response tokens, then the texts of all tokens are
+concatenated together. Double challenge texts are reduced to one text only.
+
+The *challenge_text_header* and *challenge_text_footer* may contain HTML.
+If the *challenge_text_header* ends with an ``<ul>`` or ``<ol>``, then
+all the challenge texts are formatted as an ordered or unordered list.
+In this case the *challenge_text_footer* also should contain the closing
+tag.
+
+.. note:: The footer will only be used, if the header is also set.
+
--- a/privacyidea/api/lib/postpolicy.py
+++ b/privacyidea/api/lib/postpolicy.py
@@ -775,3 +775,63 @@ def construct_radius_response(request, response):
         return make_response('', return_code)
     else:
         return response
+
+
+def mangle_challenge_response(request, response):
+    """
+    This policy decorator is used in the AUTH scope to
+    decorate the /validate/check endpoint.
+    It can modify the contents of the response "detail"->"message"
+    to allow a better readibilty for a challenge response text.
+
+    :param request:
+    :param response:
+    :return:
+    """
+    content = json.loads(response.data)
+    policy_object = g.policy_object
+    user_obj = request.User
+
+    header_pol = policy_object.get_action_values(action=ACTION.CHALLENGETEXT_HEADER,
+                                                 scope=SCOPE.AUTH,
+                                                 allow_white_space_in_action=True,
+                                                 client=g.client_ip,
+                                                 user=user_obj.login,
+                                                 realm=user_obj.realm,
+                                                 resolver=user_obj.resolver,
+                                                 audit_data=g.audit_object.audit_data)
+
+    footer_pol = policy_object.get_action_values(action=ACTION.CHALLENGETEXT_FOOTER,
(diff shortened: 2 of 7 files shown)


doc/policies/authentication.rst
```diff
@@ -304,3 +304,27 @@ dependent on the clients IP address and the user agent.
 .. note:: The AuthCache only works for user authentication, not for
    authentication with serials.
 
+
+.. _policy_challenge_text:
+
+
+challenge_text, challenge_text_header, challenge_test_footer
```
fredreichbier: Hm, I think we have no docs for the ``challenge_text`` action yet. What do you think about adding a sentence here?

privacyidea/api/lib/postpolicy.py
```diff
@@ -775,3 +775,63 @@ def construct_radius_response(request, response):
...
+
+    :param request:
+    :param response:
+    :return:
+    """
+    content = json.loads(response.data)
+    policy_object = g.policy_object
+    user_obj = request.User
+
+    header_pol = policy_object.get_action_values(action=ACTION.CHALLENGETEXT_HEADER,
```
fredreichbier: What do you think about passing ``unique=True`` here? I think we can't handle the case that we get more than one result anyway.

privacyidea/api/lib/postpolicy.py
```diff
@@ -775,3 +775,63 @@ def construct_radius_response(request, response):
...
+                                                 scope=SCOPE.AUTH,
+                                                 allow_white_space_in_action=True,
+                                                 client=g.client_ip,
+                                                 user=user_obj.login,
+                                                 realm=user_obj.realm,
+                                                 resolver=user_obj.resolver,
+                                                 audit_data=g.audit_object.audit_data)
+
+    footer_pol = policy_object.get_action_values(action=ACTION.CHALLENGETEXT_FOOTER,
```
fredreichbier: same here

tests/test_api_lib_policy.py
```diff
@@ -2048,3 +2049,65 @@ def test_17_pin_change(self):
...
+                                     "enter the TOTP from your SMS",
+                          "messages": ["enter the HOTP from your email",
+                                       "enter the HOTP from your email",
+                                       "enter the TOTP from your SMS"],
+                          "multi_challenge": ["chal1", "chal2", "chal3"]}}
+        resp = jsonify(res)
+
+        r = mangle_challenge_response(req, resp)
+        message = r.json.get("detail", {}).get("message")
+        self.assertEqual(message, "These are your options:<ul><li>enter the TOTP from your SMS</li>\n<li>enter the HOTP from your email</li>\nHappy authenticating!")
```
fredreichbier: I think we cannot be sure of the order of elements here, because we perform ``list(set(messages))`` on the messages. But I guess we'll see if the test randomly fails :-)

fredreichbier: Ah actually, just noticed that the test has already failed on Travis :-) https://travis-ci.org/privacyidea/privacyidea/jobs/493770918#L2131
So I guess we'll need a different assertion here -- something like ``self.assertIn("<li>...</li>", message)``?

privacyidea/api/lib/postpolicy.py
```diff
@@ -775,3 +775,63 @@ def construct_radius_response(request, response):
...
+    This policy decorator is used in the AUTH scope to
+    decorate the /validate/check endpoint.
+    It can modify the contents of the response "detail"->"message"
+    to allow a better readibilty for a challenge response text.
+
+    :param request:
+    :param response:
+    :return:
+    """
+    content = json.loads(response.data)
```
fredreichbier: Uh, just noticed because of Travis: In case of ``/validate/radiuscheck``, the response body is empty, so no JSON object can be decoded: https://travis-ci.org/privacyidea/privacyidea/jobs/493770918#L2185
Should we check for ``if response.data:``? Or just do nothing in case of a ValueError?

tests/test_api_lib_policy.py
plettich: This would not detect double entries. We could return `sorted(set(messages))` which returns a sorted list.

doc/policies/authentication.rst
cornelinux: Hm.
I do not care.
But will do.

privacyidea/api/lib/postpolicy.py
cornelinux: Will do

privacyidea/api/lib/postpolicy.py
cornelinux: Will do

tests/test_api_lib_policy.py
cornelinux: This is interesting.
This way it could even be sorted by adding some hidden tags.
I like this "sorted" approach.

Will do.

privacyidea/api/lib/postpolicy.py
cornelinux: What do you mean - just do nothing?

We could also do an early exit, if the endpoint is not ``/validate/check``?

fredreichbier: By "do nothing", I meant we could catch the ValueError and do an early exit:
```python
try:
    content = json.loads(response.data)
except ValueError:
    return response
```
But an early exit in case the endpoint is not ``/validate/check`` also sounds good!

doc/policies/authentication.rst
cornelinux: done in a94cc884

privacyidea/api/lib/postpolicy.py
cornelinux: done in a94cc884

privacyidea/api/lib/postpolicy.py
cornelinux: done in a94cc884

tests/test_api_lib_policy.py
cornelinux: done in a94cc884

privacyidea/api/lib/postpolicy.py
cornelinux: done in a94cc884

doc/policies/authentication.rst
fredreichbier: Thanks!

PR #1458, opened 2019-02-18
Python3 update api tests

plettich:
This PR should fix all API testcases to make them run with Python 3

Most of the changes are just string encoding fixes.
Also the admin auth-token will now be created in the `setUpClass()` method which improves the runtime for the API testcases even further (ideally we would create one auth-token for the complete test run).
Due to this, there are  changes in `test_apt_audit.py` since we save some API calls.

Some tests were changed to use the resulting json object directly instead of searching in the data-string.

Working on #676

### DIFF
--- a/privacyidea/api/before_after.py
+++ b/privacyidea/api/before_after.py
@@ -332,9 +332,9 @@ def privacyidea_error(error):
     These are not that critical exceptions.
     """
     if "audit_object" in g:
-        g.audit_object.log({"info": error.message})
+        g.audit_object.log({"info": str(error)})
         g.audit_object.finalize_log()
-    return send_error(error.message, error_code=error.id), 400
+    return send_error(str(error), error_code=error.id), 400
 
 
 # other errors
@@ -360,6 +360,6 @@ def internal_error(error):
     occurs.
     """
     if "audit_object" in g:
-        g.audit_object.log({"info": error.message})
+        g.audit_object.log({"info": str(error)})
         g.audit_object.finalize_log()
-    return send_error(error.message, error_code=-500), 500
+    return send_error(str(error), error_code=-500), 500
--- a/privacyidea/api/before_after.py
+++ b/privacyidea/api/before_after.py
@@ -332,9 +332,9 @@ def privacyidea_error(error):
     These are not that critical exceptions.
     """
     if "audit_object" in g:
-        g.audit_object.log({"info": unicode(error)})
+        g.audit_object.log({"info": error.message})
         g.audit_object.finalize_log()
-    return send_error(unicode(error), error_code=error.id), 400
+    return send_error(error.message, error_code=error.id), 400
 
 
 # other errors
@@ -360,6 +360,6 @@ def internal_error(error):
     occurs.
     """
     if "audit_object" in g:
-        g.audit_object.log({"info": unicode(error)})
+        g.audit_object.log({"info": error.message})
         g.audit_object.finalize_log()
-    return send_error(unicode(error), error_code=-500), 500
+    return send_error(error.message, error_code=-500), 500
--- a/privacyidea/api/before_after.py
+++ b/privacyidea/api/before_after.py
@@ -27,6 +27,7 @@ Flask endpoints.
 It also contains the error handlers.
 """
 
+import six
 from .lib.utils import (send_error, get_all_params)
 from ..lib.user import get_user_from_param
 import logging
@@ -332,9 +333,9 @@ def privacyidea_error(error):
     These are not that critical exceptions.
     """
     if "audit_object" in g:
-        g.audit_object.log({"info": str(error)})
+        g.audit_object.log({"info": six.text_type(error)})
         g.audit_object.finalize_log()
-    return send_error(str(error), error_code=error.id), 400
+    return send_error(six.text_type(error), error_code=error.id), 400
 
 
 # other errors
@@ -360,6 +361,6 @@ def internal_error(error):
     occurs.
     """
     if "audit_object" in g:
-        g.audit_object.log({"info": str(error)})
+        g.audit_object.log({"info": six.text_type(error)})
         g.audit_object.finalize_log()
-    return send_error(str(error), error_code=-500), 500
+    return send_error(six.text_type(error), error_code=-500), 500
--- a/privacyidea/api/lib/utils.py
+++ b/privacyidea/api/lib/utils.py
@@ -208,7 +208,7 @@ def send_csv_result(obj, data_key="tokens",
     # Do the data
     for row in obj.get(data_key, {}):
         for val in row.values():
-            if isinstance(val, six.text_type):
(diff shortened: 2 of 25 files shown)


privacyidea/api/lib/utils.py
```diff
@@ -207,7 +208,7 @@ def send_csv_result(obj, data_key="tokens",
     # Do the data
     for row in obj.get(data_key, {}):
         for val in row.values():
-            if type(val) in [str, unicode]:
+            if isinstance(val, six.text_type):
```
fredreichbier: Hm, should we change this to ``six.string_types`` for Python 2 compatibility?

privacyidea/lib/crypto.py
```diff
@@ -404,7 +404,7 @@ def aes_encrypt_b64(key, data):
     """
     iv = geturandom(16)
     encdata = aes_encrypt(key, iv, data)
-    return base64.b64encode(iv + encdata)
+    return b64encode_and_unicode(iv + encdata)
```
fredreichbier: Looks sensible! Could you just adapt the docstring? :-)

privacyidea/lib/importotp.py
```diff
@@ -568,7 +568,7 @@ def decrypt(self, input_data):
                 decrypted.status, decrypted.stderr))
             raise Exception(decrypted.stderr)
 
-        return decrypted.data
+        return to_unicode(decrypted.data)
```
fredreichbier: Originally I wanted to note that we can't be sure that we always decrypt valid UTF-8 data. However, I noticed we only use ``GPGImport`` for importing token files which we can expect to be UTF-8. So let's keep it this way, could you just add a note to the docstring?

tests/test_api_audit.py
```diff
@@ -55,7 +55,7 @@ def test_02_get_allowed_audit_realm(self):
             json_response = json.loads(res.data.decode('utf8'))
             self.assertTrue(json_response.get("result").get("status"), res)
             self.assertEqual(json_response.get("result").get("value").get(
-                "count"), 9)
+                "count"), 8)
```
fredreichbier: Just to be sure: I guess this is because we have one less ``POST /auth`` entry now?

tests/test_api_audit.py
```diff
@@ -183,7 +183,7 @@ def test_03_get_allowed_audit_realm(self):
             # We now have 3 entries, as we added one by the search in line #43
             count = json_response.get("result").get("value").get("count")
             auditdata = json_response.get("result").get("value").get("auditdata")
-            self.assertEqual(count, 22)
+            self.assertEqual(count, 20)
```
fredreichbier: Same here?

tests/test_api_system.py
```diff
@@ -55,7 +54,7 @@ def test_02_update_config(self):
                                            headers={'Authorization': self.at}):
             res = self.app.full_dispatch_request()
             self.assertTrue(res.status_code == 200, res)
-            self.assertTrue('"key3": "update"' in res.data, res.data)
+            self.assertIn('key3', res.json['result']['value'], res.json)
```
fredreichbier: Hm, I think this test doesn't check if ``res.json['result']['value']['key3'] == "update"``, like the old test, but maybe that's intentional?

privacyidea/lib/user.py
```diff
@@ -143,8 +143,6 @@ def __str__(self):
         ret = u"<empty user>"
         if not self.is_empty():
             login = self.login
-            if not isinstance(login, six.text_type):
```
fredreichbier: Can we be sure that ``login`` is always ``unicode`` in Python 2 resp. ``str`` in Python 3? It definitely should be, but I'm not sure if it's really the case currently :-)
But still, I think it is reasonable to remove these checks in 3.0. What do you think about adding a warning at the end of ``User.__init__`` like
```python
if not isinstance(login, six.text_type):
    log.warning("Unexpected bytestring as login name: {!r}".format(login))
```
?

tests/test_api_audit.py
plettich: correct

privacyidea/api/lib/utils.py
plettich: Yes, missed that.

privacyidea/lib/importotp.py
plettich: my thoughts exactly :-) I will update the docstring.

tests/test_api_audit.py
plettich: correct

tests/test_api_system.py
plettich: True, got mixed up in the json format. Will fix.

privacyidea/api/lib/utils.py
plettich: fixed in 21bdef5b

privacyidea/lib/crypto.py
plettich: fixed in 21bdef5b

privacyidea/lib/importotp.py
plettich: fixed in 21bdef5b

tests/test_api_system.py
plettich: fixed in 21bdef5b

privacyidea/lib/user.py
plettich: I think we should be on the safe side, even if it is a `str` type in Python 2. It is mostly used in logging and deleting the user cache. These should work just fine.
And `self.login` is only written in 3 places so if we run into trouble, there is a quick fix :-)

privacyidea/api/lib/utils.py
fredreichbier: Thanks!

privacyidea/lib/user.py
fredreichbier: I guess you're right -- let's do it that way :)

privacyidea/api/before_after.py
```diff
@@ -332,9 +332,9 @@ def privacyidea_error(error):
     These are not that critical exceptions.
     """
     if "audit_object" in g:
-        g.audit_object.log({"info": unicode(error)})
+        g.audit_object.log({"info": str(error)})
```
fredreichbier: Hm sorry, I missed this earlier: Might this cause problems under Python 2, because we now pass bytestrings where we used to pass unicodes? I *think* we should be fine, but I'm not really sure :-)

plettich: You are right, i tried to pass the `error.message` here but it is not the same as `str(error)`. This gets passed to the Audit-Model, which uses `db.String` for the column type instead of `db.Unicode`.
So i will change here to `six.text_type(error)` and we should change the column types in the Audit table.

plettich: fixed in dbfada63

fredreichbier: Thanks!

PR #1459, opened 2019-02-19
Update smpplib to run with Python 3

plettich:
The recent update of [python-smpplib](https://github.com/python-smpplib/python-smpplib) to version 2.0 makes it possible to run with Python 2 and 3.
Only small changes were required to make the smsprovider test cases work with Python 2 and 3.

The new [smpplib](https://pypi.org/project/smpplib/) package on pypi should be compatible with version 0.1 (https://github.com/pypa/warehouse/issues/5122)

Working on #676

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -61,7 +61,7 @@ redis==2.10.6
 requests==2.20.0
 responses==0.9.0
 six==1.11.0
-smpplib==0.1
+smpplib==2.0
 snowballstemmer==1.2.1
 Sphinx==1.7.5
 sphinxcontrib-httpdomain==1.6.1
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -61,7 +61,7 @@ redis==2.10.6
 requests==2.20.0
 responses==0.9.0
 six==1.11.0
-smpplib==0.1
+smpplib==2.0
```
cornelinux: We have not started, yet, but we should do.

So please start adding a note here,
https://github.com/privacyidea/privacyidea/blob/master/READ_BEFORE_UPDATE.md
so that we do not forget, that updates need to watch out in case of smpp.

tests/test_lib_smsprovider.py
```diff
@@ -172,7 +172,7 @@ def test_01_missing_config(self):
     def test_02_simple_config_success(self):
         smtpmock.setdata(response={"recp@example.com": (200, "OK")})
         r = self.simple_provider.submit_message("123456", "Hello")
-        self.assertRaises(r)
```
cornelinux: What was that? This worked?

plettich: in Python 2 this worked...

requirements.txt
plettich: added in fd8b091d

Issue #1457, opened 2019-02-18
Optimize hidden SQL queries


fredreichbier:
We sometimes trigger additional SQL queries even though they can be avoided. Examples:

https://github.com/privacyidea/privacyidea/blob/ec448112f01afbbdace0f37592c1a27b5eba0014/privacyidea/lib/policydecorators.py#L299
Here, the ``auth_user_passthru`` decorator searches for all tokens of the current user, even if no passthru policy is configured. We should instead check for passthru policies first, and only search the user tokens if such a policy is configured. This will increase performance because the policies are cached.

https://github.com/privacyidea/privacyidea/blob/ec448112f01afbbdace0f37592c1a27b5eba0014/privacyidea/lib/token.py#L2093
With #1368 merged, ``.user`` triggers a query even if the log message is not printed.

By fixing instances like these, we can easily save several SQL queries without constraining functionality at all.

fredreichbier:
Same in ``check_tokeninfo``:
https://github.com/privacyidea/privacyidea/blob/ec448112f01afbbdace0f37592c1a27b5eba0014/privacyidea/api/lib/postpolicy.py#L268
We should first check if the policy function applies, and then query the token.

fredreichbier:
There is always potential to optimize SQL queries :) But as #1462 fixes the three instances listed above, I'm closing this.

PR #1462, opened 2019-02-20
Avoid some hidden SQL queries in /validate/check

fredreichbier:
This PR removes the three hidden SQL queries in the /validate/check codepath mentioned in #1457.
In two cases, these queries are issued by policy functions even if there is no active matching policy.
In one case, the query is issued for a DEBUG printout.

These queries are really low-hanging fruits because optimizing them doesn't change behavior at all. Removing them might make quite a difference in case the SQL server has some latency.

### DIFF
--- a/privacyidea/api/lib/postpolicy.py
+++ b/privacyidea/api/lib/postpolicy.py
@@ -263,30 +263,32 @@ def check_tokeninfo(request, response):
     """
     content = json.loads(response.data)
     serial = content.get("detail", {}).get("serial")
-    tokeninfos_pol = g.policy_object.get_action_values(
-        ACTION.TOKENINFO,
-        scope=SCOPE.AUTHZ,
-        client=g.client_ip,
-        allow_white_space_in_action=True,
-        audit_data=g.audit_object.audit_data)
-    if serial and tokeninfos_pol:
-        tokens = get_tokens(serial=serial)
-        if len(tokens) == 1:
-            token_obj = tokens[0]
-            for tokeninfo_pol in tokeninfos_pol:
-                try:
-                    key, regex, _r = tokeninfo_pol.split("/")
-                    value = token_obj.get_tokeninfo(key, "")
-                    if re.search(regex, value):
-                        log.debug(u"Regular expression {0!s} "
-                                  u"matches the tokeninfo field {1!s}.".format(regex, key))
-                    else:
-                        log.info(u"Tokeninfo field {0!s} with contents {1!s} "
-                                 u"does not match {2!s}".format(key, value, regex))
-                        raise PolicyError("Tokeninfo field {0!s} with contents does not"
-                                          " match regular expression.".format(key))
-                except ValueError:
-                    log.warning(u"invalid tokeinfo policy: {0!s}".format(tokeninfo_pol))
+
+    if serial:
+        tokeninfos_pol = g.policy_object.get_action_values(
+            ACTION.TOKENINFO,
+            scope=SCOPE.AUTHZ,
+            client=g.client_ip,
+            allow_white_space_in_action=True,
+            audit_data=g.audit_object.audit_data)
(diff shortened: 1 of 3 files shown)


privacyidea/api/lib/postpolicy.py
```diff
@@ -262,18 +262,17 @@ def check_tokeninfo(request, response):
     :return: A new modified response
     """
     content = json.loads(response.data)
-    policy_object = g.policy_object
     serial = content.get("detail", {}).get("serial")
-    if serial:
+    tokeninfos_pol = g.policy_object.get_action_values(
```
cornelinux: In this case you will always execute ``get_action_values``.
Wouldn't it be wise to only do this if there is actually a serial? e.g. put this in the if block again.

privacyidea/api/lib/postpolicy.py
```diff
@@ -262,18 +262,17 @@ def check_tokeninfo(request, response):
...
+        ACTION.TOKENINFO,
+        scope=SCOPE.AUTHZ,
+        client=g.client_ip,
+        allow_white_space_in_action=True,
+        audit_data=g.audit_object.audit_data)
+    if serial and tokeninfos_pol:
         tokens = get_tokens(serial=serial)
```
cornelinux: Yes, we could do the get_tokens *after* the get_action_values, but imho all in a block
~~~~
if serial:
~~~~

privacyidea/lib/policydecorators.py
```diff
@@ -296,10 +296,9 @@ def auth_user_passthru(wrapped_function, user_object, passw, options=None):
     from privacyidea.lib.token import get_tokens
     options = options or {}
     g = options.get("g")
-    if get_tokens(user=user_object, count=True) == 0 and g:
```
cornelinux: I am wondering if we could get a logical problem here!

The policy is logically only checked, if the user has no token.

If the policy check would raise an exception, maybe due to conflicting policies, we would change the behaviour of the system. Even a user *with* a valid token could not login, since  a policy, that has no relevance for him, would raise an exception.

I *think* in this case there will be no policy exception, since we do not use *unique=True*, but this is something we need to consider carefully. What do you think?

privacyidea/api/lib/postpolicy.py
fredreichbier: Yep, sounds good! Will do.

privacyidea/api/lib/postpolicy.py
fredreichbier: will do!

privacyidea/lib/policydecorators.py
fredreichbier: Ah, didn't think of that! Yes, we should make sure that we don't change behavior in the case you describe. Policy conflicts are only checked if the user has no tokens (``check_for_conflicts`` in line 313), but I'll check if ``get_policies`` can throw any other exceptions.

privacyidea/api/lib/postpolicy.py
fredreichbier: fixed in a1740172ecbba64493d89bb14edbd21e2f7e08f4

privacyidea/api/lib/postpolicy.py
fredreichbier: fixed in a1740172ecbba64493d89bb14edbd21e2f7e08f4

privacyidea/lib/policydecorators.py
fredreichbier: Having thought about this some more: ``get_policies`` should normally not throw any exceptions, and conflict checking is only done if the user has, in fact, zero tokens. We use ``get_policies`` in a lot of other policy functions without checking for errors, so I think that if the call goes wrong here, this would be a hint that something is seriously wrong with the privacyIDEA instance, and having an error would be sensible in that case. What do you think?

Issue #337, opened 2016-03-08
choose crypto module


cornelinux:
We are using PyCrypto for signing and encryption.

And we are using pysaml2 for the SAML SP.

This only works with pysaml2 <= 4.0.0
Higher pysaml2 modules require pycryptodome, which replaces pycrypto and is not compatible. You can not run pycrypto and pycryptodome in the same environment.

If we switch from pycrypto to pycryptodome, we need to change all the signing and encryption in privacyidea.lib.crypto.

See https://github.com/rohe/pysaml2/issues/304

plettich:
[PyCrypto](https://github.com/dlitz/pycrypto/) is effectively [dead](https://github.com/dlitz/pycrypto#238).
[pycryptodome](https://github.com/Legrandin/pycryptodome) provides a compatibility layer to function as a drop-in replacement for PyCrypto but in our case it does not work out of the box.
pycryptodome can be installed in its own namespace/module as well and thus be installed together with PyCrypto (https://github.com/Legrandin/pycryptodome#pycryptodome).
BTW: meanwhile pysaml2 switched to [pyca/cryptography](https://github.com/pyca/cryptography).

cornelinux:
I have the feeling we should go for cryptography. In my opinion, we do not need a drop-in. We can rewrite our  complete [lib/crypto.py](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/crypto.py). If we wrap nearly everything into our own ``lib/crypto.py`` we can also assure #1212.

fredreichbier:
pyca/cryptography looks cool! But will backwards-compatibility be a problem? (e.g. using pyca/cryptography to verify audit entries signed using pycrypto).

cornelinux:
Of course we should check this. After all, the signed audit entries are RSA signatures. So it *should* not matter, which implementation we will use. (I would be very well puzzled if it would not).

(Hey I can not verify your signed email with my thunderbird, since you sent it with kmail - very unlikely)
(Hey I can not verify your signed email with my kmail, since you sent it with outlook - oups, migh happen ;-)

plettich:
Actually the signature is usually calculated using the non-standard RSA signature, not the RSA-based PKCS1-v1.5 signature scheme.
It should still be possible to calculate the signature using just RSA but we might have to dig deep into the libraries.

cornelinux:
We are already using *pyca/bcrypt* so why not use *pyca/cryptography* +1

[...]

PR #1467, opened 2019-02-25
Switch to cryptography - Part 1: AES

plettich:
- The standard AES mode is now CBC (it wasn't used otherwise anyway)
- There is only one AES encryption/decryption method in `lib/crypto.py`
  except for Yubikeys which use AES in ECB mode
- The cryptography module is used as described in
  https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/#
- Using the hashlib sha256 implementation which is a builtin module
- The `enckey.enc` in `tests/testdata/` wasn't padded correctly so
  decrypting it now lead to an error while unpadding it.

Working on #337


privacyidea/lib/crypto.py
```diff
@@ -108,15 +109,17 @@ def hmac_digest(self, data_input, hash_algo):
         return h
 
     def aes_decrypt(self, data_input):
```
cornelinux: We wondered why this is using ECB. Should we rename this function to ``aes_ecb_decrypt``, so that we know this is done intentionally?

privacyidea/lib/crypto.py
```diff
@@ -307,86 +310,92 @@ def decryptPin(cryptPin):
...
-    decrypt a variable from the given input with an initialiation vector
-
-    :param input: buffer, which contains the crypted value
-    :type  input: bytes or str
-    :param iv:    initialisation vector
-    :type  iv:    bytes or str
-    :param id:    contains the key id of the keyset which should be used
-    :type  id:    int
-    :return:      decrypted buffer
+def decrypt(enc_data, iv, key_id=0):
```
cornelinux: should we maybe call the parameters the same in all functions? Above in ``aes_decrypt`` the ciphertext parameter is called *data_input*. Here is is called *enc_data*. (Maybe we should also choose *enc_data* in the above function?)

privacyidea/lib/crypto.py
```diff
@@ -307,86 +310,92 @@ def decryptPin(cryptPin):
...
+    """
     hsm = get_hsm()
-    res = hsm.decrypt(to_bytes(input), to_bytes(iv), id)
+    res = hsm.decrypt(to_bytes(enc_data), to_bytes(iv), key_id)
     return res
 
 
 @log_with(log, log_exit=False)
-def aes_decrypt(key, iv, cipherdata, mode=AES.MODE_CBC):
+def aes_decrypt(key, iv, cipherdata):
```
cornelinux: Again, here the *data_input* or *enc_data* is called *cipherdata*. Time for clean up?

Is it worth to have this function utilize CBC and ECB?
Or is it not worth, since ECB is only used within the secretObj?
And ousite always CBC is used?

privacyidea/lib/crypto.py
```diff
@@ -656,10 +665,10 @@ def sign(self, s):
             s = s.encode('utf8')
         RSAkey = RSA.importKey(self.private)
```
cornelinux: Shouldn't we also replace the RSA signing and verification? 
See https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#signing
and https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#verification

However, we should use pkcs1, as we did before:

Verify:
~~~~python
pubkey_obj = serialization.load_pem_public_key(str(pubkey_pem), default_backend())
chosen_hash = hashes.SHA256()
hasher = hashes.Hash(chosen_hash, default_backend())
sign_data = u"{0!s}|{1!s}".format(challenge, serial)
hasher.update(sign_data.encode("utf8"))
digest = hasher.finalize()
try:
                        pubkey_obj.verify(b32decode(signature), digest,
                                          padding.PKCS1v15(),
                                          utils.Prehashed(chosen_hash))
~~~~

We use this in the new pushtoken. 
Should I refactor this Sign Object, so that it can be used in the push token?

privacyidea/lib/subscriptions.py
```diff
@@ -28,14 +28,14 @@
...
+from hashlib import sha256
 from .log import log_with
 from ..models import Subscription
 from privacyidea.lib.error import SubscriptionError
 from privacyidea.lib.token import get_tokens
 import functools
 from privacyidea.lib.framework import get_app_config_value
 import os
-from Crypto.Hash import SHA256
 from Crypto.PublicKey import RSA
```
cornelinux: RSA is issued for the second part?

tests/testdata/enckey.enc
```diff
@@ -1 +1 @@
-95c74e8a65bf9575370c07f175c9a88e:776ca294b312cb3f44b94481b3b010759912b34e2485e9f457d160b21760f7d50109fcfb4a9c0783464410a5894299e59bbf3a7ae44019bed54420a9c4ba18da5cc78dbc19bec42884831d7c3fad86695f3c629ffc41b3d7da22c53f23cf1f4b76a178ebb9636c4eb4bc454ad1e3cd47becbfd923f479b85642c93a5aa8ea2e85595fca2f45c1e0361d92748e8ccc8590f57f3d9849d7ac022183ade71263974b051574c54a636080a9bbf5bce0a09745812056ea8945666916a9019faf48ee89e5f0f8a2520ac044254524335ac3155
+95c74e8a65bf9575370c07f175c9a88e:776ca294b312cb3f44b94481b3b010759912b34e2485e9f457d160b21760f7d50109fcfb4a9c0783464410a5894299e59bbf3a7ae44019bed54420a9c4ba18da5cc78dbc19bec42884831d7c3fad86695f3c629ffc41b3d7da22c53f23cf1f4b76a178ebb9636c4eb4bc454ad1e3cd47becbfd923f479b85642c93a5aa8ea2e85595fca2f45c1e0361d92748e8ccc8590f57f3d9849d7ac022183ade71263974b051574c54a636080a9bbf5bce0a09745812056ea8945666916a9019faf48ee8fe859253e15be6404094e4f32f3db9b4
```
cornelinux: Oups, why has the key changed?
THis should not happen - what happens to existing installations?

privacyidea/lib/crypto.py
plettich: makes sense.

privacyidea/lib/crypto.py
plettich: makes sense as well

privacyidea/lib/crypto.py
plettich: on it.
This is why it is still a draft PR :-)
The `SecretObj` is used to store the cryptographic material of tokens. The material itself is encrypted/decrypted using the systems enckey or HSM.
Since the ECB-stuff is only used by the yubikey tokens, we should not mix it here.
And maybe we want to switch to a different encryption/decryption method for the cryptographic material (EC, something with MAC).

privacyidea/lib/crypto.py
plettich: This will be addressed in Part 2 :-)

tests/testdata/enckey.enc
plettich: When i am finished with the versioning of the encryption, this should be fixed.

cornelinux: OK, cool.
Can you please leave the old value as commented out, so that we know the "target-value".

plettich:
Due to some changes in the handling of AES padding, i'll close this PR.
In order to keep it simple for 3.0 we decided to skip the standard padding and keep out own for now.
The requested changes are incorporated into #1492

PR #1469, opened 2019-02-26
Backport auth cache fixes to 2.23

fredreichbier:
This PR backports the auth cache fixes from #1387 to branch-2.23. We need another compatibility commit (e172b37) to account for other changes in the master branch.

### DIFF
--- a/tests/test_api_validate.py
+++ b/tests/test_api_validate.py
@@ -2250,6 +2250,34 @@ class ValidateAPITestCase(MyTestCase):
         remove_token(serial)
         delete_privacyidea_config("no_auth_counter")
 
+    def test_33_auth_cache(self):
+        init_token({"otpkey": self.otpkey},
+                   user=User("cornelius", self.realm1))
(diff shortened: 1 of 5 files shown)


tests/test_api_validate.py
```diff
@@ -2250,6 +2251,69 @@ def test_31_count_auth(self):
...
+            self.assertTrue(res.status_code == 200, res)
+            result = json.loads(res.data.decode('utf8')).get("result")
+            self.assertTrue(result.get("status"))
+            self.assertTrue(result.get("value"))
+            detail = json.loads(res.data.decode('utf8')).get("detail")
+            self.assertEqual(detail.get("message"), u"Authenticated by AuthCache.")
+
+        delete_policy("authcache")
+
+        # If there is no authcache, the same value must not be used again!
```
cornelinux: I just see this now.
We should have added an authentication attempt with ``OTPs[1]``, which should fail after deleting the authcache policy.

fredreichbier: Whoops :-) We could add it in master?

cornelinux: Yes. Can you add the test into master like now? Then we know that it will also work! ;-)

PR #1472, opened 2019-02-27
Python 3 compatibility for tools

plettich:
The python based tools should now be able to run with python 2 and python 3.

Changes include print statement vs. print function, literal octal
representation, iteration of dict items and differences in the except
clause.

Cosmetics:
- Fix the shebang to use `/usr/bin/env python` in order to execute the
  correct python version in a virtual environment
- Add content type description if it was missing
- Add a whole lot of wrapping and indenting
- Rename some variables/functions to avoid overwriting python internals
- Cleanup of imports

Working on #676

### DIFF
--- a/pi-manage
+++ b/pi-manage
@@ -76,7 +76,7 @@ from privacyidea.lib.caconnector import (get_caconnector_list,
                                          save_caconnector)
 from privacyidea.app import create_app
 from privacyidea.lib.auth import ROLE
-from flask_script import Manager, Command
+from flask_script import Manager, Command, Option
 from privacyidea.app import db
 from flask_migrate import MigrateCommand
 # Wee need to import something, so that the models will be created.
@@ -163,8 +163,9 @@ def create_crl(ca, force=False):
 
 
 @ca_manager.option('name', help='The name of the new CA')
-@ca_manager.option('-t', '--type', help='The type of the new CA. The default '
-                                        'is "local"')
+@ca_manager.option('-t', '--type',
+                   help='The type of the new CA. The default is "local"',
+                   dest='catype')
 def create(name, catype='local'):
     """
     Create a new CA connector. In case of the "localca" also the directory
@@ -252,7 +253,7 @@ def create(directory="/var/lib/privacyidea/backup/",
     CONF_DIR = conf_dir
     DATE = datetime.datetime.now().strftime("%Y%m%d-%H%M")
     BASE_NAME = "privacyidea-backup"
-    
+
     call(["mkdir", "-p", directory])
     # set correct owner, if possible
     if os.geteuid() == 0:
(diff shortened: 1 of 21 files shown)


plettich:
We should definitely add tests for the tools, even if it is just checking if `$ script --help` works.

pi-manage
```diff
@@ -159,7 +165,7 @@ def create_crl(ca, force=False):
 @ca_manager.option('name', help='The name of the new CA')
 @ca_manager.option('-t', '--type', help='The type of the new CA. The default '
                                         'is "local"')
-def create(name, type):
+def create(name, catype='local'):
```
fredreichbier: I think renaming ``type`` to ``catype`` is a good idea to avoid shadowing the ``type`` builtin, but unfortunately it breaks the command because flask-script tries to pass a ``type`` argument :-/

```shell
$ ./pi-manage ca create foo                                                                                                                                                              
   
  File "./pi-manage", line 1200, in <module>
    manager.run()
  File "/home/fred/privacyidea/privacyidea/venv/lib/python2.7/site-packages/flask_script/__init__.py", line 417, in run
    result = self.handle(argv[0], argv[1:])
  File "/home/fred/privacyidea/privacyidea/venv/lib/python2.7/site-packages/flask_script/__init__.py", line 386, in handle
    res = handle(*args, **config)
  File "/home/fred/privacyidea/privacyidea/venv/lib/python2.7/site-packages/flask_script/commands.py", line 216, in __call__
    return self.run(*args, **kwargs)
TypeError: <flask_script.commands.Command object at 0x7f6c1569c550>: create() got an unexpected keyword argument 'type'
```

But I think it might be possible to just add ``dest="catype"`` to the @ca_manager.option decorator?

pi-manage
```diff
@@ -837,32 +855,35 @@ def list(verbose):
...
             data = resolver.get("data", {})
-            for k, v in data.iteritems():
+            for (k, v) in data.items():
                 if k.lower() in ["bindpw", "password"]:
                     v = "xxxxx"
                 print("{0!s:>24}: {1!r}".format(k, v))
             print("")
 
 
 @realm_manager.command
```
fredreichbier: I think the command is redundant because we add a ``list`` command manually below?

tools/privacyidea-standalone
```diff
@@ -174,13 +176,13 @@ def choice(question, choices, case_insensitive=True):
     :return: a value of ``choices``
     """
     while True:
-        answer = raw_input(question)
+        answer = input(question)
```
fredreichbier: Same issue with ``input`` as above

tools/privacyidea-pip-update
```diff
@@ -85,9 +86,10 @@ def main():
         update()
     else:
         answer = False
+        res = 'n'
         while not answer:
-            res = raw_input("Do you really want to update your "
-                            "python environment and the DB schema? (y/N)")
+            res = input("Do you really want to update your "
```
fredreichbier: Hm, under Python 2 we shouldn't use ``input`` because it evaluates the input: https://code-maven.com/never-use-input-in-python2
We could use six to find the right input for us:
```
from six.moves import input
```
The question is if we expect anyone to run this under Python 2?

pi-manage
plettich: That is why we need tests :-D

plettich: fixed in b065a00f

pi-manage
plettich: fixed in b065a00f

tools/privacyidea-standalone
plettich: fixed in b065a00f

tools/privacyidea-pip-update
plettich: fixed in b065a00f

fredreichbier:
Looks good to me! Thanks!

PR #1480, opened 2019-03-01
pi-manage can cleaup old entries in the authcache

### LINKED ISSUE #1471: Script for deleting expired entries from the auth cache
Similarly to the user cache, we need a script that deletes outdated entries from the auth cache. See #1466

cornelinux:
Closes #1471

### DIFF
--- a/pi-manage
+++ b/pi-manage
@@ -85,6 +85,7 @@ from sqlalchemy import create_engine, desc, MetaData
 from sqlalchemy.orm import sessionmaker
 from privacyidea.lib.auditmodules.sqlaudit import LogEntry
 from privacyidea.lib.audit import getAudit
+from privacyidea.lib.authcache import cleanup as authcache_cleanup
 from privacyidea.lib.utils import parse_timedelta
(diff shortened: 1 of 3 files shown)


pi-manage
```diff
@@ -559,6 +562,16 @@ except TypeError:
     profile = manager.command(profile)
 
 
+@authcache_manager.command
+def cleanup(minutes=480):
+    """
+    Remove entries from the authcache which last_auth entry is older than
```
plettich: which -> where

cornelinux: You see, I think I wanted to say "for which..."

cornelinux: done in 4506a7bb

cornelinux:
Hi @plettich 
we need this pull request in branch 2.23 for 2.23.5.
So we can not wait for #1472.

But: We can merge this (1480) into master *after* the PR #1472 was merged.
But we should merge this one into branch-2.23 like now.

Thanks a lot.

plettich:
I should look at the branches to merge and not assume it is always master...

PR #1481, opened 2019-03-01
Use the correct sync_window when resynching TOTP

### LINKED ISSUE #1479: TOTP token uses wrong attribute for manual sync
The TOTP token uses the wrong attribute for calculating the resync (wrong attriubte: ``timeWindow``).
It should use the ``syncWindow``.
The ``_autosync`` method (actually does it right)[https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/tokens/totptoken.py#L398]!

When the timeWindow is increased, a manual resync would work:

      update tokeninfo set Value="1800" where Key="timeWindow" and token_id=62;

This is the problematic code:
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/tokens/totptoken.py#L475

We need to use ``self.get_sync_window()`` instead of ``self.timeWindow`` for both ``res1`` and ``res2``.

As we would use the default syncwindow of 1000 just like with the autoresync this would be 1000 count, i.e. 1000*30sec = 500 Minutes --- roughly 11 hours, which is fine.

~~~~python
        # check 2nd value
        hmac2Otp = HmacOtp(secretHOtp,
                           counter,
                           otplen,
                           self.get_hashlib(self.hashlib))
        log.debug("{0!s} in otpkey: {1!s} ".format(otp2, secretHOtp))
        res2 = hmac2Otp.checkOtp(otp2,
                                 int(self.get_sync_window()),
                                 symetric=True)  # TEST -remove the 10
        log.debug("res 2: {0!r}".format(res2))
        # check 1st value
        hmac2Otp = HmacOtp(secretHOtp,
                           counter - 1,
                           otplen,
                           self.get_hashlib(self.hashlib))
        log.debug("{0!s} in otpkey: {1!s} ".format(otp1, secretHOtp))
        res1 = hmac2Otp.checkOtp(otp1,
                                 int(self.get_sync_window()),
                                 symetric=True)  # TEST -remove the 10
~~~~

We also need to add some sensible tests!

cornelinux:
Fixes #1479

### DIFF
--- a/privacyidea/lib/tokens/totptoken.py
+++ b/privacyidea/lib/tokens/totptoken.py
@@ -465,7 +465,8 @@ class TotpTokenClass(HotpTokenClass):
         oCount = self.get_otp_count()
 
         log.debug("tokenCounter: {0!r}".format(oCount))
-        log.debug("now checking window {0!s}, timeStepping {1!s}".format(self.timewindow, self.timestep))
+        sync_window = self.get_sync_window()
+        log.debug("now checking window {0!s}, timeStepping {1!s}".format(sync_window, self.timestep))
         # check 2nd value
         hmac2Otp = HmacOtp(secretHOtp,
                            counter,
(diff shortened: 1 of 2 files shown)


plettich:
should we also cherry-pick this into 2.23?

cornelinux:
This is supposed merged into 2.23. We will merge this in master again.

PR #1488, opened 2019-03-06
Migrate migration scripts to Python 3

fredreichbier:
The necessary changes are due to the "new" ``except ... as`` syntax, and due to the missing ``message`` attribute for Exceptions. The alternative ``args[0]`` should work under both Python 2 and 3.


fredreichbier:
Whoops, just noticed that actually breaks Python 2 compatibility :-)

fredreichbier:
Pushed a commit that fixes the migration scripts for Python 2.

PR #1492, opened 2019-03-07
Migrate from PyCrypto to pyca/cryptography

plettich:
Part 1: AES

- remove all AES encryption/decryption based on `PyCrypto`
- add only one AES (CBC mode) encryption/decryption function which uses
  `pyca/cryptography` without padding
- renamed yubikey AES (ECB mode) decryption function.
- currently the (un)padding must be done in the calling functions

- removed the `security/pkcs11.py` module since it is not used.
- refactored some parameter names

Working on #337


privacyidea/lib/crypto.py
```diff
@@ -107,16 +108,18 @@ def hmac_digest(self, data_input, hash_algo):
...
         return h
 
-    def aes_decrypt(self, data_input):
+    def aes_ecb_decrypt(self, enc_data):
         '''
-        support inplace aes decryption for the yubikey
+        support inplace aes decryption for the yubikey (mode ECB)
 
-        :param data_input: data, that should be decrypted
+        :param enc_data: data, that should be decrypted
```
fredreichbier: If I understand correctly, the length of ``enc_data`` should be a multiple of 16 bytes. Hm, should we maybe add this to the docstring?

privacyidea/lib/crypto.py
```diff
@@ -403,25 +413,34 @@ def aes_encrypt_b64(key, data):
     :return: base64 encrypted output, containing IV and encrypted data
     :rtype: str
     """
+    # pad data
+    block_size = algorithms.AES.block_size // 8
+    num_pad = block_size - (len(data) % block_size)
+    data = data + six.int2byte(num_pad) * num_pad
```
fredreichbier: Hmm, this would be PKCS7 padding? In this case, we could also just use the [padders from cryptography](https://cryptography.io/en/latest/hazmat/primitives/padding/#cryptography.hazmat.primitives.padding.PKCS7). What do you think?

privacyidea/lib/crypto.py
```diff
@@ -403,25 +413,34 @@ def aes_encrypt_b64(key, data):
...
-    data_bin = base64.b64decode(data_b64)
+    data_bin = base64.b64decode(enc_data_b64)
     iv = data_bin[:16]
     encdata = data_bin[16:]
-    output = aes_decrypt(key, iv, encdata)
+    output = aes_cbc_decrypt(key, iv, encdata)
+
+    # remove padding
+    padding = six.indexbytes(output, len(output) - 1)
+    output = output[0:-padding]
```
fredreichbier: same as above: could we use the cryptography unpadders?

tests/test_lib_crypto.py
```diff
@@ -36,15 +36,16 @@ def test_00_security_module_base_class(self):
 
         self.assertRaises(NotImplementedError, hsm.setup_module, {})
         self.assertRaises(NotImplementedError, hsm.random, 20)
-        self.assertRaises(NotImplementedError, hsm.encrypt, "20")
-        self.assertRaises(NotImplementedError, hsm.decrypt, "20")
+        self.assertRaises(NotImplementedError, hsm.encrypt, "20", 'abcd')
```
fredreichbier: Huh, what's this :-)

privacyidea/lib/crypto.py
fredreichbier: Oh, just noticed: Don't we have a problem if the input size is divisible by 16, because then we add a padding block of null bytes when encrypting, but we don't remove it when decrypting?
... but if this is the case, the old code had the same problem.

plettich: Exactly. The problem is the unpadding. The cryptography unpadder would error out on an incorrectly padded input.
So we would need some logic here to check both versions or make a hard cut.

tests/test_lib_crypto.py
plettich: I added the `iv` as a default parameter to the base class functions. Before it was set to `iv=None`.

fredreichbier: Ah my bad, didn't notice. Sounds good!

privacyidea/lib/crypto.py
plettich: Sorry, i didn't read carefully.
Actually we are using correct PKCS7 padding here and i should use the cryptography padders here. I'll fix that.
If the input size is divisible by 16, we would have `num_pad = 16` and thus add 16 times `b'\x16'` to the data, i.e. a whole block. And this should be correctly removed by the unpadder.

privacyidea/lib/crypto.py
plettich: will do.

privacyidea/lib/crypto.py
plettich: will do.

privacyidea/lib/crypto.py
fredreichbier: I misread too :-) You're right, I think we do correct PKCS7 padding here.

privacyidea/lib/crypto.py
plettich: actually this is the decrypt routine for yubikeys so i think we can assume that we get properly encrypted and padded data here.

privacyidea/lib/crypto.py
plettich: fixed in f2e066a6

privacyidea/lib/crypto.py
plettich: fixed in f2e066a6

fredreichbier: thanks!

privacyidea/lib/crypto.py
```diff
@@ -613,81 +630,125 @@ def zerome(bufferObject):
...
+    assert isinstance(msg, six.integer_types)
+    if hasattr(key, 'private_numbers'):
+        pn = key.private_numbers().public_numbers
+    elif hasattr(key, 'public_numbers'):
+        pn = key.public_numbers()
+    else:
+        raise TypeError('No public key')
+
+    # compute m**d (mod n)
+    return msg == pow(sig, pn.e, pn.n)
```
fredreichbier: Hm, would it be possible to use https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#cryptography.hazmat.primitives.asymmetric.rsa.RSAPublicKey.verify to verify legacy signatures?

fredreichbier: As it turns out, cryptography doesn't offer the validation of unpadded "textbook" RSA by design. Resolving this.

PR #1494, opened 2019-03-12
1342 pushtoken

### LINKED ISSUE #1342: Push Token
Define the first steps for the push token:

* [x] Define KeyURI
* [x] Define Endpoint Names
* [x] Implement Tokenclass
* [x] Implement Endpoints

concept-push

See this wiki page: https://github.com/privacyidea/privacyidea/wiki/concept%3A-PushToken

cornelinux:
This PR implements the PUSH Token.

The smartphone app and the push token object in the privacyIDEA server each have
an assymmetric key pair. They share their public keys and the app gets notified via
the Google Firebase service.

(I think currently missing: The documentation for the push policies and the push token itself!)

Closes #1342


privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,578 @@
...
+        additional ``attributes``, which are displayed in the JSON response.
+        """
+        options = options or {}
+        message = get_action_values_from_options(SCOPE.AUTH,
+                                                 ACTION.CHALLENGETEXT,
+                                                 options) or DEFAULT_CHALLENGE_TEXT
+
+        attributes = None
+        data = None
+        challenge = b32encode(geturandom())
```
fredreichbier: I think this will have unexpected effects on Python 3: ``b32encode`` returns ``bytes``, which means that ``smartphone_data["nonce"]`` will be ``bytes``, which means that ``sign_string`` will contain its string representation, which includes ``b'...'`` quotes. So we should use ``b32encode_and_unicode`` instead here.

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,578 @@
...
+            sign_string = u"{nonce}|{url}|{serial}|{question}|{title}".format(**smartphone_data)
+
+            pem_privkey = self.get_tokeninfo(PRIVATE_KEY_SERVER)
+            privkey_obj = serialization.load_pem_private_key(str(pem_privkey), None, default_backend())
+
+            # Sign the data with PKCS1 padding. Not all Androids support PSS padding.
+            signature = privkey_obj.sign(sign_string.encode("utf8"),
+                                         padding.PKCS1v15(),
+                                         hashes.SHA256())
+            smartphone_data["signature"] = b32encode(signature)
```
fredreichbier: Like above, I think this will break on Python 3 (because json.dumps in ``submit_message`` cannot encode ``bytes``), so we should use ``b32encode_and_unicode``.

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,578 @@
...
+                               "title": title,
+                               "url": url}
+            # Create the signature.
+            # value to string
+            sign_string = u"{nonce}|{url}|{serial}|{question}|{title}".format(**smartphone_data)
+
+            pem_privkey = self.get_tokeninfo(PRIVATE_KEY_SERVER)
+            privkey_obj = serialization.load_pem_private_key(str(pem_privkey), None, default_backend())
+
+            # Sign the data with PKCS1 padding. Not all Androids support PSS padding.
```
fredreichbier: I have some thoughts about the choice of RSA signatures with PKCS#1 v1.5 padding:
While PKCS#1 v1.5 signatures are not broken ([see here](https://crypto.stackexchange.com/questions/51547/security-of-rsa-signature-using-pkcs-1-5)), the general recommendation for new applications seems to be PSS padding ([see here](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#cryptography.hazmat.primitives.asymmetric.padding.PKCS1v15)).

Instead of RSA, one could even use some elliptic curve signature scheme like ed25519 (recommended by [cryptography](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ed25519/)). In our case, using ed25519 instead of RSA would reduce the size of stored keys (2048 bits to only 256 bits).

The comment suggests that we cannot use PSS because of legacy(?) Androids. I guess ed25519 isn't better in this regard :-)

But what do you think about adding a "signature" field to the ``smartphone_data``, which currently says "rsa_sha256_pkcs1v15"? This would allow us to upgrade the signature scheme without too much hassle at some point in the future.

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,578 @@
...
+        """
+        from privacyidea.lib.token import get_tokens
+        from privacyidea.lib.utils import prepare_result
+        details = {}
+        result = False
+        serial = getParam(request.all_data, "serial", optional=False)
+
+        if serial and "fbtoken" in request.all_data and "pubkey" in request.all_data:
+            # Do the 2nd step of the enrollment
+            toks = get_tokens(serial=serial, rollout_state="clientwait")
```
fredreichbier: We could use ``token_obj = get_one_token(serial=serial, rollout_state="clientwait")`` here, which makes the ``len(toks) == 0`` test obsolete :-)

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,578 @@
...
+
+            details = token_obj.get_init_detail(init_detail_dict)
+            result = True
+        elif serial and "nonce" in request.all_data and "signature" in request.all_data:
+            challenge = getParam(request.all_data, "nonce")
+            serial = getParam(request.all_data, "serial")
+            signature = getParam(request.all_data, "signature")
+
+            # get the token_obj for the given serial:
+            toks = get_tokens(serial=serial)
```
fredreichbier: same as above: ``get_one_token(serial=serial)`` saves us some work.

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,578 @@
...
+            for challengeobject in challengeobject_list:
+                # check if we are still in time.
+                if challengeobject.is_valid():
+                    _, status = challengeobject.get_otp_status()
+                    if status is True:
+                        # create a positive response
+                        otp_counter = 1
+                        # delete the challenge, should we really delete the challenge? If we do so, the information
+                        # about the successful authentication could be fetched only once!
+                        # challengeobject.delete()
```
fredreichbier: Hm, I would have thought we do have to delete the challenge. :-) But I don't have a full understanding of the protocol yet.

privacyidea/lib/tokens/pushtoken.py
fredreichbier: After a good nights sleep, I realized that ``smartphone_data`` isn't the right place to add the signature algorithm field, it would need to be part of the enrollment process.

cornelinux: I like it. But I am not sure, if this should stop our current going forward.

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,578 @@
...
+        :param g: The Flask global object g
+        :return: dictionary
+        """
+        from privacyidea.lib.token import get_tokens
+        from privacyidea.lib.utils import prepare_result
+        details = {}
+        result = False
+        serial = getParam(request.all_data, "serial", optional=False)
+
+        if serial and "fbtoken" in request.all_data and "pubkey" in request.all_data:
```
fredreichbier: We need to check that the token with serial ``serial`` is actually a push token. Otherwise, the push API endpoint may be used to update non-push tokens without authentication.

privacyidea/lib/tokens/pushtoken.py
fredreichbier: I agree that this would mean some extra work now. But including a version marker at some place now may save us time and headaches in the future when we want to migrate to some other signature algorithm.

privacyidea/lib/smsprovider/SMSProvider.py
```diff
@@ -211,6 +213,8 @@ def get_smsgateway(identifier=None, id=None):
             sqlquery = sqlquery.filter_by(id=id)
         except Exception:
             log.info("We can not filter for smsgateway {0!s}".format(id))
+    if gwtype:
+        sqlquery = sqlquery.filter_by(providermodule=gwtype)
```
fredreichbier: I haven't tested it, but if Firebase gateways are implemented as SMS gateways, don't we need to ensure that they do not turn up as SMS gateways at other places? e.g. in ``SmsTokenClass.get_class_info``.

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,578 @@
...
+
+        :return: nothing
+        """
+        upd_param = {}
+        for k, v in param.items():
+            upd_param[k] = v
+
+        if "serial" in upd_param and "fbtoken" in upd_param and "pubkey" in upd_param:
+            # We are in step 2:
+            if not self.token.rollout_state == "clientwait":
```
fredreichbier: Hm, I find ``!=`` more readable than ``not ... == ``. What do you think?

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,578 @@
...
+                raise ParameterError("Invalid state! The token you want to enroll is not in the state 'clientwait'.")
+            if upd_param.get("enrollment_credential") != self.get_tokeninfo("enrollment_credential"):
```
fredreichbier: If the token, for some reason, doesn't have an enrollment credential, this check succeeds if no ``enrollment_credential`` parameter is passed. Should we maybe ensure that ``self.get_tokeninfo("enrollment_credential")`` is not None?

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,578 @@
...
+        elif "genkey" in upd_param:
+            # We are in step 1:
+            upd_param["2stepinit"] = 1
+            self.add_tokeninfo("enrollment_credential", geturandom(20, hex=True))
+            # We also store the firebase config, that was used during the enrollment.
+            self.add_tokeninfo(PUSH_ACTION.FIREBASE_CONFIG, param.get(PUSH_ACTION.FIREBASE_CONFIG))
+        else:
+            raise ParameterError("Invalid Parameters. Either provide (genkey) or (serial, fbtoken, pubkey).")
+
+        TokenClass.update(self, upd_param, reset_failcount)
```
fredreichbier: If I understand correctly, the ``otpkey`` parameter makes no sense for the push token. However, the underlying ``TokenClass.update`` will process it if it is given, which can have weird effects. Should we maybe throw an error if it is provided?

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,578 @@
...
+        response_detail['rollout_state'] = self.token.rollout_state
+
+        extra_data = {"enrollment_credential": self.get_tokeninfo("enrollment_credential")}
+        imageurl = params.get("appimageurl")
+        if imageurl:
+            extra_data.update({"image": imageurl})
+        if self.token.rollout_state == "clientwait":
+            # Get the values from the configured PUSH config
+            fb_identifier = params.get(PUSH_ACTION.FIREBASE_CONFIG)
+            firebase_configs = get_smsgateway(identifier=fb_identifier)
```
fredreichbier: As ``get_smsgateway`` also searches for regular SMS gateways, should we add a filter for firebase gateways here?

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,578 @@
...
+                pubkey_pem = "-----BEGIN PUBLIC KEY-----\n{0!s}\n-----END PUBLIC KEY-----".format(pubkey_pem.strip().replace(" ", "+"))
+            # Do the 2nd step of the authentication
+            # Find valid challenges
+            challengeobject_list = get_challenges(serial=serial, challenge=challenge)
+
+            if challengeobject_list:
+                # There are valid challenges, so we check this signature
+                for chal in challengeobject_list:
+                    # verify the signature of the nonce
+                    pubkey_obj = serialization.load_pem_public_key(str(pubkey_pem), default_backend())
```
fredreichbier: For Python 3 compatibility, we need to use ``to_bytes(pubkey_pem)`` here instead of ``str(...)`` -- the latter will produce a (unicode) string, which is not accepted by ``load_pem_public_key``.

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,578 @@
...
+                               "question": message_on_mobile,
+                               "serial": self.token.serial,
+                               "title": title,
+                               "url": url}
+            # Create the signature.
+            # value to string
+            sign_string = u"{nonce}|{url}|{serial}|{question}|{title}".format(**smartphone_data)
+
+            pem_privkey = self.get_tokeninfo(PRIVATE_KEY_SERVER)
+            privkey_obj = serialization.load_pem_private_key(str(pem_privkey), None, default_backend())
```
fredreichbier: We need to use ``to_bytes`` instead of ``str`` for Python 3 compatibility here.

tests/test_api_lib_policy.py
```diff
@@ -1313,6 +1314,39 @@ def test_01_sms_identifier(self):
...
+        req.User = User()
+        req.all_data = {
+            "type": "push"}
+        # In this case we have no reqistration_url. We will raise an exception
+        self.assertRaises(PolicyError, pushtoken_add_config, req, "init")
+
+        # Set a policy for the registration URL
+        set_policy(name="push_pol",
+                   scope=SCOPE.ENROLL,
+                   action="{0!s}=http://test,{1!s}=11".format(PUSH_ACTION.REGISTRATION_URL,
```
fredreichbier: This test fails for me because it cannot find ``REGISTRATION_URL``:
```
======================================================================
ERROR: test_22_push_registration_url (tests.test_api_lib_policy.PrePolicyDecoratorTestCase)
----------------------------------------------------------------------
  File "/home/fred/privacyidea/privacyidea/tests/test_api_lib_policy.py", line 1339, in test_22_push_registration_url
    action="{0!s}=http://test,{1!s}=11".format(PUSH_ACTION.REGISTRATION_URL,
AttributeError: type object 'PUSH_ACTION' has no attribute 'REGISTRATION_URL'
```

privacyidea/api/lib/prepolicy.py
```diff
@@ -1246,6 +1249,46 @@ def save_client_application_type(request, action):
...
+        registration_url = None
+        user_object = request.User
+        if user_object:
+            token_user = user_object.login
+            token_realm = user_object.realm
+            token_resolver = user_object.resolver
+        else:
+            token_realm = token_resolver = token_user = None
+        # Get the TTL and the Registration URL from the configs
+        if g:
```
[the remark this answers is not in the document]
cornelinux: Thats right. ``g`` is even imported in the module :-)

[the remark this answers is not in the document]
cornelinux: done in ff56dbec

privacyidea/lib/crypto.py
```diff
@@ -750,3 +750,31 @@ def generate_password(size=6, characters=string.ascii_lowercase +
...
     """
     return ''.join(urandom.choice(characters) for _x in range(size))
+
+
+def generate_keypair(rsa_keysize=2048):
+    """
+    This create a keypair, either RSA or ECC.
+    The HSM should be used.
+
+    # TODO: This must be much nicer...
```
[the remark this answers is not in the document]
cornelinux: will do

privacyidea/lib/smsprovider/SMSProvider.py
cornelinux: Hm, this is the question. Maybe not necessarily. It could be, that you also could use Firebase to send notifications  - like SMS. (with some slight adjustments)

tests/test_api_lib_policy.py
cornelinux: You are right. The registration URL in the policies is gone, since it is part of the firebase config.
Will do.

cornelinux: done in fd13ec74

privacyidea/lib/crypto.py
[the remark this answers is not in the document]
cornelinux: done in 79b5d2cb

privacyidea/lib/tokens/pushtoken.py
cornelinux: will do

privacyidea/lib/tokens/pushtoken.py
cornelinux: will do

privacyidea/lib/tokens/pushtoken.py
cornelinux: will do.

privacyidea/lib/tokens/pushtoken.py
cornelinux: Will do.

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in ee1593b9

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in ee1593b9

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in ee1593b9

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in ee1593b9

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in ee1593b9

privacyidea/lib/tokens/pushtoken.py
cornelinux: The thing is, that we keep the challenge for a while, so that we can fetch the information several times.
The old expired challenge gets deleted by the challenge janitor, when a new challenge is created for this very token.
https://github.com/privacyidea/privacyidea/blob/1342-pushtoken/privacyidea/lib/tokens/pushtoken.py#L530

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in afca29fe

privacyidea/lib/tokens/pushtoken.py
cornelinux: Well I think this should not happen, that the token has no enrollment credential, but will do.

cornelinux: done in c0c26deb

privacyidea/lib/tokens/pushtoken.py
cornelinux: As the otpkey is not used anywhere, I think it would not be necessary.

privacyidea/lib/tokens/pushtoken.py
cornelinux: will do.

cornelinux: done in 9036c0d7

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in c3f36915

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in c3f36915

tests/test_lib_tokens_push.py
```diff
@@ -102,9 +102,9 @@ def test_01_create_token(self):
                       "pubkey": self.smartphone_public_key_pem})
         self.assertEqual(token.get_tokeninfo("firebase_token"), "firebasetoken")
         self.assertEqual(token.get_tokeninfo("public_key_smartphone"), self.smartphone_public_key_pem)
```
fredreichbier: For Python 3 compatibility, I think we either need to convert ``token.get_tokeninfo("public_key_smartphone")`` (which is a unicode string) to a bytestring, or ``self.smartphone_public_key_pem`` (a bytestring) to unicode.

tests/test_lib_tokens_push.py
```diff
@@ -102,9 +102,9 @@ def test_01_create_token(self):
...
-        self.assertTrue(token.get_tokeninfo("public_key_server").startswith("-----BEGIN RSA PUBLIC KEY-----\nMIICC"),
+        self.assertTrue(token.get_tokeninfo("public_key_server").startswith(b"-----BEGIN RSA PUBLIC KEY-----\nMIICC"),
```
fredreichbier: I think we mustn't use bytestrings here, because tokeninfo values are always unicode strings. If we use bytestrings, the test fails under Python 3.

tests/test_lib_tokens_push.py
```diff
@@ -0,0 +1,378 @@
...
+
+        def check_firebase_params(request):
+            payload = json.loads(request.body)
+            # check the signature in the payload!
+            data = payload.get("message").get("data")
+
+            sign_string = u"{nonce}|{url}|{serial}|{question}|{title}".format(**data)
+            token_obj = get_tokens(serial=data.get("serial"))[0]
+            pem_pubkey = token_obj.get_tokeninfo(PUBLIC_KEY_SERVER)
+            pubkey_obj = load_pem_public_key(str(pem_pubkey), backend=default_backend())
```
fredreichbier: This should be ``to_bytes`` instead of ``str``. Otherwise, it fails under Python 3 because ``str(pem_pubkey)`` is a (unicode) string, and ``load_pem_public_key`` expects a bytestring :-)

tests/test_lib_tokens_push.py
cornelinux: I can check this. But this whole branch will currently not run under python3, since it is based on a rather old version of master.

tests/test_lib_tokens_push.py
cornelinux: done in 9e0e477d

tests/test_lib_tokens_push.py
cornelinux: done in 9e0e477d

tests/test_lib_tokens_push.py
cornelinux: done in 9e0e477d

privacyidea/lib/tokens/pushtoken.py
cornelinux: will do

cornelinux: We will add "v=1" to the QR Code and the smartphone will check, that it is not v=2.

cornelinux: done in 21b07b3f

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,582 @@
...
+            enrollment_credential = getParam(upd_param, "enrollment_credential", optional=False)
+            if enrollment_credential != self.get_tokeninfo("enrollment_credential"):
+                raise ParameterError("Invalid enrollment credential. You are not authorized to finalize this token.")
+            self.del_tokeninfo("enrollment_credential")
+            self.token.rollout_state = "enrolled"
+            self.token.active = True
+            self.add_tokeninfo(PUBLIC_KEY_SMARTPHONE, upd_param.get("pubkey"))
+            self.add_tokeninfo("firebase_token", upd_param.get("fbtoken"))
+            # create a keypair for the server side.
+            from privacyidea.lib.crypto import generate_keypair
```
fredreichbier: Can we move this import to the top?

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,582 @@
...
+
+            # Sign the data with PKCS1 padding. Not all Androids support PSS padding.
+            signature = privkey_obj.sign(sign_string.encode("utf8"),
+                                         padding.PKCS1v15(),
+                                         hashes.SHA256())
+            smartphone_data["signature"] = b32encode_and_unicode(signature)
+
+            res = fb_gateway.submit_message(self.get_tokeninfo("firebase_token"), smartphone_data)
+            if not res:
+                from privacyidea.lib.error import ValidateError
```
fredreichbier: Can we move this import to the top?

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,582 @@
...
+                raise ValidateError("Failed to submit message to firebase service.")
+        else:
+            log.warning(u"The token {0!s} has no tokeninfo {1!s}. "
```
fredreichbier: Should we maybe exit from the function in this case?

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -0,0 +1,582 @@
...
+                    sign_data = u"{0!s}|{1!s}".format(challenge, serial)
+                    try:
+                        pubkey_obj.verify(b32decode(signature),
+                                          sign_data.encode("utf8"),
+                                          padding.PKCS1v15(),
+                                          hashes.SHA256())
+                        # The signature was valid
+                        chal.set_otp_status(True)
+                        result = True
+                    except Exception as e:
```
fredreichbier: Hm sorry, one more thing: I guess it would be a good idea to only catch ``InvalidSignature`` here? We don't notice if anything else goes wrong otherwise (e.g. the signature was valid, but ``set_otp_status(True)`` fails).

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in 30c5cd21

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in 30c5cd21

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in 7544a675

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in 0550f7f3

PR #1503, opened 2019-03-14
Add oauth depenency

cornelinux:
We need the python-oauth2client for the Firebase SMS provder a.k.a. Push
Token.

This is also available on Ubuntu 16.04 for python 2.7.
The version in Ubuntu 14.04 is too old.
But users with 14.04 should upgrade to 16.04 anyways.


fredreichbier:
Looks good! But could you add ``oauth2client`` also to ``requirements.txt`` and setup.py, such that the unit tests on Travis can run?

cornelinux:
Done in 14680cdb

fredreichbier:
Thanks!

Issue #1511, opened 2019-03-19
Certificate enrollment after token validation 


Voyrx:
**Is your feature request related to a problem? Please describe.**
What are you trying to achieve?
I want to enroll a certificate token after a successful authentication.

**Describe the solution you'd like**
I used the "event" object to enroll a certificate after successful authentication. And then add an API in ttype/certificate to get the last certificate generated for a user.

**Describe alternatives you've considered**
N/A

**Implementation 
In order to get this works, I changed the files below : 
- lib/eventhandler/tokenhandler.py
- lib/tokens/certificatetoken.py
- lib/error.py
I will pull a request and link it to that post.

cornelinux:
If you want a user to enroll a certificate if he can present an OTP value you may be considering this approach:

* The user needs to authenticate against privacyIDEA (login policy: login mode=privacyidea)
* The user calls /auth endpoint and authenticates with OTP
* The user is allowed to enroll certificates and
  * enrolls a certificate using the /token/init endpoint and a ``request``.

This is a quite normal workflow, that can be added to an application - currently I do not see the need to use event handlers for this.

If however we would want to enhance the functionality of the tokenhandler with the ``init``/``enroll`` action, we might want to choose a rather generic approach.

cornelinux:
Requesting a certificate from privacyIDEA *after* the user successfully authenticated with PIN and OTP is actually also something that Microsoft suggested for a OTP credential provider, which should receive a short living logon certificate from the OTP server to do a kerberos PKIINIT against the domain controller.

cornelinux:
I think we can close this issue. Enrolling a certificate can be done with ``/token/init?type=certificate`` after a successful ``/validate/check``.
But we can discuss certificate handling internally.

Voyrx:
Hello Cornelius,

Yes we can, thank you for help :) 
Regards,

NuvandaPV:
If not implemented, this topic should at least be discussed in milestone 3.3

cornelinux:
Hm, looking at this again I am wondering, what we want to achieve.
After a successful auth request or validate/check we can enroll a certificate with an event handler. But what should happen with the certificate, then?
I am currently not completely aware of the workflow.

Also if there is a more complex workflow, we could also trigger a script after the auth request.
We could add something to https://github.com/privacyidea/scripts.
This script could also enroll a certificate and also go on with the certificate handling.

@NuvandaPV You suggested a discussion in 3.3, what do you have in mind?

[...]

Issue #1510, opened 2019-03-19
Can't revoke certificate token


Voyrx:
# What did you try to do?

I can't revoke certificate token. 

# What outcome did you expect?

Revoke the certificate token with Openssl commands.

# What outcome did you experience?

When clicking on "revoke" button, an alert message is displayed on the screen : 
"ERR503: Using configuration from /etc/privacyidea/certificates/openssl.cnf Can't open 41.pem for reading, No such file or directory 139669700621312:error:02001002:system library:fopen:No such file or directory:crypto/bio/bss_file.c:72:fopen('41.pem','r') 139669700621312:error:2006D080:BIO routines:BIO_new_file:no such file:crypto/bio/bss_file.c:79: unable to load certificate "

# Configuration

* privacyIDEA Version: 2.23.3

* Installation method:
  (from source, github, virtualenv, ubuntu packages...)
Privacyidea-nginx using commande : "apt-get install privacyidea-nginx"

## more details: 

* OS: Ubuntu 18

* Webserver: Nginx

* Tokendatabase: (like MySQL, PostgreSQL, ...) : MySQL

# Log file

**Turn on debug level and take a look at the privacyidea.log!**
**If appropriate, attach the log file.**

## Correction 

I corrected the file privacyidea/lib/caconnectors/localca.py to get the path to certificates directory from configuration and then revoke the certificate. 
It works fine like that : 

```
cd ./privacyidea/lib/caconnectors
diff localca.py localcaORI.py
507c507
<         # Custom dev : Bug detected, path to certificate was not mentioned.
---
>
510c510
<                                certificate=self.config.get(ATTR.CERT_DIR)+'/'+filename,
---
>                                certificate=filename,
```

cornelinux:
Interesting, @Voyrx 

Could it be that you have different "WorkingDir" and "CertificateDir" in your configuration?
We probably had the same directory and thus did not stumble over this.

Voyrx:
Hello M. Kolbel,

Yes exactly.

I did have different working directory and certificates directory.

I pushed a pull request with the correction, I may have done this a little bit "ugly", it took also the code for #1511 ticket 🙁 , sorry.
________________________________
De : Cornelius Kölbel <notifications@github.com>
Envoyé : mardi 19 mars 2019 17:04
À : privacyidea/privacyidea
Cc : Tarek EL ALLAM; Mention
Objet : Re: [privacyidea/privacyidea] Can't revoke certificate token (#1510)


[...]

PR #1512, opened 2019-03-19
tickets #1510 and #1511

Voyrx:
Pull request related to https://github.com/privacyidea/privacyidea/issues/1510

### DIFF
--- a/privacyidea/lib/eventhandler/tokenhandler.py
+++ b/privacyidea/lib/eventhandler/tokenhandler.py
@@ -391,6 +391,9 @@ class TokenEventHandler(BaseEventHandler):
                                         "has no email address!".format(user))
                 elif tokentype == "motp":
                     init_param['motppin'] = handler_options.get("motppin")
+                    
+                elif tokentype == "certificate":
+                    init_param['request'] = request.all_data.get("request")
 
             t = init_token(param=init_param, user=user)
             log.info("New token {0!s} enrolled.".format(t.token.serial))
--- a/privacyidea/lib/tokens/certificatetoken.py
+++ b/privacyidea/lib/tokens/certificatetoken.py
@@ -38,11 +38,16 @@ from privacyidea.lib.tokenclass import TokenClass
 from privacyidea.lib.log import log_with
 from privacyidea.api.lib.utils import getParam
 from privacyidea.lib.caconnector import get_caconnector_object
-from privacyidea.lib.user import get_user_from_param
+#from privacyidea.lib.user import get_user_from_param
 from OpenSSL import crypto
 from privacyidea.lib.decorators import check_token_locked
 import base64
 from privacyidea.lib import _
+
+from privacyidea.lib.user import get_user_from_param, User, get_user_info
+from privacyidea.lib.token import get_tokens
+from privacyidea.lib.error import NoCertificateAvailableError
+import json
 
 optional = True
 required = False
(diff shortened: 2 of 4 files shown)


cornelinux:
Thanks a lot for the pull request, but I think we should split this in several PRs.

Please provide a PR for the fix of the certificate directory in the localca. This will be easy to merge.

I will add some further questions in your code.

privacyidea/lib/eventhandler/tokenhandler.py
```diff
@@ -391,6 +391,9 @@ def do(self, action, options=None):
                                         "has no email address!".format(user))
                 elif tokentype == "motp":
                     init_param['motppin'] = handler_options.get("motppin")
+                    
+                elif tokentype == "certificate":
+                    init_param['request'] = request.all_data.get("request")
```
cornelinux: What is your intention with enrolling a certificate.
Where does actually the "request" parameter come from, which is usually not contained in the originial http request.

The problem of this approach is, that the originial https request needs to know, that there is an event handler configured, that needs an additional parameter. 
We should think about a more generic way to solve this.

privacyidea/lib/tokens/certificatetoken.py
```diff
@@ -334,3 +339,79 @@ def revoke(self):
...
+    
+    @staticmethod
+    def api_endpoint(request, g):
+        """
+        Author : Tarek EL ALLAM
+        This provides a function to be plugged into the API endpoint
+        /ttype/<tokentype> which is defined in api/ttype.py
+        See :ref:`rest_ttype`.
+        Desc : 
+        Add API call in order to get the last generated certificate assigned, activated, unrevoked and unlocked.
```
cornelinux: actually we already have such an endpoint, it is the ``GET /token/`` endpoint, which needs an administrator or a user to authenticate. But you actually can have the same result.
This approach leaves the information, which certificate the user has, open to the world without any authentication.
Why would you do this?
Would you be able to have a service account to receive this information?

Voyrx:
Hello,

The "request" corresponds to the certificate signing request, I used it like that because I noticed that it's a native parameter for GUI certificate enrollement :
[cid:2956fb3e-b50b-44e3-922c-50c1766391ac]

But of course, if you think that it's not beautiful, I can adapt the solution. 😄

Thank you for you quick answer.

Regards,
Tarek EL ALLAM

________________________________
De : Cornelius Kölbel <notifications@github.com>
Envoyé : mardi 19 mars 2019 23:12
À : privacyidea/privacyidea
Cc : Tarek EL ALLAM; Author
Objet : Re: [privacyidea/privacyidea] tickets #1510 and #1511 (#1512)


@cornelinux commented on this pull request.

________________________________

In privacyidea/lib/eventhandler/tokenhandler.py<https://github.com/privacyidea/privacyidea/pull/1512#discussion_r267119080>:

> @@ -391,6 +391,9 @@ def do(self, action, options=None):
                                         "has no email address!".format(user))
                 elif tokentype == "motp":
                     init_param['motppin'] = handler_options.get("motppin")
+
+                elif tokentype == "certificate":
+                    init_param['request'] = request.all_data.get("request")


What is your intention with enrolling a certificate.
Where does actually the "request" parameter come from, which is usually not contained in the originial http request.

The problem of this approach is, that the originial https request needs to know, that there is an event handler configured, that needs an additional parameter.
We should think about a more generic way to solve this.

PR #1524, opened 2019-03-22
Change Yubico HTTP POST request to GET Req

### LINKED ISSUE #1500: Change yubico POST request to GET request
See: https://community.privacyidea.org/t/yubico-auth-not-working-since-a-few-hours-quickfix/977/3

We need to change the POST request to the yubicloud to a GET request.

*bummer*

cornelinux:
Fixes #1500


cornelinux:
The Yubico token does not only authenticate against the originial Yubico-Cloud, but it could also authenticate against a locally running yubico-server. I am not sure, if s.o. configured or tweeked his locally running yubico-server to (only) accept authentication via POST requests.

Issue #1505, opened 2019-03-16
passlib.utils.binary not available on Ubuntu 16.04


cornelinux:
The SQL Resolver fails to load on ubuntu 16.04, since Ubuntu 16.04 comes with passlib v 1.6.5.

This does not contain ``passlib.utils.binary``. So we can not import ``h64``.
Please check, if we can avoid to use h64.

cornelinux:
I would wait for this.

We have a lot of lot of issues on 16.04 (also #1525 and als RSA-sign/cryptography problems we have not filed, yet)
We should discuss next week, what we do in regards to this.

plettich:
already on it.

cornelinux:
Same problem as in #1525. MAybe get rid of it generically.

cornelinux:
We go for virtualenv!

PR #1527, opened 2019-03-25
Fix passlib and flask-script usage for ubuntu

### LINKED ISSUE #1525: Revert Command() in pi-manage
We need to revert the Command calls in pi-manage, since they are too new.

     ca_manager.add_command('list', Command(list_ca))

plettich:
Ubuntu 16.04 contains passlib in version 1.6.5 and flask-script in version 0.6.7
In order to support the ubuntu packages, we need to change some code:
- in `pi-manage` the 'list' commands need to be encapsulated in their own
  command classes.
- The function to calculate a password hash is called `encrypt()` in
  passlib 1.6.5 and 1.7.1 (in 1.7.1 it is marked as deprecated)
- passlib 1.6.5 does not contain a `binary` utility

Fixes #1525 #1505


plettich:
Does it make sense to pin the versions in `requirements.txt` to the module versions available in Ubuntu 16.04?

plettich:
For future versions we go for a virtualenv package and can ignore broken backwards compatibility of system packages.

PR #1532, opened 2019-03-26
security update for pyyaml

cornelinux:


requirements.txt
```diff
@@ -57,7 +57,7 @@ python-dateutil==2.7.3
 python-gnupg==0.4.3
 pytz==2018.4
 pyusb==1.0.2
-PyYAML==3.12
+PyYAML==4.2b1
```
plettich: Since we are not affected by this vulnerability (we only use `yaml.safe_load()`) maybe we should switch to the next regular release (5.1). All tests run fine with this version.

cornelinux: Perfect, will do.

cornelinux: done in  4cd040df

PR #1536, opened 2019-03-26
Allow to not check SSL for push token

### LINKED ISSUE #1534: sslverify for PUSH token enrollment
The push token enrollment should be able to tell the smartphone app if the SSL should be verified or not. 
This is done by adding a parameter ``sslverify=0/1``.
This is also part of the signature.

default is "1", verify SSL.

We need a policy to deactivate this.

cornelinux:
An enrollment policy and auth policy can define,
if the smartphone does not need to verify the SSL
certificte.

Closes #1534

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -1299,6 +1299,8 @@ def pushtoken_add_config(request, action):
         )
         if len(ssl_verify) == 1:
             request.all_data[PUSH_ACTION.SSL_VERIFY] = list(ssl_verify)[0]
+        else:
+            request.all_data[PUSH_ACTION.SSL_VERIFY] = "1"
 
 
 def u2ftoken_verify_cert(request, action):
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -1254,7 +1254,7 @@ def pushtoken_add_config(request, action):
     This is a token specific wrapper for push token for the endpoint
     /token/init
     According to the policy scope=SCOPE.ENROLL,
-    action=TTL or action=REGISTRATION_URL the parameters are added to the
+    action=SSL_VERIFY_NO or action=FIREBASE_CONFIG the parameters are added to the
     enrollment step.
     :param request:
     :param action:
@@ -1271,7 +1271,7 @@ def pushtoken_add_config(request, action):
             token_resolver = user_object.resolver
         else:
             token_realm = token_resolver = token_user = None
-        # Get the TTL and the Registration URL from the configs
+        # Get the firebase configuration from the policies
         firebase_config = g.policy_object.get_action_values(
             action=PUSH_ACTION.FIREBASE_CONFIG,
             scope=SCOPE.ENROLL,
@@ -1286,6 +1286,19 @@ def pushtoken_add_config(request, action):
         else:
             raise PolicyError("Missing enrollment policy for push token: {0!s}".format(PUSH_ACTION.FIREBASE_CONFIG))
 
+        # Get the sslverify definition from the policies
+        ssl_verify = g.policy_object.get_action_values(
+            action=PUSH_ACTION.SSL_VERIFY,
+            scope=SCOPE.ENROLL,
+            realm=token_realm,
+            user=token_user,
+            resolver=token_resolver,
+            client=g.client_ip,
+            audit_data=g.audit_object.audit_data,
+            unique=True
+        )
+        if len(ssl_verify) == 1:
+            request.all_data[PUSH_ACTION.SSL_VERIFY] = list(ssl_verify)[0]
 
 
 def u2ftoken_verify_cert(request, action):
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -1254,7 +1254,7 @@ def pushtoken_add_config(request, action):
     This is a token specific wrapper for push token for the endpoint
     /token/init
     According to the policy scope=SCOPE.ENROLL,
-    action=SSL_VERIFY_NO or action=FIREBASE_CONFIG the parameters are added to the
+    action=SSL_VERIFY or action=FIREBASE_CONFIG the parameters are added to the
     enrollment step.
     :param request:
     :param action:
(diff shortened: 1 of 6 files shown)


privacyidea/api/lib/prepolicy.py
```diff
@@ -1254,7 +1254,7 @@ def pushtoken_add_config(request, action):
     This is a token specific wrapper for push token for the endpoint
     /token/init
     According to the policy scope=SCOPE.ENROLL,
-    action=TTL or action=REGISTRATION_URL the parameters are added to the
+    action=SSL_VERIFY_NO or action=FIREBASE_CONFIG the parameters are added to the
```
plettich: Did You mean just `SSL_VERIFY`?

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -217,6 +218,12 @@ def get_class_info(key=None, ret='all'):
                            'group': "PUSH",
                            'value': [gw.identifier for gw in gws]
                        },
+                       PUSH_ACTION.SSL_VERIFY: {
+                           'type': 'str',
```
plettich: Why not use `bool` here?

cornelinux: I thought about this first. But the defaul behaviour should be that SSL is verified. If we do a "bool" we can only have a policy, that it is NOT verified. This did not look nice. Also from a code perspective I was a bit lazy to check for a str in push token and check for a bool. So I only have string configs in the push token.

privacyidea/api/lib/prepolicy.py
cornelinux: Yes, this is a remnant of my first bool try.
I will fix this.

cornelinux: done in 2637050c

privacyidea/api/lib/prepolicy.py
```diff
@@ -1286,6 +1286,19 @@ def pushtoken_add_config(request, action):
...
+            scope=SCOPE.ENROLL,
+            realm=token_realm,
+            user=token_user,
+            resolver=token_resolver,
+            client=g.client_ip,
+            audit_data=g.audit_object.audit_data,
+            unique=True
+        )
+        if len(ssl_verify) == 1:
+            request.all_data[PUSH_ACTION.SSL_VERIFY] = list(ssl_verify)[0]
```
fredreichbier: I haven't tested this, but in case no ``push_ssl_verify`` policy is defined, doesn't this allow the user to disable SSL verification by passing a ``push_ssl_verify=0`` parameter to ``/token/init``?
I think it would be nicer to only allow disabling SSL verification if a corresponding policy exists.

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -309,6 +323,7 @@ def get_init_detail(self, params=None, user=None):
         user = user or User()
         tokenlabel = params.get("tokenlabel", "<s>")
         tokenissuer = params.get("tokenissuer", "privacyIDEA")
+        sslverify = params.get(PUSH_ACTION.SSL_VERIFY, "1")
```
fredreichbier: I think it would be safer to add a check that ``sslverify`` is either "1" or "0". Otherwise, we add whatever is passed as ``push_ssl_verify`` to the query string below :)

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -474,6 +490,10 @@ def create_challenge(self, transactionid=None, options=None):
                                                  ACTION.CHALLENGETEXT,
                                                  options) or DEFAULT_CHALLENGE_TEXT
 
+        sslverify = get_action_values_from_options(SCOPE.AUTH,
+                                                   PUSH_ACTION.SSL_VERIFY,
+                                                   options) or "1"
+
```
fredreichbier: Again, I think it would be nice to check that ``sslverify`` is really either ``"1"`` or ``"0"``.

privacyidea/api/lib/prepolicy.py
cornelinux: ok, so we need to add an else-branch

~~~~python
else:
    request.all_data[PUSH_ACTION.SSL_VERIFY] = 1
~~~~
right?

fredreichbier: Yup, sounds good!

privacyidea/lib/tokens/pushtoken.py
cornelinux: OK

privacyidea/lib/tokens/pushtoken.py
cornelinux: OK

privacyidea/api/lib/prepolicy.py
cornelinux: done in f1ae0319

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in f1ae0319

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in f1ae0319

privacyidea/api/lib/prepolicy.py
fredreichbier: Thanks!

privacyidea/lib/tokens/pushtoken.py
fredreichbier: Heh, using ``getParam`` like that is clever, but in my opinion it's not very readable :-/ Also, I'm not sure if it makes sense to add the ``allowed_values`` feature to ``getParam`` just for this occasion.
What do you think about just adding a check
```python
if sslverify not in ("0", "1"):
    raise ParameterError("Invalid value for sslverify: {!r}".format(sslverify))
```
for both occasions of ``sslverify``? I realize it's a bit boilerplate, but I think it's more readable. What do you think?

cornelinux: It is more readable. And actually I already had added small function in lib/utils.py to do this - including tests. But then I though it makes more sense in the parameters, especially since this might not be the last API parameter, which we should check for valid values!

privacyidea/lib/tokens/pushtoken.py
plettich: Isn't [`is_true()`](https://github.com/privacyidea/privacyidea/blob/d1c7b37eca1f277658e68bedf9a256f206393c41/privacyidea/lib/utils.py#L725) for that?

cornelinux: No. is_true comes from the different ways a database acutally stores a boolean value in the database. In this case we do not care about database and privacyIDEA.
In this case it is important how we communitcate with the smartphone App.

PR #1541, opened 2019-03-28
Update makefile for 3

cornelinux:


Makefile
```diff
@@ -81,56 +73,6 @@ doc-man:
...
-
-builddeb:
-	make debianize
-	################## Renew the changelog
-	cp -r deploy/debian-ubuntu/* DEBUILD/privacyidea.org/debian/
-	sed -e s/"trusty) trusty; urgency"/"$(LOCAL_SERIES)) $(LOCAL_SERIES); urgency"/g deploy/debian-ubuntu/changelog > DEBUILD/privacyidea.org/debian/changelog
-	################# Build
-	(cd DEBUILD/privacyidea.org; debuild --no-lintian)
 
 venvdeb:
```
plettich: shouldn't this be removed as well? The `debianize` target is already removed.

cornelinux: The venvdeb is currently used for building debian packages as basis for UCS. 
But I think you are right, we should remove this, too, so that we have actually remove *all* distro dependent builds from the source tree. I think this makes it more clear.
Will do

PR #1546, opened 2019-03-29
Increase failcounter in case of wrong response to challenge

### LINKED ISSUE #1539: failcounter in challenge response is not increased
When authenticating with a token in challenge-response mode the failcounter when presenting a wrong OTP value is not increased.

If there is a single transaction_id we can determine the serial number, add the serial number to the audit log and also increase the failcounter of this very token.

If the transaction_id belongs to several tokens, then all failcounters should be increased.
No serial would be written to the audit log in this case.

cornelinux:
Fixes #1539


plettich:
Should we add this to 2.23 as well?

cornelinux:
I currently do not think so.

PR #1548, opened 2019-04-01
Remove last remnants of PyCrypto

plettich:
- `pi-manage` now uses cryptography to create the RSA-keys for signing audit
  entries
- remove `privacyidea-decrypt-safeword.py` (there is no feasible solution
  to calculate (single-)DES
- the import of audit signing keys is now done in the audit base class
- the PI_CHECK_OLD_SIGNATURES configuration is now read only during
  initialization of the audit class
- remove PyCrypto from `setup.py` and `requirements.txt`
- add some more description


cornelinux:
@plettich, cool! Thank you!
Would you check and close the corresponding issues?

PR #1551, opened 2019-04-02
Fix requirements for virtualenv

plettich:
(re)moved some requirements


requirements.txt
```diff
@@ -1,69 +1,44 @@
...
 Flask-Babel==0.11.2
 Flask-Migrate==2.1.1
 Flask-Script==2.0.6
 Flask-SQLAlchemy==2.3.2
-funcparserlib==0.3.6
-gunicorn==19.8.1
 huey==1.11.0
 idna==2.6
-imagesize==1.0.0
-ipaddress==1.0.22
```
cornelinux: Oh yes. We use netaddr module.

requirements.txt
```diff
@@ -1,69 +1,44 @@
...
 oauth2client>=2.0.1
 passlib==1.7.1
 Pillow==5.1.0
 psycopg2>=2.6
-py==1.5.3
 pyasn1==0.4.3
 pycparser==2.18
 PyJWT==1.6.4
 PyMySQL==0.8.1
-pymysql-sa==1.0
```
cornelinux: I think we need this for sqlalchemy when running in a virtualenv on a productive system with MySQL.

requirements.txt
```diff
@@ -1,69 +1,44 @@
...
-pytest==3.6.1
-pytest-cov==2.5.1
-pytest-runner==4.2
 python-dateutil==2.7.3
 python-gnupg==0.4.4
 pytz==2018.4
 pyusb==1.0.2
 PyYAML==5.1
 qrcode==6.0
-redis==2.10.6
```
fredreichbier: Ah I think we actually need redis to use the huey-based queue.
I didn't know, but apparently redis is only an extra requirement of huey :)

requirements.txt
plettich: i tested with `mysql-python` and `pymysql` under python 2 and python 3. We don't need this package anymore.

requirements.txt
plettich: so it doesn't get installed by huey?

cornelinux: Wait - as it is the requirements, we will add this, even if it would get installed by huey. The requirements contain *all* packages.

requirements.txt
cornelinux: Perfect! THanks!

requirements.txt
fredreichbier: I *thought* redis is a direct dependency of huey. But as it turns out, it is technically only an optional dependency, so it doesn't get installed automatically when we install huey :)

requirements.txt
```diff
@@ -1,69 +1,44 @@
-alabaster==0.7.10
```
fredreichbier: I just noticed one missing package: ``setup.py`` requires ``argparse``, which we apparently need for some scripts in ``tools/``. Should we add pin a specific version in requirements.txt as well?

cornelinux: imho yes please

requirements.txt
```diff
@@ -1,69 +1,44 @@
...
 alembic==0.9.9
 Babel==2.6.0
 bcrypt==3.1.4
 beautifulsoup4==4.6.0
 cffi>=1.7.0
```
fredreichbier: Ah, sorry, one more thing :) I think it would be nice if we could replace the ``>=`` conditions with specific versions -- what do you think?

requirements.txt
plettich: It looks like `argparse` is part of the python base distribution, so we don't need to add it explicitly.

fredreichbier: You're right, I missed this. Should we then just remove it from ``setup.py``?

plettich: on it.

requirements.txt
plettich: fixed in d4ae1646

requirements.txt
plettich: fixed in d4ae1646

requirements.txt
plettich: fixed in d4ae1646

requirements.txt
fredreichbier: Thanks!

requirements.txt
```diff
@@ -1,69 +1,56 @@
...
-mock==2.0.0
+MarkupSafe==1.1.1
 netaddr==0.7.19
-nose==1.3.7
-numpy==1.14.3
-oauth2client>=2.0.1
+oauth2client==2.0.1
 passlib==1.7.1
 Pillow==5.1.0
-psycopg2>=2.6
```
cornelinux: I would like to have psycopg2 in the ubuntu packages.
To my understanding this would have to be compiled if installed manually in a virtualenv?
So we obviously should patch this into ubuntu packages but not have this in the requirements.txt?

PR #1556, opened 2019-04-03
Add installation of requirements.txt during upgrade

### LINKED ISSUE #1555: Modify privacyidea-pip-update to install pinned versions from requirements.txt
We should change the behavior of privacyidea-pip-update such that it installs the pinned versions from the requirements.txt of the latest privacyIDEA version.

See #1127 and #1554

cornelinux:
Closes #1555

### DIFF
--- a/tools/privacyidea-pip-update
+++ b/tools/privacyidea-pip-update
@@ -23,27 +23,11 @@ privacyidea-pip-update
     """)
 
 
-def get_installed_packages():
-    """
-    Get a list of names of all locally installed packages by
-    parsing the output of ``pip freeze --local --all``.
-    """
-    packages = []
-    pip_output = check_output(PIP_EXECUTABLE + ["freeze", "--local", "--all"])
-    for line in pip_output.splitlines():
-        line = line.strip()
-        if b"==" in line:
-            package_name = line.split(b"==", 1)[0]
-            packages.append(package_name)
-    return packages
-
-
-def update():
-    for package_name in get_installed_packages():
-        call(PIP_EXECUTABLE + ["install", "--upgrade", package_name])
-    # Now we need again to upgrade privacyIDEA, so that we get pinned
-    # dependencies like ldap3 or pyasn1. This call will downgrade packages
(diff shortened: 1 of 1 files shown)


tools/privacyidea-pip-update
```diff
@@ -23,27 +23,11 @@ privacyidea-pip-update
...
-def update():
-    for package_name in get_installed_packages():
-        call(PIP_EXECUTABLE + ["install", "--upgrade", package_name])
-    # Now we need again to upgrade privacyIDEA, so that we get pinned
-    # dependencies like ldap3 or pyasn1. This call will downgrade packages
+def update(environment):
     call(PIP_EXECUTABLE + ["install", "--upgrade", "privacyidea"])
+    # Now we need to run the new requirements of privacyidea
+    # dependencies like ldap3 or pyasn1. This call could downgrade packages
+    call(PIP_EXECUTABLE + ["install", "-r", "{0!s}/lib/privacyidea/requirements.txt".format(environment)])
```
fredreichbier: Looks good, but could you add a call of ``pipes.quote()`` here to quote the directory name (like in ``update_db_schema`` below)?

cornelinux: This is not necessary!

The pipes.quote() is used when the call() statement gets one single string with ``shell=True``.

However, we pass in this call a list!

    ["python", "-m", "pip", "install", "-r", "/directory with blanks/requirements.txt"]

Using the list takes care of blanks.
I successfully ran this script on a virtualenv at "/opt/privacy IDEA":

~~~~python
import pipes
from subprocess import call, check_output
import getopt
import sys
import os
from six.moves import input

PIP_EXECUTABLE = [sys.executable, '-m', 'pip']
environment = os.environ.get("VIRTUAL_ENV")
call(PIP_EXECUTABLE + ["install", "-r", "{0!s}/lib/privacyidea/requirements.txt".format(environment)])
~~~~

fredreichbier: Whoops! You're totally right :-)

PR #1563, opened 2019-04-08
Fix Unicode error in audit log

### LINKED ISSUE #1560: UnicdeDecodeError on Ubuntu 16.04 in audit log
A 3.0dev8 has an LDAP resolver talking to 2012.
A user has a german umlaut.

In this method we get a UnicodeDecodeError:
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/auditmodules/sqlaudit.py#L310

indeed it is due to ``le.user``.

plettich:
When using mysql with the audit log, there is an issue if one of the
columns of an entry contains non ascii symbols.
To avoid writing invalid data to mysql, the column types in the SQLAlchemy
model need to be set to `Unicode`.
Some possible regressions need to be caught when updating the code.

fixes #1560

### DIFF
--- a/privacyidea/lib/auditmodules/sqlaudit.py
+++ b/privacyidea/lib/auditmodules/sqlaudit.py
@@ -250,6 +250,8 @@ class Audit(AuditBase):
                 self.session.merge(le)
                 self.session.commit()
         except Exception as exx:  # pragma: no cover
+            # in case of a Unicode Error in _log_to_string() we won't have
+            # a signature, but the log entry is available
             log.error("exception {0!r}".format(exx))
             log.error("DATA: {0!s}".format(self.audit_data))
             log.debug("{0!s}".format(traceback.format_exc()))
@@ -266,10 +268,10 @@ class Audit(AuditBase):
         the given id.
         
         TODO: We can not check at the moment if the first or the last entries
-        were deleted. If we want to do this, we need to store some signed
-        meta information
-        1. Which one was the first entry. (use initialize_log)
-        2. Which one was the last entry.
+              were deleted. If we want to do this, we need to store some signed
+              meta information:
+              1. Which one was the first entry. (use initialize_log)
+              2. Which one was the last entry.
         """
         res = False
         try:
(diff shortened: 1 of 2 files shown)


privacyidea/lib/auditmodules/sqlaudit.py
```diff
@@ -469,9 +479,16 @@ def clear(self):
...
+            try:
+                sig = self.sign_object.verify(self._log_to_string(audit_entry),
+                                              audit_entry.signature,
+                                              self.verify_old_sig)
+            except UnicodeDecodeError as _e:
+                # TODO: Unless we trace and eliminate the broken unicode in the
+                #  audit_entry, we will get issues when packing the response.
+                log.warning('Could not verify log entry! We get invalid values '
+                            'from the database, please check the encoding.')
+                log.debug('{0!s'.format(traceback.format_exc()))
```
cornelinux: There is a curly closing bracked missing.

privacyidea/lib/auditmodules/sqlaudit.py
```diff
@@ -405,14 +407,22 @@ def search(self, search_dict, page_size=15, page=1, sortorder="asc",
         auditIter = self.search_query(search_dict, page_size=page_size,
                                       page=page, sortorder=sortorder,
                                       timelimit=timelimit)
-        try:
-            le = next(auditIter)
-            while le:
```
cornelinux: Why did you change the ``while le`` loop to a ``while True`` loop?
Would there be an issue with having a "double" loop condition of "while le" and "StopIteration"?

plettich: My first approach was to catch the error for each individual `auditIter` entry. Unfortunately the first call to `next(auditIter)` throws the error _and_ stops the iteration.
And while looking through examples to call `next(auditIter)` only once, the  `while True:` loop seemed to be the most pythonic way.

privacyidea/lib/auditmodules/sqlaudit.py
plettich: fixed in 7484253d

PR #1576, opened 2019-04-16
enhance pi-manage to allow additional radius directory

### LINKED ISSUE #1575: pi-manage backup should be able to backup the freeradius config
``pi-manage backup`` should also backup the freeradius configuration.

We should check if we add a switch/paramter so that it works nice with the pi-appliance.

cornelinux:
Closes #1575

### DIFF
--- a/pi-manage
+++ b/pi-manage
@@ -247,12 +247,16 @@ def change(username, email=None, password_prompt=False):
 @backup_manager.command
 def create(directory="/var/lib/privacyidea/backup/",
            conf_dir="/etc/privacyidea/",
+           radius_directory=None,
            enckey=False):
     """
     Create a new backup of the database and the configuration. The default
     does not include the encryption key. Use the 'enckey' option to also
     backup the encryption key. Then you should make sure, that the backups
     are stored safely.
+
+    If you want to also include the RADIUS configuration into the backup
+    specify a directory using 'radius_directory'.
     """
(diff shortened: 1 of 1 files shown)


pi-manage
```diff
@@ -275,11 +279,13 @@ def create(directory="/var/lib/privacyidea/backup/",
...
+        m = re.match(".*mysql://(.*):(.*)@(.*)/(\w*)\??(.*)", sqluri)
         username = m.groups()[0]
         password = m.groups()[1]
         datahost = m.groups()[2]
         database = m.groups()[3]
+        # We strip parameters, but we do not use them
+        _parameters = m.groups()[4]
         defaults_file = "{0!s}/mysql.cnf".format(conf_dir)
         _write_mysql_defaults(defaults_file, username, password)
         call("mysqldump --defaults-file=%s -h %s %s > %s" % (
```
fredreichbier: Change looks good to me, but I think it would be nice to get rid of the unescaped variables in the subprocess calls with ``shell=True`` in this function. We could do this using ``pipes.quote``.
In order to stay compatible to Python 2+3, we could use ``from six.moves import shlex_quote``.

But I can also do this in a separate PR.

cornelinux: As I do not exactly understand what you mean, an addtional PR would be great! ;-)

fredreichbier: Will do!

PR #1577, opened 2019-04-16
Use resolver in UI-user-policies

### LINKED ISSUE #1496: UI rights do not check for user resolver filters
# What did you try to do?

Display UI rights correctly according to policies and their filters.
For example:
In the same realm realmA, there are 2 resolvers resolverX and resolverY. In policies setup, resolverX has assign action enabled, whereas resolverY does not have assign action enabled.

When logging to the UI, users in resolverX should see assign option in the menu. Users in resolverY should not see assign option in the UI.

Users in both resolvers of realmA see the assign option. Although when actually self-assigning a token, it will fail for user in resolverY.

# Configuration

* privacyIDEA Version: 2.23.3

* Installation method: from source

## more details: 

* OS: Amazon Linux

* Webserver: Nginx

* Tokendatabase: PostgreSQL

# Log file

N/A

# Side notes:

I checked this method and it seems that we do not pass the resolver info when checking for UI rights, and may result in the described outcome?

```
# privacyidea/privacyidea/lib/policy.py
def ui_get_rights(self, scope, realm, username, client=None):
[...]
```
This is not a very critical bug. However, it affects user experience and causes some confusion. It would be nice to have it adjusted!

Cheers,

Quynh Nguyen

cornelinux:
Closes #1496

### DIFF
--- a/privacyidea/lib/policy.py
+++ b/privacyidea/lib/policy.py
@@ -750,6 +750,7 @@ class PolicyClass(with_metaclass(Singleton, object)):
         rights = set()
         userrealm = None
         adminrealm = None
+        resolver = None
         logged_in_user = {"username": username,
                           "realm": realm}
         if scope == SCOPE.ADMIN:
@@ -758,9 +759,11 @@ class PolicyClass(with_metaclass(Singleton, object)):
         elif scope == SCOPE.USER:
             userrealm = realm
             logged_in_user["role"] = ROLE.USER
+            resolver = User(username, userrealm).resolver
         pols = self.get_policies(scope=scope,
                                  adminrealm=adminrealm,
                                  realm=userrealm,
+                                 resolver=resolver,
                                  user=username, active=True,
                                  client=client)
         for pol in pols:
--- a/tests/test_lib_policy.py
+++ b/tests/test_lib_policy.py
@@ -19,7 +19,9 @@ from privacyidea.lib.resolver import (save_resolver, get_resolver_list,
 from privacyidea.lib.error import ParameterError
 from privacyidea.lib.user import User
 import datetime
-PWFILE = "tests/testdata/passwords"
+from .base import PWFILE as FILE_PASSWORDS
+from .base import PWFILE2 as FILE_PASSWD
+
 
 
 def _check_policy_name(polname, policies):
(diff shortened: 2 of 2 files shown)


privacyidea/lib/policy.py
```diff
@@ -758,9 +759,11 @@ def ui_get_rights(self, scope, realm, username, client=None):
         elif scope == SCOPE.USER:
             userrealm = realm
             logged_in_user["role"] = ROLE.USER
+            resolver = User(username, userrealm).resolver
```
fredreichbier: I was thinking: The ``User(...).resolver`` line results in a query of the Id Resolver. ``ui_get_rights`` is ultimately called by ``get_auth_token``, and there, we already have a user object:
https://github.com/privacyidea/privacyidea/blob/5e43b23f40cbf304a6246d05589fe3f4d1b280b7/privacyidea/api/auth.py#L255
So there isn't really a need to query the Id Resolver again. We could just add the ``resolver`` parameter to the ``ui_get_rights`` (and ``ui_get_main_menus``) to avoid the additional call.

On the other hand, this code is only called once at each user login, so the impact of the additional calls might be small, so we could also just keep it as above. What do you think?

privacyidea/lib/policy.py
```diff
@@ -758,9 +759,11 @@ def ui_get_rights(self, scope, realm, username, client=None):
...
         pols = self.get_policies(scope=scope,
                                  adminrealm=adminrealm,
                                  realm=userrealm,
+                                 resolver=resolver,
```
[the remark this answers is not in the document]
cornelinux: Yes.
There is only an *adminrealm*, but no *adminresolver*.
This was not necessary till now. i think this would lead to confusion, when an administrator logs in.
Usually the realms are called like "admins" or "helpdesk". This way the logging in admin knows, what policies or rights he will get.
Otherwise the admin would login as "weber@admin", but could not be sure if he has rights as a superuser or as a servicedesk user.

privacyidea/lib/policy.py
cornelinux: You are right. But I am not quite sure.
We only have a resolver, if a user from a resolver authenticates. We do not have  a resolver, if an internal admin authenticates! 
This way to ui_get_rights looks more self contained to me.

PR #1581, opened 2019-04-17
Set the logo to vertical center

cornelinux:
The logo is contained in the height of the text of the navbar.
Due to this I can not scale the logo larger than the text of
the menu.
This is why the logo was a bit below the center in the first place.

Working on #1559

image

PR #1582, opened 2019-04-17
Vertically center the dropdown menu on the right

### LINKED ISSUE #1559: Vertical center in top bar
Bildschirmfoto vom 2019-04-05 10-22-28
In the attached screenshot, you can see the top bar of privacyIDEA. You see, the logo on the left and the button on the right side. These two objects should be centered vertically, as the other elements do.

cornelinux:
image

Closes #1559

### DIFF
--- a/privacyidea/static/css/signin.css
+++ b/privacyidea/static/css/signin.css
@@ -126,8 +126,3 @@ select.ng-invalid:focus:not(#username):not(#password){
   -webkit-box-shadow: inset 0 1px 1px rgba(0, 0, 0, .075), 0 0 6px #ce8483;
           box-shadow: inset 0 1px 1px rgba(0, 0, 0, .075), 0 0 6px #ce8483;
 }
-
-.pidropdown {
-    padding-top: 7px;
-    padding-bottom: 7px;
-}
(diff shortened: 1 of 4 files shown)


privacyidea/static/css/signin.css
```diff
@@ -126,3 +126,8 @@ select.ng-invalid:focus:not(#username):not(#password){
   -webkit-box-shadow: inset 0 1px 1px rgba(0, 0, 0, .075), 0 0 6px #ce8483;
           box-shadow: inset 0 1px 1px rgba(0, 0, 0, .075), 0 0 6px #ce8483;
 }
+
+.pidropdown {
+    padding-top: 7px;
+    padding-bottom: 7px;
+}
```
plettich: We could also add
`.navbar-nav .btn-group {
    padding-top: 7px;
}`
to avoid adding an extra class.

cornelinux: @Mipronimo what do you think?

plettich: And since my comment on #1581 didn't appear: isn't the `signin.css` for the login page? Shouldn't we use a generic bootstrap override?

cornelinux: I do not think we should overcomplicate things ;-)

cornelinux: I will put the navbar css in an extra css file ``menu.css``.
We already have a few own css files. This way it might make more sense, that users can more easily adopt the css (customization) without the need to maintain the complete file.

cornelinux: @plettich I added your comments. Please review and merge in branch-3.0
By introducing the menu.css we also get rid of merge conflicts. And it makes customization easier.

PR #1588, opened 2019-04-23
Fix IE11 rendering issue with 3.0

droobah:
IE11 is still used in many corporate environments and does not render the WebUI due to an unsupported for..of loop.  Replaced for..of loop with simple iterative loop.


plettich:
If IE11 doesn't support u2f anyway, can't we just exclude the complete file if the user-agent is =<IE11?

droobah:
> If IE11 doesn't support u2f anyway, can't we just exclude the complete file if the user-agent is =<IE11?

I tried renaming the file to simulate this, but the U2F factory is referenced in other scripts so the page still fails to load.

cornelinux:
@Mipronimo Can you please check this change?
And can you please also check, if we need to adapt this in other plugins?

PR #1602, opened 2019-04-30
Default usercache time to 0 seconds

### LINKED ISSUE #1596: Setting and then Clearing 'User Cache expiration in seconds' Breaks Authentication
# What did you try to do?

Clear/delete value for 'User Cache expiration in seconds' under System Config.

Value reset to default of 0.

Authentication breaks after value is cleared.  Error is logged to log file.

`
[2019-04-26 14:31:42,511][ERROR][privacyidea.lib.token:493] User information can not be retrieved: invalid literal for int() with base 10: ''
`

Changing UserCacheExpiration back to 0 directly in the config table in the pi database and restarting the Apache2 service resolves the issue.

# Configuration

* privacyIDEA Version: 3.0
* Installation method: ubuntu packages

## more details: 

* OS: Ubuntu 16.04LTS
* Webserver: Apache2
* Tokendatabase: MySQL

cornelinux:
If invalid data is saved in the usercase timeout,
we default to 0.

Fix #1596

### DIFF
--- a/privacyidea/lib/usercache.py
+++ b/privacyidea/lib/usercache.py
@@ -70,7 +70,12 @@ def get_cache_time():
     :return: UserCacheExpiration config value as a timedelta
     :rtype: timedelta
     """
-    seconds = int(get_from_config(EXPIRATION_SECONDS, '0'))
+    seconds = 0
+    try:
+        seconds = int(get_from_config(EXPIRATION_SECONDS, '0'))
+    except:
+        log.info(u"Non-Integer value stored in system config {0!s}".format(EXPIRATION_SECONDS))
+
     return datetime.timedelta(seconds=seconds)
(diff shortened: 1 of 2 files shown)


privacyidea/lib/usercache.py
```diff
@@ -70,7 +70,12 @@ def get_cache_time():
     :return: UserCacheExpiration config value as a timedelta
     :rtype: timedelta
     """
-    seconds = int(get_from_config(EXPIRATION_SECONDS, '0'))
+    seconds = 0
+    try:
+        seconds = int(get_from_config(EXPIRATION_SECONDS, '0'))
+    except:
```
fredreichbier: Could we change this to ``except ValueError:``?

PR #1606, opened 2019-05-02
Fix logic of auth counters and registration token

### LINKED ISSUE #1587: Registration token validity not working
# What did you try to do?

Using Registration token with validity period

Registration token that is outside of validity period should not work.

Registration token that is outside of validity period is still successfully authenticated.

# Configuration

* privacyIDEA Version: 2.23.5

* Installation method:
 from source

## more details: 

* OS: Amazon Linux

* Webserver: Nginx

* Tokendatabase: PostgreSQL

# Note:
I suspect that because in privacyidea/lib/token.py method `check_token_list` we delete the token during the `inc_count_auth_success()` before the `check_all()`, resulting in the incorrect behavior mentioned in this ticket.

```
        # One ore more successfully authenticating tokens found
        # We need to return success
        message_list = ["matching {0:d} tokens".format(len(valid_token_list))]
        # write serial numbers or something to audit log
        for token_obj in valid_token_list:
            if increase_auth_counters:
                token_obj.inc_count_auth_success()
            # Check if the max auth is succeeded
            if token_obj.check_all(message_list):
```

I think this will be a security issue, and would like to ask if we can enhance the logic? If not maybe I can use event handler to handle this issue?

Best regards,

Quynh Nguyen

cornelinux:
The check for the auth counters and max_auth_counters is fixed.
The order or checking the counters and increasing the counters (which
implies deleting the registration token) is changed and fixed.

We now first check if the token is allowed to be used for authentication
("check_all") and then increase the counters (which deletes the
registration token).
This way we need to add an offset to the counter checking.

Fixes #1587

### DIFF
--- a/doc/configuration/system_config.rst
+++ b/doc/configuration/system_config.rst
@@ -72,7 +72,8 @@ clear the failcounter again:
 
 * A successful authentication with correct PIN and correct OTP value
 * A successfully triggered challenge (Usually this means a correct PIN)
-* An authentication with a correct PIN, but a wrong OTP value
+* An authentication with a correct PIN, but a wrong OTP value (Only if
+  :ref:`reset_failcounter_on_correct_pin` is set).
 
 A "0" means automatically clearing the fail counter is not used.
 
@@ -80,6 +81,9 @@ A "0" means automatically clearing the fail counter is not used.
    update the mentioned timestamp.
 
 Also see :ref:`brute_force`.
+
+
+.. _reset_failcounter_on_correct_pin:
 
 Resetting Failcounter on correct PIN
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/doc/configuration/system_config.rst
+++ b/doc/configuration/system_config.rst
@@ -81,6 +81,19 @@ A "0" means automatically clearing the fail counter is not used.
 
 Also see :ref:`brute_force`.
 
+Resetting Failcounter on correct PIN
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+After the above mentioned timeout the failcounter is reset by a successful
+authentication (correct PIN and OTP value) or by the correct PIN of a challenge
+response token.
+
+It can be also reset by the correct PIN of any token, when setting
+``ResetFailcounterOnPIN`` to True.
+The default behaviour is, that the correct PIN of a normal token will *not*
+reset the failcounter after the clearing timeout.
+
+
 Prepend PIN
 ~~~~~~~~~~~
 
--- a/privacyidea/lib/token.py
+++ b/privacyidea/lib/token.py
@@ -2264,8 +2264,7 @@ def check_token_list(tokenobject_list, passw, user=None, options=None, allow_res
         # So we increase the failcounter. Return failure.
         for tokenobject in pin_matching_token_list:
             tokenobject.inc_failcount()
-            if get_from_config(SYSCONF.RESET_FAILCOUNTER_ON_PIN_ONLY, False, return_bool=True):  # pragma: no cover
-                tokenobject.check_reset_failcount()
+            tokenobject.check_reset_failcount()
             reply_dict["message"] = "wrong otp value"
             if len(pin_matching_token_list) == 1:
                 # If there is only one pin matching token, we look if it was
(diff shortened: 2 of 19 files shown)


fredreichbier:
I think this PR actually changes behavior in two aspects:
* With this PR, we do not increase ``count_auth`` and ``count_auth_success`` if a user tries to authenticate with a token outside of the validity period, or with a token for which the maximum number of authentication has been exceeded.
* Without this PR, setting ``count_auth_max = 3`` for a token allows *two* authentications. With this PR, ``count_auth_max = 3`` allows *three* authentications.

I think both changes make a lot of sense, but we should be aware of them :)

However, I find the parameters ``offset`` and ``offset_success`` a bit confusing, and I think it's hard to tell what ``check_auth_counter`` actually checks. The only invocation of ``check_auth_counter`` (besides tests) is in ``check_all``. ``check_all`` is called in two places, and I think in both cases the only sensible values for ``offset`` and ``offset_success`` are 1:

https://github.com/privacyidea/privacyidea/blob/d6462062ddfe01faa43430ed2806a3e280f0525a/privacyidea/lib/token.py#L2005

https://github.com/privacyidea/privacyidea/blob/d6462062ddfe01faa43430ed2806a3e280f0525a/privacyidea/lib/token.py#L2165

If ``offset`` and ``offset_success`` are 1, the new ``check_auth_counter`` is equivalent to the old ``check_auth_counter``. So I would propose removing the offset parameters from ``check_all`` and ``check_auth_counter``. Instead, ``check_auth_counter`` should just check if the current counters are *less than and not equal to* the maximally allowed counters: If this is the case, we allow at least one more authentication.

What do you think?

cornelinux:
I think my latest commit will please you! :-)

privacyidea/lib/token.py
```diff
@@ -2160,12 +2160,14 @@ def check_token_list(tokenobject_list, passw, user=None, options=None, allow_res
...
-            # Reset the failcounter, if there is a timeout set
-            token_obj.check_reset_failcount()
-            # Check if the max auth is succeeded
+            # Check if the max auth is succeeded.
+            # We need to set the offsets, since we are in the n+1st authentication.
             if token_obj.check_all(message_list):
+                if increase_auth_counters:
+                    token_obj.inc_count_auth_success()
+                # Reset the failcounter, if there is a timeout set
+                token_obj.check_reset_failcount()
```
fredreichbier: Sorry I'm only noticing this now, but I think this change breaks the ``failcounter_clear_timeout`` feature.
Assume we set ``failcounter_clear_timeout`` to 3 minutes and there is a token whose failcounter is exceeded.
After 5 minutes, we try to authenticate with a correct OTP.
We should be able to authenticate because the last unsuccessful authentication was >3 minutes ago.
But because the fail counter is exceeded, ``check_all`` returns False, so we cannot authenticate.

I think we have to move ``token_obj.check_reset_failcount()`` before ``check_all()`` again?

cornelinux: Thanks a lot for pointing this out.
But this is strange, since I *think* we have test cases for the resetting of the failcounter.

I will check this, add a testcase and adapt this if necessary!

cornelinux:
I rebased this on branch-3.0, to be sure to NOT to run into any issues.
I am not sure now, if this was such a glorious idea.
However, @fredreichbier , please take a look at my latest (2, including the merge) commit today.
This is, what acutally has changed.

doc/configuration/system_config.rst
```diff
@@ -81,6 +81,19 @@ A "0" means automatically clearing the fail counter is not used.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+After the above mentioned timeout the failcounter is reset by a successful
+authentication (correct PIN and OTP value) or by the correct PIN of a challenge
+response token.
+
+It can be also reset by the correct PIN of any token, when setting
+``ResetFailcounterOnPIN`` to True.
+The default behaviour is, that the correct PIN of a normal token will *not*
+reset the failcounter after the clearing timeout.
```
fredreichbier: Looks good, but should we then remove the third bullet point from the section "Automatically clearing Failcounter"?

cornelinux: done in 90c868d8

cornelinux:
> Code looks OK to me, I have one comment regarding documentation.

> I'm not sure what happened with the commits here :) My test case commit appears twice with different > commit hashes now:
> 628443d
> 8e16fba

> I think we should try a new cleaned-up PR based on branch-3.0 :) I can do this once we're ready.

This would be the same as sqaushing this PR. I am not sure if this is really necessary.

PR #1607, opened 2019-05-02
Improve the Firebase configuration

### LINKED ISSUE #1592: Configure the push token with two json files
Instead of asking for one json file and all the data that is relevant for the client, we should ask for two json files and read the relevant data from the two files.

The two files would be

* google-services.json
* and the json file for the service account

cornelinux:
The firebase configuration is a bit tricky.
To avoid uploading the wrong json file or entering wrong data
SMSprovider now can have a config-check-method, which is run
after the SMS provider is created.

This way privacyIDEA checks the sanity of the firebase configuration.
It checks if the correct json file is specified and if
the correct project number is configured.

Closes #1592

### DIFF
--- a/privacyidea/lib/error.py
+++ b/privacyidea/lib/error.py
@@ -55,8 +55,6 @@ class ERROR:
     SERVER = 903
     USER = 904
     PARAMETER = 905
-    FIREBASE_JSON = 1001
-    FIREBASE_PROJECT = 1002
 
 
 @six.python_2_unicode_compatible
--- a/privacyidea/lib/error.py
+++ b/privacyidea/lib/error.py
@@ -55,6 +55,8 @@ class ERROR:
     SERVER = 903
     USER = 904
     PARAMETER = 905
+    FIREBASE_JSON = 1001
+    FIREBASE_PROJECT = 1002
 
 
 @six.python_2_unicode_compatible
--- a/privacyidea/lib/smsprovider/FirebaseProvider.py
+++ b/privacyidea/lib/smsprovider/FirebaseProvider.py
@@ -23,8 +23,8 @@ Firebase Cloud Messaging Service.
 This provider is used for the push token and can be used for SMS tokens.
 """
 
-from privacyidea.lib.smsprovider.SMSProvider import (ISMSProvider, SMSError)
-from privacyidea.lib.error import ERROR
+from privacyidea.lib.smsprovider.SMSProvider import (ISMSProvider)
+from privacyidea.lib.error import ConfigAdminError
 from privacyidea.lib import _
 import logging
 from oauth2client.service_account import ServiceAccountCredentials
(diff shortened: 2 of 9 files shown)


privacyidea/lib/error.py
```diff
@@ -55,6 +55,8 @@ class ERROR:
     SERVER = 903
     USER = 904
     PARAMETER = 905
+    FIREBASE_JSON = 1001
```
fredreichbier: Hm, I'm not sure if we should add new error codes specifically for the Firebase config validation? Maybe we should use some more generic error code like "CONFIG"?

privacyidea/lib/smsprovider/FirebaseProvider.py
```diff
@@ -98,6 +99,28 @@ def submit_message(self, firebase_token, data):
 
         return res
 
+    def check_configuration(self):
+        """
+        This method checks the sanity of the configuration of this provider.
+        If there is a configuration error, than an exception is raised.
+        :return:
+        """
+        json_file = self.smsgateway.option_dict.get(FIREBASE_CONFIG.JSON_CONFIG)
```
fredreichbier: Should we check if the configuration file exists?

privacyidea/lib/smsprovider/SMSProvider.py
```diff
@@ -170,6 +179,7 @@ def set_smsgateway(identifier, providermodule, description=None,
     smsgateway = SMSGateway(identifier, providermodule,
```
fredreichbier: ``SMSGateway.__init__`` already calls ``.save()`` to write the entry to the database, even if ``check_configuration`` fails later. This means that even though an error is raised, we will end up with invalid database entries. Should we change that, or just keep this behavior?

privacyidea/lib/smsprovider/FirebaseProvider.py
plettich: isn't it (implicitly) checked in the `with open(...` part?

cornelinux: Yes, If the file does not exist, imho an exception is raised and we get an error anyways.
Would work like expected.

privacyidea/lib/smsprovider/SMSProvider.py
cornelinux: We need to save this, otherwise we can not check the config.
Before the configuration was always saved without checking it.
Yes, we need to keep it that way.

privacyidea/lib/error.py
cornelinux: I am not sure. 
Imho the idea of the error codes is, that you can find a dedicated place in the code, where the error occurred.
So I really want to have a dedicated firebase error code.
The question is, if we should define these error codes in the firebase modules with its own number range - this would be a general question.

fredreichbier: Hm, I'm not sure if we have a connection between error codes and code places currently. For example, ``PARAMETER`` is used in the exception class ``ParameterError``, and ``ParameterError`` is raised at a lot of different places.

If I compare the new constants ``FIREBASE_JSON`` and ``FIREBASE_PROJECT`` with the existing constants like ``SERVER``, ``PARAMETER``, the new constants seem a lot more specific to me, because conceptually the firebase provider is just one of multiple SMS providers.

So I'm not strongly opposed to the new error codes, they just seem a bit inconsistent with the existing error codes to me. :)

privacyidea/lib/smsprovider/FirebaseProvider.py
fredreichbier: OK, yes, we get a HTTP 500 if the file does not exist. The ``SMSError`` exceptions below also create HTTP 500 responses, so we can keep it that way.

privacyidea/lib/smsprovider/SMSProvider.py
fredreichbier: My issue with this is as follows: I enter an invalid firebase configuration. The check fails, which causes the HTTP request to fail with a 500 response. Because of this response, I would expect that the failed request does not actually save the invalid configuration.

I think checking before saving would be nice. But it seems it would require moving around a lot of code, so let's keep it like above for now.

privacyidea/lib/error.py
cornelinux: > So I'm not strongly opposed to the new error codes, they just seem a bit inconsistent with the existing error codes to me. :)

This might be due to the fact, that we did not finally rework the error codes completely.

You could argue, that these errors are no "ParameterError"s, sinde ParameterErrors are errors in the API. But these errors, that are raised here, are directly **logical** errors.

So somehow something new, right?
Maybe we should do a meeting or talk. :-)

cornelinux: Hi @fredreichbier 
I reworked to use the ConfigAdminError. I think this makes a lot of sense now.

fredreichbier: Looks good, thx!

PR #1614, opened 2019-05-07
Add parameters for iOS Push

### LINKED ISSUE #1608: Allow to configure push for iOS
We need to be able to configure push for iOS.
We probalby need to add some APP ID and API KEy.

cornelinux:
Closes #1608

### DIFF
--- a/privacyidea/lib/smsprovider/FirebaseProvider.py
+++ b/privacyidea/lib/smsprovider/FirebaseProvider.py
@@ -121,6 +121,18 @@ class FirebaseProvider(ISMSProvider):
         else:
             raise ConfigAdminError(description="Please check your configuration. Can not load JSON file.")
 
+        # We need at least
+        #         FIREBASE_CONFIG.API_KEY_IOS and FIREBASE_CONFIG.APP_ID_IOS
+        # or
+        #         FIREBASE_CONFIG.API_KEY and FIREBASE_CONFIG.APP_ID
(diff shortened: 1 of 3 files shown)


privacyidea/lib/smsprovider/FirebaseProvider.py
```diff
@@ -150,12 +152,22 @@ def parameters(cls):
...
+                              "The App ID, that the Android client should use. Get it from your Firebase console.")
                       },
                       FIREBASE_CONFIG.API_KEY: {
                           "required": True,
                           "description": _(
-                              "The API Key, that the client should use. Get it from your Firebase console.")
+                              "The API Key, that the Android client should use. Get it from your Firebase console.")
+                      },
+                      FIREBASE_CONFIG.APP_ID_IOS:{
+                          "required": True,
```
plettich: Should we really require the user to create a firebase app for Android **and** iOS? Would it be possible to use only Android **or** iOS?

cornelinux: Logically we could only tell *after* the firebase provider is created, that **either** iOS or Andoird parameters are required.

It does not do any harm, if you simply copy values from android to ios or vice versa.

But there might also be "enough" scenarios, where **both** values are needed.

cornelinux: Hi @plettich I added this in 9f7f3fc2

PR #1628, opened 2019-05-13
Branch 3.0

cornelinux:
We fixed all open issues for 3.0.1, so we can merge it again into master.

Well, in master we will probably set the version in setup.py to 3.1dev1 then.


Changelog
```diff
@@ -1,3 +1,24 @@
...
+  Fixes:
+  * Fix PUSH token issues:
+    * Add logic checking to setup of PUSH token (#1592)
+    * Remove double enrollment notification of PUSH token in WebUI (#1598)
+    * Fix to allow spaces in Firebase configuration (#1599)
+    * Add support for iOS Firebase configuration (#1608)
+    * Fix to allow PUSH token enrollment, even with Label-policy (#1589)
+    * Fix to mark PUSH token challenge answered in the database (#1584)
+  * Fix the validity period of the registration token (#1587)
+  * Beautify the vertial alignment in the Web UI top menu (#1559)
```
fredreichbier: Typo: *vertical

doc/conf.py
```diff
@@ -17,7 +17,7 @@
 # built documents.
 #
 # The short X.Y version.
-version = '3.0'
+version = '3.0.1dev1'
```
fredreichbier: Should this also be ``3.0.1dev2``, like in ``setup.py``?

Changelog
cornelinux: Will do

doc/conf.py
cornelinux: Yes. But the most important part is, that set this to 3.0.1 when we are done! ;-)
Will do.

cornelinux: Hm, lets see if this appears, when i push on the branch-3.0

cornelinux: done in 656fe468

Changelog
cornelinux: done in 656fe468

fredreichbier:
Travis has run successfully, but github doesn't know. Merging!

PR #1633, opened 2019-05-15
Remove audit from remote user fetching

### LINKED ISSUE #1632: audit_data not available, when requiring a remote user
When using remote user the following error can occur:

~~~~log
   File "/opt/privacyidea/local/lib/python2.7/site-packages/privacyidea/api/lib/prepolicy.py", line 1226, in is_remote_user_allowed
     audit_data=g.audit_object.audit_data)
   File "/opt/privacyidea/local/lib/python2.7/site-packages/werkzeug/local.py", line 347, in __getattr__
     return getattr(self._get_current_object(), name)
 AttributeError: '_AppCtxGlobals' object has no attribute 'audit_object'
~~~~

cornelinux:
Fetching the remote user config also occurs by simply
calling the login dialog. At this point, we do not have
an audit object. The pure calling of the login dialog
also does not create any audit entry, so we do not
pass the audit_object.

fixes #1632

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -1222,8 +1222,7 @@ def is_remote_user_allowed(req):
                                                          scope=SCOPE.WEBUI,
                                                          user=loginname,
                                                          realm=realm,
-                                                         client=g.client_ip,
-                                                         audit_data=g.audit_object.audit_data)
+                                                         client=g.client_ip)
 
         res = bool(ruser_active)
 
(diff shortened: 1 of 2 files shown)


privacyidea/api/lib/prepolicy.py
```diff
@@ -1222,8 +1222,7 @@ def is_remote_user_allowed(req):
                                                          scope=SCOPE.WEBUI,
                                                          user=loginname,
                                                          realm=realm,
-                                                         client=g.client_ip,
-                                                         audit_data=g.audit_object.audit_data)
+                                                         client=g.client_ip)
```
fredreichbier: ``is_remote_user_allowed`` is not only called by ``single_page_application`` but also by ``get_auth_token``, which implements the ``/auth`` endpoint. In the latter case, I think we do have an audit object and we should pass it to ``get_action_values`` to extend the audit log entry with the information that the REMOTE_USER policy is in effect.

cornelinux: Done in 0cbab21b

PR #1638, opened 2019-05-21
Fix sending mails in combination with CRAM-MD5 authentication

### LINKED ISSUE #1637: Sending mails fails if SMTP server advertises CRAM-MD5 authentication
Hello,

I've installed privacyidea 3.0 on a debian stretch in an virtualenv with python 2.7. There were a problem so the smtp module did not works as exprected.

If you try to use a smtp server privacyidea throws the exception: "character mapping must return integer, None or unicode"

Here's the stack:
```
[2019-05-21 09:59:31,017] ERROR in app: Exception on /smtpserver/send_test_email [POST]
 127.0.0.1:34212] Traceback (most recent call last):
 127.0.0.1:34212]   File "/opt/privacyidea/local/lib/python2.7/site-packages/flask/app.py", line 2292, in wsgi_app
 127.0.0.1:34212]     response = self.full_dispatch_request()
 127.0.0.1:34212]   File "/opt/privacyidea/local/lib/python2.7/site-packages/flask/app.py", line 1815, in full_dispatch_request
 127.0.0.1:34212]     rv = self.handle_user_exception(e)
 127.0.0.1:34212]   File "/opt/privacyidea/local/lib/python2.7/site-packages/flask/app.py", line 1718, in handle_user_exception
 127.0.0.1:34212]     reraise(exc_type, exc_value, tb)
 127.0.0.1:34212]   File "/opt/privacyidea/local/lib/python2.7/site-packages/flask/app.py", line 1813, in full_dispatch_request
 127.0.0.1:34212]     rv = self.dispatch_request()
 127.0.0.1:34212]   File "/opt/privacyidea/local/lib/python2.7/site-packages/flask/app.py", line 1799, in dispatch_request
 127.0.0.1:34212]     return self.view_functions[rule.endpoint](**req.view_args)
 127.0.0.1:34212]   File "/opt/privacyidea/local/lib/python2.7/site-packages/privacyidea/api/lib/prepolicy.py", line 129, in policy_wrapper
 127.0.0.1:34212]     return wrapped_function(*args, **kwds)
 127.0.0.1:34212]   File "/opt/privacyidea/local/lib/python2.7/site-packages/privacyidea/lib/log.py", line 154, in log_wrapper
 127.0.0.1:34212]     return func(*args, **kwds)
 127.0.0.1:34212]   File "/opt/privacyidea/local/lib/python2.7/site-packages/privacyidea/api/smtpserver.py", line 157, in test
 127.0.0.1:34212]     u"The configuration {} is working.".format(identifier))
 127.0.0.1:34212]   File "/opt/privacyidea/local/lib/python2.7/site-packages/privacyidea/lib/smtpserver.py", line 145, in send_or_enqueue_email
 127.0.0.1:34212]     return send(config, recipient, subject, body, sender, reply_to, mimetype)
 127.0.0.1:34212]   File "/opt/privacyidea/local/lib/python2.7/site-packages/privacyidea/lib/smtpserver.py", line 116, in test_email
 127.0.0.1:34212]     mail.login(config['username'], password)
 127.0.0.1:34212]   File "/usr/lib/python2.7/smtplib.py", line 607, in login
 127.0.0.1:34212]     (code, resp) = self.docmd(encode_cram_md5(resp, user, password))
 127.0.0.1:34212]   File "/usr/lib/python2.7/smtplib.py", line 571, in encode_cram_md5
 127.0.0.1:34212]     response = user + " " + hmac.HMAC(password, challenge).hexdigest()
 127.0.0.1:34212]   File "/usr/lib/python2.7/hmac.py", line 75, in __init__
 127.0.0.1:34212]     self.outer.update(key.translate(trans_5C))
 127.0.0.1:34212] TypeError: character mapping must return integer, None or unicode
 127.0.0.1:34212] /opt/privacyidea/local/lib/python2.7/site-packages/sqlalchemy/engine/default.py:536: Warning: Data truncated for column 'info' at row 1
 127.0.0.1:34212]   cursor.execute(statement, parameters)
```

fredreichbier:
Under Python 2, we need to pass passwords as bytestrings, while Python 3 expects passwords as (unicode) strings.

In order to test this, I used a modified version of a [fake SMTP server](https://github.com/ReachFive/fake-smtp-server):
```
git clone https://github.com/fredreichbier/fake-smtp-server
cd fake-smtp-server
npm install .
node ./index.js
```
starts a SMTP server on port 1025. This server advertises CRAM-MD5:
```
$ (sleep 1; echo 'EHLO test') | nc localhost 1025
220 teapot ESMTP
250-teapot Nice to meet you, localhost
250-PIPELINING
250-8BITMIME
250-SMTPUTF8
250 AUTH PLAIN LOGIN CRAM-MD5
```
In privacyIDEA, I now configure a SMTP server on localhost:1025 with username ``test`` and password ``huhu5``, and test the configuration.

Without this PR, I get a privacyIDEA error "TypeError: character mapping must return integer, None or unicode". With this PR, privacyIDEA says "Test Email sent successfully", and the fake SMTP server prints:
```
INFO: CRAM-MD5 SMTP login for user: test password: undefined
INFO: CRAM-MD5 authentication -- is password huhu5 correct?  true
```

Fixes #1637

### DIFF
--- a/privacyidea/lib/smtpserver.py
+++ b/privacyidea/lib/smtpserver.py
@@ -17,11 +17,13 @@
 # License along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 #
+import six
+
 from privacyidea.lib.queue import job, wrap_job, has_job_queue
 from privacyidea.models import SMTPServer as SMTPServerDB
 from privacyidea.lib.crypto import (decryptPassword, encryptPassword,
                                     FAILED_TO_DECRYPT_PASSWORD)
-from privacyidea.lib.utils import fetch_one_resource
+from privacyidea.lib.utils import fetch_one_resource, to_utf8
 import logging
 from privacyidea.lib.log import log_with
 from time import gmtime, strftime
@@ -113,6 +115,10 @@ class SMTPServer(object):
             password = decryptPassword(config['password'])
             if password == FAILED_TO_DECRYPT_PASSWORD:
                 password = config['password']
+            # Under Python 2, we pass passwords as bytestrings to get CRAM-MD5 to work.
+            # Under Python 3, we pass passwords as unicode.
+            if six.PY2:
+                password = to_utf8(password)
             mail.login(config['username'], password)
         r = mail.sendmail(mail_from, recipient, msg.as_string())
         log.info("Mail sent: {0!s}".format(r))
--- a/privacyidea/lib/smtpserver.py
+++ b/privacyidea/lib/smtpserver.py
@@ -19,11 +19,12 @@
 #
 import six
 
+from privacyidea.lib.framework import get_app_config_value
 from privacyidea.lib.queue import job, wrap_job, has_job_queue
 from privacyidea.models import SMTPServer as SMTPServerDB
 from privacyidea.lib.crypto import (decryptPassword, encryptPassword,
                                     FAILED_TO_DECRYPT_PASSWORD)
-from privacyidea.lib.utils import fetch_one_resource, to_utf8
+from privacyidea.lib.utils import fetch_one_resource, to_bytes
 import logging
 from privacyidea.lib.log import log_with
 from time import gmtime, strftime
@@ -116,9 +117,10 @@ class SMTPServer(object):
             if password == FAILED_TO_DECRYPT_PASSWORD:
                 password = config['password']
             # Under Python 2, we pass passwords as bytestrings to get CRAM-MD5 to work.
+            # We add a safeguard config option to disable the conversion.
             # Under Python 3, we pass passwords as unicode.
-            if six.PY2:
-                password = to_utf8(password)
+            if six.PY2 and get_app_config_value("PI_SMTP_PASSWORD_AS_BYTES", True):
+                password = to_bytes(password)
             mail.login(config['username'], password)
         r = mail.sendmail(mail_from, recipient, msg.as_string())
         log.info("Mail sent: {0!s}".format(r))


privacyidea/lib/smtpserver.py
```diff
@@ -113,6 +115,10 @@ def test_email(config, recipient, subject, body, sender=None,
             password = decryptPassword(config['password'])
             if password == FAILED_TO_DECRYPT_PASSWORD:
                 password = config['password']
+            # Under Python 2, we pass passwords as bytestrings to get CRAM-MD5 to work.
+            # Under Python 3, we pass passwords as unicode.
+            if six.PY2:
+                password = to_utf8(password)
```
cornelinux: I can not tell you why, I feel slightly unpleasent with that.
Could we maybe add a config value, so that the password will *not* be converted, if the value is set - like

~~~~python
from privacyidea.lib.framework import get_app_config_value

if six.PY2 and get_app_config_value().get("PI_MAIL_UTF8_PASSWORDS", True):
    password = to_utf8(password)
~~~~
This way could "switch this off", by setting in pi.cfg

    PI_MAIL_UTF8_PASSWORDS = False

plettich: Maybe use `to_bytes()`, it basically does the same and the name is more intuitive (i think).
I think we should remove `to_utf8()` in the long run.

fredreichbier: I think always converting the password to a bytestring under Python 2 should be fine for all three supported authentication mechanisms PLAIN, LOGIN and CRAM-MD5. But OK, better be safe than sorry, I'll add the config option :) But I guess it's ok if we keep this option undocumented for now?

fredreichbier: Yes, will do!

fredreichbier: BTW, just wanted to note that even with this change, passwords with *non-ASCII characters* (like ``passwörd``) won't work, because Python's smtplib doesn't really support them: https://bugs.python.org/issue29750.

fredreichbier: I changed ``to_utf8`` to ``to_bytes`` in c298946.

fredreichbier: I added an option PI_SMTP_PASSWORD_AS_BYTES in c298946748f95ff7ede3f901d7cfdbbb0d9f600d. What do you think? :)

cornelinux: If smtplib would one day support non-ASCII characters, to_utf8 would immediately work, while to_bytes would not. Right?

cornelinux: thanks, @plettich for the explanation. So this can be merged.

PR #1647, opened 2019-05-24
Check resolvers in enrollTYPE policy

### LINKED ISSUE #1646: Add resolvers to some policies
The resolver of the user should be added to the following policies:

* api/lib/prepolicy.py:check_token_init
* api/lib/prepolicy.py:check_max_token_user

cornelinux:
When enrolling different token types the policies can
now differentiate according to resolvers.

Closes #1646

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -697,6 +697,7 @@ def check_max_token_user(request=None, action=None):
         limit_list = policy_object.get_action_values(ACTION.MAXTOKENUSER,
                                                      scope=SCOPE.ENROLL,
                                                      realm=user_object.realm,
+                                                     resolver=user_object.resolver,
                                                      user=user_object.login,
                                                      client=g.client_ip)
         if limit_list:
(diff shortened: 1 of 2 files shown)


privacyidea/api/lib/prepolicy.py
```diff
@@ -1085,14 +1086,24 @@ def check_token_init(request=None, action=None):
     role = g.logged_in_user.get("role")
     admin_realm = g.logged_in_user.get("realm")
     scope = SCOPE.ADMIN
+    # We need this user, if an admin enrolls a token for a user
+    user_obj = get_user_from_param(params)
+    resolver = realm = None
     if role == ROLE.USER:
         scope = SCOPE.USER
+        realm = admin_realm
```
plettich: Shouldn't the user have his own realm here?

fredreichbier: +1 -- to me it looks like we're changing behavior here, but I haven't yet fully understood this :)

cornelinux: "admin_realm" contains the realm of the logged in user. (line 1087.)
I some cases the variable name is confusing.
THis is why - in case of a user - we set the user-realm name and reset the admin_realm name.

PR #1649, opened 2019-05-25
1241/max active token per user

### LINKED ISSUE #1241: max_active_token_per_user
# What did you try to do?
Assign only one (01) active token, while the user has multiple disabled tokens.

Please provide your top level intent!
privacyIDEA provides configuration option to set max_token_per_user, but doesn't allow to set only one (01) active token per user, while old tokens are disabled.

Please provide steps to reproduce!

Admin can  set max_token_per_user, and while to avoid misuse if admin sets the max_token_per_user = 01, user can generate a token and uses it for a while. But in an event user loses the phone and wants helpdesk/admin to disable old token and wants to enroll a new token, privacyIDEA policy will restrict the user from enrolling the token due to token limit set in policy. Admin doesn't want to delete the disable token to keep the token history and audit log.

Hence there should be an option to set max_active_token_per_user, which means a user can have several disabled tokens and only 01 token can be active at a time. max_active_token_value should be less than max_token_per_user.

Another safety guard should be that once a new token is generated for the user, privacyIDEA must not allow activation of disabled tokens for the same user.

What did acutally happen?
Once user's old token is disabled, and while token limit is set to 01 in config, user's can't register a new token.
# Configuration

* privacyIDEA Version:

* Installation method:
  (from source, github, virtualenv, ubuntu packages...)

## more details: 

* OS:

* Webserver:

* Tokendatabase: (like MySQL, PostgreSQL, ...)

# Log file

**Turn on debug level and take a look at the privacyidea.log!**
**If appropriate, attach the log file.**

cornelinux:
Closes #1241

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -679,6 +679,7 @@ def check_max_token_user(request=None, action=None):
     Pre Policy
     This checks the maximum token per user policy.
     Check ACTION.MAXTOKENUSER
+    Check ACTION.MAXACTIVETOKENUSER
 
     This decorator can wrap:
         /token/init  (with a realm and user)
@@ -694,9 +695,11 @@ def check_max_token_user(request=None, action=None):
     serial = getParam(params, "serial")
     if user_object.login:
         policy_object = g.policy_object
+        # check maximum tokens of user
         limit_list = policy_object.get_action_values(ACTION.MAXTOKENUSER,
                                                      scope=SCOPE.ENROLL,
                                                      realm=user_object.realm,
+                                                     resolver=user_object.resolver,
                                                      user=user_object.login,
                                                      client=g.client_ip)
         if limit_list:
@@ -711,6 +714,30 @@ def check_max_token_user(request=None, action=None):
             if already_assigned_tokens >= max_value:
                 g.audit_object.add_policy(limit_list.get(str(max_value)))
                 raise PolicyError(ERROR)
+
+        # check maximum active tokens of user
+        limit_list = policy_object.get_action_values(ACTION.MAXACTIVETOKENUSER,
+                                                     scope=SCOPE.ENROLL,
+                                                     realm=user_object.realm,
+                                                     resolver=user_object.resolver,
+                                                     user=user_object.login,
+                                                     client=g.client_ip)
+        if limit_list:
+            # we need to check how many active tokens the user already has assigned!
+            tokenobject_list = get_tokens(user=user_object, active=True)
+            if serial:
+                for tok in tokenobject_list:
+                    if tok.token.serial == serial and tok.is_active():
+                        # If a serial is provided and this token already exists (and is active), the
+                        # token can be regenerated. If the token would be inactive, regenerating this
+                        # token would reactivate it and thus the user would have more tokens!
+                        return True
+            already_assigned_tokens = len(tokenobject_list)
+            max_value = max([int(x) for x in limit_list])
+            if already_assigned_tokens >= max_value:
+                g.audit_object.add_policy(limit_list.get(str(max_value)))
+                raise PolicyError(ERROR)
+
     return True
 
 
--- a/doc/policies/enrollment.rst
+++ b/doc/policies/enrollment.rst
@@ -48,6 +48,17 @@ Limit the maximum number of tokens per user in this realm.
    allowed number of tokens among the matching policies is enforced.
    Policy priorities are ignored.
 
+max_active_token_per_user
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+type: int
+
+Limit the maximum number of active tokens per user.
+
+.. note:: Inactive tokens will not be taken into account.
+   If the token already exists, it can be recreated if the token
+   is already active.
+
 tokenissuer
 ~~~~~~~~~~~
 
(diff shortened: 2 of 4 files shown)


privacyidea/api/lib/prepolicy.py
```diff
@@ -711,6 +714,30 @@ def check_max_token_user(request=None, action=None):
...
+                    if tok.token.serial == serial and tok.is_active():
+                        # If a serial is provided and this token already exists (and is active), the
+                        # token can be regenerated. If the token would be inactive, regenerating this
+                        # token would reactivate it and thus the user would have more tokens!
+                        return True
+            already_assigned_tokens = len(tokenobject_list)
+            max_value = max([int(x) for x in limit_list])
+            if already_assigned_tokens >= max_value:
+                g.audit_object.add_policy(limit_list.get(str(max_value)))
+                raise PolicyError(ERROR)
```
fredreichbier: If we use ``ERROR``, the error message is "The number of tokens for this user is limited!". Should we change that to "The number of *active* tokens for this user is limited!"?

privacyidea/api/lib/prepolicy.py
```diff
@@ -679,6 +679,7 @@ def check_max_token_user(request=None, action=None):
     Pre Policy
     This checks the maximum token per user policy.
     Check ACTION.MAXTOKENUSER
+    Check ACTION.MAXACTIVETOKENUSER
```
fredreichbier: I think we also need to add a check to ``enable_token``? Otherwise, users can circumvent the "max active tokens per user" limit by disabling their tokens, enrolling new tokens, and then enabling all of their tokens.

privacyidea/api/lib/prepolicy.py
```diff
@@ -711,6 +714,30 @@ def check_max_token_user(request=None, action=None):
...
+        # check maximum active tokens of user
+        limit_list = policy_object.get_action_values(ACTION.MAXACTIVETOKENUSER,
+                                                     scope=SCOPE.ENROLL,
+                                                     realm=user_object.realm,
+                                                     resolver=user_object.resolver,
+                                                     user=user_object.login,
+                                                     client=g.client_ip)
+        if limit_list:
+            # we need to check how many active tokens the user already has assigned!
+            tokenobject_list = get_tokens(user=user_object, active=True)
```
fredreichbier: If there are both a ``MAXTOKENUSER`` and a ``MAXACTIVETOKENUSER`` policy, we have the following issues:
* We call ``get_tokens`` two times (in lines 707 and 727), which results in database queries. We can avoid this by using:
```python
tokenobject_list = get_tokens(user=user_object)
active_tokenobject_list = [token for token in tokenobject_list if token.is_active()]
```
* We have a ``return True`` in line 707. If we hit this line, the ``MAXACTIVETOKENUSER`` policy isn't taken into account. I think this might lead to situations in which the ``MAXACTIVETOKENUSER`` policy can be bypassed?

Also, we use the "fetch number of tokens, get maximum number of allowed tokens, throw error if maximum is exceeded" pattern three times already (1x ``check_max_token_realm``, 2x ``check_max_token_user``). We could move the code to its own function:
```python
    def check_limit_exceeded(tokenobject_list, limit_list):
        already_assigned_tokens = len(tokenobject_list)
        max_value = max([int(x) for x in limit_list])
        if already_assigned_tokens >= max_value:
             g.audit_object.add_policy(limit_list.get(str(max_value)))
             raise PolicyError(ERROR)
```

This would simplify the code. What do you think?

privacyidea/api/lib/prepolicy.py
```diff
@@ -711,6 +714,30 @@ def check_max_token_user(request=None, action=None):
...
+            if serial:
+                for tok in tokenobject_list:
+                    if tok.token.serial == serial and tok.is_active():
```
fredreichbier: I think ``tok.is_active()`` should always be true, because we pass ``active=True`` to ``get_tokens`` above.

privacyidea/api/lib/prepolicy.py
cornelinux: Right - I forgot this. Will do.

privacyidea/api/lib/prepolicy.py
cornelinux: You are right. Will do.

privacyidea/api/lib/prepolicy.py
cornelinux: In regards to get_tokens it is not that simple.
The first get_tokens could not be called, since there is no MAX_TOKEN policy.
So we would have to check, if a tokenlist is already filled and then either reduce the list to an active-token-list or do a new get_tokens call. The code looks a bit cluttered. Given that most people probably will either use a max token or a max-active-token I would like to keep two get_tokens calls. (Open for discussion)

3 comments in a comment is a bit difficult to handle.

privacyidea/api/lib/prepolicy.py
cornelinux: Will do

privacyidea/api/lib/prepolicy.py
cornelinux: Done the early True-exist in 3ad5a51d

privacyidea/api/lib/prepolicy.py
cornelinux: Done in 3ad5a51d

privacyidea/api/lib/prepolicy.py
cornelinux: Done in 3ad5a51d

privacyidea/api/lib/prepolicy.py
cornelinux: done in 8d037b56

PR #1651, opened 2019-05-28
Fix the logic that max_token and max_active_token

cornelinux:
will work at the same time.

Added to #1241

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -690,6 +690,7 @@ def check_max_token_user(request=None, action=None):
     :return: True otherwise raises an Exception
     """
     ERROR = "The number of tokens for this user is limited!"
+    ERROR_ACTIVE = "The number of active tokens for this user is limited!"
     params = request.all_data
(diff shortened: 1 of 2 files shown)


privacyidea/api/lib/prepolicy.py
```diff
@@ -690,6 +690,7 @@ def check_max_token_user(request=None, action=None):
     :return: True otherwise raises an Exception
     """
     ERROR = "The number of tokens for this user is limited!"
+    ERROR_ACTIVE = "The number of active tokens for this user is limited!"
     params = request.all_data
     user_object = get_user_from_param(params)
```
fredreichbier: There is one problem here: If an admin requests ``/enable/<serial>``, we do not have a ``user`` in the ``request.all_data``, so ``user_object.login`` is empty and the maximum number of tokens is not checked.

cornelinux: You are right. That's pitty.

fredreichbier: OK, I opened #1652 to keep track of the issue.

PR #1654, opened 2019-05-31
Make LDAP resolver server pool persistent per-process

### LINKED ISSUE #1396: The Serverpool in LDAP resolver is not persistent
The serverpool in the ldap resolver is not persistent but rather created on each ``checkPass``.

As far as normal quering is concerned, it is persistent if the bind is still active. Nevertheless we should move the serverpool into a member ``self.serverpool``.

fredreichbier:
This implements a per-process LDAP server pool.

Until now, every resolver object uses its own ``ldap3.ServerPool``.  But resolver objects are only alive for the duration of the current request. So if a LDAP server goes offline at some point, this information gets lost when the request ends, and subsequent requests will re-try to contact the LDAP server.

With this PR, the admin can check the "Per-process server pool" checkbox in the LDAP resolver configuration. If it is checked, all LDAP resolvers with the same server pool configuration will use the same shared server pool instance. This PR defines a ``LockingServerPool`` class which uses locks to synchronize access to the server pool state, in order to rule out race conditions if two threads concurrently manipulate the server pool state.

I decided to implement this as an opt-in feature to avoid unforeseen issues with existing installations.

I'm happy about any comments :)

Closes #1396.

### DIFF
--- a/privacyidea/static/components/config/views/config.resolvers.ldap.html
+++ b/privacyidea/static/components/config/views/config.resolvers.ldap.html
@@ -163,13 +163,13 @@
         </div>
     </div>
     <div class="form-group">
-        <label for="serverpool-persistent" class="col-sm-3 control-label" translate>
-            Per-process server pool
-        </label>
-        <div class="col-sm-9">
-            <input name="serverpool-persistent"
-                   ng-model="params.SERVERPOOL_PERSISTENT"
-                   type="checkbox" />
+        <label for="sizelimit" class="col-sm-3 control-label"
+                translate>Size Limit</label>
+
+        <div class="col-sm-3">
+            <input name="sizelimit" class="form-control"
+                   ng-model="params.SIZELIMIT" required
+                   placeholder="500"/>
         </div>
     </div>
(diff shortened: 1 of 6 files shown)


privacyidea/static/components/config/views/config.resolvers.ldap.html
```diff
@@ -162,6 +162,16 @@
                    placeholder="120"/>
         </div>
     </div>
+    <div class="form-group">
+        <label for="serverpool-persistent" class="col-sm-3 control-label" translate>
+            Per-process server pool
```
cornelinux: Put this checkbox under the form-group of retry-rounds and pool skip.

And the sizelimit above the retry-rounds and pool skip.

Also add a ``help-block`` under the checkbox (like in "editable") with the text:

     This activates a persistent server pool between requests.

fredreichbier: Done in add0f03

fredreichbier:
I added some documentation in 39fa5d3 -- I hope it's understandable :)

PR #1656, opened 2019-06-03
Fix height of checkboxes

### LINKED ISSUE #1655: Height of checkboxes
The checkboxes are higher than the texts. This should be fixed

Bildschirmfoto vom 2019-06-03 12-44-45

Mipronimo:
fixes #1655


fredreichbier:
Unfortunately this doesn't look very nice on Firefox :-( It seems like fixing this properly would involve way more CSS magic than we thought, so I'm closing this.

PR #1659, opened 2019-06-04
Add focus for filter input

Mipronimo:
If you enable a filter, the input field should be focused automatically.

(Tested with Google Chrome)


fredreichbier:
Tested in Firefox, looks good to me!

Fixes #1661

Issue #1436, opened 2019-02-08
Policies could be dependent on any user attribute


cornelinux:
The conditions for a policy and event handler to apply could depend on any arbitrary user attribute (like a group membership)

See https://community.privacyidea.org/t/resolver-and-user-token-relation/941

cornelinux:
Also the event handler conditions should be triggered via such user attributes.
The interesting thing is, that then there is no need to be depent on the resolver, since the resolver can very well also be such an attribute.

cornelinux:
To be more flexible in policies we would need to untangle the policy-table, which currently contains

* The policy skeleton
   * the name
   * the scope
* all the conditions
  * user
  * time
  * client IP
  * active
* the action ("action")

We should probably split it up like the eventhandler.
THis way it would be much easier, to add new conditions for policies.
Split into the DB tables:

* Policy
* PolicyCondition
* PolicyAction (Already now a policy can contain several actions, comma separated - how nasty is that!)

fredreichbier:
I think, conceptually, wouldn't it be cool if policy conditions and event handler conditions could have the same logic? In the end, they both match on some properties of the current request.

In terms of DB tables, I agree with your thoughts above: I think it would be more sensible to store the policy conditions and actions in their own tables.

... but this could potentially make policy queries a bit slower. I don't know :-)

cornelinux:
As far as the speed is concerned: Currently the policies are fetched at the beginning of the request. This problably would not be changed. We then have a request wide policy object. (We can even keep the policy-object without rereading it from the database too often - I think we already do this)

Condition Handling: At the first glance it might look the same. But at a second it is rather different since the logic of how events are triggered and how policies are fetched (think of get_action_value) are different. So we need to take a third look! ;-)

Marc-Axiad:
Hi Cornelinux, I think I am partly at the origin of the request, a customer wants to apply different policies to end users based on group membership (which can be translated indeed as a user attribute if looking at "memberof"). Currently we implemented this using resolvers but the problem comes when the customer move users from one group to another, users lose their device assignment. Anyway, if you have any questions about the customer use case please let me know!

fredreichbier:
I still suspect that policy matching and event handler matching are very close to one another -- maybe we should try to unify the two concepts. This could help with several issues, e.g. #1102, #1645.

As a first step, I think it may be useful to split the policy table as @cornelinux suggested above (especially because performance isn't really a concern here, as you said).

[...]

PR #1662, opened 2019-06-05
Move policy conditions to a separate table

fredreichbier:
As drafted in #1436, this PR splits off policy conditions to a separate table. Now, every policy holds a collection of policy conditions. Each condition consists of a key, comparator and value.

This PR does not change the API, and changes as little code as possible:
* A migration script converts the so-called "legacy condition" ``user``, ``realm``, ... fields of the policy table to policy conditions.
* When constructing the policy dictionary in ``Policy.get()``, we just add legacy conditions to the policy dictionary. So the return value of ``Policy.get()`` does not change, and we don't have to adjust policy matching at all.
* Then, we only need to change ``set_policy`` such that it properly creates policy conditions.

Some open questions:
* The policy table contains the ``check_all_resolvers`` flag. I kept this for now, because I'm not sure how it would translate to a policy condition.
* Analogously to the event handler conditions, the policy conditions contain a ``comparator`` field. Currently, it is always set to ``"equal"``, but it could allow more sophisticated comparators in the future. However, it complicates the code a bit.
This PR represents conditions as a list of tuples:
```python
[("user", "equal", "foo, bar"),("realm", "equal", "baz")]
```
If we ditch comparators, we could just represent conditions with a dictionary:
```python
{"user": "foo, bar", "realm": "baz"}
```
... also, the event handlers do not seem to use comparators anyway :-)

### DIFF
--- a/migrations/versions/7ac61e64e16d_.py
+++ b/migrations/versions/7ac61e64e16d_.py
@@ -1,0 +1,112 @@
+"""Move policy conditions to a separate table
+
+Revision ID: 7ac61e64e16d
+Revises: 849170064430
+Create Date: 2019-06-04 16:06:22.890250
+
+"""
+
+# revision identifiers, used by Alembic.
+from sqlalchemy import Sequence, orm
+from sqlalchemy.ext.declarative import declarative_base
+
+from privacyidea.models import Policy
+
+revision = '7ac61e64e16d'
+down_revision = '849170064430'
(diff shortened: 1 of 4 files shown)


fredreichbier:
Closing this, because we will keep the ``Policy`` table as it is, and only keep *additional, more complex conditions* in ``PolicyCondition``.

PR #1664, opened 2019-06-05
Add MAC calculation for pksc export

### LINKED ISSUE #1663: pskc export does not calculate mac
According to [RFC6030](https://tools.ietf.org/html/rfc6030#section-6.1.1) we need to calculate a MAC if an encryption algorithm without integrity checks is use (like `AES-128-CBC`).
We already create a MAC key and encrypt it accordingly but the MAC for the token secrets was not calculated.

plettich:
Also fixed a missing import in the token janitor.

Fixes #1663

### DIFF
--- a/tools/privacyidea-token-janitor
+++ b/tools/privacyidea-token-janitor
@@ -47,7 +47,8 @@ from privacyidea.lib.policy import ACTION
 from privacyidea.lib.utils import parse_legacy_time
 from privacyidea.lib.importotp import export_pskc
 from privacyidea.lib.token import (get_tokens, remove_token, enable_token,
-                                   unassign_token, import_token)
+                                   unassign_token, import_token,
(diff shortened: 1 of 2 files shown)


tools/privacyidea-token-janitor
```diff
@@ -47,7 +47,8 @@ from privacyidea.lib.policy import ACTION
 from privacyidea.lib.utils import parse_legacy_time
 from privacyidea.lib.importotp import export_pskc
 from privacyidea.lib.token import (get_tokens, remove_token, enable_token,
-                                   unassign_token, import_token)
+                                   unassign_token, import_token,
+                                   get_tokens_paginated_generator)
```
cornelinux: What do we get_tokens_paginated_generator for?
Is it used somewhere?

cornelinux: OK, I See.

PR #1665, opened 2019-06-06
CSS code cleanup

### LINKED ISSUE #1657: CSS cleanup
As mentioned in https://github.com/privacyidea/privacyidea/pull/1656/commits/34b97d179fab0979db412fc96affa2e22f8abfdc, we should change the way how we edit our css files.

An idea is to create a file for each region like menu, sidebar, content, etc. and put the changes where they belong to.

Mipronimo:
As mentioned in #1657 we want to edit the style in a better way.
I started to clean our css directory a bit:

* moved the images to img/
* added css files for regions
  * content
  * menu / navbar
  * papertoken
  * signin
  * table ui
* put changes from bootstrap.css in the correct file

closes #1657


doc/faq/customization.rst
```diff
@@ -130,7 +130,7 @@ do it like this::
 Logo
 ~~~~
 
-The default logo is located at ``privacyidea/static/css/privacyIDEA1.png``.
+The default logo is located at ``privacyidea/static/img/privacyIDEA1.png``.
```
cornelinux: I would argue that a logo is part of the style.
Currently we do not have an image directory.
Also this change requires, that all users, who customized their logo (which is actually severals) will have to manually change the location, where they dropped their logo file.
I recommend to not change the logo location!

privacyidea/static/templates/header.html
```diff
@@ -7,19 +7,19 @@
     <meta name="description" content="">
     <meta name="author" content="Cornelius Kölbel" >
     <link rel="icon" type="image/png" href="{{ instance }}/static/favicon.png">
+    <link href="{{ instance }}/static/contrib/css/bootstrap.css" rel="stylesheet">
```
cornelinux: OK, I see. You moved this to the top, so that other stylesheets may overwrite the original bootstrap css. +1

doc/faq/customization.rst
Mipronimo: Fixed in c66a29f7

PR #1669, opened 2019-06-11
Fix U2F enrollment for Chrome 72+

### LINKED ISSUE #1636: Failed to enroll U2F token on latest Chrome
# What did you try to do?

Enroll U2F token on Chrome

Being able to enroll U2F token on Chrome

Failed to enroll U2F token on Chrome

# Configuration

* privacyIDEA Version: 2.23.5

* Installation method: from source

## more details: 

* OS: Amazon LInux

* Webserver: nginx

* Tokendatabase: postgresql

# Log file

U2F recently fails to enroll with error pop up: "Bad U2F Request / challenge must be base64url encoded"

My Chrome version is Version 74.0.3729.157 (Official Build) (64-bit)

I think this maybe related to this change in Chrome Crypto:
https://github.com/chromium/chromium/commit/ceb92979d1f3a77146ed68f8484636ce63073625#diff-bbd4efb6053a9f322d1ba351e036bd71

I am using PrivacyIDEA 2.23.5. Just want to make sure this is already fixed in 3.0?

Thank you and best regards,

Quynh Nguyen

Mipronimo:
In Chrome 72, the base64url decoding behaviour changed.
This could be fixed, if we use the base64_encode function for u2f registration and not only for signing.

I did not test it in a good way yet. But I noticed that this works in my scenario with both, Google Chrome and Mozilla Firefox.

resolves #1636

### DIFF
--- a/privacyidea/lib/tokens/u2ftoken.py
+++ b/privacyidea/lib/tokens/u2ftoken.py
@@ -348,7 +348,7 @@ class U2fTokenClass(TokenClass):
             if not app_id:
                 raise TokenAdminError(_("You need to define the appId in the "
                                         "token config!"))
-            nonce = urlsafe_b64encode_and_unicode(geturandom(32))
+            nonce = url_encode(geturandom(32))
             response_detail = TokenClass.get_init_detail(self, params, user)
             register_request = {"version": U2F_Version,
                                 "challenge": nonce,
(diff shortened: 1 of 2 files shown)


Mipronimo:
> You need to fix the one test. It checks for the challenge to be 44 chars long.
> But it seems the trailing "=" is not "missing", so the length of the challenge is only 43 chars long.

done in 382ba41

PR #1671, opened 2019-06-11
Allow to use push token in normal /validate/check

### LINKED ISSUE #1583: Allow push token to delay the response
When authenticating with a push token, the response of the API request to ``validate/check`` is directly returned - for good reasons.
Applications then should [poll for the confirmation of the push token](https://github.com/privacyidea/privacyidea/wiki/concept%3A-PushToken#polling-endpoint-for-application).
Although this is a cleaner design, we can offer an alternative.

The response of the ``validate/check`` request is held back until the push token was confirmed.

Some problems: If the user has several tokens, how should we react:
1. another token successfully authenticated: return directly?
2. all other tokens failed to authenticate: wait?

We need at least the following settings:
1. activate this behaviour
2. timeout, how long privacyIDEA should wait for an answer of the push token

cornelinux:
Setting a policy the push token can be used in one
single /validate/check request.
The request will simply wait to return until the challenge
is answered.

Closes #1583

### DIFF
--- a/doc/policies/authentication.rst
+++ b/doc/policies/authentication.rst
@@ -362,6 +362,32 @@ This is the title of the push notification that is displayed
 on the user's smartphone during the login process with
 a :ref:`push_token`.
 
+.. _policy_push_wait:
+
+push_wait
+~~~~~~~~~
+
+.. index:: push token, push direct authentication
+
+type: int
+
+This can be set to a number of seconds. If this is set, the authentication
+with a push token is only performed via one request to ``/validate/check``.
+The HTTP request to ``/validate/check`` will wait up to this number of
+seconds and check, if the push challenge was confirmed by the user.
+
+This way push tokens can be used with any non-push-capable applications.
+
+Sensible numbers might be 10 or 20 seconds.
+
+.. note:: This behaviour can interfere with other tokentypes. Even if
+   the user also has a normal HOTP token, the ``/validate/check`` request
+   will only return after this number of seconds.
+
+.. warning:: Using simple webserver setups like Apache WSGI this actually
+   can block all available worker threads!
+
+
 .. _policy_challenge_text:
 
 challenge_text, challenge_text_header, challenge_test_footer
--- a/doc/policies/authentication.rst
+++ b/doc/policies/authentication.rst
@@ -385,7 +385,9 @@ Sensible numbers might be 10 or 20 seconds.
    will only return after this number of seconds.
 
 .. warning:: Using simple webserver setups like Apache WSGI this actually
-   can block all available worker threads!
+   can block all available worker threads, which will cause privacyIDEA
+   to become unresponsive if the number of open PUSH challenges exceeds
+   the number of available worker threads!
 
 
 .. _policy_challenge_text:
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -1295,6 +1295,39 @@ def save_client_application_type(request, action):
     return True
 
 
+def pushtoken_wait(request, action):
+    """
+    This is a auth specific wrapper to decorate /validate/check
+    According to the policy scope=SCOPE.AUTH, action=push_wait
+
+    :param request:
+    :param action:
+    :return:
+    """
+    user_object = request.User
+    if user_object:
+        token_user = user_object.login
+        token_realm = user_object.realm
+        token_resolver = user_object.resolver
+    else:
+        token_realm = token_resolver = token_user = None
+
+    waiting = g.policy_object.get_action_values(
+        action=PUSH_ACTION.WAIT,
+        scope=SCOPE.AUTH,
+        realm=token_realm,
+        user=token_user,
+        resolver=token_resolver,
+        client=g.client_ip,
+        audit_data=g.audit_object.audit_data,
+        allow_white_space_in_action=True,
+        unique=True)
+    if len(waiting) >= 1:
+        request.all_data[PUSH_ACTION.WAIT] = max([int(i) for i in list(waiting)])
+    else:
+        request.all_data[PUSH_ACTION.WAIT] = False
+
+
 def pushtoken_add_config(request, action):
     """
     This is a token specific wrapper for push token for the endpoint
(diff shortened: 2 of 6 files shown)


privacyidea/lib/tokens/pushtoken.py
```diff
@@ -561,6 +571,47 @@ def create_challenge(self, transactionid=None, options=None):
...
+        if pin_match:
+            waiting = int(options.get(PUSH_ACTION.WAIT, 20))
+            # Trigger the challenge
+            _t, _m, transaction_id, _attr = self.create_challenge(options=options)
+            # now we need to check and wait for the response to be answered in the challenge table
+            starttime = time.time()
+            while True:
+                # TODO: It is not clear if in a process/threaded environment like apache the challenges are updated in one thread!
+                otp_counter = self.check_challenge_response(options={"transaction_id": transaction_id})
+                if otp_counter >= 0 or (time.time() - starttime) > waiting:
```
fredreichbier: This may be nitpicking, but: ``time.time()`` is not guaranteed to always increase. So if the system time is reset while we are in this loop, the loop will never terminate.
So we could do something like
```python
elapsed_time = time.time() - starttime
if otp_counter >=0 or elapsed_time > waiting or elapsed_time < 0:
    break
```
If the system time is reset, this will cause the authentication to fail, but at least we don't loop forever.

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -561,6 +571,47 @@ def create_challenge(self, transactionid=None, options=None):
...
+        reply = None
+        pin_match = self.check_pin(passw, user=user, options=options)
+        if pin_match:
+            waiting = int(options.get(PUSH_ACTION.WAIT, 20))
+            # Trigger the challenge
+            _t, _m, transaction_id, _attr = self.create_challenge(options=options)
+            # now we need to check and wait for the response to be answered in the challenge table
+            starttime = time.time()
+            while True:
+                # TODO: It is not clear if in a process/threaded environment like apache the challenges are updated in one thread!
```
fredreichbier: We should definitely test what happens in a real-world environment before merging this :)

tests/test_lib_tokens_push.py
```diff
@@ -392,6 +393,44 @@ def test_03_api_authenticate_client(self):
...
+                                               data={"user": "cornelius",
+                                                     "realm": self.realm1,
+                                                     "pass": "pushpin"}):
+                res = self.app.full_dispatch_request()
+                self.assertTrue(res.status_code == 200, res)
+                jsonresp = json.loads(res.data.decode('utf8'))
+                # We successfully authenticated! YEAH!
+                self.assertTrue(jsonresp.get("result").get("value"))
+                self.assertTrue(jsonresp.get("result").get("status"))
+                self.assertEqual(jsonresp.get("detail").get("serial"), tokenobj.token.serial)
```
fredreichbier: What do you think about adding a test for the case that the challenge is not marked as accepted during the timeout? Maybe with a shorter timeout then, like 1 or 2 seconds :)

... and maybe also for the case that the PIN is not correct.

Mipronimo:
I tested it with httpie on Ubuntu 19.04 and it did not work for me.
After adding the policy push_wait: 29 (30 is the timeout for httpie :smile: ), the Android application is not able to reach the server.

doc/policies/authentication.rst
```diff
@@ -362,6 +362,32 @@ This is the title of the push notification that is displayed
...
+This way push tokens can be used with any non-push-capable applications.
+
+Sensible numbers might be 10 or 20 seconds.
+
+.. note:: This behaviour can interfere with other tokentypes. Even if
+   the user also has a normal HOTP token, the ``/validate/check`` request
+   will only return after this number of seconds.
+
+.. warning:: Using simple webserver setups like Apache WSGI this actually
+   can block all available worker threads!
```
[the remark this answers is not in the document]
cornelinux: Will do

privacyidea/api/lib/prepolicy.py
```diff
@@ -1295,6 +1295,39 @@ def save_client_application_type(request, action):
...
+        scope=SCOPE.AUTH,
+        realm=token_realm,
+        user=token_user,
+        resolver=token_resolver,
+        client=g.client_ip,
+        audit_data=g.audit_object.audit_data,
+        allow_white_space_in_action=True,
+        unique=True)
+    if len(waiting) >= 1:
+        request.all_data[PUSH_ACTION.WAIT] = max([int(i) for i in list(waiting)])
```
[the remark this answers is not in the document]
cornelinux: I think you are totally right!

privacyidea/lib/tokens/pushtoken.py
cornelinux: Now I understood. I like it. Will do.

tests/test_lib_tokens_push.py
cornelinux: Will do

doc/policies/authentication.rst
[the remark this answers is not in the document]
cornelinux: done in d370c8f9

privacyidea/api/lib/prepolicy.py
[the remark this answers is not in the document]
cornelinux: done in d370c8f9

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in d370c8f9

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in d370c8f9

tests/test_lib_tokens_push.py
cornelinux: done in d370c8f9

privacyidea/lib/tokens/pushtoken.py
cornelinux: ...well...

doc/policies/authentication.rst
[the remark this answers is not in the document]
fredreichbier: thanks!

PR #1672, opened 2019-06-12
Add possibility to use user object for policy matching

fredreichbier:
This adds a ``user_object`` parameter to ``get_policies`` and ``get_action_values``.

One can either pass ``user``/``realm``/``resolver`` or ``user_object``. In the latter case, user/realm/resolver are read from the user object.

Working on #1436/#1645, as this will allow us to implement conditions on arbitrary user attributes in a simple way.

This also fixes some cases in which the user resolver is not taken into account when matching policies (see #1646)

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -146,8 +146,7 @@ def init_random_pin(request=None, action=None):
     # get the length of the random PIN from the policies
     pin_pols = policy_object.get_action_values(action=ACTION.OTPPINRANDOM,
                                                scope=SCOPE.ENROLL,
-                                               user=user_object.login,
-                                               realm=user_object.realm,
+                                               user_object=user_object,
                                                client=g.client_ip,
                                                unique=True,
                                                audit_data=g.audit_object.audit_data)
@@ -159,7 +158,7 @@ def init_random_pin(request=None, action=None):
         # handle the PIN
         handle_pols = policy_object.get_action_values(
             action=ACTION.PINHANDLING, scope=SCOPE.ENROLL,
-            user=user_object.login, realm=user_object.realm,
+            user_object=user_object,
             client=g.client_ip,
             audit_data=g.audit_object.audit_data)
         # We can have more than one pin handler policy. So we can process the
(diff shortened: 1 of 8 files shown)


privacyidea/api/lib/prepolicy.py
```diff
@@ -1431,19 +1404,10 @@ def u2ftoken_allowed(request, action):
...
-        else:
-            token_realm = token_resolver = token_user = None
-
         allowed_certs_pols = policy_object.get_action_values(
             U2FACTION.REQ,
             scope=SCOPE.ENROLL,
-            realm=token_realm,
-            user=token_user,
-            resolver=token_resolver,
+            user_object=request.User if request.User else None,
```
cornelinux: One thing: It ``request`` for some reason would not contain the attribute ``User`` I think this would also raise an exception?

If ``request.User`` is empty like ``User()`` I think we can also pass it this way.

So, if we are **certain** that ``request`` **always** has an attribute ``User`` I simply would pass this

    user_object=request.User

and handle the empty User object in the policy methods.

privacyidea/lib/policy.py
```diff
@@ -493,6 +501,13 @@ def get_policies(self, name=None, scope=None, realm=None, active=None,
         :return: list of policies
         :rtype: list of dicts
         """
+        if user_object is not None:
```
cornelinux: We could say

    if user_object:

this would also be fales, if user_object would be an empty user_object - right? (see my comment above)

fredreichbier: I'll try this!

fredreichbier:
We define the meaning of the arguments to ``get_policies`` as follows: ``get_policies`` filters the available policies according to a filter criterion given by the arguments. If we pass ``None`` for a specific argument, we don't care about the attributes specified in the policies, we want them all. If we pass a value for a specific argument, we only want policies with empty attributes OR matching attributes.

Examples:

| | ``get_policies(resolver=None)`` | ``get_policies(resolver='')`` | ``get_policies(resolver='reso')`` |
| --- | ----- | ------ | ----- |
| ``policy.resolver = ''`` | match | match | match |
| ``policy.resolver = 'reso'`` | match | no match | match |

With the new ``user_object`` argument, we want to implement the same behavior:
* ``get_policies(user_object=None)``: we don't care about the user attributes of policies
* ``get_policies(user_object=User())``: we only want policies with empty user, resolver, realm
* ``get_policies(user_object=User("foo", "bar", "baz"))`` we only want policies matching the user (or the empty user), resolver (or the empty resolver), realm (or the empty realm)

TODO for this PR:
* Check that ``get_policies`` actually implements this behavior :)
* If we don't already have tests for these cases, write some
* Document this in the ``get_policies`` docstring

TODO after this PR.
* Check that all callers of ``get_policies`` and ``get_action_values`` actually call the methods according to the specification above

PR #1675, opened 2019-06-12
Add invalid char message for policy name

### LINKED ISSUE #1674: Show disallowed characters in policy name
If you create a policy, privacyIDEA does not check its name. You could enter a name like "foo/bar", but privacyIDEA will say 404, because "/policy/foo/bar" could not be found.

So we should may add a message, which says something like "The character '/' is not allowed here."

Mipronimo:
If a user changes the policy name and adds an '/', it will fail.
This commit checks it and prints an error message.
closes #1674


Mipronimo:
We may should discuss it a bit in #1674 
I will close it here for now.

PR #1678, opened 2019-06-12
Fix db migration with postgres

### LINKED ISSUE #1676: Database migration to PI 3.0.1 fails with postgres db
When the db upgrade script creates a new table with a postgres backend, the required sequence will not be added, instead a default sequence is created which breaks further use of the table.
We also need to make sure, that the corresponding dialect supports sequences before creating them.
Actually this breaks all upgrades with postgres where tables are created in the migration script (or if tables are created before we added sequences to the tables).

plettich:
PostgreSQL (and Oracle and SQLite) use sequences for autoincrement columns.
If no sequence is given, one will be created automatically which clashes
with the required one in `models.py`
We have to create the required sequence manually but only if the backend
supports it.
Actually this affects all migration scripts which create columns with an
auto_increment attribute.

Fixes #1676


plettich:
I'll delete this PR and create a new one based on branch-3.0

PR #1683, opened 2019-06-13
Fix empty client_ip condition for pre handler

### LINKED ISSUE #1658: Configuring pre-eventhandler with empty condition breaks authentication
When adding a pre-eventhandler and activating a condition **without** specifying a value, we'll get an entry in the eventhandlercondition table without a value as well.
If this eventhandler is triggered, it fails with a `IndexError: string index out of range`.
In case of a **post**-eventhandler, the empty condition value is silently ignored and no entry in the eventhandlercondition table is created.
I guess, this is an UI issue, but we should check the the table entries for consistency anyway before using the values (See #1596).

cornelinux:
If the client_ip in the condition would be empty
we would get a string index out of range error.

Fixes #1658

### DIFF
--- a/privacyidea/lib/utils.py
+++ b/privacyidea/lib/utils.py
@@ -650,6 +650,8 @@ def check_ip_in_policy(client_ip, policy):
     """
     client_found = False
     client_excluded = False
+    # Remove empty strings from the list
+    policy = list(filter(None, policy))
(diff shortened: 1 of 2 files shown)


cornelinux:
@fredreichbier I think you had this request. Please take a look and merge, if ok.

PR #1687, opened 2019-06-14
Add welcome message if user has no tokens

### LINKED ISSUE #1074: Show welcome message if user has no token
The token list (https://github.com/privacyidea/privacyidea/blob/master/privacyidea/static/components/token/views/token.list.html) could show a welcome message for a normal user (not an admin), if this user does not have any tokens, yet.

This is the first view the user sees int the Web UI, so we could add a configurable template, that gets displayed.
Maybe we simply use a template like the *top* and *bottom* templates in the wizard.

cornelinux:
It a user logs in, we check if a welcome policy is set.
If the user has not tokens, a welcome message is displayed.

Closes #1074

### DIFF
--- a/privacyidea/static/components/dialogs/views/dialog.no.token.html
+++ b/privacyidea/static/components/dialogs/views/dialog.no.token.html
@@ -1,0 +1,47 @@
+<!-- Modal -->
+
+<!-- this dialog is displayed, if the logged in user is or role "user" and he has no tokens. -->
+
+<div class="modal fade lock-dialog" id="dialogNoToken" tabindex="-1"
+     role="dialog"
+     ng-show="dialogNoToken == true"
+     data-backdrop="static" data-keyboard="false"
+     aria-labelledby="myModalNoToken"
+     aria-hidden="true">
+    <div class="modal-dialog">
+
+        <div class="modal-content">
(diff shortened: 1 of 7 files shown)


privacyidea/static/components/dialogs/views/dialog.no.token.html
```diff
@@ -0,0 +1,47 @@
...
+        <div class="modal-content">
+            <div class="modal-header">
+                <h4 class="modal-title" id="myModalNoToken" translate
+                >Welcome</h4>
+            </div>
+
+            <div class="modal-body" ng-show="browserLanguage == 'en'">
+                privacyIDEA is a system to manage your tokens.
+                You currently have no token assigned.
+                You can manage your own tokens here - you can create new tokens, set the PIN of the tokens, disable tokens
```
plettich: I guess, evaluating all the user permissions is a bit too much but we should at least mention, that the allowed actions could depend on the configuration.

cornelinux: I assumed that, if this feature is used, with a 99.8% propability the admin would create his own template anyways. This is why I chose a very bad text! ;-)

PR #1690, opened 2019-06-18
Allow to use transaction_id as parameter for GET /token/challenges

### LINKED ISSUE #1689: Add transaction_id to GET token/challenges
Certain tokens (like Push, or maybe tiqr) work better, when a service account can fetch the challenges to verify, if the response was answered correctly.

Thus we should add a parameter ``transaction_id`` to the API endpoint.

cornelinux:
Closes #1689


fredreichbier:
looks good!

PR #1696, opened 2019-06-19
Add additional sanity check for responses to challenges

### LINKED ISSUE #1695: Allow additional checks while answering challenge response
During a challenge response the user can sends possible responses to privacyIDEA. 
These get checked against the challenge table.

The tokentype should define, which checks are performed in addition to checking if the challenge matches and is still valid.

The tokenclass can define a method ``check_challenge_response()`` which would be added in this if-statement.

https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/token.py#L2213

Within this method a tokenclass could call the existing check_failcount or run additional checks depending on the tokentype. The challenge table also counts the response tries, so this could be checked as well.

A certificate or smartcard token in the future could also use this hook to pull CRLs...

cornelinux:
Closes #1695

### DIFF
--- a/privacyidea/lib/token.py
+++ b/privacyidea/lib/token.py
@@ -2082,7 +2082,7 @@ def check_token_list(tokenobject_list, passw, user=None, options=None, allow_res
     valid_token_list = []
 
     # Remove locked tokens from tokenobject_list
-    if len(tokenobject_list) > 1:
+    if len(tokenobject_list) > 0:
         tokenobject_list = [token for token in tokenobject_list if not token.is_revoked()]
 
         if len(tokenobject_list) == 0:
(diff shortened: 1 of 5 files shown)


privacyidea/lib/token.py
```diff
@@ -2210,8 +2210,18 @@ def check_token_list(tokenobject_list, passw, user=None, options=None, allow_res
                                                     options=options) >= 0:
                 reply_dict["serial"] = tokenobject.token.serial
                 matching_challenge = True
-                if tokenobject.is_active():
-                    # OTP matches
+                messages = []
```
fredreichbier: ``messages`` is unused currently. I guess we should either remove it, or pass it to ``is_fit_for_challenge`` and add the messages to the reply. What do you think?

cornelinux: It thinkg the messages was not filled anyways, this is why I removed it.
And I forgot to remove it from the function definition of ``is_fit_for_challenge``.

I think: Othe than dict, a list is not passed as reference and will not be filled in the function.

cornelinux: So I think I should remove the messages.

cornelinux: Obviously I am wrong. It would. But it looks like it does not (in our tests). I will add the messges

PR #1698, opened 2019-06-19
Add a tokenmode "outofband"

### LINKED ISSUE #1697: Do not increase PIN when checking for an answered challenge.
This refers to
https://community.privacyidea.org/t/tiqr-authentication-fail-counter-increase-if-we-query-the-status/1061

There are certain tokentypes, that run a bit different authentication workflow.
The TiQR and the PUSH token are both challenge response token.
The challenge is sent by issuing the PIN to the ``/validate/check`` endpoint.
But the repsonse to the challenge is not sent to the /validate/check endpoint but to another endpoint of privacyIDEA.
/validate/check can simply act as an interface to check, if the challenge has been asnwered correctly. So the /validate/check in this case is not prune to brute force and we do not need to increase the PIN.

This was tried in PR #1684 but in a not quite nice and generic way.

Tokenclasses need to have a flag, if they are actually such tokens.

cornelinux:
challenge response tokens like "PUSH" and "TiQR" do the authentication
out of band. I.e. the 2nd /validate/check request or /auth request is
not used to send the response to the challenge but to verify, if the
challenge was answered correct. Therefore we must not increase the
failcounter in this case.

Closes #1697

### DIFF
--- a/privacyidea/lib/token.py
+++ b/privacyidea/lib/token.py
@@ -2237,7 +2237,8 @@ def check_token_list(tokenobject_list, passw, user=None, options=None, allow_res
             # We did not find any successful response, so we need to increase the
             # failcounters
             for token_obj in challenge_response_token_list:
-                token_obj.inc_failcount()
+                if not token_obj.is_outofband():
+                    token_obj.inc_failcount()
             if not matching_challenge:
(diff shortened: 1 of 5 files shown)


Mipronimo:
lgtm. :+1:

PR #1700, opened 2019-06-20
Add additional test in is_challenge_response

### LINKED ISSUE #1699: add more sophisticated check in is_challenge_response
In the lib/tokenclass.py the method ``is_challenge_response`` only checks, if the request (options) contain a transaction_id to identify if this is supposed to be a response to a challenge.

We need to investigate, if we also should check, if the given transaction_id actually is a running transaction_id to this very token.

cornelinux:
In addition to check if the request contains a
transaction_id, we also check, if the
transaction_id actually matches an existing challenge
of this very token.

Closes #1699

### DIFF
--- a/privacyidea/lib/tokenclass.py
+++ b/privacyidea/lib/tokenclass.py
@@ -1360,8 +1360,12 @@ class TokenClass(object):
         """
         options = options or {}
         challenge_response = False
-        if "state" in options or "transaction_id" in options:
-            challenge_response = True
+        transaction_id = options.get("transaction_id") or options.get("state")
+        if transaction_id:
+            # Now we also need to check, if the transaction_id is an entry to the
+            # serial number of this token
+            chals = get_challenges(serial=self.token.serial, transaction_id=transaction_id)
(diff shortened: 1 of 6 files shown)


fredreichbier:
This looks very reasonable for me! Consider the following situation:
* A user has two EMail tokens A and B with PINs ``pina`` and ``pinb``
* The user triggers a challenge for token A (by sending the PIN ``pina``), with transaction ID ``14191784131548810258``
* The user sends an incorrect OTP value while specifying the transaction ID ``14191784131548810258``

On current master: The failcounters of A *and* B are increased
With this PR: Only the failcounter of A is increased

I think the new behavior with this PR makes more sense, but we should add an API-level test for that.

I can also do that, in order to save you some work.

fredreichbier:
My commit 56e5fb6 adds a test for the scenario above (just with challenge-response HOTP tokens instead of email tokens). I hope it makes sense :)

cornelinux:
nice test! :-)
So I merge ;-)

PR #1704, opened 2019-06-21
Allow several tags in email token

### LINKED ISSUE #1703: Allow more tags in email token
The email token sends an emal with an OTP value to the user's email address.
The email text can contain the tags *otp* and *serial*.

We should use the same tags and the same code as we are using in the email notification handler!
Including esepcially the **username**.

cornelinux:
The email token now can also contain the tokenowner

Closes #1703


privacyidea/lib/utils.py
```diff
@@ -1192,4 +1192,56 @@ def split_pin_pass(passw, otplen, prependpin):
...
+                tokentype=tokentype,
+                registrationcode=registrationcode,
+                recipient_givenname=recipient.get("givenname"),
+                recipient_surname=recipient.get("surname"),
+                googleurl_value=googleurl_value,
+                time=time,
+                date=date,
+                client_ip=client_ip,
+                ua_browser=request.user_agent.browser if request else "",
+                ua_string=request.user_agent.string if request else "")
```
fredreichbier: I think defining a ``create_tag_dict`` function is a good idea!
But I was thinking: The Email token and the user notification handler are able to send HTML mails, and here we fill the HTML template with user-controlled data, e.g. ``ua_string``. So we should probably escape all HTML tags in the user-controlled dictionary values, e.g. ``ua_string``, ``ua_browser``, ``url``, ``user``, ``surname`` ... 
What do you think?

cornelinux: The surname is not user controlled. It is admin controlled. 
I think the username "user" is also not user controlled.
I think it is only the componentes of ``request``, right?

Do we have a method for this?

fredreichbier: We can use [flask.escape](http://flask.pocoo.org/docs/1.0/api/#flask.escape) or [cgi.escape](https://docs.python.org/2/library/cgi.html#cgi.escape), if we want to avoid the flask import.

Yes, we should definitely escape the components of ``request``.

But can we be sure that the surname, user, ... are always trusted, or may there be cases in which users can edit their own names? I was thinking if we should just escape *all* values, just to be sure:
```python
escaped_tags = {}
for key, value in tags:
    escaped_tags[key] = cgi.escape(value) if value is not None else value
```
What do you think?

cornelinux: Hm I do not like being address in the email as "Cornelius K\xf6lbel".

cornelinux:
@fredreichbier when merging, can you please squash?

PR #1706, opened 2019-06-24
Enable python 3.5

plettich:


privacyidea/api/lib/postpolicy.py
```diff
@@ -170,14 +170,14 @@ def after_request(response):
     else:
         response_object = response
     try:
-        content = json.loads(response_object.data)
```
cornelinux: In other versions than 2.5 the json object already exists in the ``response_object``.
What would be the speed difference, when we do an additional jsonify?
Could it make sense to only do a ``loads`` if the json object does not exist, yet?

cornelinux:
@plettich When should this draft be considered a full pull request to be merged?
If you want this to be merged, could you please add an issue, add this to milestone 3.1 and reference the issue in this PR. Thanks a lot!

plettich:
Actually i wanted You to have a look first You would consider this approach. If we want to go forward with the Python 3.5 support, i'll add an issue.

cornelinux:
using the ``.json`` attribute looks good to me.
So also this PR looks good to me. I would consider merging it! :-)

plettich:
> using the `.json` attribute looks good to me.

but be aware, the `.json` attribute is read-only. To modify the response we still need to do
```python
response.data = json.dumps(data)
```

cornelinux:
Hm, do you have any possible code location in mind?
It looks like we usually modify the ``content`` and pass it back to ``response.data`` - not to ``response.json``.

plettich:
> Hm, do you have any possible code location in mind?
> It looks like we usually modify the `content` and pass it back to `response.data` - not to `response.json`.

The current way is:
```python
content = json.loads(response.data.decode('utf8'))
...
response.data = json.dumps(content)
```
We could keep this functionality or use:
```python
content = response.json
...
response.data = json.dumps(content)
```

privacyidea/api/validate.py
```diff
@@ -198,10 +198,8 @@ def offlinerefill():
                     otps = MachineApplication.get_refill(tokenobj, password, mdef.get("options"))
                     refilltoken = MachineApplication.generate_new_refilltoken(tokenobj)
                     response = send_result(True)
-                    content = json.loads(response.data)
-                    content["auth_items"] = {"offline": [{"refilltoken": refilltoken,
-                                                          "response": otps}]}
-                    response.data = json.dumps(content)
+                    response.json["auth_items"] = {"offline": [{"refilltoken": refilltoken,
```
cornelinux: You said [here](https://github.com/privacyidea/privacyidea/pull/1706#issuecomment-507190292) the json attribute would be read-only  - so this should not work, righ!?

And I now realize, why we are using ``.data`` and not ``.json`` (Or am I getting somthing rong here?)

privacyidea/lib/eventhandler/base.py
```diff
@@ -299,7 +299,7 @@ def _get_tokenowner(request):
     @staticmethod
     def _get_response_content(response):
         if response:
-            content = json.loads(response.data)
+            content = response.get_json(force=True)
```
cornelinux: Why don't we do ``.json`` here?

tests/test_api_lib_policy.py
```diff
@@ -1569,7 +1569,7 @@ def test_01_check_tokentype(self):
         # The token type SPASS is not allowed on this client, so an exception
         #  is raised.
         r = check_tokentype(req, resp)
-        jresult = json.loads(r.data)
+        jresult = r.json
```
cornelinux: I am starting to doubt, if we can use ``.json``.
Since we might modify ``.data`` in decorators and like this.
Or is ``.json`` simply a reprensation of  ``.data``, even if it was modifies?

plettich: `.json` is a (cached) JSON representation of `.data`, implemented in [`werkzeug`](https://github.com/pallets/werkzeug/blob/a41a3e06ec1403f464d23431495ba59b41484e58/src/werkzeug/wrappers/json.py#L119)

privacyidea/api/validate.py
plettich: Yes, You are right, missed this (and the tests apparently as well)

cornelinux: Strange enough, lately I had the impression it *would* be possible. Please check this with @fredreichbier

plettich:
Since there were several changes, i created a new PR: (#1762 ) and close this one.
I fixed the mentioned issues as well.
Some information for the `json` attribute in [`Response`](https://github.com/pallets/flask/blob/b7f6fae9b34341b9be7742b86f6caffe07fc6f25/src/flask/wrappers.py#L103):
The `json` attribute is actually a read-only property, so assigning a new `dict` to that property fails. But editing the property (adding/removing an entry) did work. Unfortunately this is a volatile change which survived during a test request context but not in a full wsgi call.
So when changing the `json`-data in the Response, we need to dump it and use `set_data()` to correctly update the Response object.

PR #1708, opened 2019-06-24
Fix empty audit entry in case of error

### LINKED ISSUE #1707: Fix empty line in audit log
If an error occurs during a request, both the error handler and the `after_request` handler call [`finalize_log()`](https://github.com/privacyidea/privacyidea/blob/9704cbc1c9df14d97d01a3fd902cdecb87408523/privacyidea/lib/auditmodules/sqlaudit.py#L217) which results in an empty line in the audit log.

plettich:
In case an error occurred during handling of a request, the error handler
would finalize the audit log entry and clean the audit data.
But after the error handler, the general `after_request` handler was called
as well, which called `finalize_log()` again (with an empty audit_data)
resulting in an empty line in the audit log.

Also remove the `after_request` handler in `validate.py` and `ttype.py` and
use the handler in `before_after.py` since they all do the same.

Fixes #1707

### DIFF
--- a/privacyidea/api/before_after.py
+++ b/privacyidea/api/before_after.py
@@ -38,9 +38,11 @@ from flask import current_app
 from privacyidea.lib.policy import PolicyClass
 from privacyidea.lib.event import EventConfiguration
 from privacyidea.lib.lifecycle import call_finalizers
-from privacyidea.api.auth import (user_required, admin_required)
+from privacyidea.api.auth import (user_required, admin_required, jwtauth)
 from privacyidea.lib.config import get_from_config, SYSCONF, update_config_object
 from privacyidea.lib.token import get_token_type
+from privacyidea.api.ttype import ttype_blueprint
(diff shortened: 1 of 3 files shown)


privacyidea/api/before_after.py
```diff
@@ -226,7 +233,7 @@ def after_request(response):
     """
     # In certain error cases the before_request was not handled
     # completely so that we do not have an audit_object
-    if "audit_object" in g:
+    if "audit_object" in g and bool(g.audit_object.audit_data):
```
fredreichbier: We could shorten this to ``if "audit_object" in g and g.audit_object.audit_data:`` as the ``bool()`` isn't necessary here. What do you think?

plettich: Done in 541cfc7c

cornelinux: lgtm. Thanks!

PR #1712, opened 2019-06-24
Add distinction between list_policies/get_policies

fredreichbier:
The former should be used to retrieve a list of policy definitions, e.g. to display them in the WebUI. The latter should be used to get a list of policies that should take effect when handling the current request.

We already (kind of) have this distinction with the parameter ``all_times``: If it is set to ``True``, we ignore the ``time`` of policies. This PR just makes this distinction more explicit by introducing a parameter ``match_context`` and a new method ``list_policies`` as a shorthand.

This will be useful when implementing policy matching based on arbitrary user info attributes (#1436).

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -465,9 +465,7 @@ def enroll_pin(request=None, action=None):
                                           client=g.client_ip,
                                           active=True,
                                           audit_data=g.audit_object.audit_data)
-    action_at_all = policy_object.get_policies(scope=scope,
-                                               active=True,
-                                               all_times=True)
+    action_at_all = policy_object.list_policies(scope=scope, active=True)
 
     if action_at_all and not pin_pols:
         # Not allowed to set a PIN during enrollment!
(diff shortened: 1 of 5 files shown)


cornelinux:
If we would design this one method the correct way I think there is no need to have two methods, which would only lead to confusio in my opinion.

I think there is no difference, if we want to get policies to be displayed in the web ui or to get policies, that would acutally currently match!

Get all policies for the web ui:

    get_policies()

THen we could filter:

    get_policies(scope="admin")

And when fetching policies to evaluate the current request:

    get_policies(scope=..., action=..., client_ip=...., active=True, time=datetime.now())

I would rather recommend to slightly redesign the logic of the parameters, so that **one** method can be used always.
The ``any_time`` parameter was a quick, simple and poor decision.

But I think a sensible parameter redesign helps us to end up with **one** method.

https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/policy.py#L465

Why not e.g. rework the policies, so that:

~~~~python
def get_policies(self, .... time=None, ...) 
~~~~
could give us all policies independt of the time - just like all filtereing works.
``time=True`` could give us the policies for the current time.

I am not sure when to use ``list_policies`` and when to use ``get_policies``.
When I list the realms in the UI, this calls get_realms. This is the normal naming convention. 

Still I am not happing with having two methods, but naming them like ``get_policies`` (which would be more consistent with all other get_ methods) and maybe ``get_matching_policies`` would sound more intuitive to me.

fredreichbier:
Thanks for the thoughts! I'm closing this for now, until we've figured out if we really need two methods for fetching policies.

PR #1714, opened 2019-06-26
Add realmadmin, only tokens in certain realms visible

### LINKED ISSUE #1713: restrict the admins tokenview

Administrators should be able to manage tokens of different realms.
Admin A should be able to manage tokens in realm 1 and
Admin B should be able to mnaage tokens in realm 2.
Admin A should not see tokens in realm 2 and vice versa.

We should add a new policy action into the scope *admin*.
You *could* however deduce from the existing policies if an admin should be allowed to view which tokens and which tokens not.
But reality might be a bit more difficult.

E.g. An admin might very well be allowed to enroll a token in realm 1, but he might not be allowed to list the tokens in realm 1.
Therefore I would like to implement a new policy-action **tokenlist** in lib/policy.py

~~~~python
class ACTION(object):
    __doc__ = """This is the list of usual actions."""
    TOKENLIST = "tokenlist"



   
ACTION.USERLIST: {'type': 'bool',
                              'desc': _(
                                  'Admin is allowed to view tokens''),
                              'mainmenu': [MAIN_MENU.TOKENS],
                              'group': GROUP.TOKEN},
~~~~

In the api/token.py we need a decorator, that reads the allowed realms and sets the parameter ``filterRealms``.
These get then processed in the _create_token_query in lib/token.py using the SQLAlchemy ``in_`` statement.

None, today.

cornelinux:
The admin who has a "tokenlist" policy can only
see tokens in certain realms.

Closes #1713

### DIFF
--- a/doc/policies/admin.rst
+++ b/doc/policies/admin.rst
@@ -44,6 +44,20 @@ to enable tokens in the user-realm *sales*.
 The following actions are available in the scope
 *admin*:
 
+tokenlist
+~~~~~~~~~
+
+type: bool
+
+This allows the administrator to list existing tokens in the specified user realm.
+Note, that the resolver in this policy is ignored.
+
+If the policy with the action ``tokenlist`` is not bound to any user realm, this acts
+as a wild card and the admin is allowed to list all tokens.
+
+If the action ``tokenlist`` is not active, but admin policies exist, then the admin
+is not allowed to list any tokens.
+
 init
 ~~~~
 
--- a/doc/policies/admin.rst
+++ b/doc/policies/admin.rst
@@ -58,6 +58,12 @@ as a wild card and the admin is allowed to list all tokens.
 If the action ``tokenlist`` is not active, but admin policies exist, then the admin
 is not allowed to list any tokens.
 
+.. note:: As with all boolean policies, multiple *tokenlist* policies add up to
+   create the resulting rights of the administrator.
+   So if there are multiple matching policies for different realms,
+   the admin will have list rights on all mentioned realms
+   independent on the priority of the policies.
+
 init
 ~~~~
 
--- a/migrations/versions/b9131d0686eb_.py
+++ b/migrations/versions/b9131d0686eb_.py
@@ -1,0 +1,85 @@
+"""Add a generic admin policy for tokenlist
+
+Revision ID: b9131d0686eb
+Revises: 849170064430
+Create Date: 2019-06-28 07:37:59.224088
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = 'b9131d0686eb'
+down_revision = '849170064430'
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy import orm
+from sqlalchemy.schema import Sequence
+from privacyidea.lib.policy import SCOPE, ACTION
+
+Base = declarative_base()
+
+# dashes are not allowed, when creating policies in the WebUI
+# or via the library. So we are sure, that in normal operation
+# this policy can never be created.
+POLICYNAME = u"pi-update-policy-b9131d0686eb"
+
+
+class Policy(Base):
+    """
+    The policy table contains policy definitions which control
+    the behaviour during
+     * enrollment
+     * authentication
+     * authorization
+     * administration
+     * user actions
+    """
+    __tablename__ = "policy"
+    __table_args__ = {'mysql_row_format': 'DYNAMIC'}
+    id = sa.Column(sa.Integer, Sequence("policy_seq"), primary_key=True)
+    active = sa.Column(sa.Boolean, default=True)
+    check_all_resolvers = sa.Column(sa.Boolean, default=False)
+    name = sa.Column(sa.Unicode(64), unique=True, nullable=False)
+    scope = sa.Column(sa.Unicode(32), nullable=False)
+    action = sa.Column(sa.Unicode(2000), default=u"")
+    realm = sa.Column(sa.Unicode(256), default=u"")
+    adminrealm = sa.Column(sa.Unicode(256), default=u"")
+    resolver = sa.Column(sa.Unicode(256), default=u"")
+    user = sa.Column(sa.Unicode(256), default=u"")
+    client = sa.Column(sa.Unicode(256), default=u"")
+    time = sa.Column(sa.Unicode(64), default=u"")
+    condition = sa.Column(sa.Integer, default=0, nullable=False)
+    # If there are multiple matching policies, choose the one
+    # with the lowest priority number. We choose 1 to be the default priotity.
+    priority = sa.Column(sa.Integer, default=1, nullable=False)
+
+
+def upgrade():
+    """
+    During upgrade we check, if admin policies exist.
+    If so, we add a generic policy for all admins, that allows "tokenlist".
+    This mimicks the pervious behaviour.
+    :return:
+    """
+    bind = op.get_bind()
+    session = orm.Session(bind=bind)
+    if session.query(Policy).filter(Policy.scope == u"{0!s}".format(SCOPE.ADMIN),
+                                    Policy.active.is_(True)).all():
+        # add policy
+        tokenlist_pol = Policy(name=POLICYNAME, scope=u"{0!s}".format(SCOPE.ADMIN),
+                               action=u"{0!s}".format(ACTION.TOKENLIST))
+        session.add(tokenlist_pol)
+        print("Added '{0!s}' action for admin policies.".format(ACTION.TOKENLIST))
+    else:
+        print("No admin policy active. No need to create '{0!s}' action.".format(ACTION.TOKENLIST))
+
+    session.commit()
+
+
+def downgrade():
+    # Delete the policy, if it still exists
+    bind = op.get_bind()
+    session = orm.Session(bind=bind)
+    session.query(Policy).filter(Policy.name == u"{0!s}".format(POLICYNAME)).delete()
+    session.commit()
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -994,10 +994,10 @@ def check_anonymous_user(request=None, action=None):
 def check_admin_tokenlist(request=None, action=None):
     """
     Depending on the policy scope=admin, action=tokenlist, the
-    filterRealms parameter is set to define, the token of which
+    allowed_realms parameter is set to define, the token of which
     realms and administrator is allowed to see.
 
-    Sets the filterRealm
+    Sets the allowed_realms
     None: means the admin has no restrictions
     []: the admin can not see any realms
     ["realm1", "realm2"...]: the admin can see these realms
@@ -1005,7 +1005,7 @@ def check_admin_tokenlist(request=None, action=None):
     :param request:
     :return:
     """
-    filterRealm = None
+    allowed_realms = None
     wildcard = False
     role = g.logged_in_user.get("role")
     if role == ROLE.USER:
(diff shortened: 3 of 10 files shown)


cornelinux:
This pull request basically works!

I think there still might be some things to consider - either in this PR or in another one:

1. are there quicker ways to implement this? the create_sql_query could be a bit slow. I have not tested this!
2. We still need to figure out the migration path from installations where admins in their policies do not have the "tokenlist" action set.
Either we need to
  * a) modify existing policies
  * b) document this well, so that users need ot manually change their policies
  * c) add a config setting, that the tokenlist will only be evaluated, if the config setting is present.

I think I would prefer a) or b)

doc/policies/admin.rst
```diff
@@ -44,6 +44,20 @@ to enable tokens in the user-realm *sales*.
...
+
+type: bool
+
+This allows the administrator to list existing tokens in the specified user realm.
+Note, that the resolver in this policy is ignored.
+
+If the policy with the action ``tokenlist`` is not bound to any user realm, this acts
+as a wild card and the admin is allowed to list all tokens.
+
+If the action ``tokenlist`` is not active, but admin policies exist, then the admin
```
fredreichbier: Should we add a sentence what happens if multiple policies with a ``tokenlist`` action are defined? I think in this case, priorities are ignored and the admin is allowed to list tokens in all mentioned realms.

privacyidea/api/lib/prepolicy.py
```diff
@@ -991,6 +991,58 @@ def check_anonymous_user(request=None, action=None):
...
+            if not pol.get("realm"):
+                # if there is no realm set in a tokenlist policy, then this is a wildcard!
+                wildcard = True
+            else:
+                filterRealm.extend(pol.get("realm"))
+
+        if wildcard:
+            filterRealm = None
+
+    request.filterRealm = filterRealm
```
fredreichbier: I'm not very happy about the name ``filterRealm`` :-) because it uses camel-case and isn't very descriptive. What do you think about ``allowed_realms`` or ``restrict_realms``?

privacyidea/api/token.py
```diff
@@ -292,7 +293,7 @@ def init():
 
 @token_blueprint.route('/challenges/', methods=['GET'])
 @token_blueprint.route('/challenges/<serial>', methods=['GET'])
-@prepolicy(check_base_action, request, action=ACTION.GETCHALLENGES)
+@prepolicy(check_base_action, request, action=ACTION.TOKENLIST)
```
fredreichbier: Hm, is this intentional?

privacyidea/api/lib/prepolicy.py
```diff
@@ -991,6 +991,58 @@ def check_anonymous_user(request=None, action=None):
...
+    []: the admin can not see any realms
+    ["realm1", "realm2"...]: the admin can see these realms
+
+    :param request:
+    :return:
+    """
+    filterRealm = None
+    wildcard = False
+    role = g.logged_in_user.get("role")
+    if role == ROLE.USER:
```
fredreichbier: We can keep it like that, I just had an idea: but what do you think about restructuring the whole function like this
```python
if g.logged_in_user.get("role") != ROLE.USER:
    # everything else
return True
```
? I think this makes it clearer that it only takes effects for admins.

privacyidea/api/token.py
```diff
@@ -379,13 +381,13 @@ def list_api():
...
 
-    # filterRealm determines, which realms the admin would be allowed to see
-    filterRealm = ["*"]
-    # TODO: Userfields
+    if hasattr(request, "filterRealm"):
+        # filterRealm determines, which realms the admin would be allowed to see
+        filterRealm = request.filterRealm
+    else:
+        # In certain cases like for users, we do not have filterRealms  
+        filterRealm = None
```
fredreichbier: If you prefer, we could use ``filterRealm = getattr(request, "filterRealm", None)``. But we can also keep it like it above.

privacyidea/lib/token.py
```diff
@@ -140,7 +140,7 @@ def create_tokenclass_object(db_token):
 def _create_token_query(tokentype=None, realm=None, assigned=None, user=None,
                         serial_exact=None, serial_wildcard=None, active=None, resolver=None,
                         rollout_state=None, description=None, revoked=None,
-                        locked=None, userid=None, tokeninfo=None, maxfail=None):
+                        locked=None, userid=None, tokeninfo=None, maxfail=None, filterRealm=None):
```
fredreichbier: Also here, I would prefer to rename the parameter to something more descriptive.

privacyidea/lib/token.py
```diff
@@ -413,7 +418,7 @@ def get_tokens(tokentype=None, realm=None, assigned=None, user=None,
 def get_tokens_paginate(tokentype=None, realm=None, assigned=None, user=None,
                 serial=None, active=None, resolver=None, rollout_state=None,
                 sortby=Token.serial, sortdir="asc", psize=15,
-                page=1, description=None, userid=None):
+                page=1, description=None, userid=None, filterRealm=None):
     """
     This function is used to retrieve a token list, that can be displayed in
     the Web UI. It supports pagination.
```
fredreichbier: Hm, from a library design level, I'm not *too* happy about ``get_tokens_paginate`` having parameters ``user`` which holds a realm, ``realm`` *and* ``filterRealm`` :-) 
But I'm not sure if we can make this nicer easily, so maybe we just have to document this properly.

doc/policies/admin.rst
cornelinux: I think this is as with all simple bool policies. Yes, I can add something.

privacyidea/api/lib/prepolicy.py
cornelinux: Yes, will do.

privacyidea/api/token.py
cornelinux: Argh! Very much not!
Will do!

privacyidea/api/lib/prepolicy.py
cornelinux: Will check and decide.

privacyidea/api/token.py
cornelinux: Sounds good. Usually I do not program things, I do not understand! ;-)

privacyidea/lib/token.py
cornelinux: yes, yes, we will get rid of this "filterRealm". Everywhere!

privacyidea/lib/token.py
cornelinux: The problem is that the function calls would get far more complex!

The thing is that "realm" and "allowed_realms" do not simply sum up. If "realm" is not contained in "allowed_realms", than such token will not be returned!

If you want to simplify the library call, the code, that is calling the library will get much more complicated and very prone to errors!

fredreichbier: Yes. I mean, we could adjust ``realm`` such that it accepts a list of realms (and returns tokens which are in one of these realms), but this would complicate the calling locations.
So yeah, let's add a ``allowed_realms`` parameter and document it nicely.

privacyidea/lib/token.py
cornelinux: done in 80747ff6

privacyidea/api/token.py
cornelinux: done in 80747ff6

privacyidea/api/lib/prepolicy.py
cornelinux: done in 80747ff6

privacyidea/api/token.py
cornelinux: done in 80747ff6

privacyidea/api/lib/prepolicy.py
cornelinux: done in 80747ff6

doc/policies/admin.rst
cornelinux: done in 80747ff6

migrations/versions/b9131d0686eb_.py
```diff
@@ -0,0 +1,85 @@
...
+from sqlalchemy import orm
+from sqlalchemy.schema import Sequence
+from privacyidea.lib.policy import SCOPE, ACTION
+
+Base = declarative_base()
+
+# dashes are not allowed, when creating policies in the WebUI
+# or via the library. So we are sure, that in normal operation
+# this policy can never be created.
+POLICYNAME = u"pi-update-policy-b9131d0686eb"
```
fredreichbier: This idea to avoid name clashes is clever :) But I think we should avoid using dashes, because it contradicts our assumption that policy names never contain dashes, and this may cause trouble in the future.
As a current side effect, we cannot edit this policy via the web UI.
I think something like "policy_migration_b9131d0686eb" should be OK too.

privacyidea/api/lib/prepolicy.py
```diff
@@ -991,6 +991,58 @@ def check_anonymous_user(request=None, action=None):
     return True
 
 
+def check_admin_tokenlist(request=None, action=None):
+    """
+    Depending on the policy scope=admin, action=tokenlist, the
+    allowed_realms parameter is set to define, the token of which
```
fredreichbier: Hm, I think since we're now calling the parameter ``pi_allowed_realms``, we should also use this name in the docstring. Could you adjust the docstring accordingly?

privacyidea/lib/token.py
fredreichbier: The parameter looks good to me now! Could you add it to the docstring? :)

migrations/versions/b9131d0686eb_.py
cornelinux: I did not plan on editing the policy.
But I would think of deleting the policy.
This is a nice side effect, that the policy can only be deleted. Do you not like this?

privacyidea/lib/token.py
cornelinux: Will do

privacyidea/api/lib/prepolicy.py
cornelinux: will do

migrations/versions/b9131d0686eb_.py
fredreichbier: I guess the side effect is OK, but I fear that there may be some point in the future at which we do something relying on the assumption that all policy names only contain letters, digits, underscores, ... and then run into trouble because there are some installations with a policy not matching this rule. :-/

privacyidea/lib/token.py
cornelinux: done in 09065737

migrations/versions/b9131d0686eb_.py
cornelinux: ...and are you sure, that no customer until now has created a policy named "policy_migration_b9131d0686eb" because this would realy be problematic since we overwrite an existing policy.

fredreichbier: Not 100% of course :) But maybe we should only create the migration policy in case a policy with that name doesn't exist yet anyway?

cornelinux: If you do not create the policy, then migration will fail, since the policy-logic is missing.
With the current setup we will only create the policy, if the migration has not yet occurred, i.e. the data base schema is on an older version.

fredreichbier: Hm well, I think *if* someone has configured a policy named ``policy_migration_b9131d0686eb`` and we do not create our policy during migration, the migration itself will work, but the admins would have a restricted token list view afterwards?
I think if this unlikely case happens, it would be OK to require manually adding another migration policy.

cornelinux: Should we just go with the way to allow "-" in the policy name as discussed in person?

fredreichbier: As it turns out, dashes in policy names are not really a problem for privacyIDEA. We are not allowed to modify them, but we can enable/disable/delete them.

As it's nice to be able to create policies in migration scripts that don't clash with user-defined policies, let's keep the dashes in the policy name.

Do we have a good place to document this behavior? e.g. with a sentence
"Dashes are normally not allowed in policy names. However, privacyIDEA might create policies whose names contain dashes during upgrades".

cornelinux: I think we should document this at the update section and at the policies.

cornelinux: Look, it is bbc24af5

fredreichbier: Thanks!

PR #1718, opened 2019-06-28
Allow users to change the description

### LINKED ISSUE #1717: Allow users to change token description

Users have several tokens and need to distinguish the tokens by changing the description.

An additional user policy should allow users to change the description of their own tokens.

Description is currently set via an endpoint, which is "too mighty", ``/token/set`` and can also be used to set a lot of other token options. THis is why users were not allowed in the past to do so.

So we should add a new endpoint like ``/token/description`` to set the description of the token.

cornelinux:
User can now use the "setdescription" policy
to set the description of their tokens

Closes #1717

### DIFF
--- a/privacyidea/api/token.py
+++ b/privacyidea/api/token.py
@@ -673,6 +673,31 @@ def setpin_api(serial=None):
     return send_result(res)
 
 
+@token_blueprint.route('/description', methods=['POST'])
+@token_blueprint.route('/description/<serial>', methods=['POST'])
+@prepolicy(check_base_action, request, action=ACTION.SETDESCRIPTION)
+@event("token_set", request, g)
+@log_with(log)
+def set_description_api(serial=None):
+    """
+    This endpoint can be used by the user or by the admin to set
+    the description of a token.
+
+    :jsonparam basestring description: The description for the token
+    :param serial:
+    :return:
+    """
+    user = request.User
+    if not serial:
+        serial = getParam(request.all_data, "serial", required)
+    g.audit_object.log({"serial": serial})
+    description = getParam(request.all_data, "description", optional=required)
+    g.audit_object.add_to_log({'action_detail': u"description={0!r}".format(description)})
+    res = set_description(serial, description, user=user)
+    g.audit_object.log({"success": True})
+    return send_result(res)
+
+
 @token_blueprint.route('/set', methods=['POST'])
 @token_blueprint.route('/set/<serial>', methods=['POST'])
 @prepolicy(check_base_action, request, action=ACTION.SET)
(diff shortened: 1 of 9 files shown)


privacyidea/api/token.py
```diff
@@ -673,6 +673,31 @@ def setpin_api(serial=None):
     return send_result(res)
 
 
+@token_blueprint.route('/description', methods=['POST'])
+@token_blueprint.route('/description/<serial>', methods=['POST'])
+@prepolicy(check_base_action, request, action=ACTION.SETDESCRIPTION)
+@event("token_set", request, g)
+@log_with(log)
+def set_description_api(serial=None):
```
cornelinux: Till now there is only one policy ``ACTION.SET`` that allows an admin to change all settings of a token.
We do not want to get there into more details. Users till now were not able to change anything.
THis is why we add a new endpoint to change the description.

privacyidea/static/components/directives/controllers/directives.js
```diff
@@ -30,6 +30,7 @@ myApp.directive('tokenDataEdit', function(AuthFactory, instanceUrl) {
             setRight: '@',
             inputPattern: '@',
             inputType: '@',
+            editableAsUser: '@',
```
cornelinux: This passes the parameter to the Angular directive

privacyidea/static/components/directives/views/directive.tokendata.html
```diff
@@ -10,24 +10,27 @@
...
-            translate>Cancel
-    </button>
-    <button ng-hide="editFlag || (setRight == 'false' && hideButtons == 'true')"
-            class="btn btn-primary"
-            ng-click="editFlag = true;"
-            ng-disabled="setRight == 'false'"
-            translate>Edit
-    </button>
+<td>
+    <div ng-hide="loggedInUser.role == 'user' && editableAsUser != 'true' ">
```
cornelinux: To make the logic work more reliable I split this in two div's.
If a directive is ediable for a user, we will *not* hide the edit button!

privacyidea/static/components/token/controllers/tokenDetailController.js
```diff
@@ -138,6 +138,9 @@ myApp.controller("tokenDetailController", function ($scope,
     $scope.set = function (key, value) {
         TokenFactory.set($scope.tokenSerial, key, value, $scope.get);
     };
+    $scope.setdescription = function (description) {
```
cornelinux: This is the scope function, that calls the tokenFactory.

privacyidea/static/components/token/views/token.details.html
```diff
@@ -142,6 +143,18 @@ <h3 name="Token Details"><translate>Token details for {{ tokenSerial }}
             callback-cancel="get()"
             callback="set(key, value)"
             logged-in-user="loggedInUser"></tr>
+        <tr ng-show="loggedInUser.role==='user'"
```
cornelinux: We dublicate the row for the description, one time for the admin adnd the other time for the user.
This is simpler, since there are different conditions when the row and the edit button is visible and which function gets called.

cornelinux:
@plettich I also added some comments in the PR to explain the changes.

PR #1726, opened 2019-07-03
Add a script that exports OTP counters from privacyIDEA

### LINKED ISSUE #1728: Add export-counter script
We add an existing script, that can export the otp counter into a CSV with (serial, counter).

cornelinux:
Just for sake of completeness I add this script
that helped in different migration scenarios with LinOTP.
It exports only the counter values from all tokens
in the database. (relevant for HOTP and TOTP)

Closes #1728


cornelinux:
added issue

PR #1740, opened 2019-07-09
Fix TiQR token

### LINKED ISSUE #1739: TiQR token fails if user has multiple tokens
If a user has multiple challenge response token (e.g. TiQR and PUSH) he gets a QR-Code to scan. After entering the PIN in the TiQR app, a message like "Challenge not valid" appears.
If the user does not have other tokens, everything works as expected. The user will be successfully authenticated.

Mipronimo:
closes #1739


fredreichbier:
@Mipronimo Apparently I cannot request a review from you because you're the original author :) (it makes sense, I guess) I just restructured your fix a bit and added a test. Could you have a look at the new logic and check if it still fixes the problem?

Mipronimo:
lgtm thanks

PR #1741, opened 2019-07-09
Ideas for improved issue templates

fredreichbier:
Some ideas for shorter/more concise issue templates. Happy about any comments!


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,67 +1,55 @@
...
+### Configuration
 
-## more details: 
+**privacyIDEA Version**:
 
-* OS:
+**Installation method**: (from source, github, virtualenv, ubuntu packages...)
 
-* Webserver:
+**Operating System**:
```
Mipronimo: We should add
`**Python version**:`

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,67 +1,55 @@
...
-  (from source, github, virtualenv, ubuntu packages...)
+What did acutally happen?
 
+### Configuration
 
-## more details: 
+**privacyIDEA Version**:
 
-* OS:
+**Installation method**: (from source, github, virtualenv, ubuntu packages...)
```
plettich: pip?
what is the difference between source and github?

.github/ISSUE_TEMPLATE/bug_report.md
fredreichbier: thanks, added in the next revision

.github/ISSUE_TEMPLATE/bug_report.md
fredreichbier: good point, I reformulated the points a bit

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -35,21 +35,18 @@ What do you think, how the system should have worked?
...
+* **Python version**:
+* **Operating system**:
+* **Webserver**:
+* **Token database**: (MySQL, PostgreSQL, ...)
 
 ### Log file
 
-**Turn on debug level and take a look at the privacyidea.log!**
-**If appropriate, attach the log file.**
+**Set PI_LOGLEVEL = logging.DEBUG in pi.cfg and take a look at the privacyidea.log!**
```
plettich: Can we use `Code`-format in these templates? Might make the text more readable.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,67 +1,52 @@
 ---
-name: Bug report
-about: Create a report to help us improve
```
cornelinux: I would like to provide a wording that tells all contributors that their filing a bug actually is a contribution which comes with a certain responsibility and not simply a complaint that something works wrong in their opinion.
Filing a bug is not fingerpointing but helping.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,67 +1,52 @@
...
   (https://netknights.it/en/leistungen/one-time-services/)
 - or the privacyIDEA Enterprise Edition 
   (https://netknights.it/en/leistungen/service-level-agreements/)
-
-If you still think, you should open an issue, please do so:
 -->
 
-# What did you try to do?
-
-Please provide your top level intent!
```
cornelinux: When filing bug reports or support requests people often forget to actually tell what their intention is. They might say "i did this and that", but this often did not help to understand their situation. They try to be precise but often forget things.
I found it very helpful to ask them, what they actually want to achieve! This is why there is the question for the "top level intent": What was your intention?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,67 +1,52 @@
...
-Please provide steps to reproduce!
-
-# What outcome did you expect?
+### Top-level intent
```
cornelinux: Hm, here it is - the top level intent. I think I got lost in the github diff-view.

fredreichbier: Yup, it apparently just slipped away a bit :)

.github/ISSUE_TEMPLATE/bug_report.md
fredreichbier: Hm, I'm not sure if this would help, because users only see the text box without any formatting anyway (plus the preview, of course). Or maybe I am misinterpreting your comment :)

.github/ISSUE_TEMPLATE/bug_report.md
fredreichbier: Good point, I reformulated the "about" section in 4ed8d970

.github/ISSUE_TEMPLATE/are-you-lost-.md
```diff
@@ -1,7 +1,6 @@
 ---
-name: Are you lost?
-about: If you did not find a bug and do not want to suggest a new feature, you probably
-  have some questions.
+name: Ask a quetion
+about: ... if you need some explanations TODO
```
cornelinux: I would not propose "explanations". This is not support forum. We should be aware to split this. So I would prefer a text like:

"If you have an idea or an issue, that does not fit in the other mentioned....".

.github/ISSUE_TEMPLATE/are-you-lost-.md
```diff
@@ -1,7 +1,6 @@
 ---
-name: Are you lost?
-about: If you did not find a bug and do not want to suggest a new feature, you probably
-  have some questions.
+name: Ask a quetion
```
cornelinux: For asking questions we should point the users to https://community.privacyidea.org.

PR #1742, opened 2019-07-11
Add WebUI for policy conditions

fredreichbier:
This PR makes it possible to edit policy conditions in the WebUI.

Working on #1436

### DIFF
--- a/privacyidea/static/components/directives/views/directive.policyconditions.html
+++ b/privacyidea/static/components/directives/views/directive.policyconditions.html
@@ -1,6 +1,15 @@
 <div class="table-responsive">
     <table class="table">
-       <tr ng-repeat="condition in policyConditions">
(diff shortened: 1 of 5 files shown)


privacyidea/static/components/directives/views/directive.policyconditions.html
```diff
@@ -0,0 +1,62 @@
+<div class="table-responsive">
+    <table class="table">
+       <tr ng-repeat="condition in policyConditions">
```
plettich: Maybe we should add a table header to describe the entries in the columns in more detail, so that a user knows, what to put where.

fredreichbier: Good idea! I added table headings in 65ee04ee

PR #1753, opened 2019-07-22
Finetuning policy matching against arbitrary userinfo attributes

fredreichbier:
This PR implements some of the open questions mentioned in #1436:
* We rename ``==`` to ``equals``: I think that's more consistent because the other comparators (see below) are English words too
* We add comparators ``matches`` (for regexes) and ``in``
* We add negated comparators, which are named like ordinary comparators prefixed with ``!``, e.g. ``!contains``

### DIFF
--- a/privacyidea/lib/utils/compare.py
+++ b/privacyidea/lib/utils/compare.py
@@ -63,8 +63,6 @@ def parse_comma_separated_string(input_string):
         rows = list(reader)
     except Exception as exx:
         raise CompareError(u"Malformed comma-separated value: {!r}".format(input_string, exx))
-    if len(rows) != 1:
-        raise CompareError(u"Newlines in comma-separated values are unsupported: {!r}".format(input_string))
     return rows[0]
 
 
--- a/privacyidea/lib/utils/compare.py
+++ b/privacyidea/lib/utils/compare.py
@@ -66,7 +66,7 @@ def _compare_contains(left, comparator, right):
 #: In order to add a comparator to this module, add a suitable member to COMPARATORS
 #: and suitable entries to COMPARATOR_FUNCTIONS and COMPARATOR_DESCRIPTIONS.
 class COMPARATORS(object):
-    EQUALS = "=="
+    EQUALS = "equals"
     CONTAINS = "contains"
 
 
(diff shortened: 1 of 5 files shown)


privacyidea/lib/utils/compare.py
```diff
@@ -62,25 +87,92 @@ def _compare_contains(left, comparator, right):
...
+    """
+    Return True if ``left`` matches the regular expression given in ``right``.
+    Raise a CompareError if ``right`` is not a valid regular expression, or
+    if any other matching error occurs.
+    :param left: a string
+    :param right: a regular expression
+    :return: True or False
+    """
+    try:
+        return re.match("^" + right + "$", left) is not None
```
plettich: Maybe we should make clear that the regexp must describe the complete string, not just a part (since we add `^` and `$`).

privacyidea/lib/utils/compare.py
```diff
@@ -26,21 +26,46 @@
...
+    Whitespace immediately following a delimiter is skipped.
+    Raise a CompareError if the input is malformed.
+    :param input_string: an input string
+    :return: a list of strings
+    """
+    # We use Python's csv module because it supports quoted values
+    try:
+        reader = csv.reader([input_string], strict=True, skipinitialspace=True, doublequote=False, escapechar="\\")
+        rows = list(reader)
+    except Exception as exx:
```
plettich: can we have a more explicit Exception here? What does the `csv.reader` throw in case of an error?

privacyidea/lib/utils/compare.py
fredreichbier: Sounds good! I adjusted the docstring and description in c93c15aa1f1c2118cd5bfa60828364001537b7b

fredreichbier: That was a lie, it is actually 9c93c15aa1f1c2118cd5bfa60828364001537b7b

privacyidea/lib/utils/compare.py
fredreichbier: Fair point :-) Fixed this in 9c93c15aa1f1c2118cd5bfa60828364001537b7b

PR #1754, opened 2019-07-22
Remove whitespace from key during Yubikey token initialization

### LINKED ISSUE #1735: usability adding yubikey tokens: automatically remove spaces?
I'm adding a new yubikey token and use ykpersonalize and privacyidea's webui. In our documentation https://privacyidea.readthedocs.io/en/latest/configuration/tokens/yubikey.html we have:
(Remove possible white spaces!)
That's something a computer can do very easily - we should allow to use cut&paste without manual editing to avoid possible errors.
* privacyIDEA Version: 3.0.1

* Installation method: ubuntu packages

fredreichbier:
When enrolling Yubikey (AES mode) tokens, we now strip whitespaces from the OTP key.

I decided to implement the whitespace stripping on the server side because I couldn't think of a nice way to do it in the frontend :-) As it is implemented in the ``update`` method of the Yubikey token class, it doesn't interfere with other token classes.

Closes #1735

### DIFF
--- a/doc/configuration/tokens/yubikey.rst
+++ b/doc/configuration/tokens/yubikey.rst
@@ -49,8 +49,7 @@ initialized the yubikey with the external *ykpersonalize* tool.
 
 When using the yubikey personalization GUI you need to copy the value of
 "Secret Key (16 bytes Hex)". This is the secret OTP key, which you need to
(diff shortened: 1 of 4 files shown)


plettich:
In the long run we should move all API test-cases to the `test_api_...` tests and avoid cluttering the lib test-cases with API initialization stuff.

Issue #1631, opened 2019-05-15
check key length when creating tokens in the UI.


cornelinux:
check for the length of the otpkey when creating an HOTP or TOTP token.

Based on the hash algo we can check the otp length to be correct.
I think we should do this in javascript in the webui.

I am not sure if we should **avoid** creating a token with a wrong key length or if we only should **warn**.

plettich:
We should also check if the given/default hash-algorithm matches the key length or set the appropriate hash-algorithm based on the key length (like in [`importotp.py`](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/importotp.py))
This should be done by the back end, i think.

cornelinux:
Yes, we should do this in the API, not UI.

cornelinux:
I was trying to start to implement this.
But things can get arbitrary complicated.

We have "base32check", when the otpkey is passed base32 encoded.
We have a "shorter" client component of e.g. only 11 bytes, if we do a 2step enrollment. :-/

I wonder if we should add a classmethod, that checks for parameter aspects during enrollment.
I am currently not quite sure, how we should implement this.

For HOTP I added here https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/tokens/hotptoken.py#L337

~~~~python
if "otpkey" in upd_param and upd_param.get("otpkeyformat", "hex") == "hex":
            # If the user provides an otpkey, we check the length unless base32check
            check_hash_keylength(hashlibStr, len(otpKey)/2)
~~~~

But it breaks with 2step enrollment. We could also exclude ``is_true(params.get("2stepinit"))``, but a more generic approach would be great.

cornelinux:
I wonder, if we should put this in the UI.

plettich:
Maybe we can do the check before the key is written to the db. At this point, the key should be complete.
See [`tokenclass.py`](https://github.com/privacyidea/privacyidea/blob/059e6b3ef10518595cfb5dfbf3918cc205099d79/privacyidea/lib/tokenclass.py#L540)

cornelinux:
This is still a bit nasty, since

1. each token type can have different rules. But this is fine, this can be implemented in the tokenclasses.
2. the much bigger challenge is, that due to different rollout mechansims like two step etc. there are a lot of different rules for the key length. :-/

I think we should not check the keylength all the time, but maybe only if certain conditions are met like:

* ``genkey`` is not set
[...]

PR #1755, opened 2019-07-23
Add optional token-specific plausibility checks during enrollment

fredreichbier:
This PR implements a possible solution of #1631. See the issue for details.


fredreichbier:
Closing this, as we will do a more thorough rework of the token initialization logic for 3.2. See #1631.

Issue #1691, opened 2019-06-18
Check callers of match_policies and get_action_values


fredreichbier:
In #1672 (last comment), we clarified and documented the meaning of parameters for ``match_policies`` and ``get_action_values``. We should now ensure that all calling locations of these methods actually pass the parameters correctly (i.e. if they pass ``resolver=None``, they really don't care about the resolver).

cornelinux:
Hi @fredreichbier I guess you know what to do? ;-)

fredreichbier:
I think in the medium run, we should refactor our policy matching routines a bit. #1757 contains a first idea. However, as this is quite a big change, I would suggest we do this after 3.1?

fredreichbier:
Refactoring policy matching sounds nice, but we'll do it in 3.2 at the earliest.

cornelinux:
We should also consider the following in this issue:

In certain cases we only have a serial number in the request but we would also have to match for realm and resolver. Imagine an administrator who has a policy that allows the admin to unassign tokens in ``realmA`` in ``resolver2``. We can define such an admin policy.

But the endpoint ``/token/unassign?serial=<serial>`` is called from the webui.
The check_base_action decorator (https://github.com/privacyidea/privacyidea/blob/master/privacyidea/api/token.py#L438) thus only received the serial number, since the [request.User](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/api/token.py#L450) is None.

So we either need to add the tokenowner in such cases or we should add the ``request.User`` in general if an API endpoint only contains  the serial.

fredreichbier:
This is an interesting point! I think in case of unassigning a token, i.e. when the ``check_base_action`` function only receives a serial and no realm or resolver, the function actually fetches the first token realm and uses it for policy matching:

https://github.com/privacyidea/privacyidea/blob/f1543ced95df9dd7214dbd2499e3f57bd3990f98/privacyidea/api/lib/prepolicy.py#L1092

... but ``resolver`` is still None.

I'm currently not sure how this should behave. Should we set ``resolver`` to the tokenowner's resolver? Or should we also account for the future in which one token might have multiple tokenowners?

cornelinux:
Also take into consideration the changes in PR #1870

Issue #1425, opened 2019-02-06
Honour HTTP_AGENT in policies


cornelinux:
It could make sense to also make policies dependent on the HTTP-AGENT. This of course is often difficult, since the client can change its HTTP-AGENT, so this could only be possible if the client acutally is a trusted system.

But this way, we would be independent of the client IP.

@fredreichbier @plettich what do you think?

plettich:
How do we trust a client system?

cornelinux:
If the client is a server, this could be trusted. This way on one machine there could be several different applications, that query privacyIDEA with different HTTP_AGENTs.

At the moment there would be no (easy) way for us, to have different policies for different applications originating from the same IP.

fredreichbier:
Ah yeah, I guess matching the HTTP user agent makes sense in order to differentiate applications from the same IP. But this wouldn't work in all cases, would it? e.g. two ownCloud instances at the same IP would still send the same user agent and would be indistinguishable?

Instead of focusing on the user agent, we could also implement matching arbitrary HTTP request headers?

cornelinux:
arbitrary HTTP headers sound nice!

cornelinux:
see #1260

cornelinux:
* [ ] We need to pass the REQUEST information to the policy handling in [get_policies](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/policy.py#L464) and [get_action_values](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/policy.py#L638)

* [ ] We need to add a database field or whatever for the HTTP_Agent. Maybe we should already add a table for conditions?

cornelinux:
I am moving this to a later milestone. In 3.1 we should concentrate on the user attributes!

fredreichbier:
If #1757 gets into 3.2, this could be implemented quite easily for 3.2. Otherwise, we might move this to 3.3.

cornelinux:
Somewhere around the client IP we could add the ``HTTP header``.

![image](https://user-images.githubusercontent.com/1908620/66775629-c6557200-eec4-11e9-9dea-ce06b514f123.png)

We probably want to do some regular expression matching, so that we would match different headers with e.g. different version numbers. However, if we want to allow a **list** of different headers, this may get difficult. One header is simple:

    User-Agent=/Mozilla.*/

matching all mozillas for this policy.

If we want a logical **and**, that this policy does only match for

[...]

PR #1757, opened 2019-07-24
Refactor policy matching, introduce high-level matching API

fredreichbier:
This implements a high-level policy matching API on top of ``match_policies``/``get_action_values`` that aims to simplify the most common policy matching operations.

An example:
```python
login_mode_dict = Match.simple(g, scope=SCOPE.WEBUI, action=ACTION.LOGINMODE,
                               realm=None, user=user_object).action_values(unique=True)
```
The line reads as follows:
* Retrieve all active policies with scope ``WEBUI`` and action ``loginmode`` that match the user object ``user_object`` (i.e. its username, realm, resolver, userinfo) and the current ``g.client_ip``.
* For these policies, extract the action values of ``loginmode``, check for them for uniqueness, write matching policies to the audit log and return the action values dict.

Why do this? :-) Some reasons:
* When using ``match_policies``/``get_action_values`` directly, we almost always pass the same set of arguments: e.g. ``client_ip=g.client_ip``, ``audit_data=g.audit_object.audit_data``, or ``active=True`` (for ``match_policies``). With the new API, these arguments are passed implicitly. If we do not want the matching policies to be written to the audit log (which doesn't happen often), we explicitly pass ``write_to_audit_log=False``.
* ``match_policies`` and ``get_action_values`` take very similar arguments. The only difference is that ``get_action_values`` performs some postprocessing. With the new API, we first call ``Match.func(g, ...)`` to specify our parameters for matching, and *then* call ``.policies()`` or ``.action_values()``, depending on our needs.
* There are more complicated cases of policy matching, e.g. if we want to match ``scope=ADMIN`` or ``scope=USER`` policies depending on the current logged-in user's role. Up until now, we had to manually put together the respective values of ``scope``, ``user``, ``realm`` and ``adminrealm``. The new API provides a method for that:
  ```python
    audit_age = Match.admin_or_user(g, action=ACTION.AUDIT_AGE, realm=user_object.realm).action_values(unique=True)
  ```
* The new API makes it easier to extend policy matching in the future: e.g. if we want to allow matching of arbitrary request parameters (#1425), we only need to modify the ``Match`` class.
* For cases of policy matching that do not fit in the provided methods (there are a few), ``match_policies``/``get_action_values`` still works :)

Do you think this is a reasonable refactoring? What do you think of the API?

I realize this is quite a big change, so I would be ok with postponing this until after privacyIDEA 3.1.
 
ref #1691


fredreichbier:
@plettich @cornelinux  As this is quite a large change, I would be interested in your opinions about the API! :)

privacyidea/lib/policymatch.py
```diff
@@ -0,0 +1,175 @@
...
+    @classmethod
+    def admin(cls, g, action, realm):
+        """
+        Matching admin policies with an action and, optionally, a realm.
+        Assumes that the currently logged-in user is an admin, and throws an error otherwise.
+        Policies will be matched against the admin's username and adminrealm,
+        and optionally also the provided realm.
+        :param g: ``flask.g`` object
+        :param action: the policy action
+        :param realm: the realm against which policies should be matched. Can be None.
```
cornelinux: Please add in docs, that this is a userrealm

privacyidea/lib/policymatch.py
```diff
@@ -0,0 +1,175 @@
...
+                                                                    allow_white_space_in_action=
+                                                                    allow_white_space_in_action)
+        if write_to_audit_log:
+            for action_value, policy_names in action_values.items():
+                for p_name in policy_names:
+                    self._g.audit_object.audit_data.setdefault("policies", []).append(p_name)
+        return action_values
+
+    @classmethod
+    def simple(cls, g, scope, action, realm, user):
```
cornelinux: as realm and user are exclusive pleases split this in two methods *user* and *realm*.

privacyidea/lib/policymatch.py
```diff
@@ -0,0 +1,175 @@
...
+#
+# You should have received a copy of the GNU Affero General Public
+# License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+from privacyidea.lib.user import User
+from privacyidea.lib.error import ServerError
+from privacyidea.lib.policy import SCOPE
+
+
+class MatchingError(ServerError):
```
cornelinux: Please add this to ``lib/policy.py``.

privacyidea/lib/policymatch.py
```diff
@@ -0,0 +1,175 @@
...
+            username = g.logged_in_user["username"]
+            adminrealm = None
+            # TODO: resolver?
+        else:
+            # TODO: API key?
+            raise MatchingError("Unknown role")
+        return cls(g, name=None, scope=scope, realm=realm, active=True,
+                   resolver=None, user=username, user_object=None,
+                   client=g.client_ip, action=action, adminrealm=adminrealm, time=None,
+                   sort_by_priority=True)
```
cornelinux: Please add a further constructor, that reflects the interface of ``match_policies`` so that it can also be used e.g. in ``check_base_action`` called "generic".

privacyidea/lib/policymatch.py
```diff
@@ -0,0 +1,175 @@
...
+            for action_value, policy_names in action_values.items():
+                for p_name in policy_names:
+                    self._g.audit_object.audit_data.setdefault("policies", []).append(p_name)
+        return action_values
+
+    @classmethod
+    def simple(cls, g, scope, action, realm, user):
+        """
+        Simple policy matching with a scope, an action and optionally either a realm or a user.
+        :param g: ``flask.g`` object
```
cornelinux: We should not call this "flask.g", since we do not want to have flask dependencies in the lib level.
It is an object with "certain attributes" ;-)

privacyidea/lib/policy.py
```diff
@@ -837,6 +819,45 @@ def get_action_values(self, action, scope=SCOPE.AUTHZ, realm=None,
...
+        :param unique: if set, the function will only consider the policy with the
+            highest priority and check for policy conflicts.
+        :param allow_white_space_in_action: Some policies like emailtext
+            would allow entering text with whitespaces. These whitespaces
+            must not be used to separate action values!
+        :type allow_white_space_in_action: bool
+        :param audit_data: This is a dictionary, that can take audit_data in the g object.
+            If set, this dictionary will be filled with the list of triggered policynames in the
+            key "policies". This can be useful for policies like ACTION.OTPPIN - where it is clear, that the
+            found policy will be used. I could make less sense with an aktion like ACTION.LASTAUTH - where
```
plettich: ... It could...

privacyidea/lib/policymatch.py
fredreichbier: done in a11ccd0d1a2a1d94e5c2be6a02fae4454f636279

privacyidea/lib/policymatch.py
fredreichbier: done in a11ccd0d1a2a1d94e5c2be6a02fae4454f636279.
I also added a classmethod ``action_only``, which is also used in some places.

privacyidea/lib/policymatch.py
fredreichbier: done in a11ccd0d1a2a1d94e5c2be6a02fae4454f636279

privacyidea/lib/policymatch.py
fredreichbier: We can do that! But as this PR is already pretty big, I'd like to move this to a later PR which would then also remove the remaining usages of ``match_policies``/``get_action_values``. What do you think?

privacyidea/lib/policymatch.py
fredreichbier: Sounds good! However, I'd like to do that in a later PR in order to not make this PR even bigger :)

privacyidea/lib/policy.py
fredreichbier: fixed in 29916930

fredreichbier:
Fair point! I'd like to address this after having migrated the ``check_base_action`` policy matching to the new ``Match.generic`` interface -- what do you think?

fredreichbier:
OK, I think I finally got rid of policymatch.py. :)
So this PR should be good now. Not sure if it makes more sense to merge or squash though.

cornelinux:
You get the credits of 21 commits! That's ok! :-)

PR #1762, opened 2019-07-30
Enable Python 3.5

### LINKED ISSUE #1616: Python 3.5 does not work with privacyIDEA
Currently privacyIDEA does not work with Python 3.5, which is the default python 3 version in Ubuntu 16.04.
The `json.loads()` function expects a unicode string in python 3.5. In later versions, unicode and byte strings are supported.
The Flask-Response `data`-attribute, which we use to feed `json.loads()`, is unfortunately a byte string.
We could work around this problem by
1. converting all data passed to `json.loads()` to unicode first (which is done in most of the tests) or
2. use the Flask-Response `get_json()` function.

plettich:
privacyIDEA did not work correctly with python 3.5 due to some json
encoding issues.
Since we are using Flask 1.0, the Response class provides a `json` property
which we can use to read the corresponding data.

Fixes #1616

### DIFF
--- a/privacyidea/api/lib/postpolicy.py
+++ b/privacyidea/api/lib/postpolicy.py
@@ -170,7 +170,7 @@ def sign_response(request, response):
     else:
         response_object = response
     if response_object.is_json:
-        content = response_object.json
+        content = response_object.get_json(cache=False)
         nonce = request.all_data.get("nonce")
         if nonce:
             content["nonce"] = nonce
@@ -201,7 +201,7 @@ def check_tokentype(request, response):
     :type response: Response object
     :return: A new (maybe modified) response
     """
-    tokentype = response.json.get("detail", {}).get("type")
+    tokentype = response.get_json(cache=False).get("detail", {}).get("type")
     policy_object = g.policy_object
     user_object = request.User
     allowed_tokentypes = policy_object.get_action_values(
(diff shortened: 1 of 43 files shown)


privacyidea/api/lib/postpolicy.py
```diff
@@ -339,7 +336,7 @@ def add_user_detail_to_response(request, response):
     :param response:
     :return:
     """
-    content = json.loads(response.data)
+    content = response.json
```
fredreichbier: I like the usage of ``.json`` in postpolicies and tests!

However, I think we have a caching problem: Flask 1.0.2 seems to cache the object at ``response.json``, and the object does not seem to be updated when calling ``response.set_data(...)``. In our case, this does not seem to cause problems because we always modify the object at ``response.json`` in-place and then also call ``response.set_data(json.dumps(content))`` -- so the cached object and ``response.data`` are always in sync.

But if we would decide to write a new JSON object to ``response.data`` at some point in the future, we would run into weird problems because the cached object at ``response.json`` is not invalidated.

In order to avoid these problems, we should probably use ``response.get_json(cache=False)`` instead of ``response.json``. But if we want to keep using the nice ``response.json`` attribute, we could also implement a custom ``Response`` class that disables caching.

plettich:
Created a new PR: #1783 due to the new Response class.

PR #1763, opened 2019-07-30
Do not accept a push_wait parameter for /auth

### LINKED ISSUE #1743: /auth endpoint allows specifying a push_wait parameter
This is a follow-up to #1583 which introduced the ``push_wait`` policy action. This action allows to trigger a challenge and wait for the confirmation by the user in a single request to ``/validate/check``. Internally, the policy decorator function ``pushtoken_wait`` sets a ``push_wait`` option, either to a number of seconds:
https://github.com/privacyidea/privacyidea/blob/a2174a0260a3ec4484caacfe0bc6a0f8a67a2937/privacyidea/api/lib/prepolicy.py#L1353
or False:
https://github.com/privacyidea/privacyidea/blob/a2174a0260a3ec4484caacfe0bc6a0f8a67a2937/privacyidea/api/lib/prepolicy.py#L1355
The push token later reads the ``push_wait`` option value to find out whether to wait for the user confirmation or not:
https://github.com/privacyidea/privacyidea/blob/a2174a0260a3ec4484caacfe0bc6a0f8a67a2937/privacyidea/lib/tokens/pushtoken.py#L481

As it turns out, the same code location is hit by an ``/auth`` request. But as this endpoint is not decorated with ``pushtoken_wait``, we never explicitly set the ``push_wait`` option. As a result, the user can manually pass a ``push_wait`` option in the request body.

As the documentation explicitly states that ``push_wait`` only takes effect for ``/validate/check``, the ``push_wait`` option should be ignored in the ``/auth`` request.

fredreichbier:
Fixes #1743

### DIFF
--- a/tests/test_api_lib_policy.py
+++ b/tests/test_api_lib_policy.py
@@ -27,7 +27,7 @@ from privacyidea.api.lib.prepolicy import (check_token_upload,
                                            u2ftoken_verify_cert,
                                            tantoken_count, sms_identifiers,
                                            pushtoken_add_config, pushtoken_wait,
-                                           check_admin_tokenlist)
+                                           check_admin_tokenlist, pushtoken_disable_wait)
 from privacyidea.lib.realm import set_realm as create_realm
 from privacyidea.lib.realm import delete_realm
 from privacyidea.api.lib.postpolicy import (check_serial, check_tokentype,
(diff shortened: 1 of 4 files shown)


tests/test_api_lib_policy.py
```diff
@@ -1523,6 +1523,31 @@ def test_24_push_wait_policy(self):
 
         delete_policy("push1")
 
+    def test_24b_push_disable_wait_policy(self):
+        # We send a fake push_wait that is not in the policies
+        builder = EnvironBuilder(method='POST',
```
plettich: Do we need to build a complete environment? If we call the function directly, there won't be any request handling, correct?
So here we can just test if the function does the right thing (adjusting the `all_data` entry).

fredreichbier: That is true :) Fixed in 3769619.

PR #1765, opened 2019-08-01
Fix {caconnector,privacyideaserver,radiusserver}_read policies

### LINKED ISSUE #1749: Users cannot enroll RADIUS tokens in the WebUI if a policy with scope=user is defined
Steps to reproduce:
* Define a policy with ``scope=user`` and ``action=enrollRADIUS``
* Log in as a user, click "Enroll token", select RADIUS token
* We get an error "User actions are defined, but the action radiusserver_read is not allowed!"
* Even if RADIUS servers are defined, the dropdown list "RADIUS server configuration" presents no choices
* Hence, we cannot select a RADIUS server, and submitting the form says "ERR905: Missing parameter: 'radius.server'"

This is because the WebUI requests ``GET /radiusserver/`` when the user tries to enroll a RADIUS token. As a side effect of #1495, this endpoint is not allowed for users by default anymore: We would need a policy with scope user and action ``radiusserver_read`` to allow users to retrieve the list of RADIUS servers.

However, such a policy cannot be created via the WebUI.

I guess the right way to fix this would be to add a ``radiusserver_read`` action to the ``user`` scope. Maybe we should also automatically create it in the migration script, like we did for admin policies in #1721.

fredreichbier:
This PR fixes #1749:
* reintroduces enforcing the ``caconnector_read`` policy action for admins and users
* adds the possibility to define ``caconnector_read``, ``privacyideaserver_read`` and ``radiusserver_read`` policies for users
* fixes the WebUI, such that CA connectors are only retrieved when the user tries to enroll a certificate token.

Open question:

Should we add the respective user policy to the policy migration script here? https://github.com/privacyidea/privacyidea/blob/master/migrations/versions/3d7f8b29cbb1_.py

I think it's not as urgent as with the admin policies in this case, as admins can simply add the respective ``_read`` policies by hand.

### DIFF
--- a/privacyidea/api/caconnector.py
+++ b/privacyidea/api/caconnector.py
@@ -45,7 +45,7 @@ caconnector_blueprint = Blueprint('caconnector_blueprint', __name__)
 @caconnector_blueprint.route('/<name>', methods=['GET'])
 @caconnector_blueprint.route('/', methods=['GET'])
 @log_with(log)
-#@prepolicy(check_base_action, request, ACTION.CACONNECTORREAD)
+@prepolicy(check_base_action, request, ACTION.CACONNECTORREAD)
 def get_caconnector_api(name=None):
     """
     returns a json list of the available CA connectors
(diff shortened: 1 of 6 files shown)


cornelinux:
see https://github.com/privacyidea/privacyidea/issues/1749#issuecomment-520814887

PR #1767, opened 2019-08-02
Add a policy to enforce setting an app PIN

### LINKED ISSUE #1295: Add policy to set Smartphone App PIN
The privacyIDEA Authenticator supports forcing to set a PIN for the token.

We need to add a policy in the scope **enrollment**, to set a PIN for the token. The policies could be:

* hotp_force_pin (bool)
* totp_force_pin (bool)

The PIN is set in the QR code here:
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/apps.py#L164

It is set via ``pin=true``.

plettich:
The privacyIDEA Authenticator App has the ability to enforce setting a PIN
for a token. With this policy the appropriate parameter can be set during
rollout of a TOTP/HOTP Token

Fixes #1295

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -502,6 +502,9 @@ def init_tokenlabel(request=None, action=None):
     In addtion it adds the tokenissuer to the params like this:
     params : { "tokenissuer": "privacyIDEA instance" }
 
+    It also checks if the force_app_pin policy is set and adds the corresponding
+    value to params.
+
     It uses the policy SCOPE.ENROLL, ACTION.TOKENLABEL and ACTION.TOKENISSUER
     to set the tokenlabel and tokenissuer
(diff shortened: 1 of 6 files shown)


privacyidea/api/lib/prepolicy.py
```diff
@@ -541,7 +545,19 @@ def init_tokenlabel(request=None, action=None):
...
+    app_pin_pols = policy_object.match_policies(action='{0!s}_force_app_pin'.format(token_type),
+                                                scope=SCOPE.ENROLL,
+                                                user_object=user_object,
+                                                client=g.client_ip,
+                                                active=True,
+                                                audit_data=g.audit_object.audit_data)
+    if app_pin_pols:
+        request.all_data["force_app_pin"] = True
+    else:
+        request.all_data["force_app_pin"] = False
```
cornelinux: Is there a reason this is set to ``False``?  I think if it was not set, we would have the same outcome.

PR #1770, opened 2019-08-05
Check if a given token belongs to a given user

### LINKED ISSUE #1768: Enhance validate/check logic
Check that a token belongs to a user if `user` and `serial` were given.

plettich:
The validate/check endpoint could be used with a given user or a given
(token) serial. In case both were given, only the token was used for
authentication.

Fixes #1768

### DIFF
--- a/privacyidea/api/validate.py
+++ b/privacyidea/api/validate.py
@@ -325,8 +325,7 @@ def check():
     if serial:
         if user:
             # check if the given token belongs to the user
-            token_list = get_tokens(user=user, serial=serial)
-            if not token_list:
+            if not get_tokens(user=user, serial=serial, count=True):
                 raise ParameterError('Given serial does not belong to given user!')
         if not otp_only:
             result, details = check_serial_pass(serial, password, options=options)
--- a/privacyidea/api/validate.py
+++ b/privacyidea/api/validate.py
@@ -323,6 +323,11 @@ def check():
                         "realm": user.realm})
 
     if serial:
+        if user:
+            # check if the given token belongs to the user
+            token_list = get_tokens(user=user, serial=serial)
+            if not token_list:
+                raise ParameterError('Given serial does not belong to given user!')
         if not otp_only:
             result, details = check_serial_pass(serial, password, options=options)
         else:
--- a/tests/test_api_validate.py
+++ b/tests/test_api_validate.py
@@ -2555,7 +2555,7 @@ class ValidateAPITestCase(MyApiTestCase):
                              result)
 
         # try to authenticate with a token assigned to a different user
-        token.add_user(User("nönäscii", self.realm1))
+        token.add_user(User(u"nönäscii", self.realm2))
         token.set_pin("pin")
         self.assertEqual(token.token.owners.first().user_id, "1116")
         with self.app.test_request_context('/validate/check',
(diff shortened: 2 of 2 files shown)


plettich:
I am not sure about throwing the parameter error (which gets logged in the audit log) and the extra db request in the API function.

privacyidea/api/validate.py
```diff
@@ -323,6 +323,11 @@ def check():
                         "realm": user.realm})
 
     if serial:
+        if user:
+            # check if the given token belongs to the user
+            token_list = get_tokens(user=user, serial=serial)
```
fredreichbier: I think this should work! Maybe we could even pass ``count=True`` to only issue a ``SELECT COUNT(*)`` query, which could help performance . But I'm not so sure it would be noticeable :)

plettich: fixed in 23816b3a

plettich:
Actually other `ParameterError` are logged the same way (they all result in a `400 BAD REQUEST`) so i'll keep it that way.

tests/test_api_validate.py
```diff
@@ -2533,6 +2533,36 @@ def test_33_auth_cache(self):
...
+        self.assertEqual(token.token.serial, self.serials[1], token)
+        # try to authenticate a given user with a given unassigned token serial
+        with self.app.test_request_context('/validate/check',
+                                           method='POST',
+                                           data={"user": "cornelius",
+                                                 "realm": self.realm1,
+                                                 "serial": self.serials[1],
+                                                 "pass": OTPs[3]}):
+            res = self.app.full_dispatch_request()
+            self.assertEqual(res.status_code, 400, res)
```
cornelinux: Can you please also check for the error message "Given serial does not belong to given user!"

tests/test_api_validate.py
```diff
@@ -2533,6 +2533,36 @@ def test_33_auth_cache(self):
...
+        token.set_pin("pin")
+        self.assertEqual(token.token.owners.first().user_id, "1116")
+        with self.app.test_request_context('/validate/check',
+                                           method='POST',
+                                           data={"user": "cornelius",
+                                                 "realm": self.realm1,
+                                                 "serial": self.serials[1],
+                                                 "pass": OTPs[3]}):
+            res = self.app.full_dispatch_request()
+            self.assertEqual(res.status_code, 400, res)
```
cornelinux: Can you please also check for the error message "Given serial does not belong to given user!"

privacyidea/api/validate.py
cornelinux: I talke the liberty of resolving this conversation! Looks good.

tests/test_api_validate.py
plettich: fixed in ed931650

tests/test_api_validate.py
plettich: s.o.

cornelinux:
Oups, we have an error: "UserError: ERR904: The user can not be found in any resolver in this realm!" in the tests

plettich:
> Oups, we have an error: "UserError: ERR904: The user can not be found in any resolver in this realm!" in the tests

Good thing we have travis :-)
Fixed in f8473efa
Maybe we should squash all these little commits...

cornelinux:
@plettich  Will squash when the tests have passed. THanks!

PR #1771, opened 2019-08-05
TIQR token enhancements

### LINKED ISSUE #1777: The TiQR Token should return the remaining attempts before the token is blocked
The INVALID_RESPONSE in the TiQR token returns the number of remaining PIN attempts in ` ``INVALIDE_RESPONSE:<n>``. Currently the ``:<n>`` is not implemented.

basvandervlies:
When we unlock the TIQR token with the App and use the wrong PIN/biometric code. Respond
back how many retries we have left before the token is locked/blocked

the TIQR app can not handle challenges bigger or equal then  `2^31 = 2147483648. Check if the
value is bigger if yes adjust it.

Closes #1777

### DIFF
--- a/privacyidea/lib/tokens/tiqrtoken.py
+++ b/privacyidea/lib/tokens/tiqrtoken.py
@@ -331,7 +331,6 @@ class TiqrTokenClass(OcraTokenClass):
                     token = get_one_token(serial=challenge.serial)
                     if token.type.lower() == "tiqr":
                         # We found a TiQR token with a valid challenge with the given transaction ID
-                        res = "INVALID_RESPONSE"
                         r = token.verify_response(
                             challenge=challenge.challenge, passw=passw)
                         if r > 0:
@@ -340,6 +339,14 @@ class TiqrTokenClass(OcraTokenClass):
                             challenge.set_otp_status(True)
                             # We have found a valid TiQR token transaction, we break out of the loop
                             break
+                        else:
+                            # Send back how may retries there are left for the token is blocked
+                            token.inc_failcount()
+                            fail = token.get_failcount()
+                            maxfail = token.get_max_failcount()
+                            res = "INVALID_RESPONSE:{0!s}".format(maxfail - fail)
+                            break
+
             cleanup_challenges()
 
             return "plain", res
@@ -378,6 +385,8 @@ class TiqrTokenClass(OcraTokenClass):
 
         service_identifier = get_from_config("tiqr.serviceIdentifier") or \
                              "org.privacyidea"
+        service_displayname = get_from_config("tiqr.serviceDisplayname") or \
+                              "privacyIDEA"
 
         # Get the OCRASUITE from the token information
         ocrasuite = self.get_tokeninfo("ocrasuite") or OCRA_DEFAULT_SUITE
@@ -396,11 +405,13 @@ class TiqrTokenClass(OcraTokenClass):
 
         # Encode the user to UTF-8 and quote the result
         encoded_user_identifier = quote_plus(user_identifier.encode('utf-8'))
-        authurl = u"tiqrauth://{0!s}@{1!s}/{2!s}/{3!s}".format(
+        authurl = u"tiqrauth://{0!s}@{1!s}/{2!s}/{3!s}/{4!s}".format(
                                               encoded_user_identifier,
                                               service_identifier,
                                               db_challenge.transaction_id,
-                                              challenge)
+                                              challenge,
+                                              service_displayname
+                                              )
         attributes = {"img": create_img(authurl, width=250),
                       "value": authurl,
                       "poll": True,
--- a/tests/test_lib_tokens_tiqr.py
+++ b/tests/test_lib_tokens_tiqr.py
@@ -126,7 +126,7 @@ class OCRASuiteTestCase(MyTestCase):
         # test creation of hex challenge
         os = OCRASuite("OCRA-1:HOTP-SHA1-6:QH10-S128")
         c = os.create_challenge()
-        self.assertEqual(len(c), 20)
+        self.assertEqual(len(c), 10)
         self.assertTrue("G" not in c, c)
 
         # test creation of alphanum challenge
@@ -529,7 +529,7 @@ class TiQRTokenTestCase(MyApiTestCase):
                         "operation": "login"}
         r = TiqrTokenClass.api_endpoint(req, g)
         self.assertEqual(r[0], "plain")
-        self.assertEqual(r[1], "INVALID_RESPONSE")
+        self.assertRegexpMatches(r[1], r"INVALID_RESPONSE:[0-9]+")
 
         # Check that the OTP status is still incorrect
         r = token.check_challenge_response(options={"transaction_id":
@@ -684,7 +684,7 @@ class TiQRTokenTestCase(MyApiTestCase):
                         "operation": "login"}
         r = TiqrTokenClass.api_endpoint(req, g)
         self.assertEqual(r[0], "plain")
-        self.assertEqual(r[1], "INVALID_RESPONSE")
+        self.assertRegexpMatches(r[1], r"INVALID_RESPONSE:[0-9]+")
 
         # Check that the OTP status is still incorrect
         r = token.check_challenge_response(options={"transaction_id":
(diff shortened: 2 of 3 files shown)


privacyidea/lib/tokens/tiqrtoken.py
```diff
@@ -384,6 +391,10 @@ def create_challenge(self, transactionid=None, options=None):
         # Depending on the OCRA-SUITE we create the challenge
         os = OCRASuite(ocrasuite)
         challenge = os.create_challenge()
+        if int(challenge) >= 2147483648:
+            # TIQR app has problems with value greater the 2^31
+            challenge = '1' + challenge[1:]
```
cornelinux: Should we downshift all values >=2147483648? Then there are certain values with a higher probability (1147483648, 1147483649 have double the probability like 1147483647, 1147483646) or should we rather reshuffle:
~~~~python
challenge = 2147483648
while challenge >= 2147483648:
    challenge = os.create_challenge()
~~~~

privacyidea/lib/tokens/tiqrtoken.py
```diff
@@ -384,6 +391,10 @@ def create_challenge(self, transactionid=None, options=None):
...
+            log.info("Challenge value is too big changing it to: {0!s}".format(challenge))
```
cornelinux: Depending on the decision above I would downgrade the to ``log.debug``.

privacyidea/lib/tokens/tiqrtoken.py
basvandervlies: My concern was that creating a challenge is an "expensive" operation and opted for this approach. We can also use a flip coin approach.  Replace the first digit with a `0` or a `1`.   Then the probability will lower a bit.

@cornelinux you make the decision what is the right approach and I will implement it.

cornelinux: I looked into it again. The default OCRAsuite for the TiQR token  is currently:
"OCRA-1:HOTP-SHA1-6:QN10"

This is a ten digit number ranging from 0000000001 to 999999999.
So roughly 78% are bigger than 2147483648.
I am not quite sure of the limitations here. Is the the TiQR protocol in itself or the implementation of the App? Maybe it might make more sense to simply change the default_suite, which defines the challenge to OCRA-1:HOTP-SHA1-6:QN9 (only 9 digits)

cornelinux: @basvandervlies where do you get the 31bit limitation from?

cornelinux: Hm, maybe we should add TiQR to the privacyIDEA authenticator to get rid of this limitation.
I think there is no need to change this in the code. You could also specify another OCRASuite (like QN9) in the WebUI in Config->Token->TiQR

tests/test_lib_tokens_tiqr.py
```diff
@@ -684,7 +684,7 @@ def test_05_api_endpoint_with_multiple_tokens(self):
                         "operation": "login"}
         r = TiqrTokenClass.api_endpoint(req, g)
         self.assertEqual(r[0], "plain")
-        self.assertEqual(r[1], "INVALID_RESPONSE")
+        self.assertRegexpMatches(r[1], r"INVALID_RESPONSE:[0-9]+")
```
plettich: I think, we should also test, that the counter is decreased correctly. I can provide a test for that since this might be a bit tricky.

privacyidea/lib/tokens/tiqrtoken.py
basvandervlies: @cornelinux I just stumbled on this by Trial and Error because the default is `OCRA-1:HOTP-SHA1-6:QN10` for PrivacyIdea.  See also:
 * https://community.privacyidea.org/t/tiqr-response-problems/1100/8

I changed the `QN` part of the ocrasuite to `9`, `09`, `8` and `08` and they do not work at all. There is something wrong the official app.   I can only test the IOS app. I have updated the issue:
 * https://github.com/SURFnet/tiqr-client-ios/issues/39

The only code that works for me is `QN:10` with the check of value is not too big

basvandervlies: I am also in contact with the TIQR app developers and our sister organization SURFnet. They say the app is tested with suite: `OCRA-1:HOTP-SHA1-6:QH10-S512`

But the `S` is not implemented in privacyidea:
```
NotImplementedError: OCRA Session not implemented, yet.
```

Also tested other suites without the `S` option, but they also fail

martijnkruiten: We have now tested with `OCRA-1:HOTP-SHA1-6:QH10`, which seems to work. We had one enrolled token that failed 100% of the time, but after that we couldn't reproduce it with other tokens. A 10 digit hexadecimal number has a high probability to be larger than 31 bit, so the same limitation does not seem to apply.

tests/test_lib_tokens_tiqr.py
cornelinux: @plettich Would you enhance the test in a seperate pull request?

privacyidea/lib/tokens/tiqrtoken.py
cornelinux: @martijnkruiten, @basvandervlies What now?
I would rather not add this challenge-reduce-code into the upstream code, since it seems to be a special fix for the TiQR app. So I think there are two possibilities:

1. Identify the an OCRASuite that works for you and configure it that way, or
2. run the "challenge-reduce-code"  only if a special config value in pi.cfg is set like
    ~~~~python
    if int(challenge) >= 2147483648 and get_app_config_value("PI_REDUCE_TIQR_CHALLENGE") is True:
            # TIQR app has problems with value greater the 2^31
            challenge = '1' + challenge[1:]
    ~~~~
3. Have the TiQR developer fix the app

basvandervlies:
This check is wrong:
```
  def test_02_create_challenge(self):
        # test creation of hex challenge
        os = OCRASuite("OCRA-1:HOTP-SHA1-6:QH10-S128")
        c = os.create_challenge()
>       self.assertEqual(len(c), 20)
E       AssertionError: 10 != 20
```
We ask for length `10` and we get `20`. It is `QH` challenge. The conversion from byte to hex is wrong 1 byte is 2 nibbles. See https://tools.ietf.org/html/rfc6287
```
o  "OCRA-1:HOTP-SHA1-4:QH8-S512" means version 1 of OCRA with HMAC-
      SHA1 function, truncated to a 4-digit value, using a random
      hexadecimal challenge up to 8 nibbles and a session value of 512
      bytes
```

privacyidea/lib/crypto.py
```diff
@@ -469,6 +469,8 @@ def geturandom(length=20, hex=False):
         
     if hex:
         ret = to_unicode(binascii.hexlify(ret))
+        # 1 bytes is 2 nibbles
+        ret = ret[:length]
```
cornelinux: To my understanding this must be fixed in the TiQR and not here! Please remove this.
If you do a geturandom(length=2, Hex=True) the intended behaviour is to get two random bytes in hex format! like ``AB12``.

privacyidea/lib/tokens/tiqrtoken.py
basvandervlies: @cornelinux I have removed the fix for the this special case. Its indeed depended on the OCRAsuite used. It can be a bug in the app or in the server code with conversion.  The developpers of the APP will also take a look at it.

We have now switched to: `OCRA-1:HOTP-SHA1-6:QH10` and that works much better. But we encountered something else that the length of the challenge is incorrect. Is must be 10 instead of 20. That is the other improvement. But my colleague just informed my and we will put it in the TIQR token module

cornelinux: Yes, see my other comment. This must be fixed in lib/tokens/ocra.py.

privacyidea/lib/crypto.py
martijnkruiten: The length can also be an odd number, so we could pass length/2 rounded up and then truncate it to the correct length afterwards.

cornelinux: OK. fine! :-)

tests/test_lib_tokens_tiqr.py
plettich: will do.

Issue #1641, opened 2019-05-22
Some JS-functions in the UI appear as links but don't change the pointer symbol


plettich:
The About-link in the footer is basically just a JS-function to open a pop-up. When hovering, the cursor changes to `text` instead of `pointer`. `pointer` would better indicate a clickable link.
The same applies to the `Lock Screen` menu entry in the user menu.

NuvandaPV:
Fixed #1772

PR #1772, opened 2019-08-06
Change the cursor to `pointer` where needed.

NuvandaPV:
This commit will set the cursor style to `pointer` for any element that
has an Angular-callback attached to it.


NuvandaPV:
fixes #1641

Issue #1484, opened 2019-03-04
Checkbox layer cuts off values


plettich:
The Checkbox layer, which is displayed i.e. during creation of an event handler, is cut off by the bottom bar:
![pi_checkbox_hidden](https://user-images.githubusercontent.com/37443810/53746781-3a955500-3ea2-11e9-8a5d-eb347553ce52.png)
The value `validate_triggerchallenge` is not visible anymore.

Changing the CSS value `position` to `static` for the Checkbox layer might improve this issue:
![pi_checkbox_float](https://user-images.githubusercontent.com/37443810/53747020-a972ae00-3ea2-11e9-8f94-a0d14bf33335.png)

Issue #955, opened 2018-02-28
Overlay issue in privacyIDEA 2.21


ali-axiadids:
Platform: Amazon Linux
Database: PostgreSQL 9.5
privacyIDEA version: 2.21
Browsers: Chrome Version 64.0.3282.186 (Official Build) (64-bit) and FireFox 58.0.2 (64 bit)

Steps to reproduce:

1- Log in to privacyIDEA web portal
2- Enroll HOTP token (or TOTP)
3- Set validity in Extended attributes section
4- Additionally, re-adjusting the page window causes the header and footer to cover the body

Please see screenshot for additional information.

![errors](https://user-images.githubusercontent.com/35825556/36813146-7f6fc3dc-1c88-11e8-954d-84c6074b7d09.png)

cornelinux:
This occurs when the browser is set to a size, when the top most menu creates a line break but the responsive design does not yet hide the menu in the hamburger item.

NuvandaPV:
I think the nicest way to fix this, would be to hide the labels for the navbar at the top, if the device is too large for the mobile navigation, but too small to fit the full desktop navigation (i.e. less then 1024px) – something like this:

![localhost_(iPad) (2)](https://user-images.githubusercontent.com/13070524/64854550-2767fc80-d61e-11e9-92f9-eaa17f0a7cbd.png)

@cornelinux what do you think?

cornelinux:
cool!

NuvandaPV:
Fixed #1852

PR #1776, opened 2019-08-08
Baseline styles

NuvandaPV:
Add some protected space around floating input elements to keep them from clashing with the baseline.

### DIFF
--- a/privacyidea/static/css/baseline.css
+++ b/privacyidea/static/css/baseline.css
@@ -1,10 +1,21 @@
+body {
+    padding-bottom: 70px;
+}
+
+body > .container > .slide > .well {
+    margin-bottom: 0;
+}
+
 #baseline {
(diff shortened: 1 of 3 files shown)


NuvandaPV:
relates #955

NuvandaPV:
fixes #1484

plettich:
I thought about something like this: https://jsfiddle.net/kqhj2s3p/

NuvandaPV:
I don't really wanna go there, since making this work in practice will require `calc()` which will do some freaky stuff in IE.

NuvandaPV:
@plettich how is this? Works in all browsers I have tested and at least shouldn't make things any worse in browsers where it is not supported (IE doesn't allow `*::after { display: block; }`, but that just means the whole fix does nothing).

PR #1785, opened 2019-08-14
Add /system/names/{caconnector,radius} endpoints

### LINKED ISSUE #1749: Users cannot enroll RADIUS tokens in the WebUI if a policy with scope=user is defined

fredreichbier:
This PR adds endpoints ``/system/names/{caconnector,radius}`` which are accessible for users and admins with the respective ``enrollCERTIFICATE``/``enrollRADIUS`` rights. These are now used to retrieve the CA connectors and RADIUS servers during enrollment of CERTIFICATE and RADIUS tokens.

As a consequence, the ``/radius/`` and ``/caconnector/`` endpoints are not accessible for users anymore.

Likewise, the ``/privacyideaserver/`` endpoints are not accessible to users anymore, because REMOTE tokens cannot be enrolled by users anyway.

This PR also fixes the enrollment of CERTIFICATE tokens: Without this PR, users cannot choose a certificate template.

Closes #1749

### DIFF
--- a/privacyidea/static/components/token/controllers/tokenControllers.js
+++ b/privacyidea/static/components/token/controllers/tokenControllers.js
@@ -287,6 +287,9 @@ myApp.controller("tokenEnrollController", function ($scope, TokenFactory,
             // because the user might not be allowed to list RADIUS servers
             $scope.getRADIUSIdentifiers();
         }
+        if ($scope.form.type === "certificate") {
+            $scope.getCAConnectors();
+        }
         // preset twostep enrollment
         $scope.setTwostepEnrollmentDefault();
     };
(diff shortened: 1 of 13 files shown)


privacyidea/static/components/token/controllers/tokenControllers.js
```diff
@@ -287,6 +287,9 @@ myApp.controller("tokenEnrollController", function ($scope, TokenFactory,
             // because the user might not be allowed to list RADIUS servers
             $scope.getRADIUSIdentifiers();
         }
+        if ($scope.form.type === "certificate") {
+            $scope.getCAConnectors();
```
cornelinux: Is it intentionally that you want to reload the CA Connectors each time you change the tokentype?
Before the caconnector list was loaded only once.

fredreichbier: You're right, we now reload the CA connectors each time we change the tokentype to "Certificate". But before, we tried to load the CA connectors even if the user is not allowed to enroll certificate tokens, and this causes a red error popup :/

PR #1788, opened 2019-08-15
Implement Challenge/Response for RADIUS token

cornelinux:
Improving radiusmock to allow AccessAccept, AccessReject and
AccessChallenge even with additional attributes.

### DIFF
--- a/privacyidea/lib/tokens/radiustoken.py
+++ b/privacyidea/lib/tokens/radiustoken.py
@@ -59,6 +59,7 @@ from privacyidea.lib.decorators import check_token_locked
 from privacyidea.lib.radiusserver import get_radius
 from privacyidea.models import Challenge
 from privacyidea.lib.challenge import get_challenges
+from privacyidea.lib.policydecorators import challenge_response_allowed
 
 import pyrad.packet
 from pyrad.client import Client
@@ -149,10 +150,12 @@ class RadiusTokenClass(RemoteTokenClass):
         self.add_tokeninfo("tokenkind", TOKENKIND.VIRTUAL)
 
     @log_with(log)
+    @challenge_response_allowed
     def is_challenge_request(self, passw, user=None, options=None):
         """
         This method checks, if this is a request, that triggers a challenge.
-        It depends on the way, the pin is checked - either locally or remote
+        It depends on the way, the pin is checked - either locally or remotely.
+        In addition, the RADIUS token has to be configured to allow challenge response.
 
         communication with RADIUS server: yes
         modification of options: The communication with the RADIUS server can
--- a/privacyidea/lib/tokens/radiustoken.py
+++ b/privacyidea/lib/tokens/radiustoken.py
@@ -1,5 +1,9 @@
 # -*- coding: utf-8 -*-
 #
+#  2019-08-15 Cornelius Kölbel <cornelius.koelbel@netknights.it>
+#             Allow RADIUS challenge / response
+#             Credits to @droobah, who provided the first pull request
+#             https://github.com/privacyidea/privacyidea/pull/1389
 #  2018-01-21 Cornelius Kölbel <cornelius.koelbel@netknights.it>
 #             Add tokenkind
 #  2016-02-22 Cornelius Kölbel <cornelius@privacyidea.org>
@@ -46,19 +50,21 @@ import logging
 import traceback
 import binascii
 from privacyidea.lib.utils import is_true, to_bytes, hexlify_and_unicode, to_unicode
-from privacyidea.lib.tokenclass import TokenClass, TOKENKIND
 from privacyidea.lib.tokens.remotetoken import RemoteTokenClass
+from privacyidea.lib.tokenclass import TokenClass, TOKENKIND
 from privacyidea.api.lib.utils import getParam, ParameterError
 from privacyidea.lib.log import log_with
 from privacyidea.lib.config import get_from_config
 from privacyidea.lib.decorators import check_token_locked
 from privacyidea.lib.radiusserver import get_radius
+from privacyidea.models import Challenge
+from privacyidea.lib.challenge import get_challenges
 
 import pyrad.packet
 from pyrad.client import Client
 from pyrad.dictionary import Dictionary
+from pyrad.packet import AccessChallenge, AccessAccept, AccessReject
 from privacyidea.lib import _
-
 
 optional = True
 required = False
@@ -99,7 +105,7 @@ class RadiusTokenClass(RemoteTokenClass):
                'title': 'RADIUS Token',
                'description': _('RADIUS: Forward authentication request to a '
                                 'RADIUS server.'),
-               'user':  ['enroll'],
+               'user': ['enroll'],
                # This tokentype is enrollable in the UI for...
                'ui_enroll': ["admin", "user"],
                'policy': {},
@@ -113,6 +119,7 @@ class RadiusTokenClass(RemoteTokenClass):
 
         return ret
 
+    @log_with(log)
     def update(self, param):
         # New value
         radius_identifier = getParam(param, "radius.identifier")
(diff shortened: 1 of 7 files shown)


cornelinux:
In the first commit I simply replace the radiusmock so that it can also return an AccessChallenge with RADIUS attributes. I did not change any productive code, simply prepare the new mock for the new code.
@fredreichbier maybe you already want to take a look at it, so the PR will not be too big ;-)

privacyidea/lib/tokens/radiustoken.py
```diff
@@ -168,24 +338,91 @@ def split_pin_pass(self, passw, user=None, options=None):
...
+        state = options.get('radius_state')
+        result = options.get('radius_result')
+        if result is None:
+            radius_response = self._check_radius(otpval, options=options, radius_state=state)
+        else:
+            radius_response = result
+
+        if radius_response == AccessAccept:
+            res = True
+            reply = {'message': 'matching 1 tokens',
```
fredreichbier: I don't think we need to set the reply here because it is set in the ``check_token_list`` loop

privacyidea/lib/tokens/radiustoken.py
```diff
@@ -168,24 +338,91 @@ def split_pin_pass(self, passw, user=None, options=None):
...
+                     'serial': self.get_serial(),
+                     'type': self.get_tokentype()}
+            otp_counter = 1
+
+        return res, otp_counter, reply
+
     @log_with(log)
     @check_token_locked
     def check_otp(self, otpval, counter=None, window=None, options=None):
+        """
```
fredreichbier: Is this function used in our production code? If I see correctly, it is currently only used in our RADIUS token tests?

privacyidea/lib/tokens/radiustoken.py
```diff
@@ -168,24 +338,91 @@ def split_pin_pass(self, passw, user=None, options=None):
...
+        if res == AccessAccept:
+            return 1
+        else:
+            return -1
+
+    @log_with(log)
+    @check_token_locked
+    def _check_radius(self, otpval, options=None, radius_state=None):
         """
         run the RADIUS request against the RADIUS server
```
fredreichbier: Could you add some information to the various methods' docstrings how they use the ``options`` dictionary? e.g. they write something, they read something ...

privacyidea/lib/tokens/radiustoken.py
```diff
@@ -137,11 +144,174 @@ def update(self, param):
...
+                        otp_counter = 1
+                        break
+                    elif radius_response == AccessChallenge:
+                        # The response was valid but triggered a new challenge
+                        # Note: The second challenge currently does not work correctly
+                        # see https://github.com/privacyidea/privacyidea/issues/1792
+                        challengeobject.delete()
+                        _, _, transaction_id, _ = self.create_challenge(options=options)
+                        options["transaction_id"] = transaction_id
+                        otp_counter = 1
```
fredreichbier: We should return ``-1`` in this case, as the RADIUS server doesn't want to authenticate us yet

privacyidea/lib/tokens/radiustoken.py
```diff
@@ -137,11 +144,174 @@ def update(self, param):
...
+        if options is None:
+            options = {}
+        message = options.get('radius_message') or "Enter your RADIUS tokencode:"
+        state = options.get('radius_state')
+        attributes = {'state': transactionid}
+        validity = int(get_from_config('DefaultChallengeValidityTime', 120))
+
+        db_challenge = Challenge(self.token.serial,
+                                 transaction_id=transactionid,
+                                 data=state,
```
fredreichbier: As ``state`` is probably a bytestring, I think we should hexlify it before writing it to the database.

privacyidea/lib/tokens/radiustoken.py
```diff
@@ -137,11 +144,174 @@ def update(self, param):
...
+        transaction_id = options.get('transaction_id') or options.get('state')
+
+        # get the challenges for this transaction ID
+        if transaction_id is not None:
+            challengeobject_list = get_challenges(serial=self.token.serial,
+                                                  transaction_id=transaction_id)
+
+            for challengeobject in challengeobject_list:
+                if challengeobject.is_valid():
+                    state = challengeobject.data
```
fredreichbier: We should unhexlify here

privacyidea/lib/tokens/radiustoken.py
```diff
@@ -248,34 +485,40 @@ def check_otp(self, otpval, counter=None, window=None, options=None):
                                        NAS_Identifier=nas_identifier.encode('ascii'))
 
             req["User-Password"] = req.PwCrypt(otpval)
-            if "transactionid" in options:
-                req["State"] = str(options.get("transactionid"))
+
+            if radius_state:
+                req["State"] = str(radius_state)
+                log.info("Sending saved challenge to radius server: {0} ".format(radius_state))
```
fredreichbier: I think this should be ``{0!r}``

privacyidea/lib/tokens/radiustoken.py
```diff
@@ -248,34 +485,40 @@ def check_otp(self, otpval, counter=None, window=None, options=None):
                                        NAS_Identifier=nas_identifier.encode('ascii'))
 
             req["User-Password"] = req.PwCrypt(otpval)
-            if "transactionid" in options:
-                req["State"] = str(options.get("transactionid"))
+
+            if radius_state:
+                req["State"] = str(radius_state)
```
fredreichbier: I think ``radius_state`` should be a bytestring anyway, so we don't need ``str()`` here (and it would break under Python 3)

privacyidea/lib/tokens/radiustoken.py
```diff
@@ -137,11 +144,174 @@ def update(self, param):
...
+        else:
+            state = options.get('radius_state')
+            # The pin is checked remotely
+            res = options.get('radius_result')
+            if res is None:
+                res = self._check_radius(passw, options=options, radius_state=state)
+
+            return res == AccessChallenge
+
+        return res
```
fredreichbier: Could we move this to the ``if`` branch body?

privacyidea/lib/tokens/radiustoken.py
fredreichbier: Seems like we should treat the state as a bytestring: https://tools.ietf.org/html/rfc2865#section-5.24
So as the ``state`` field of the ``Challenge`` table is defined as ``db.Unicode``, it seems safer to ``hexlify_and_unicode`` the state before writing it to the database.

tests/test_api_validate.py
```diff
@@ -3239,6 +3241,79 @@ def test_10_unique_transaction_id(self):
...
+                                                 "pass": "correctPW",
+                                                 "transaction_id": transaction_id}):
+            res = self.app.full_dispatch_request()
+            self.assertTrue(res.status_code == 200, res)
+            data = json.loads(res.data)
+            self.assertTrue(data.get("result").get("status"))
+            self.assertTrue(data.get("result").get("value"))
+            t = data.get("detail").get("transaction_id")
+            # No transaction_id
+            self.assertIsNone(t)
```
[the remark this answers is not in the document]
cornelinux: done in 061ee4ad

privacyidea/lib/tokens/radiustoken.py
cornelinux: done in 13d94769

privacyidea/lib/tokens/radiustoken.py
cornelinux: done in 13d94769

privacyidea/lib/tokens/radiustoken.py
cornelinux: probably yes. But it looks fine, right?

privacyidea/lib/tokens/radiustoken.py
cornelinux: done. removed

privacyidea/lib/tokens/radiustoken.py
cornelinux: done.

privacyidea/lib/tokens/radiustoken.py
fredreichbier: OK, let's keep it

privacyidea/lib/tokens/radiustoken.py
cornelinux: done

privacyidea/lib/tokens/radiustoken.py
cornelinux: done

tests/test_lib_tokens_radius.py
```diff
@@ -187,5 +188,199 @@ def test_12_non_ascii(self):
...
+        r = add_radius(identifier="myserver", server="1.2.3.4",
+                       secret="testing123", dictionary=DICT_FILE)
+        self.assertTrue(r > 0)
+        token = init_token({"type": "radius",
+                            "radius.identifier": "myserver",
+                            "radius.local_checkpin": False,
+                            "radius.user": u"nönäscii"})
+
+        # Check if the remote PIN would create a RADIUS challenge
+        state1 = ["123456"]
```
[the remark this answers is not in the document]
cornelinux: done

privacyidea/lib/tokens/radiustoken.py
cornelinux: done

privacyidea/lib/tokens/radiustoken.py
```diff
@@ -168,24 +352,92 @@ def split_pin_pass(self, passw, user=None, options=None):
...
+
+        This is only called after it is verified, that the upper level is no challenge-request
+        or challenge-response
+
+        The "options" are read-only in this method. They are not modified here. authenticate
+        is the last method in the loop ``check_token_list``.
+
+        communication with RADIUS server: yes, if is no previous "radius_result"
+            If there is a "radius" result in the options, we do not query the radius server
+        modification of options: options are not modified
```
fredreichbier: I think they are modified if we query the radius server -- could you add that?

privacyidea/lib/tokens/radiustoken.py
fredreichbier: thanks! I only have one last remark regarding the docstrings, see above

cornelinux:
> I have some small remarks, but it looks good so far, and I've tested it with a `httpie -> privacyIDEA RADIUS token -> FreeRADIUS -> privacyIDEA email token` setup :)
> But I've noticed that we do change behavior if we have a RADIUS token with a locally checked PIN, because we now trigger a (privacyIDEA) challenge in this case, but don't do that on current master.

What do we do on current master? Do we fail?

fredreichbier:
Aha! It seems that on current master, we only issue a challenge if a policy allows the RADIUS token to be used as a challenge-response token. On this branch, we always issue a challenge, regardless of the policy.

We should probably change that? And use the ``challenge_response_allowed`` decorator like here: https://github.com/privacyidea/privacyidea/blob/f32626e0eb168004f039928163aaeb70bf007848/privacyidea/lib/tokens/remotetoken.py#L279

privacyidea/lib/tokens/radiustoken.py
cornelinux: yeah, I changes this before...

privacyidea/lib/tokens/radiustoken.py
cornelinux: done in 20504484

fredreichbier:
Yes -- I think this behavior is reasonable. Some documentation also sounds good :)

cornelinux:
@fredreichbier  Wait - at least due to configuration the RADIUS token always *has been* a challenge response token: https://github.com/privacyidea/privacyidea/blob/v3.0.2/privacyidea/lib/tokens/radiustoken.py#L75

fredreichbier:
I think the ``mode`` attribute only says that it *might* be used as a challenge-response token ... I'm not totally sure though :) The HOTP token class also has the mode ``["authenticate", "challenge"]``.

cornelinux:
...whoever knows this software!

cornelinux:
You are right. The "challenge" mode means it can be used as challenge response. BUt the decorator before ``is_challenge_request`` defines if it will work as challenge response out of the box (see sms and email) or if you need to set a policy to do so.
And to my understanding there was no need to set a policy for the RADIUS token.

cornelinux:
OK, RADIUS does not inherit the ``is_challenge_request`` from the remote token anymore...

fredreichbier:
According to [codecov](https://codecov.io/gh/privacyidea/privacyidea/compare/c4d352f6fcad1ea58b2e5e0525829316d38bcc20...498b75d5ab5dd7273f7c1d8be43ec8e83df43d46/diff#D2-325), our unit tests do not seem to cover the case that the RADIUS server rejects a challenge response. Could we add a test for that? I can do this too.

cornelinux:
@fredreichbier  This would be great! Thank you!

Issue #1789, opened 2019-08-15
User TIQR enroll page


basvandervlies:
**Is your feature request related to a problem? Please describe.**
Not to bother the user with this message:
```
For TiQR tokens to work properly you need to set the:
TIQR Token Config (URL)
```
The user can not click on URL link and it is confusing the user.

**Describe the solution you'd like**
Only the help message nothing about how to config TIQR

**Describe alternatives you've considered**
Tried to delete the lines, but then I can not enroll the ticket anymore

**Additional context**
See screenshot
![tiqr](https://user-images.githubusercontent.com/431367/63081474-fb385d80-bf43-11e9-989b-8d6bca584234.png)

fredreichbier:
I was wondering why the message is displayed even if the TiQR token is configured. The reason is that the ``GET /system/`` endpoint only returns the ``tiqr.regServer`` config option for admins, not for users. So if a user is logged in, the view does not see the ``tiqr.regServer`` option and assumes the TiQR config is incomplete.
We could either make the ``tiqr.regServer`` setting public (its value is contained in the QR code anyway), or modify the view such that it only displays the warning for admins.

basvandervlies:
@fredreichbier I also found it strange that the mesage was displayed for the User.   The solution to only enable it for admin is the best one.  I do not know if you want to publish your `tiqr.regServer` for all users or is this only accessible for authenticated users?

cornelinux:
I think these messages should not be displayed to the normal user in the first place!
A normal user could not configure the system anyways. So we should not bother them, that the admins have messed up! ;-)

So: Lets remove all those messages during token enrollment.

basvandervlies:
@cornelinux I totally agree ;-) 

I have removed the message but we need somehow execute the code to enroll the token:
 * `<div ng-hide="config.tokens({tokentype:'tiqr'})" class="ng-hide"> </div>` 

If I am correct, but I am not am expert in this area.

cornelinux:
> @cornelinux I totally agree ;-)
> 
> I have removed the message but we need somehow execute the code to enroll the token:
> 
> * `<div ng-hide="config.tokens({tokentype:'tiqr'})" class="ng-hide"> </div>`
> 
> If I am correct, but I am not am expert in this area.

This has to be user-role-dependent.

PR #1791, opened 2019-08-15
TIQR enrollment page for user

basvandervlies:
Do not display the `TIQR Config' url, see:
 * https://github.com/privacyidea/privacyidea/issues/1789

### DIFF
--- a/privacyidea/static/components/token/views/token.enroll.email.html
+++ b/privacyidea/static/components/token/views/token.enroll.email.html
@@ -3,7 +3,7 @@
 </p>
 
 <h4 translate>Token data</h4>
-<div ng-hide="form['email.mailserver']" translate class="bg-danger">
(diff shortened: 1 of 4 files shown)


cornelinux:
The whole info has to be hidden for the user like this:

~~~~html
<div ng-hide="form['tiqr.regServer'] || loggedInUser.role==='user'" translate class="bg-danger">
    For TiQR tokens to work properly you need to set the
    <a ui-sref="config.tokens({tokentype:'tiqr'})">TiQR Token Config</a>!
</div>
~~~~

I think this has to be done in:

* token.enroll.email.html:<div ng-hide="form['email.mailserver']" translate class="bg-danger">
* token.enroll.tiqr.html:<div ng-hide="form['tiqr.regServer']" translate class="bg-danger">
* token.enroll.yubico.html:<div ng-hide="form['yubico.id']" translate class="bg-danger">
* token.enroll.sms.html:<div translate class="bg-info">

Issue #1780, opened 2019-08-09
Add testing artefacts to gitignore


NuvandaPV:
Running nosetests(1) will generate a handful of files that are not yet caught by gitignore.

PR #1794, opened 2019-08-15
Update .gitignore

NuvandaPV:
Ignore a few files that will appear if the machine running the tests has a newer version of GPG.


NuvandaPV:
fixes #1780

PR #1796, opened 2019-08-16
Reusing an existing firebase access_token

### LINKED ISSUE #1795: Hanging at contacting firebase service
When testing the iOS Push token App the communication to the firebase service hangs some times. It is not clear why.

We could however improve the communication here:
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/smsprovider/FirebaseProvider.py#L76

And reuse an existing access_token as long as it has not expired.

cornelinux:
We are reusing an access token if it has not expired, yet.

Closes #1795

### DIFF
--- a/privacyidea/lib/smsprovider/FirebaseProvider.py
+++ b/privacyidea/lib/smsprovider/FirebaseProvider.py
@@ -78,7 +78,7 @@ def get_firebase_access_token(config_file_name):
         At = AccessToken(access_token_info.access_token, access_token_info.expires_in)
         app_store[fbt][config_file_name] = At
         readable_time = datetime.datetime.fromtimestamp(At.expires_at).isoformat()
-        log.debug(u"Setting the expiration for {!r} of the new access_token to {0!s}.".format(config_file_name, readable_time))
+        log.debug(u"Setting the expiration for {!r} of the new access_token to {!s}.".format(config_file_name, readable_time))
 
     return app_store[fbt][config_file_name].access_token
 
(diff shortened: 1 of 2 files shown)


privacyidea/lib/smsprovider/FirebaseProvider.py
```diff
@@ -73,12 +80,17 @@ def submit_message(self, firebase_token, data):
...
                                        SCOPES)
 
-        access_token_info = credentials.get_access_token()
-
-        # Should we do something with expires in?
-        # expires_in = access_token_info.expires_in
-
-        bearer_token = access_token_info.access_token
+        now = time.time()
+        if self.access_token_info is None or now > self.access_token_expires_at:
```
fredreichbier: I think this doesn't hurt, but I'm not sure if it will have an effect: Caching the access token would only speed up things if we call the ``submit_message`` method multiple times on the same ``FirebaseProvider`` object. As the ``FirebaseProvider`` objects do not persist between requests, this would only help if we call ``submit_message`` multiple times during the same request. I'm not sure if we do that?

cornelinux: To bad the object gets destroyed. No, to my knowledge we do not call several submit_messages in one requst.

So how should we persist the access_token. Does only look like, as if we could persist it in the database. But to me it feels a bit awkward writing it to the SQL database.

cornelinux: done, check it out :-)

fredreichbier:
I'll add a test and merge afterwards.

PR #1815, opened 2019-08-22
Allow "-" and " " in policy names

### LINKED ISSUE #1813: Allow "-" in policy names
It is totally annoying to me personally. ;-)
We should allow "-" in policy names.

cornelinux:
I personally find it easier to use these two characters in
policy names.

Closes #1813

### DIFF
--- a/privacyidea/api/lib/utils.py
+++ b/privacyidea/api/lib/utils.py
@@ -34,6 +34,7 @@ import json
 import jwt
 import threading
 import six
+import re
 from flask import (jsonify,
                    current_app)
 
@@ -267,3 +268,18 @@ def verify_auth_token(auth_token, required_role=None):
                         "this resource!").format(required_role),
                         id=ERROR.AUTHENTICATE_MISSING_RIGHT)
     return r
+
+
+def check_policy_name(name):
+    """
+    This function checks, if the given name is a valid policy name.
+
+    :param name: The name of the policy
+    :return: Raises a ParameterError in case of an invalid name
+    """
+    if not re.match('^[a-zA-Z0-9_.\- ]*$', name):
+        raise ParameterError(_("The name of the policy may only contain "
+                               "the characters a-zA-Z0-9_.- "))
+
+    if name.lower() == "check":
+        raise ParameterError(_("T'check' is an invalid policy name."))
--- a/privacyidea/api/lib/utils.py
+++ b/privacyidea/api/lib/utils.py
@@ -277,9 +277,14 @@ def check_policy_name(name):
     :param name: The name of the policy
     :return: Raises a ParameterError in case of an invalid name
     """
+    disallowed_patterns = [("^check$", re.IGNORECASE),
+                           ("^pi-update-policy-", re.IGNORECASE)]
+    for disallowed_pattern in disallowed_patterns:
+        if re.search(disallowed_pattern[0], name, flags=disallowed_pattern[1]):
+            raise ParameterError(_(u"'{0!s}' is an invalid policy name.").format(name))
+
     if not re.match('^[a-zA-Z0-9_.\- ]*$', name):
         raise ParameterError(_("The name of the policy may only contain "
                                "the characters a-zA-Z0-9_.- "))
 
-    if name.lower() == "check":
-        raise ParameterError(_("T'check' is an invalid policy name."))
+
--- a/privacyidea/api/policy.py
+++ b/privacyidea/api/policy.py
@@ -40,7 +40,8 @@ from .lib.utils import (getParam,
                         getLowerParams,
                         optional,
                         required,
-                        send_result)
+                        send_result,
+                        check_policy_name)
 from ..lib.log import log_with
 from ..lib.policy import (set_policy,
                           PolicyClass, ACTION,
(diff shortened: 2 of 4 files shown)


privacyidea/api/policy.py
```diff
@@ -175,12 +176,7 @@ def set_policy_api(name=None):
     """
     res = {}
     param = request.all_data
-    if not re.match('^[a-zA-Z0-9_.]*$', name):
-        raise ParameterError(_("The name of the policy may only contain "
-                               "the characters a-zA-Z0-9_."))
-
-    if name.lower() == "check":
-        raise ParameterError(_("T'check' is an invalid policy name."))
+    check_policy_name(name)
```
plettich: shouldn't we put this in the lib-layer?

privacyidea/api/lib/utils.py
```diff
@@ -267,3 +268,18 @@ def verify_auth_token(auth_token, required_role=None):
...
+
+    :param name: The name of the policy
+    :return: Raises a ParameterError in case of an invalid name
+    """
+    if not re.match('^[a-zA-Z0-9_.\- ]*$', name):
+        raise ParameterError(_("The name of the policy may only contain "
+                               "the characters a-zA-Z0-9_.- "))
+
+    if name.lower() == "check":
+        raise ParameterError(_("T'check' is an invalid policy name."))
```
plettich: maybe we can define a list of invalid names (there are a few IIRC).
And is this `T` in the error message necessary? It looks a bit like [Klingon](https://en.wikipedia.org/wiki/Klingon_language).

privacyidea/api/lib/utils.py
```diff
@@ -267,3 +268,18 @@ def verify_auth_token(auth_token, required_role=None):
...
+    This function checks, if the given name is a valid policy name.
+
+    :param name: The name of the policy
+    :return: Raises a ParameterError in case of an invalid name
+    """
+    if not re.match('^[a-zA-Z0-9_.\- ]*$', name):
+        raise ParameterError(_("The name of the policy may only contain "
+                               "the characters a-zA-Z0-9_.- "))
+
+    if name.lower() == "check":
```
fredreichbier: Until now, we used dashes in names of migration policies, because this way we were sure that they don't clash with user-defined policies. If we now allow dashes in policy names, we lose this loophole. An idea: Should we disallow to create policies whose names start with ``pi-update-policy-``? This way, we could still create migration policies without worrying about name clashes in the future.

privacyidea/api/policy.py
cornelinux: It has not been in the lib layer the whole time.
The documentation points out, that it is actually a cool thing, that you *can* create policies with other names in the lib layer.

We used the possibility to create other-named policies in the migration scripts:
We created policies with "not-allowed" characters in the migration script. This way we could be sure, that those policies do not exist.

privacyidea/api/lib/utils.py
cornelinux: We sill can use "~" or "*" or whatever in future migration policies.
Do you really think we need to disallow that? Would of course play into @plettich's hand/suggestion with the disallowed list ;-)

fredreichbier: Well, I think just disallowing ``pi-update-policy-`` prefixes can be easily implemented now, and it would save us from hassle in the future :) And then we have the advantage that all migration policies have the same prefix.

tests/test_api_lib_utils.py
```diff
@@ -25,3 +25,11 @@ def test_01_getParam(self):
 
         v = getParam({}, "sslverify", allowed_values=["0", "1"], default="1")
         self.assertEqual("1", v)
+
+    def test_02_check_policy_name(self):
+        check_policy_name("This is a new valid Name")
+        check_policy_name("THis-is-a-valid-Name")
+        # The name "check" is reserved
+        self.assertNotEqual(ParameterError, check_policy_name, "check")
```
fredreichbier: Oh, this should probably be ``assertRaises``. According to codecov, we actually don't hit the error case.

privacyidea/api/lib/utils.py
cornelinux: check this out. 
0ea6ecfd

privacyidea/api/lib/utils.py
cornelinux: check this out. 
0ea6ecfd

tests/test_api_lib_utils.py
```diff
@@ -25,3 +25,11 @@ def test_01_getParam(self):
...
+        # This is an invalid name
+        self.assertNotEqual(ParameterError, check_policy_name, "~invalid name")
```
fredreichbier: Same here

privacyidea/api/policy.py
cornelinux: I want to keep it at this place.

tests/test_api_lib_utils.py
cornelinux: Now we shoud, right?
check this out. 
0ea6ecfd

tests/test_api_lib_utils.py
cornelinux: ?

cornelinux: 8c1a3ea4

tests/test_api_lib_utils.py
cornelinux: 8c1a3ea4

privacyidea/api/lib/utils.py
fredreichbier: thanks!

PR #1820, opened 2019-08-22
Add the rollout_state condition to eventhandlers

### LINKED ISSUE #1801: Add rollout_state to event handlers

When enrolling a 2step HOTP or TOTP token, we would like to define different types of event handlers. And even handler should be triggered in the first step, but not in the second step.

The conditions of the event handler should be able to be filtered for the ``rollout_state`` of a token.

As an alternative we would have to define a script.

cornelinux:
Closes #1801


cornelinux:
This helps with the notification in the scenario where users should get the first QR code with the first part of 2step via email.
I am not sure, if we shoud add this to branch-3.1.

PR #1821, opened 2019-08-23
Translate the event handler help for "rollout_state".

cornelinux:
We should add the translation for the rollout_state in version 3.1, so that it looks nicer.


cornelinux:
I am sorry. Due to the autogeneration of the pot file and changes in the source files a lot of unchanged things have changed. :-/

PR #1824, opened 2019-08-27
Only re-read event handler definitions if the configuration has changed

### LINKED ISSUE #1823: Read event handler definitions from database only if config changed
Currently, privacyIDEA reads all event handler definitions from the database (along with their conditions and options) on each request. This is because each request instantiates a new ``EventConfiguration`` object:

https://github.com/privacyidea/privacyidea/blob/26088893aa3089e8461a70e02246d447c9d4fc04/privacyidea/lib/event.py#L244

Analogously to policies, we could only read event handler definitions from the database if the config timestamp has changed. This would save us at least one SQL query per request.

Then, event handler definitions would also respect the ``PI_CHECK_RELOAD_CONFIG`` setting.

fredreichbier:
This is achieved by storing the currently defined event handlers in the
shared config object.

Fixes #1823

### DIFF
--- a/privacyidea/lib/config.py
+++ b/privacyidea/lib/config.py
@@ -41,7 +41,7 @@ import threading
 
 from .log import log_with
 from ..models import (Config, db, Resolver, Realm, PRIVACYIDEA_TIMESTAMP,
-                      save_config_timestamp, Policy)
+                      save_config_timestamp, Policy, EventHandler)
 from privacyidea.lib.framework import get_request_local_store, get_app_config_value, get_app_local_store
 from .crypto import encryptPassword
 from .crypto import decryptPassword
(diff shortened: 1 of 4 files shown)


fredreichbier:
Some performance numbers on a system with a local MySQL and 2 defined event handlers, 1000 successful requests to ``/validate/check``.
On master:
```
Transactions:		        1000 hits
Availability:		      100.00 %
Elapsed time:		      237.65 secs
Data transferred:	        0.80 MB
Response time:		        0.95 secs
Transaction rate:	        4.21 trans/sec
Throughput:		        0.00 MB/sec
Concurrency:		        3.98
Successful transactions:        1000
Failed transactions:	           0
Longest transaction:	        2.15
Shortest transaction:	        0.54
```

On ``1823/event-handler-config``:
```
Transactions:		        1000 hits
Availability:		      100.00 %
Elapsed time:		      225.86 secs
Data transferred:	        0.80 MB
Response time:		        0.89 secs
Transaction rate:	        4.43 trans/sec
Throughput:		        0.00 MB/sec
Concurrency:		        3.96
Successful transactions:        1000
Failed transactions:	           0
Longest transaction:	        1.92
Shortest transaction:	        0.43
```

So, it helps a bit, but not that much :) The effect may be greater with a slower database though.

PR #1829, opened 2019-08-28
The getUserId() function returns only strings

### LINKED ISSUE #862: UUIDs instead of Strings for userid and username
# What did you try to do?

I want to setup the SQLResolver to resolve `userid` and `username` which are Postgres UUID instead of the required VARCHAR type.
The SqlAlchemy `.like` syntax used throughout the `SQLIdResolver.py` code obviously doesn't work well with UUIDs. 
I got it working to some extent by patching up those `.like`s with ` == ` but I'd prefer a more general solution.
Moreover it looks like the size of UUIDs doesn't play nicely with the `VARCHAR(20)` in the audit log. 

# Configuration

* privacyIDEA Version: 2.20.1

* Installation method: pip (inside docker)

## more details: 

* OS: Debian Jessie (Dockerized)

* Webserver: uwsgi

* Tokendatabase: PostgreSQL

Do you have a simple solution for this, or should we think about changing the SQLIdResolver?

### LINKED ISSUE #1825: User ID type INT screws up postgres
If the UserIDResolver returns a UID of type int, the resulting query on the TokenOwner table (where the UID is of type varchar) messes up postgres.
Since we use a UID of type string, the [`getUserId()`](https://github.com/privacyidea/privacyidea/blob/26088893aa3089e8461a70e02246d447c9d4fc04/privacyidea/lib/resolvers/UserIdResolver.py#L118) function should return only strings.

plettich:
privacyIDEA works with a user ID type string so we enforce that the
`getUserId()`-function only returns strings.

Fixes #1825
Also fixes #862 (yay!)

### DIFF
--- a/privacyidea/lib/resolvers/LDAPIdResolver.py
+++ b/privacyidea/lib/resolvers/LDAPIdResolver.py
@@ -60,12 +60,13 @@ import logging
 import yaml
 import threading
 import functools
+import six
 
 from .UserIdResolver import UserIdResolver
 
 import ldap3
 from ldap3 import MODIFY_REPLACE, MODIFY_ADD, MODIFY_DELETE
-from ldap3 import Server, Tls, Connection
+from ldap3 import Tls
 from ldap3.core.exceptions import LDAPOperationResult
 from ldap3.core.results import RESULT_SIZE_LIMIT_EXCEEDED
 import ssl
@@ -76,7 +77,6 @@ import traceback
 from passlib.hash import ldap_salted_sha1
 import hashlib
 import binascii
-from privacyidea.lib.crypto import urandom, geturandom
 from privacyidea.lib.utils import is_true
 from privacyidea.lib.framework import get_app_local_store
 import datetime
@@ -87,7 +87,6 @@ from privacyidea.lib.error import privacyIDEAError
 import uuid
 from ldap3.utils.conv import escape_bytes
 from operator import itemgetter
-from six import string_types
 
 CACHE = {}
 
@@ -539,7 +538,7 @@ class IdResolver (UserIdResolver):
                 if ldap_k == map_v:
                     if ldap_k == "objectGUID":
                         # An objectGUID should be no list, since it is unique
-                        if isinstance(ldap_v, string_types):
+                        if isinstance(ldap_v, six.string_types):
                             ret[map_k] = ldap_v.strip("{").strip("}")
                         else:
                             raise Exception("The LDAP returns an objectGUID, that is no string: {0!s}".format(type(ldap_v)))
@@ -573,6 +572,7 @@ class IdResolver (UserIdResolver):
         :param LoginName: The login name from the credentials
         :type LoginName: string
         :return: UserId as found for the LoginName
+        :rtype: str
         """
         userid = ""
         self._bind()
@@ -626,11 +626,11 @@ class IdResolver (UserIdResolver):
                             LoginName))
 
         for entry in r:
-            userid = self._get_uid(entry, self.uidtype)
+            userid = str(self._get_uid(entry, self.uidtype))
 
         return userid
 
-    def getUserList(self, searchDict):
+    def getUserList(self, searchDict=None):
         """
         :param searchDict: A dictionary with search parameters
         :type searchDict: dict
--- a/privacyidea/lib/resolvers/LDAPIdResolver.py
+++ b/privacyidea/lib/resolvers/LDAPIdResolver.py
@@ -82,7 +82,7 @@ from privacyidea.lib.framework import get_app_local_store
 import datetime
 
 from privacyidea.lib import _
-from privacyidea.lib.utils import to_utf8, to_unicode
+from privacyidea.lib.utils import to_utf8, to_unicode, convert_column_to_unicode
 from privacyidea.lib.error import privacyIDEAError
 import uuid
 from ldap3.utils.conv import escape_bytes
@@ -418,7 +418,7 @@ class IdResolver (UserIdResolver):
                 # we strip the curly braces from objectGUID values.
                 # If we are using ldap3 <= 2.4.1, there are no curly braces and we leave the value unchanged.
                 uid = uid.strip("{").strip("}")
-        return uid
+        return convert_column_to_unicode(uid)
 
     def _trim_user_id(self, userId):
         """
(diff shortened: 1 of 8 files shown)


privacyidea/lib/resolvers/LDAPIdResolver.py
```diff
@@ -626,11 +626,11 @@ def getUserId(self, LoginName):
                             LoginName))
 
         for entry in r:
-            userid = self._get_uid(entry, self.uidtype)
+            userid = str(self._get_uid(entry, self.uidtype))
```
fredreichbier: I guess we need to use ``six.text_type`` here (or ``to_unicode``?). If we use ``str()``, this breaks under Python 2.7 with uids that contain non-ASCII characters (like ``dn`` with an umlaut):
```
  File "/home/fred/privacyidea/privacyidea/venv/lib/python2.7/site-packages/flask/app.py", line 2446, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/fred/privacyidea/privacyidea/venv/lib/python2.7/site-packages/flask/app.py", line 1951, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/fred/privacyidea/privacyidea/venv/lib/python2.7/site-packages/flask/app.py", line 1820, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/fred/privacyidea/privacyidea/venv/lib/python2.7/site-packages/flask/app.py", line 1947, in full_dispatch_request
    rv = self.preprocess_request()
  File "/home/fred/privacyidea/privacyidea/venv/lib/python2.7/site-packages/flask/app.py", line 2241, in preprocess_request
    rv = func()
  File "/home/fred/privacyidea/privacyidea/privacyidea/api/auth.py", line 350, in decorated_function
    return f(*args, **kwargs)
  File "/home/fred/privacyidea/privacyidea/privacyidea/api/before_after.py", line 99, in before_user_request
    before_request()
  File "/home/fred/privacyidea/privacyidea/privacyidea/api/before_after.py", line 143, in before_request
    request.User = get_user_from_param(request.all_data)
  File "/home/fred/privacyidea/privacyidea/privacyidea/lib/log.py", line 154, in log_wrapper
    return func(*args, **kwds)
  File "/home/fred/privacyidea/privacyidea/privacyidea/lib/user.py", line 582, in get_user_from_param
    resolver=param.get("resolver"))
  File "/home/fred/privacyidea/privacyidea/privacyidea/lib/log.py", line 154, in log_wrapper
    return func(*args, **kwds)
  File "/home/fred/privacyidea/privacyidea/privacyidea/lib/user.py", line 98, in __init__
    self._get_user_from_userstore()
  File "/home/fred/privacyidea/privacyidea/privacyidea/lib/usercache.py", line 63, in cache_wrapper
  File "/home/fred/privacyidea/privacyidea/privacyidea/lib/user.py", line 106, in _get_user_from_userstore
    self._get_resolvers()
  File "/home/fred/privacyidea/privacyidea/privacyidea/lib/user.py", line 207, in _get_resolvers
    if self._locate_user_in_resolver(resolvername):
  File "/home/fred/privacyidea/privacyidea/privacyidea/lib/user.py", line 227, in _locate_user_in_resolver
    uid = y.getUserId(self.login)
  File "/home/fred/privacyidea/privacyidea/privacyidea/lib/resolvers/LDAPIdResolver.py", line 258, in cache_wrapper
    f_result = func(self, *args, **kwds)
  File "/home/fred/privacyidea/privacyidea/privacyidea/lib/resolvers/LDAPIdResolver.py", line 629, in getUserId
    userid = str(self._get_uid(entry, self.uidtype))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 14: ordinal not in range(128)
```

fredreichbier: Also, what do you think about moving the stringification into ``_get_uid``?

privacyidea/lib/resolvers/PasswdIdResolver.py
```diff
@@ -269,7 +269,7 @@ def getUserId(self, LoginName):
         # We do not encode the LoginName anymore, as we are
         # storing unicode in nameDict now.
         if LoginName in self.nameDict:
-            return self.nameDict[LoginName]
+            return str(self.nameDict[LoginName])
```
fredreichbier: I think we have the same problem here with Python 2.7, we should use ``six.text_type``.

privacyidea/lib/resolvers/SCIMIdResolver.py
```diff
@@ -133,7 +133,7 @@ def getUserId(self, loginName):
         #                                          ("userName", loginName)})
         #return res.get("Resources", [{}])[0].get("externalId")
         # It seems that the userName is the userId
-        return loginName
+        return str(loginName)
```
fredreichbier: I haven't tested it, but this may also break under Python 2.7 -- we should use ``text_type`` instead.

privacyidea/lib/resolvers/SQLIdResolver.py
```diff
@@ -395,7 +399,7 @@ def getUserId(self, LoginName):
                     raise Exception("More than one user with loginname"
                                     " %s found!" % LoginName)
                 user = self._get_user_from_mapped_object(r)
-                userid = user["id"]
+                userid = str(user["id"])
```
fredreichbier: Also here we might have the same problem, I guess ``six.text_type would be safer``

privacyidea/lib/resolvers/LDAPIdResolver.py
plettich: Done in 61743ea

privacyidea/lib/resolvers/PasswdIdResolver.py
plettich: Fixed in 61743ea

privacyidea/lib/resolvers/SCIMIdResolver.py
plettich: Fixed in 61743ea

privacyidea/lib/resolvers/SQLIdResolver.py
plettich: Fixed in 61743ea

privacyidea/lib/resolvers/LDAPIdResolver.py
plettich: Since we do the unicode conversion in LDAP-Resolver now, this should be fixed in 61743ea

Issue #1819, opened 2019-08-22
Store event counters and client application timestamps per node to improve performance of replication


fredreichbier:
In replicated setups like Galera, we sometimes have problems with multiple nodes trying to write to the same database row at the same time. This happens especially in case of the ``clientapplication`` table, which records the last seen date of the client application and is modified in each ``/validate/check`` request. If counting event handlers are defined, the same happens for the ``eventcounter`` tables. In case of Galera, multiple nodes trying to write to the same row might cause deadlocks (see #1268).

We could try to mitigate the issue by having each privacyIDEA node only write counters to its own private table row. As an example, with ``eventcounter``: We could add a column ``node`` to the table, which holds the name of the privacyIDEA node that the counter belongs to. Assume that we have three privacyIDEA nodes (A, B and C), each connected to its own Galera node.

Then, A updates an event counter like this: ``UPDATE eventcounter SET counter_value = counter_value + 1 WHERE counter_name = 'mycounter' AND node = 'A'``

And B updates an event counter like this: ``UPDATE eventcounter SET counter_value = counter_value + 1 WHERE counter_name = 'mycounter' AND node = 'B'``

C updates an event counter like this: ``UPDATE eventcounter SET counter_value = counter_value + 1 WHERE counter_name = 'mycounter' AND node = 'C'``

Then, when retrieving the event counter value with a given name, we would need to sum up all rows of all nodes: ``SELECT SUM(counter_value) FROM eventcounter WHERE counter_name = 'mycounter'``.

This way, each Galera node would only try to write to its own private row of ``eventcounter``, which would reduce the number of deadlocks.

However, this would only work if each privacyIDEA node has its own Galera node. It wouldn't work, for example, if the privacyIDEA nodes connect to a load balancer that selects the Galera node to use.

cornelinux:
I think it would even work, if there are several privacyIDEA Applications servers all writing to the same database, since each application server has it's own node-specifer:

nodeA, nodeB, nodeC writing to one database would result in each node updating it's own row.
Or am I missing something?

I like this idea a lot and I think we should address this in version 3.2.

fredreichbier:
Do you mean the scenario that we have no database-level replication in place, but three privacyIDEA servers ``nodeA``, ``nodeB`` and ``nodeC`` all writing to the same database ``db1``?
I also think this should work, but I'm not sure how much of a performance improvement we would get in this case -- but I guess it can't hurt :)

fredreichbier:
[...]

PR #1833, opened 2019-09-03
Store event counters separately for each privacyIDEA node

fredreichbier:
This PR implements one half of #1819: It modifies the ``eventcounter`` table such that each privacyIDEA node writes to its own "private" row. For retrieving the value of a given counter, the counter values of all rows of this counter are summed up.

This hopefully reduces the number of deadlocks in a replicated setup that occur because multiple nodes simultaneously try to write to the same table row. This does actually work, see #1819.

Writing to the correct row is completely handled by the lib level, so the API level does not need to be changed.

This also modifies the ``increase``/``decrease`` API a bit: We now don't return the new counter value -- this would require an additional SQL query, and the result is actually only used to write to the debug log. 

Writing a migration script turned out to be a bit complicated because ``counter_name`` was the primary key before, and we now need an primary ``id`` key. But alembic apparently cannot add a new primary key to an existing table, so we have to create a new table, migrate the data, drop the old table and rename the old one.


privacyidea/lib/counter.py
```diff
@@ -82,8 +105,5 @@ def read(counter_name):
...
     :return: The value of the counter
     """
-    counter = EventCounter.query.filter_by(counter_name=counter_name).first()
-    if not counter:
-        return None
-    else:
-        return counter.counter_value
\ No newline at end of file
+    return db.session.query(func.sum(EventCounter.counter_value))\
+        .filter(EventCounter.counter_name == counter_name).one()[0]
```
cornelinux: Thats a lot of logic in one line! ;-)

privacyidea/lib/eventhandler/counterhandler.py
```diff
@@ -99,14 +99,12 @@ def do(self, action, options=None):
         counter_name = handler_options.get("counter_name")
 
         if action == "increase_counter":
-            r = increase(counter_name)
-            log.debug(u"Increased the counter {0!s} to {1!s}.".format(counter_name,
-                                                                      r))
+            increase(counter_name)
+            log.debug(u"Increased the counter {0!s}.".format(counter_name))
```
cornelinux: We could have also returned the overall new counter value (over all nodes). This would have kept the log entry the same. But it is not that important to me.

tests/test_lib_counter.py
```diff
@@ -2,12 +2,25 @@
...
 
 from .base import MyTestCase
 from privacyidea.lib.counter import increase, decrease, reset, read
 from privacyidea.models import EventCounter
 
 
+def increase_and_read(name):
+    """ helper function that increases the event counter and returns the new value """
+    increase(name)
+    return read(name)
```
cornelinux: Why wouldn't we do this in ``increase`` or ``decrease``? Performance and avoiding deadlocks?

cornelinux:
@fredreichbier I had some minor comments. But none of them which keep me from merging this. Maybe you get some a 2nd thought or me some hint :-)

privacyidea/lib/eventhandler/counterhandler.py
fredreichbier: Yes, I agree that would be nice. But this results in another ``SELECT`` query -- which, I think, would be totally fine if the result is used for something important. But as it was only written to the debug log, I thought we could save one query and speed up things a bit :)

tests/test_lib_counter.py
fredreichbier: For performance, mostly :) Also I didn't want to change all occurrences of
```r = increase(...)`` in the tests to
```
increase(...)
r = read(...)
```
so I decided to go for the helper method.

Issue #1810, opened 2019-08-21
mangle event handler


cornelinux:
# Background pain

In certain scenarios the input data is not given in an optimal way.

Especially in the case of usernames in realm this can be very complicated.

E.g. the UPN of a user may be ``givenname.surname@company.com`` and a user would authenticate with this UPN at a given application. But if this username hits privacyidea

1. in case of splitAtSign=False, the user ``givenname.surname@company.com`` can not be found in the default realm, since this user does not exist in the default realm, but in the realm ``company.com``.

2. In case of splitAtSign=True, the user might be searched in the realm ``company.com`` but the user ``givenname.surname`` can not be found in this realm, because the attribte ``givenname.surname`` does not exist. The sAMAccountname might be ``username`` and the UPN is ``givenname.surname@company.com``.

Thus we need a more flexible way to tweek or mangle attributes. In this case a request with 
* user = ``givenname.surname@company.com``
* realm = ""

must result in 

* user = ``givenname.surname@company.com``
* realm = ``company.com``
This way the UPN could be found in the non-default realm ``company.com``.

# Approach of the solution

We can create a mangle event handler, that acts as a pre-Event handler, that can 
* modify existing attributes
* delete existing attributes
* create new attributes based on existing attributes.

The "allowed_positions" would probably on be ``[pre]``.
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/eventhandler/base.py#L102

Possible actions could be:
* modify
  * options:
    * name of the parameter to match
       * regular expression to match the parameter  
    * name of the parameter to modify 
       * result how the changed value should look like
  (This way we could also either overwrite the existing parameter or create a new parameter)
* add
  * options:
     * name of the parameter to add
       *value to add
* delete
  * options:
    * name of the parameter to match
      * regular expression, when to delete
    * name of the parameter to delete
   (This way we could also either delete the matching parameter or check the matching of Paramter A and delete Parameter B)

cornelinux:
I think "add" could also modify an existing parameter a.k.a. "modify" could add the new parameter, if it does not exist yet. So I would go with:

* add
  * options:
    * name of the parameter to match
       * regular expression to match the parameter  
    * name of the parameter to modify 
       * result how the changed value should look like
  (This way we could also either overwrite the existing parameter or create a new parameter)
* delete
  * options:
    * name of the parameter to match
      * regular expression, when to delete
    * name of the parameter to delete
[...]

PR #1841, opened 2019-09-04
Add events to user API

### LINKED ISSUE #1822: Event handler for user API endpoint
Add event handler hook to user endpoints, to do something when a user is created or modified.

This is basically the result of #455

cornelinux:
We can now attach events the the user API calls.

Closes #1822

### DIFF
--- a/privacyidea/api/user.py
+++ b/privacyidea/api/user.py
@@ -144,7 +144,7 @@ def delete_user(resolvername=None, username=None):
 @user_blueprint.route('/', methods=['POST'])
 @admin_required
 @prepolicy(check_base_action, request, ACTION.ADDUSER)
-@event("user_create", request, g)
+@event("user_add", request, g)
 def create_user_api():
     """
     Create a new user in the given resolver.
--- a/privacyidea/api/user.py
+++ b/privacyidea/api/user.py
@@ -3,6 +3,8 @@
 # http://www.privacyidea.org
 # (c) cornelius kölbel, privacyidea.org
 #
+# 2019-09-04 Cornelius Kölbel <cornelius.koelbel@netknights.it>
+#            Add decorators for event handling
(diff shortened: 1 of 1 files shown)


privacyidea/api/user.py
```diff
@@ -138,6 +144,7 @@ def delete_user(resolvername=None, username=None):
 @user_blueprint.route('/', methods=['POST'])
 @admin_required
 @prepolicy(check_base_action, request, ACTION.ADDUSER)
+@event("user_create", request, g)
```
fredreichbier: Should we call this ``user_add`` instead of ``user_create``, to have naming consistent with the ``adduser`` policy?

plettich:
should we add some basic testing to see if the events get triggered?

privacyidea/api/user.py
fredreichbier: I'm wondering if we could somehow test the functionality of the ``user_?`` events? I think it would be interesting to see if e.g. the ``TokenEventHandler`` would allow to enroll a token for the newly created user.

cornelinux:
@plettich @fredreichbier  You are basically right in regards to the tests. I have no time to write such tests. We only have tested a few decorations with events. I.e. not all occurences of the decoration, since we know, that these basically work.
If you want to, you are welcome to add corresponding tests.

privacyidea/api/user.py
cornelinux: see below.

cornelinux: you're right. Will do.

plettich:
> We only have tested a few decorations with events. I.e. not all ocurences of the decoration, since we know, that these basically work.

True, we can not test all possible events with all handlers. And checking if a handler was triggered requires some additional setup (but soon we have #1810 and can just add something to the response).

PR #1844, opened 2019-09-05
ResponseMangler handler to modify the JSON response

### LINKED ISSUE #1138: Event handler to remove response parameters to allow enrolling softtokens without seeing the QR Code
The helpdesk user or admin could be able to enroll a token for a user, without seeing the qrcode.

The qrcode could be sent to a user via an event handler.
Or the qrcode could be made available in the self service portal of the user, so that the user can "pick up" his token in the selfservice portal.

cornelinux:
This handler can be used to delete JSON entries from
the response or to add new ones.

This helps to define very specific workflows.

Closes #1138

### DIFF
--- a/privacyidea/lib/event.py
+++ b/privacyidea/lib/event.py
@@ -159,6 +159,7 @@ def get_handler_object(handlername):
     from privacyidea.lib.eventhandler.federationhandler import \
         FederationEventHandler
     from privacyidea.lib.eventhandler.counterhandler import CounterEventHandler
+    from privacyidea.lib.eventhandler.responsemangler import ResponseManglerHandler
     h_obj = None
     if handlername == "UserNotification":
         h_obj = UserNotificationEventHandler()
@@ -170,6 +171,8 @@ def get_handler_object(handlername):
         h_obj = FederationEventHandler()
     if handlername == "Counter":
         h_obj = CounterEventHandler()
+    if handlername == "ResponseMangler":
+        h_obj = ResponseManglerHandler()
     return h_obj
 
 
--- a/privacyidea/lib/event.py
+++ b/privacyidea/lib/event.py
@@ -163,15 +163,15 @@ def get_handler_object(handlername):
     h_obj = None
     if handlername == "UserNotification":
         h_obj = UserNotificationEventHandler()
-    if handlername == "Token":
+    elif handlername == "Token":
         h_obj = TokenEventHandler()
-    if handlername == "Script":
+    elif handlername == "Script":
         h_obj = ScriptEventHandler()
-    if handlername == "Federation":
+    elif handlername == "Federation":
         h_obj = FederationEventHandler()
-    if handlername == "Counter":
+    elif handlername == "Counter":
         h_obj = CounterEventHandler()
-    if handlername == "ResponseMangler":
+    elif handlername == "ResponseMangler":
         h_obj = ResponseManglerHandler()
     return h_obj
 
--- a/privacyidea/lib/event.py
+++ b/privacyidea/lib/event.py
@@ -159,7 +159,7 @@ def get_handler_object(handlername):
     from privacyidea.lib.eventhandler.federationhandler import \
         FederationEventHandler
     from privacyidea.lib.eventhandler.counterhandler import CounterEventHandler
-    from privacyidea.lib.eventhandler.responsemangler import ResponseManglerHandler
+    from privacyidea.lib.eventhandler.responsemangler import ResponseManglerEventHandler
     h_obj = None
     if handlername == "UserNotification":
         h_obj = UserNotificationEventHandler()
@@ -172,7 +172,7 @@ def get_handler_object(handlername):
     elif handlername == "Counter":
         h_obj = CounterEventHandler()
     elif handlername == "ResponseMangler":
-        h_obj = ResponseManglerHandler()
+        h_obj = ResponseManglerEventHandler()
     return h_obj
 
 
--- a/privacyidea/lib/eventhandler/responsemangler.py
+++ b/privacyidea/lib/eventhandler/responsemangler.py
@@ -40,7 +40,7 @@ class ACTION_TYPE(object):
     """
     Allowed actions
     """
-    ADD = "add"
+    SET = "set"
     DELETE = "delete"
 
 
@@ -81,11 +81,11 @@ class ResponseManglerEventHandler(BaseEventHandler):
                              "description": _("The JSON pointer (key) that should be deleted. Please "
                                               "specify in the format '/detail/message'.")}
                         },
-                   ACTION_TYPE.ADD:
+                   ACTION_TYPE.SET:
                        {"JSON pointer":
                             {"type": "str",
                              "required": True,
-                             "description": _("The JSON pointer (key) that should be added. "
+                             "description": _("The JSON pointer (key) that should be set. "
                                               "Please specify in the format '/detail/message'.")
                              },
                         "type":
@@ -97,7 +97,7 @@ class ResponseManglerEventHandler(BaseEventHandler):
                         "value":
                             {"type": "str",
                              "required": True,
-                             "description": _("The value of the JSON key that should be added.")
+                             "description": _("The value of the JSON key that should be set.")
                             }
                         }
                    }
@@ -137,7 +137,7 @@ class ResponseManglerEventHandler(BaseEventHandler):
                 options.get("response").data = json.dumps(content)
             except KeyError:
                 log.warning("Can not delete response JSON Pointer {0!s}.".format(json_pointer))
-        elif action.lower() == ACTION_TYPE.ADD:
+        elif action.lower() == ACTION_TYPE.SET:
             try:
                 if type == "integer":
                     value = int(value)
@@ -159,6 +159,6 @@ class ResponseManglerEventHandler(BaseEventHandler):
                     log.warning("JSON pointer of length {0!s} not supported.".format(len(comp)))
                 options.get("response").data = json.dumps(content)
             except KeyError:
-                log.warning("Can not add JSON Pointer {0!s}.".format(json_pointer))
+                log.warning("Can not set JSON Pointer {0!s}.".format(json_pointer))
 
         return ret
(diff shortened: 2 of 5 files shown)


privacyidea/lib/eventhandler/responsemangler.py
```diff
@@ -0,0 +1,176 @@
...
+            try:
+                if len(comp) == 1:
+                    content[comp[0]] = value
+                elif len(comp) == 2:
+                    content[comp[0]] = content.get(comp[0]) or {}
+                    content[comp[0]][comp[1]] = value
+                elif len(comp) == 3:
+                    content[comp[0]] = content.get(comp[0]) or {}
+                    content[comp[0]][comp[1]] = content[comp[0]].get(comp[1]) or {}
+                    content[comp[0]][comp[1]][comp[2]] = value
```
cornelinux: I checked some things with python modules like ``jsonpatch`` but they all usually have the problem, that they do not "create" the missing path.

~~~~python
jp = jsonpatch.JsonPatch([{'op': 'add', 'path': '/detail/message', 'value': "Hallo"}])
original_dict = {"original": "dict"}
jp.apply(original_dict)
~~~~
In this example I would like to add a pointer ``/detail/message`` to the "original_dict". But if the path to this pointer does not exist, yet, it will bail out with an error.
So even when using external modules, we would have to create the missing path manually which creates some nasty code.

privacyidea/lib/event.py
```diff
@@ -159,17 +159,20 @@ def get_handler_object(handlername):
...
+    elif handlername == "Script":
         h_obj = ScriptEventHandler()
-    if handlername == "Federation":
+    elif handlername == "Federation":
         h_obj = FederationEventHandler()
-    if handlername == "Counter":
+    elif handlername == "Counter":
         h_obj = CounterEventHandler()
+    elif handlername == "ResponseMangler":
+        h_obj = ResponseManglerHandler()
```
plettich: I think we should keep the naming scheme and call it ``ResponseManglerEventHandler()``. That way we can later add arbitrary handler more easy by using the ``handlername + 'EventHandler'`` object.

privacyidea/lib/eventhandler/responsemangler.py
```diff
@@ -0,0 +1,164 @@
+# -*- coding: utf-8 -*-
```
plettich: maybe we should add the ``handler`` suffix to the filename as well to make auto-discovery easier in the future.

privacyidea/lib/eventhandler/responsemangler.py
```diff
@@ -0,0 +1,164 @@
...
+                elif len(comp) == 2:
+                    content[comp[0]] = content.get(comp[0]) or {}
+                    content[comp[0]][comp[1]] = value
+                elif len(comp) == 3:
+                    content[comp[0]] = content.get(comp[0]) or {}
+                    content[comp[0]][comp[1]] = content[comp[0]].get(comp[1]) or {}
+                    content[comp[0]][comp[1]][comp[2]] = value
+                else:
+                    log.warning("JSON pointer of length {0!s} not supported.".format(len(comp)))
+                options.get("response").data = json.dumps(content)
```
plettich: Now this gets a little tricky: In case we have an endpoint that does not return JSON data (`/validate/radiuscheck`, `/token` with `outform=csv`) we would overwrite possible data here.
Maybe we should only trigger the mangler if the content-type is `application/json`?

privacyidea/lib/eventhandler/responsemangler.py
```diff
@@ -0,0 +1,164 @@
...
+                    value = int(value)
+                elif type == "bool":
+                    value = is_true(value)
+            except ValueError:
+                log.warning("Failed to convert value")
+            try:
+                if len(comp) == 1:
+                    content[comp[0]] = value
+                elif len(comp) == 2:
+                    content[comp[0]] = content.get(comp[0]) or {}
```
plettich: `dict.setdefault()` might make this a bit clearer.

tests/test_lib_events.py
```diff
@@ -909,6 +911,169 @@ def test_02_forward_admin_request(self):
...
+                   "request": req,
+                   "response": resp,
+                   "handler_def": {"options":
+                                       {"JSON pointer": "/detail/message"}
+                                   }
+                   }
+        r_handler = ResponseManglerHandler()
+        res = r_handler.do("delete", options=options)
+        self.assertTrue(res)
+        self.assertEqual(json.loads(resp.data.decode("utf-8")).get("detail").get("error"), 1)
```
plettich: since #1783 You can use `resp.json` directly like `resp.json['detail']['error']`

privacyidea/lib/eventhandler/responsemangler.py
```diff
@@ -0,0 +1,164 @@
...
+
+        :param action:
+        :param options: Contains the flask parameters g, request, response
+            and the handler_def configuration
+        :type options: dict
+        :return:
+        """
+        ret = True
+        response = options.get("response")
+        content = self._get_response_content(response)
```
plettich: We should check if we got a dictionary here before manipulating it.
Actually i broke this in 3fdac34 since i try to enforce parsing the `data` attribute as JSON.

tests/test_lib_events.py
```diff
@@ -909,6 +911,169 @@ def test_02_forward_admin_request(self):
...
+                                 headers={})
+        env = builder.get_environ()
+        # Set the remote address so that we can filter for it
+        env["REMOTE_ADDR"] = "10.0.0.1"
+        g.client_ip = env["REMOTE_ADDR"]
+        req = Request(env)
+        req.all_data = {"serial": "SPASS01", "type": "spass"}
+        resp = Response()
+
+        # delete JSON pointer with two components
```
plettich: We should also check deleting a JSON pointer with only one component.

tests/test_lib_events.py
```diff
@@ -909,6 +911,169 @@ def test_02_forward_admin_request(self):
...
+                                       {"JSON pointer": "/result"}
+                                   }
+                   }
+        r_handler = ResponseManglerHandler()
+        res = r_handler.do("delete", options=options)
+        self.assertTrue(res)
+        self.assertIn("message", json.loads(resp.data.decode("utf-8")).get("detail"))
+        self.assertNotIn("result", json.loads(resp.data.decode("utf-8")))
+
+        # JSON pointer with more than 3 components not supported
```
plettich: We should also check deleting a JSON pointer with 3 components.

tests/test_lib_events.py
```diff
@@ -909,6 +911,169 @@ def test_02_forward_admin_request(self):
...
+                                        "value": "1",
+                                        "type": "integer"}
+                                   }
+                   }
+        r_handler = ResponseManglerHandler()
+        res = r_handler.do("add", options=options)
+        self.assertTrue(res)
+        self.assertNotIn("comp1", json.loads(resp.data.decode("utf-8")))
+
+
```
plettich: We should also check adding a JSON pointer with 1 component.
And what happens if the "type" and "value" mismatch?

privacyidea/lib/eventhandler/responsemangler.py
cornelinux: I dont think so.
I would discover by inherited classes and not by filenames.

privacyidea/lib/event.py
cornelinux: Done in 070a10f9
Please check and resolve the conversation.

privacyidea/lib/eventhandler/responsemangler.py
cornelinux: Not to me ;-)

tests/test_lib_events.py
cornelinux: Good idea, please change it.

tests/test_lib_events.py
cornelinux: Good idea. Please add it.

tests/test_lib_events.py
cornelinux: Good idea, please add the test.

privacyidea/lib/eventhandler/responsemangler.py
cornelinux: The ``self._get_response_content(response)`` actually allows parsing either real json response or text/html as json. 
So I decided to catch the exception in case the response is not jsonable.
Done in 0e16f5f5

privacyidea/lib/eventhandler/responsemangler.py
cornelinux: TO my understanding this is not relevant anymore, since we catched it further upwards, if it is not json.

(see)

The ``self._get_response_content(response)`` actually allows parsing either real json response or text/html as json. 
So I decided to catch the exception in case the response is not jsonable.
Done in 0e16f5f5

privacyidea/lib/eventhandler/responsemangler.py
cornelinux: If for any reason we have a json response like:

~~~~json
{"comp0": ""}
~~~~

Setting ``/comp0/comp1 = test`` would work, (since the content would be set to ``{ "comp0": {} }``.
Whereas with setdefault() this would not work out, since the empty string would not be replaced by the dictionary. 

~~~~python
>>> d.get("hallo") or {}
{}
>>> d.setdefault("hallo", {})
''
~~~~

Thus I will keep it this way, since this gives you a greater flexibility in overwriting empty strings.

plettich: Ok, makes sense. But in this case we can remove the `try ... except KeyError` block since we would never run into one.

privacyidea/lib/eventhandler/responsemangler.py
plettich: How do you discover by inherited classes?
Anyway, i'm ok with that.

privacyidea/lib/eventhandler/responsemangler.py
plettich: I've drafted PR #1851 to return json by default. This would also change the behavior of `_get_response_content()`. But we can discuss this later.

tests/test_lib_events.py
plettich: Done in c7225471

tests/test_lib_events.py
plettich: Don in c7225471

tests/test_lib_events.py
plettich: Done in c7225471

privacyidea/lib/eventhandler/responsemangler.py
cornelinux: OK removed it.

Issue #1838, opened 2019-09-04
Add an endpoint for polling the transaction status which automatically deletes answered challenges


fredreichbier:
Assume a user has enrolled a push token and tries to log in via the owncloud app. The app triggers a challenge in the privacyIDEA server, the user smartphone receives a push notification, the user accepts the challenge. The smartphone app sends a challenge response to the ``/ttype/push`` endpoint, which sets the ``otp_valid`` flag of the challenge in the database. According to the [push token protocol](https://github.com/privacyidea/privacyidea/wiki/concept%3A-PushToken#polling-endpoint-for-application), the owncloud app polls the ``/token/challenges`` endpoint and authenticates the user if it finds a challenge with ``otp_valid`` set to true.

After authentication, the accepted challenge *stays* in the database (https://github.com/privacyidea/privacyidea-owncloud-app/issues/73). Eventually, it will be removed by the challenge janitor, but the cleanup timestamp is not really defined: The challenge janitor is, for example, invoked when a new challenge is triggered. But the challenge could potentially stay in the database for a long time.

This means that theoretically, one could still authenticate with the old transaction ID until the challenge is cleaned up. An app should not allow the user to specify a transaction ID, but privacyIDEA should not allow this behavior nonetheless.

An initial idea was to add a boolean flag to the response of ``/token/challenges`` that indicates the validity of the challenge (#1837). However, a better idea might be the following:

We add a new endpoint, like ``POST /token/finalize_transaction`` (or so), which takes a parameter ``transaction_id`` and

1) checks for a valid challenge with the given transaction ID and ``otp_valid=true``
2) if such a challenge exists, the endpoint returns ``true`` and removes *all* challenges with the given transaction ID

This endpoint could be polled by apps to check the challenge response status of out-of-band tokens (currently push + tiqr).

fredreichbier:
The documentation of the TiQR token actually proposes that applications poll the ``/validate/check`` endpoint with the transaction ID and an empty password: https://privacyidea.readthedocs.io/en/latest/modules/lib/tokentypes/tiqr.html#authentication

The ``/validate/check`` endpoint then calls ``TiqrTokenClass.check_challenge_response``, which checks for a valid challenge with ``otp_valid=true``. If it finds such a challenge, it returns true and then *actually deletes* the challenge (but other challenges with the same transaction ID stay in the database).

[...]

PR #1845, opened 2019-09-06
RequestMangler handler allows to modify Reqeust params

### LINKED ISSUE #1810: mangle event handler

cornelinux:
The request mangler can delete, add or modify
parameters in a request. The parameters can be modified
with substrings from other parameters.

Closes #1810

### DIFF
--- a/doc/eventhandler/requestmangler.rst
+++ b/doc/eventhandler/requestmangler.rst
@@ -31,7 +31,8 @@ This action is used to add or modify additional request parameters.
 
 You can set a parameter with the value or substrings of another parameter.
 
-This is why this action takes the additional options *value*, *match_parameter* and *match_pattern*.
+This is why this action takes the additional options *value*, *match_parameter* and
+*match_pattern*. *match_pattern* always needs to match the *complete* value of the *match_parameter*.
 
 If you simply want to set a parameter to a fixed value you only need the options:
 
@@ -80,6 +81,8 @@ will be modified to::
 
    user=surname.givenname@newcompany.com
 
+.. note:: The *match_pattern* in the above example will not match "surname.givenname@example.company",
+   since it always matches the complete value as mentioned above.
 
 Code
 ~~~~
--- a/doc/eventhandler/requestmangler.rst
+++ b/doc/eventhandler/requestmangler.rst
@@ -10,8 +10,8 @@ the request parameters of an HTTP request.
 This way privacyIDEA can change the data that is processed within the request.
 
 Usually this handler is used in the **pre** location. However there might be occasions
-when you want to modify paramters only *before* passing them to the next handler. In this case
-you can also use the RequestMangler handler in the **post** location.
+when you want to modify parameters only *before* passing them to the next **post** handler.
+In this case you can also use the RequestMangler handler in the **post** location.
 
 Possible Actions
 ~~~~~~~~~~~~~~~~
@@ -24,12 +24,12 @@ This action simply deletes the given parameter from the request.
 E.g. you could in certain cases delete the ``transaction_id`` from a
 ``/validate/check`` request. This way you would render challenge response inactive.
 
-add
+set
 ...
 
 This action is used to add or modify additional request parameters.
 
-You can set a parameter with the value or substrings of another paramter.
+You can set a parameter with the value or substrings of another parameter.
 
 This is why this action takes the additional options *value*, *match_parameter* and *match_pattern*.
 
@@ -43,40 +43,40 @@ If you can to set a parameter based on the value of another parameter, you can u
 
 **Example 1**
 
-To set the realm based on the user parameter:
+To set the realm based on the user parameter::
 
    parameter: realm
    match_parameter: user
    match_pattern: .*@(.*)
    value: {0}
 
-A request like:
+A request like::
 
    user=surname.givenname@example.com
    realm=
 
-with an empty realm will be modified to:
+with an empty realm will be modified to::
 
    user=surname.givenname@example.com
    realm=example.com
 
-since, the pattern <pre>.*@(.*)</pre> will match the email address and extract the domain after the "@"
+since, the pattern ``.*@(.*)`` will match the email address and extract the domain after the "@"
 sign. The python tag "{0}" will be replaced with the matching domainname.
 
 **Example 2**
 
-To simply change the domain name in the very same parameter:
+To simply change the domain name in the very same parameter::
 
    paramter: user
    match_parameter: user
    match_pattern: (.*)@example.com
    value: {0}@newcompany.com
 
-A request like:
+A request like::
 
    user=surname.givenname@example.com
 
-will be modified to
+will be modified to::
 
    user=surname.givenname@newcompany.com
 
--- a/doc/eventhandler/requestmangler.rst
+++ b/doc/eventhandler/requestmangler.rst
@@ -1,0 +1,90 @@
+.. _requestmanglerhandler:
+
+RequestMangler Handler Module
+-----------------------------
+
+.. index:: RequestMangler, Handler Modules
+
+The RequestMangler is a special handler module, that can modify
+the request parameters of an HTTP request.
+This way privacyIDEA can change the data that is processed within the request.
+
+Usually this handler is used in the **pre** location. However there might be occasions
+when you want to modify paramters only *before* passing them to the next handler. In this case
+you can also use the RequestMangler handler in the **post** location.
+
+Possible Actions
+~~~~~~~~~~~~~~~~
+
+delete
+......
+
+This action simply deletes the given parameter from the request.
+
+E.g. you could in certain cases delete the ``transaction_id`` from a
+``/validate/check`` request. This way you would render challenge response inactive.
+
+add
+...
+
+This action is used to add or modify additional request parameters.
+
+You can set a parameter with the value or substrings of another paramter.
+
+This is why this action takes the additional options *value*, *match_parameter* and *match_pattern*.
+
+If you simply want to set a parameter to a fixed value you only need the options:
+
+* *parameter*: as the name of the parameter you want to set and
+* *value*: to set to a fixed value.
+
+If you can to set a parameter based on the value of another parameter, you can use the regex notation
+**()** and the python string formatting tags **{0}**, **{1}**.
+
+**Example 1**
+
+To set the realm based on the user parameter:
+
+   parameter: realm
+   match_parameter: user
+   match_pattern: .*@(.*)
+   value: {0}
+
+A request like:
+
+   user=surname.givenname@example.com
+   realm=
+
+with an empty realm will be modified to:
+
+   user=surname.givenname@example.com
+   realm=example.com
+
+since, the pattern <pre>.*@(.*)</pre> will match the email address and extract the domain after the "@"
+sign. The python tag "{0}" will be replaced with the matching domainname.
+
+**Example 2**
+
+To simply change the domain name in the very same parameter:
+
+   paramter: user
+   match_parameter: user
+   match_pattern: (.*)@example.com
+   value: {0}@newcompany.com
+
+A request like:
+
+   user=surname.givenname@example.com
+
+will be modified to
+
+   user=surname.givenname@newcompany.com
+
+
+Code
+~~~~
+
+
+.. automodule:: privacyidea.lib.eventhandler.requestmangler
+   :members:
+   :undoc-members:
--- a/privacyidea/lib/eventhandler/requestmangler.py
+++ b/privacyidea/lib/eventhandler/requestmangler.py
@@ -148,8 +148,8 @@ class RequestManglerHandler(BaseEventHandler):
                 value = handler_options.get("value")
                 match_parameter = handler_options.get("match_parameter")
                 match_pattern = handler_options.get("match_pattern")
-                if value:
-                    # We only take action, if we have a value
+                if value is not None:
+                    # We only take action, if we have a value, even an empty string "".
                     if not match_parameter:
                         # simple setting a parameter
                         request.all_data[parameter] = value
--- a/privacyidea/lib/eventhandler/requestmangler.py
+++ b/privacyidea/lib/eventhandler/requestmangler.py
@@ -156,6 +156,12 @@ class RequestManglerHandler(BaseEventHandler):
                     elif match_pattern and match_parameter in request.all_data:
                         # setting a parameter depending on another value,
                         # but only set it, if match_parameter exists
+                        """
+                        Note: Beware user supplied  format-string like "match_pattern", it can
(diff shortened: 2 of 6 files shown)


doc/eventhandler/requestmangler.rst
```diff
@@ -0,0 +1,90 @@
...
+-----------------------------
+
+.. index:: RequestMangler, Handler Modules
+
+The RequestMangler is a special handler module, that can modify
+the request parameters of an HTTP request.
+This way privacyIDEA can change the data that is processed within the request.
+
+Usually this handler is used in the **pre** location. However there might be occasions
+when you want to modify paramters only *before* passing them to the next handler. In this case
```
fredreichbier: Typo: *parameters

doc/eventhandler/requestmangler.rst
```diff
@@ -0,0 +1,90 @@
...
+you can also use the RequestMangler handler in the **post** location.
```
fredreichbier: If I understand it correctly, having a **post** RequestMangler handler makes sense if one wants to modify parameters only before passing them to the next **post** event handler. Maybe we should make this clearer by writing "passing them to the next **post** handler" above?

doc/eventhandler/requestmangler.rst
```diff
@@ -0,0 +1,90 @@
...
+
+* *parameter*: as the name of the parameter you want to set and
+* *value*: to set to a fixed value.
+
+If you can to set a parameter based on the value of another parameter, you can use the regex notation
+**()** and the python string formatting tags **{0}**, **{1}**.
+
+**Example 1**
+
+To set the realm based on the user parameter:
```
fredreichbier: I think we should use a double colon like *... user parameter::* here to avoid weird formatting issues with the match pattern below.

doc/eventhandler/requestmangler.rst
```diff
@@ -0,0 +1,90 @@
...
+
+   parameter: realm
+   match_parameter: user
+   match_pattern: .*@(.*)
+   value: {0}
+
+A request like:
```
fredreichbier: for consistency, we should also use ``::``

doc/eventhandler/requestmangler.rst
```diff
@@ -0,0 +1,90 @@
...
+
+   user=surname.givenname@example.com
+   realm=
+
+with an empty realm will be modified to:
```
fredreichbier: also here

doc/eventhandler/requestmangler.rst
```diff
@@ -0,0 +1,90 @@
...
+
+   user=surname.givenname@example.com
+   realm=example.com
+
+since, the pattern <pre>.*@(.*)</pre> will match the email address and extract the domain after the "@"
```
fredreichbier: Does reST support ``<pre>``? In case it doesn't we should use <pre>\`\`.\*@(.\*)\`\`</pre>

privacyidea/lib/eventhandler/requestmangler.py
```diff
@@ -0,0 +1,163 @@
...
+from dateutil.tz import tzlocal
+
+log = logging.getLogger(__name__)
+
+
+class ACTION_TYPE(object):
+    """
+    Allowed actions
+    """
+    ADD = "add"
```
fredreichbier: As `ADD` also overwrites existing entries, what do you think about renaming it to ``SET``?

privacyidea/lib/eventhandler/requestmangler.py
```diff
@@ -0,0 +1,163 @@
...
+
+        if parameter:
+            if action.lower() == ACTION_TYPE.DELETE:
+                if parameter in request.all_data:
+                    del(request.all_data[parameter])
+            elif action.lower() == ACTION_TYPE.ADD:
+                value = handler_options.get("value")
+                match_parameter = handler_options.get("match_parameter")
+                match_pattern = handler_options.get("match_pattern")
+                if value:
```
fredreichbier: This wouldn't allow us to add a parameter whose value is the empty string. Do you think we would ever need this?

privacyidea/lib/eventhandler/requestmangler.py
```diff
@@ -0,0 +1,163 @@
...
+                    # We only take action, if we have a value
+                    if not match_parameter:
+                        # simple setting a parameter
+                        request.all_data[parameter] = value
+                    elif match_pattern:
+                        # setting a parameter depending on another value
+                        m = re.match(match_pattern, request.all_data.get(match_parameter))
```
fredreichbier: By default, match only searches for a match at the beginning of the match parameter value, i.e. it does not require the whole string to match. Should we maybe require ``match_pattern`` to match the whole string by passing ``"^" + match_pattern + "$"`` to ``re.match``? This seems safer to me.

privacyidea/lib/eventhandler/requestmangler.py
```diff
@@ -0,0 +1,163 @@
...
+                        if m:
+                            # Now we set the new value with the matching tuple
+                            request.all_data[parameter] = value.format(*m.groups())
```
fredreichbier: In general, one should be careful with user-supplied format strings: http://lucumr.pocoo.org/2016/12/29/careful-with-str-format/
because passing Python objects may make it possible to leak secrets. But in this case, I think we are fine because we only pass strings. Maybe we should add a comment about that?

privacyidea/lib/eventhandler/requestmangler.py
fredreichbier: This will cause an exception ``TypeError: expected string or buffer``  if ``request.all_data`` does not contain ``match_parameter``. What should we do in that case?

Maybe we should add a flag ``abort_on_error`` that describes what to do in case of error?

privacyidea/lib/eventhandler/requestmangler.py
fredreichbier: This will cause an exception if ``value`` refers to a group that ``m`` doesn't contain. Should we catch this exception, or allow the request to be aborted?

privacyidea/lib/eventhandler/requestmangler.py
cornelinux: OK, I understand. Thought about it, too, that it is not always an "add".

Will do

privacyidea/lib/eventhandler/requestmangler.py
cornelinux: I thought about this, if a value should have an empty parameter. Probably not. So you can delete the parameter.
I am not sure, if there are any cases where would need an empty-string-parameter.

privacyidea/lib/eventhandler/requestmangler.py
cornelinux: ok, i think I wanted to use ``re.search``, right?

cornelinux: So you say, I need to check if ``request.all_data(match_parameter)`` is ``None``.

I think in this case, we simply will not modify the other parameter.

privacyidea/lib/eventhandler/requestmangler.py
cornelinux: We probably should catch.

privacyidea/lib/eventhandler/requestmangler.py
fredreichbier: Just replacing ``re.match`` with ``re.search`` isn't right either because this searches for a match *anywhere* in the input. I only recently learned that ``re.match`` only requires a match at the beginning at the string -- any trailing characters *after* the match are just ignored.
So we either need to use ``re.match("^" + match_pattern + "$")`` or ``re.search("^" + match_pattern + "$")``, I think both would work.

doc/eventhandler/requestmangler.rst
cornelinux: done in a8f1447c

doc/eventhandler/requestmangler.rst
cornelinux: done in a8f1447c

doc/eventhandler/requestmangler.rst
cornelinux: done in a8f1447c

doc/eventhandler/requestmangler.rst
cornelinux: done in a8f1447c

doc/eventhandler/requestmangler.rst
cornelinux: done in a8f1447c

doc/eventhandler/requestmangler.rst
cornelinux: done in a8f1447c

privacyidea/lib/eventhandler/requestmangler.py
cornelinux: done in 4fd308ea

privacyidea/lib/eventhandler/requestmangler.py
cornelinux: Hm, this is a question of definition. It is not defined, yet, that we always want to match the **complete** value of the match_parameter.

We could very well match a substring in the middle like: ``\w(.*)@(.*)\w`` to match something like an email address anywhere in the value.

The admin can decide to put the beinnging and ending "^", "$" in the match pattern himself.

My original intention was to use ``re.search``. Why shouldn't we?

cornelinux: done in 6cc4d4a2

privacyidea/lib/eventhandler/requestmangler.py
cornelinux: done in 9d3d19ca

cornelinux: I am not quite sure, how this would be relevant to us. 
In the new parameter we can only add what was already in the old parameter.
I.e. we can only format-in what was already passed in the old parameter.

So you want me to add a comment at this code part, that using a user defined format string in this case is fine?

privacyidea/lib/eventhandler/requestmangler.py
fredreichbier: I think matching the whole string is a safer default: Assume we take the use ``re.match`` behavior, i.e. only look for a match at the beginning of the string. An admin wants to add a field "foo=bar" if the ``realm`` parameter is ``realm1``. The admin defines:
<pre>
parameter: foo
value: bar
match_parameter: realm
match_value: realm1
</pre>
This works for a while. Now the setup grows, and there is a new realm ``realm12``. Now, the rule above suddenly matches ``realm12`` as well, which is probably not intended by the admin.

If we require the pattern to match the whole string instead, the rule wouldn't match ``realm12``. In order to get the old ``re.match`` behavior, the admin could still define a pattern ``realm1.*``.

privacyidea/lib/eventhandler/requestmangler.py
fredreichbier: The problem with a user-supplied format string in combination with object arguments is that a rogue admin could try to extract e.g. the ``SECRET_KEY`` with a carefully crafted format string.

So yes, adding a comment would be nice I think, like:
```python
# Using a user-supplied format string can be dangerous: http://lucumr.pocoo.org/2016/12/29/careful-with-str-format/
# but in this case it is fine because no objects are involved, as re.groups() always returns a tuple of strings
```
so if someone decides to extend this in the future, there is a warning :)

privacyidea/lib/eventhandler/requestmangler.py
fredreichbier: I am also not sure. But I think there are cases in which the API differentiates between an empty and a non-existent parameter, e.g. with the parameter ``pass`` in ``/validate/check``: The endpoint complains if ``pass`` is not supplied, but works fine if ``pass`` is empty (as seen in #1838)

cornelinux: OK, I will change it to ``is not None``.

cornelinux: done in e1e123d0

privacyidea/lib/eventhandler/requestmangler.py
cornelinux: Thank you. You convinced me!

cornelinux: done in 5bd08dd5

privacyidea/lib/eventhandler/requestmangler.py
cornelinux: Done in b534ba2e

privacyidea/lib/eventhandler/requestmangler.py
fredreichbier: thanks!

privacyidea/lib/eventhandler/requestmangler.py
fredreichbier: thanks!

privacyidea/lib/eventhandler/requestmangler.py
fredreichbier: thanks!

PR #1848, opened 2019-09-09
Fix the key of the tokentype log entry

### LINKED ISSUE #1846: token_type typo in validate.py audit
Hi @cornelinux,

It seems that there is a typo in the validate.py for `token_type` param, where it is spelled as `tokentype` instead.

For example:
```
    g.audit_object.log({"info": log_used_user(user, details.get("message")),
                        "success": result,
                        "serial": serial or details.get("serial"),
                        "tokentype": details.get("type")})
```

Correct one should be:
```
    g.audit_object.log({"info": log_used_user(user, details.get("message")),
                        "success": result,
                        "serial": serial or details.get("serial"),
                        "token_type": details.get("type")})
```

This cause the token type is not in our audit. There are multiple occurrence of this typos not just the above block of code.

Can we please confirm?

Best regards,

Quynh Nguyen

cornelinux:
Fixes #1846

We could also through an assertion if we get a wrong key. The question is, if our tests would cover all possible occurrences. This is why I chose the log-warning. Open to discussions.


privacyidea/lib/auditmodules/sqlaudit.py
```diff
@@ -223,6 +223,11 @@ def finalize_log(self):
             self.audit_data["policies"] = ",".join(self.audit_data.get("policies",[]))
             if self.config.get("PI_AUDIT_SQL_TRUNCATE"):
                 self._truncate_data()
+            if "tokentype" in self.audit_data:   # pragma: no cover
+                log.warning("We have a wrong 'tokentype' key. This should not happen. Fix it!. "
+                            "Error occurs in action: {0!r}.".format(self.audit_data.get("action")))
+                if not "token_type" in self.audit_data:
+                    self.audit_data["token_type"] = self.audit_data.get("tokentype")
```
fredreichbier: I think this approach is fine, but I think it would be nice if you could add a small test for the branch that rewrites ``tokentype`` -> ``token_type``. It could be as easy as passing a ``tokentype`` parameter and checking that the resulting ``LogEntry`` has the ``token_type`` field set correctly. What do you think?

cornelinux: Will do.

cornelinux: done in 3442b31f

fredreichbier: thanks! seems like the tests are failing now because the commit happens to contain ``audit.sqlite``. I'll fix this.

cornelinux: darn. I always try to remove these changing testdata/ Thanks!

PR #1849, opened 2019-09-09
Store clientapplication "last seen" info per privacyIDEA node

### LINKED ISSUE #1819: Store event counters and client application timestamps per node to improve performance of replication

fredreichbier:
This also adds a corresponding migration script and tests and reduces the number of deadlocks to zero in the setup mentioned in #1819.

The migration script has two disadvantages:
* It doesn't work on SQLite, because it doesn't allow removing constraints
* Downgrading probably doesn't work, because removing the ``node`` field may violate UNIQUE constraints. But I'm not sure if downgrading currently works at all? :)

Fixes #1819

### DIFF
--- a/tests/test_lib_clientapplication.py
+++ b/tests/test_lib_clientapplication.py
@@ -1,8 +1,12 @@
 """
 This test file tests the lib.clientapplicaton.py
 """
+import mock
+from datetime import datetime, timedelta
+from contextlib import contextmanager
+
+from privacyidea.models import ClientApplication
 from .base import MyTestCase
-from datetime import datetime
 from privacyidea.lib.clientapplication import (get_clientapplication,
                                                save_clientapplication)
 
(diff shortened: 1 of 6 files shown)


tests/test_lib_clientapplication.py
```diff
@@ -38,3 +42,86 @@ def test_01_save_and_get(self):
...
+        self.assertEqual(row3.lastseen, t2)
+        # ... but on pinode2, it was t1!
+        row4 = ClientApplication.query.filter_by(ip="1.2.3.4", clienttype="RADIUS", node="pinode2").one()
+        self.assertEqual(row4.lastseen, t1)
+
+        # check that the apps are returned correctly
+        apps = get_clientapplication(ip="1.2.3.4")
+        self.assertEqual(set(apps.keys()), {"PAM", "RADIUS"})
+        self.assertEqual(apps["PAM"], [{"ip": "1.2.3.4", "hostname": None, "lastseen": t2}])
+        self.assertEqual(apps["RADIUS"], [{"ip": "1.2.3.4", "hostname": None, "lastseen": t2}])
```
plettich: Can we assume, that the dictionary order does not change here?

tests/test_lib_clientapplication.py
```diff
@@ -38,3 +42,86 @@ def test_01_save_and_get(self):
...
+
+        apps = get_clientapplication(group_by="ip")
+        self.assertEqual(set(apps.keys()), {"1.2.3.4", "2.3.4.5"})
+        self.assertEqual(len(apps["1.2.3.4"]), 2)
+        self.assertIn({"clienttype": "PAM", "hostname": None, "lastseen": t2}, apps["1.2.3.4"])
+        self.assertIn({"clienttype": "RADIUS", "hostname": None, "lastseen": t2}, apps["1.2.3.4"])
+        self.assertEqual(apps["2.3.4.5"], [{"clienttype": "PAM", "hostname": None, "lastseen": t1}])
```
plettich: same here. Or does `assertEqual()` ignore dictionary order?

tests/test_lib_clientapplication.py
fredreichbier: I think even though dictionaries officially remember the insertion order in Python 3.7, this doesn't affect equality checks: so ``{"a": "b", "c":"d"}`` will always be equal to ``{"c": "d", "a": "b"}``, and this is what ``assertEqual`` checks :)

PR #1851, opened 2019-09-10
Set the default return type to 'application/json'

### LINKED ISSUE #1850: Fix JSON content detection
While working on #1616 i broke the [`_get_response_content()`](https://github.com/privacyidea/privacyidea/blob/a128999d61f7d6fed114ed371528ab7d766be4d8/privacyidea/lib/eventhandler/base.py#L303) function in 3fdac34.
In case a response does not contain any JSON data, this would result in an error.
This would happen if a post-event was bound to an endpoint like `/validate/radiuscheck`.

plettich:
Since we mostly return JSON data, it makes sense to the set it as the
default return type in out response class.
Also added a `send_file()` utility function which adjusts content type and
header accordingly.
Other mimetypes are set explicitly.

Fixes #1850

### DIFF
--- a/privacyidea/api/lib/utils.py
+++ b/privacyidea/api/lib/utils.py
@@ -148,6 +148,18 @@ def send_error(errstring, rid=1, context=None, error_code=-311, details=None):
     return ret
 
 
+def send_html(output):
+    """
+    Send the ouput as HTML to the client with the correct mimetype.
+
+    :param output: The HTML to send to the client
+    :type output: str
+    :return: The generated response
+    :rtype: flask.Response
+    """
+    return current_app.response_class(output, mimetype='text/html')
+
+
 def send_file(output, filename, content_type='text/csv'):
     """
     Send the output to the client with the "Content-disposition" header to
--- a/privacyidea/api/lib/utils.py
+++ b/privacyidea/api/lib/utils.py
@@ -148,6 +148,12 @@ def send_error(errstring, rid=1, context=None, error_code=-311, details=None):
     return ret
 
 
+def send_file(output, filename):
+    content_type = "application/force-download"
+    headers = {'Content-disposition': 'attachment; filename={0!s}'.format(filename)}
+    return current_app.response_class(output, headers=headers, mimetype=content_type)
+
+
 def send_csv_result(obj, data_key="tokens",
                     filename="privacyidea-tokendata.csv"):
     """
@@ -169,8 +175,6 @@ def send_csv_result(obj, data_key="tokens",
     :rtype: Response object
     """
     delim = "'"
-    content_type = "application/force-download"
-    headers = {'Content-disposition': 'attachment; filename={0!s}'.format(filename)}
     output = u""
     # Do the header
     for k, _v in obj.get(data_key, {})[0].items():
@@ -187,7 +191,7 @@ def send_csv_result(obj, data_key="tokens",
             output += "{0!s}{1!s}{2!s}, ".format(delim, value, delim)
         output += "\n"
 
-    return current_app.response_class(output, mimetype=content_type)
+    return send_file(output, filename)
 
 
 @log_with(log)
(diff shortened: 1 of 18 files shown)


privacyidea/api/lib/utils.py
```diff
@@ -148,6 +148,12 @@ def send_error(errstring, rid=1, context=None, error_code=-311, details=None):
     return ret
 
 
+def send_file(output, filename):
+    content_type = "application/force-download"
```
fredreichbier: I'm not sure about the ``application/force-download`` mimetype here. I didn't know it, so I googled it, and [it seems like](https://stackoverflow.com/a/10616753) only shouldn't really use it, but instead set the actual content mimetype. I'm wondering -- should we do that instead? What do you think?

privacyidea/api/lib/utils.py
```diff
@@ -148,6 +148,12 @@ def send_error(errstring, rid=1, context=None, error_code=-311, details=None):
     return ret
 
 
+def send_file(output, filename):
```
fredreichbier: For the final PR, could you add a small docstring?

tests/test_api_audit.py
```diff
@@ -25,6 +28,38 @@ def test_00_get_audit(self):
             self.assertTrue(json_response.get("result").get("value").get(
                 "current") == 1, res)
 
+    def test_01_get_audit_csv(self):
```
fredreichbier: nice testcase!

privacyidea/api/lib/utils.py
plettich: Done in ba5ab0b

privacyidea/api/lib/utils.py
plettich: Fixed in ba5ab0b

privacyidea/webui/login.py
```diff
@@ -58,7 +58,8 @@ def single_page_application():
 
     if current_app.config.get("PI_UI_DEACTIVATED"):
         # Do not provide the UI
-        return render_template("deactivated.html")
+        return current_app.response_class(render_template("deactivated.html"),
```
cornelinux: We often use the ``current_app.respsonse_class(render_template...)`` which return an html mimetype. Would it be nice to have a wrapper funktion for this? 
(e.g. the same in webui/certificate.py)

privacyidea/webui/certificate.py
```diff
@@ -98,14 +100,14 @@ def cert_enroll():
...
-
+    render_context = {'instance': instance,
+                      'backendUrl': backend_url,
+                      'username': "{0!s}@{1!s}".format(request.PI_username,
+                                                       request.PI_realm),
+                      'role': request.PI_role,
+                      'serial': serial,
+                      'certificate': certificate,
+                      'cert_pem': cert_pem }
+    return current_app.response_class(render_template("token_enrolled.html",
```
cornelinux: see below

privacyidea/webui/certificate.py
```diff
@@ -58,9 +58,11 @@ def cert_form():
     backend_url = ""
     authtoken = request.all_data.get("authtoken")
     ca = request.all_data.get("ca")
-    return render_template("cert_request_form.html", instance=instance,
-                           backendUrl=backend_url, ca=ca,
-                           authtoken=authtoken)
+    return current_app.response_class(render_template("cert_request_form.html",
```
cornelinux: see below

privacyidea/webui/login.py
plettich: Fixed in 12f8421e

privacyidea/webui/certificate.py
plettich: see above...

privacyidea/webui/certificate.py
plettich: see above...

PR #1854, opened 2019-09-14
Notification Handler can write to a file

### LINKED ISSUE #717: Enhance Notification Handler to write information to a file in spool directory
The enrollment tab for HOTP and TOTP token could provide a letter template, when the QR Code is enrolled.

A policy could define the template file.
The template file could read contact data from the user source.
The created QR Code letter should be editable.

cornelinux:
The notification handler can create a file with
a lot of tags in a spool directory.

Closes #717

### DIFF
--- a/privacyidea/lib/eventhandler/usernotification.py
+++ b/privacyidea/lib/eventhandler/usernotification.py
@@ -361,7 +361,7 @@ class UserNotificationEventHandler(BaseEventHandler):
                 random = get_alphanum_str(16)
                 filename = filename.format(random=random, **tags)
                 try:
-                    with open(spooldir + "/" + filename, "wb") as f:
+                    with open(spooldir + "/" + filename, "w") as f:
                         f.write(body)
                 except Exception as err:
                     log.error(u"Failed to write notification file: {0!s}".format(err))
--- a/privacyidea/lib/eventhandler/usernotification.py
+++ b/privacyidea/lib/eventhandler/usernotification.py
@@ -48,6 +48,7 @@ from privacyidea.lib.smtpserver import get_smtpservers
 from privacyidea.lib.smsprovider.SMSProvider import get_smsgateway
 from privacyidea.lib.user import User, get_user_list
 from privacyidea.lib.utils import create_tag_dict
+from privacyidea.lib.crypto import get_alphanum_str
 from privacyidea.lib import _
 import logging
 import datetime
@@ -182,7 +183,25 @@ class UserNotificationEventHandler(BaseEventHandler):
                                       "description": _("Send notification to "
                                                        "this user."),
                                       "value": [NOTIFY_TYPE.TOKENOWNER]}
-                               }
+                               },
+                   "savefile": {"spooldirectory":
+                                    {"type": "str",
+                                     "required": True,
+                                     "description": _("Save the notification file to "
+                                                      "this directory.")},
+                                "body":
+                                    {"type": "text",
+                                     "required": True,
+                                     "description": _("This is the template content of "
+                                                      "the new file. Can contain the tags "
+                                                      "as specified in the documentation.")},
+                                "filename":
+                                    {"type": "str",
+                                     "required": True,
+                                     "description": _("The filename of the notification. Existing files "
+                                                      "are overwritten. The name can contain tags as specified "
+                                                      "in the documentation and can also contain the tag {random}.")}
+                   }
                    }
         return actions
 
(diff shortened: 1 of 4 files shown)


privacyidea/lib/eventhandler/usernotification.py
```diff
@@ -335,6 +355,17 @@ def do(self, action, options=None):
                     log.warning("Failed to send a notification email to user "
                                 "{0}".format(recipient))
 
+            elif action.lower() == "savefile":
+                spooldir = handler_options.get("spooldirectory", ".")
+                filename = handler_options.get("filename")
+                random = get_alphanum_str(16)
+                filename = filename.format(random=random, **tags)
+                try:
+                    with open(spooldir + "/" + filename, "w") as f:
```
fredreichbier: We should make sure that one cannot "break out" of ``spooldir`` by specifying a filename like ``../../foo/bar``.

plettich: currently this works like a charm... But seriously, neither filename nor directory are user-controlled and if the admin defining the handler wants to do harm...
Anyway we should sanitize it in case it gets used for something else in the future.

fredreichbier: I agree that protecting against rogue admins isn't as urgent as protecting against rogue users :) But still, we should sanitize the filename here.
@cornelinux suggested that we define a root spool directory in ``pi.cfg`` (instead of the event handler configuration) and only allow filenames without ``..``. What do you think?

plettich: You can use the following snipped to check the filename:
```python
# expand tags and remove possible leading '/'
filename = filename.format(random=random, **tags).lstrip(os.path.sep)
outfile = os.path.normpath(os.path.join(spooldir, filename))
if not outfile.startswith(spooldir):
    log.error('Cannot write outside of spooldir {0!s}!'.format(spooldir))
```

plettich: I'll write a test to check for directory traversal.

cornelinux: done in ac1c326d

tests/test_lib_events.py
```diff
@@ -2867,4 +2867,86 @@ def test_20_sendmail_googleurl(self):
...
+                   "handler_def": {"options":
+                                       {"filename": "test{serial}.txt",
+                                        "body": "{serial}, {user}"}
+                                   }
+                   }
+
+        # create a file, that is not writable
+        with open("tests/testdata/testOATH123456.txt", "w") as f:
+            f.write("empty")
+        os.chmod("tests/testdata/testOATH123456.txt", 0x400)
```
fredreichbier: ``0x400`` is hexadecimal, but I think file modes are octal? should probably be ``0o400``

tests/test_lib_events.py
```diff
@@ -2867,4 +2867,86 @@ def test_20_sendmail_googleurl(self):
...
+                   "handler_def": {"options":
+                                       {"filename": "../../../test{serial}.txt",
+                                        "body": "{serial}, {user}"}
+                                   }
+                   }
+
+        un_handler = UserNotificationEventHandler()
+        un_handler.do("savefile", options=options)
+        # Check the log file for an error. There might be faster possibilities but more robust
+        with open("privacyidea.log") as f:
```
fredreichbier: I like that you check that a warning was actually written, but I think we shouldn't read the complete logfile for that. Under Python 3, we could use the pretty cool ``assertLogs`` method: https://stackoverflow.com/a/34920727
But unfortunately that hasn't been backported to Python 2.7.

Here, I would prefer a solution using mocks. I can also modify this if you prefer.

tests/test_lib_events.py
```diff
@@ -2867,4 +2867,86 @@ def test_20_sendmail_googleurl(self):
...
+                   }
+
+        # create a file, that is not writable
+        with open("tests/testdata/testOATH123456.txt", "w") as f:
+            f.write("empty")
+        os.chmod("tests/testdata/testOATH123456.txt", 0x400)
+        un_handler = UserNotificationEventHandler()
+        un_handler.do("savefile", options=options)
+        # Check the log file for an error. There might be faster possibilities but more robust
+        with open("privacyidea.log") as f:
```
fredreichbier: same as above

privacyidea/lib/eventhandler/usernotification.py
fredreichbier: Nice

tests/test_lib_events.py
cornelinux: done in 2b85a2fe

tests/test_lib_events.py
cornelinux: You are probably right. But I would also like to keep it in a way, that others understand what is happening ;-)
I think in the test environment the log files are not that long.

If it is understandable for other, do it, in the other case rather leave it.

tests/test_lib_events.py
cornelinux: see above ;-)

tests/test_lib_events.py
fredreichbier: Fair enough. I pushed a commit that uses mocks here: b746ac31e4b6eae181df71bccba93235f20c12b4
I find it readable :-) What do you think?

PR #1859, opened 2019-09-19
Add a /validate/polltransaction endpoint for out-of-band tokens

fredreichbier:
This implements the endpoint specified in #1838. I'm happy about any comments, especially regarding naming :)

I'd like to move the documentation part mentioned in #1838 to another PR.

### DIFF
--- a/privacyidea/api/validate.py
+++ b/privacyidea/api/validate.py
@@ -542,10 +542,11 @@ def trigger_challenge():
 
 
 @validate_blueprint.route('/polltransaction', methods=['GET'])
+@validate_blueprint.route('/polltransaction/<transaction_id>', methods=['GET'])
 @prepolicy(mangle, request=request)
 @CheckSubscription(request)
 @prepolicy(api_key_required, request=request)
-def poll_transaction():
+def poll_transaction(transaction_id=None):
     """
     Given a mandatory transaction ID, check if any non-expired challenge for this transaction ID
     has been answered. In this case, return true. If this is not the case, return false.
@@ -556,7 +557,8 @@ def poll_transaction():
 
     :jsonparam transaction_id: a transaction ID
     """
-    transaction_id = getParam(request.all_data, "transaction_id", required)
+    if transaction_id is None:
+        transaction_id = getParam(request.all_data, "transaction_id", required)
     # Fetch a list of challenges with the given transaction ID
     # and determine whether it contains at least one non-expired answered challenge.
     matching_challenges = get_challenges(transaction_id=transaction_id)
--- a/privacyidea/api/validate.py
+++ b/privacyidea/api/validate.py
@@ -71,7 +71,7 @@ from ..lib.decorators import (check_user_or_serial_in_request)
 from .lib.utils import required
 from privacyidea.lib.error import ParameterError
 from privacyidea.lib.token import (check_user_pass, check_serial_pass,
-                                   check_otp, create_challenges_from_tokens)
+                                   check_otp, create_challenges_from_tokens, get_one_token)
 from privacyidea.api.lib.utils import get_all_params
 from privacyidea.lib.config import (return_saml_attributes, get_from_config,
                                     return_saml_attributes_on_fail,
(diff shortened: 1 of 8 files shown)


privacyidea/api/validate.py
```diff
@@ -539,3 +540,56 @@ def trigger_challenge():
...
+    # and determine whether it contains at least one non-expired answered challenge.
+    matching_challenges = get_challenges(transaction_id=transaction_id)
+    answered_challenges = extract_answered_challenges(matching_challenges)
+
+    if answered_challenges:
+        result = True
+        log_challenges = answered_challenges
+    else:
+        result = False
+        log_challenges = matching_challenges
```
fredreichbier: I just noticed something: In case there *are* challenges with the given transaction ID, but *none* of these challenges have been answered, this will write the serials of all challenges with the given transaction ID to the audit log, including serials of expired challenges. Should we only consider non-expired challenges?

privacyidea/api/validate.py
```diff
@@ -539,3 +540,56 @@ def trigger_challenge():
 
     return send_result(result_obj, details=details)
 
+
+@validate_blueprint.route('/polltransaction', methods=['GET'])
```
cornelinux: To be more *RESTish* with GET-Requests we could add a second route-decorator:

@validate_blueprint.route('/polltransaction/<transacion_id>', methods=['GET'])

(compare GET requests in the token-endpoint)

privacyidea/api/validate.py
cornelinux: You are right, I think we should **not** log all the challenges, that are not relevant.
Or we should only log this, if a valid/answered challenge was found.

privacyidea/api/validate.py
```diff
@@ -539,3 +540,56 @@ def trigger_challenge():
...
+
+    This is mostly useful for out-of-band tokens that should poll this endpoint
+    to determine when to send an authentication request to ``/validate/check``.
+
+    :jsonparam transaction_id: a transaction ID
+    """
+    transaction_id = getParam(request.all_data, "transaction_id", required)
+    # Fetch a list of challenges with the given transaction ID
+    # and determine whether it contains at least one non-expired answered challenge.
+    matching_challenges = get_challenges(transaction_id=transaction_id)
```
cornelinux: Wouldn't it be nicer to enhance get_challenges like

    get_challenges(transaction_id=transaction_id, otp_valid=True)

so that we only have one call?

privacyidea/lib/challenge.py
```diff
@@ -150,3 +151,21 @@ def _create_challenge_query(serial=None, transaction_id=None):
             sql_query = sql_query.filter(Challenge.transaction_id == transaction_id)
 
     return sql_query
+
+
+def extract_answered_challenges(challenges):
```
cornelinux: See my comment above. Shouldn't we skip this method?

privacyidea/api/validate.py
fredreichbier: Hm yes, this actually seems more RESTish. Maybe we should *only* implement the ``/polltransaction/<transaction_id>`` endpoint?

privacyidea/api/validate.py
fredreichbier: I guess the question is what information would be useful in the audit log: If there *are* challenges with the given transaction ID, but *none* of them has been answered -- should the audit log contain the serials of the unanswered challenges, and the respective user? i.e. would this be useful information to the helpdesk user?

privacyidea/api/validate.py
fredreichbier: This would be a possibility! But then we would have two SQL queries -- one for *all* challenges of the transaction, and another one for all challenges of the transaction with ``otp_valid=True``. So I did this differently here, in order to have only one SQL query.

But if we want to change this to not log the unanswered challenges (see discussion below), I agree your solution seems better.

cornelinux:
In the past usually we implemented both ways.I would like to do so in this case, too.
null

privacyidea/api/validate.py
fredreichbier: I've added a second endpoint in 250a3cf.

privacyidea/api/validate.py
fredreichbier: I modified the endpoint to only consider *non-expired* challenges in 7e16c1a.

privacyidea/lib/challenge.py
fredreichbier: I think as we want to keep logging serials of non-answered challenges to the audit log, we should keep this function to save one database query.

fredreichbier:
As it turns out, ``/validate/triggerchallenge`` currently does not write the serials of triggered tokens to the audit log. I have a local branch that adds this -- I can open a new PR once this one is merged :)

privacyidea/lib/challenge.py
cornelinux: ok.

Issue #1261, opened 2018-09-25
Versioning in js and css


cornelinux:
Introduce version numbers for .js and .css files (/file.js?ver1.2.3) so they can be easily incremented after a modification or an update of PrivacyIDEA. This would cause the browser to reload those instead of
reading them from the local cache and will avoid potential issues on the client side.

fredreichbier:
We could use [flask-versioned](https://pythonhosted.org/Flask-Versioned/) for that.

PR #1860, opened 2019-09-19
1261/versioning in js and css

NuvandaPV:
This will version all stylesheets and scripts (also the favicon, for good measure) with their respective timestamps, so caches will get invalidated properly.


NuvandaPV:
Fixes #1261

NuvandaPV:
Nvm. Seems I made Travis unhappy somehow.

fredreichbier:
I think this would be very nice to have! Could this also implement versioning in angular html templates?

Travis seems to break because some endpoints do not pass the ``theme`` variable to ``render_template``, i.e. ``cert_form`` in ``api/certificate.py``, which makes it ``None`` in the template, which makes the ``|versioned`` filter unhappy. I'm currently unsure how this would be fixed best.

privacyidea/webui/login.py
```diff
@@ -62,6 +62,7 @@ def single_page_application():
 
     # The default theme. We can change this later
     theme = current_app.config.get("PI_CSS", DEFAULT_THEME)
+    theme = theme.strip('/');
```
fredreichbier: A bit nitpicky, but could you remove the ``;``? :)

NuvandaPV: That's totally not nitpicking. That shouldn't have been there and has been exorcized.

NuvandaPV:
I have implemented your suggestion and wrapped theme loading in an `ifdef`. Tests should be all green now.

Versioning the Angular templates would be definitely possible, however I don't think there is much merit to implementing a cache invalidation strategy for these, since in the current architecture they are dynamically generated and thus cannot be cached anyways (in fact the versioning as implemented is among the things that get dynamically added to these templates). Ideally I'd like for that to change, to allow for greater performance, security and deployment flexibility, however that would probably have to go along with a major rewrite of large parts of the webui.

fredreichbier:
Thanks for the fix!

This *does* add a new dependency on flask-versioned, but I think having versioned .js and .css is worth it. @cornelinux what do you think?

Regarding the Angular templates versioning: I *think* there are two different kinds of templates in privacyIDEA. The templates in ``static/templates/`` are rendered on the server-side by Flask+Jinja (so they dynamically generated), and they now use Flask-Versioned to version .js and .css files. The templates in ``static/components`` are not rendered server-side (so they are static), but fetched and rendered on the client-side by Angular. If I recall correctly, we sometimes have problems with updated  privacyIDEA instances, because "old" angular templates are still cached by the browser, and users need to clear their cache to get the "new" angular templates. I think theoretically, it should be possible to version these templates too, but I agree that it would probably be quite hard. So let's only implement this for .js and .css for now.

cornelinux:
I am not sure if we ever had problems with .js and .css. 
I am fine with an additional dependency.

The only thing I like to point out: An installation should not run into problems, when they customized their stuff - i.e. would this mean if an installation customized a css (like a rewrite rule in apache) they would have to adapt this with each and every update? We can not allow this!

fredreichbier:
> The only thing I like to point out: An installation should not run into problems, when they customized their stuff - i.e. would this mean if an installation customized a css (like a rewrite rule in apache) they would have to adapt this with each and every update? We can not allow this!

That's a valid concern! @NuvandaPV But I think this should not cause problems, because we merely append the version to the query string?

NuvandaPV:
@fredreichbier oh, right. I got mixed up there, sorry. You are right, adding some sort of versioning to these would be useful as well. However it seems that would require employing a different approach. I will open an issue.

@cornelinux This PR doesn't actually modify the caching behavior (or any other Apache configuration) at all, actually. So it should not create any new problems. If a user has a custom theme, or uses our custom styles feature, that will now be automagically versioned as well, fixing any caching issues they may have had, without requiring any changes on their side. Nothing will have to be adapted for every update, since the „version“ is simply the timestamp of the last modification of the respective files and gets generated automatically on the server. Changes to webserver configuration will not be required, as the query parameter will invalidate all caches on the way, but will otherwise be ignored on the server (as @fredreichbier already pointed out). In the worst case, a rewrite rule could be causing the webserver to deliver entirely different files, than the ones Python is seeing, with would cause the cache to not be invalidated reliably. This scenario seems unlikely to me, but if it is the case for one of our customers, this pull request will not cause any problems for them, it will just fail to fix the problems it is designed to address.

requirements.txt
```diff
@@ -54,3 +54,4 @@ SQLAlchemy==1.3.0
 sqlsoup==0.9.1
 urllib3==1.24.3
 Werkzeug==0.14.1
+flask-versioned==0.9.4.post20101221
```
cornelinux: This code is from 2010?

Is there a newer possibiliy? I think new versions of flask actually come with components and features, that had to be added with extenions earlier. I would be curious, if there is a newer possibility.

PR #1861, opened 2019-09-19
Style checkboxes to match up with their labels.

NuvandaPV:


NuvandaPV:
Fixes #1655

NuvandaPV:
What you are seeing is not how this is intended to look. The checkboxes should be centered on the line. Firefox will render things slightly from platform to platform and depending on the DE and toolkits you have installed and the adjustments to the visual preferences to these various toolkits and you seem to have just found an edge case.

The fix actually works for you, in that it fixes a bug in PrivacyIDEA, that would shift down the checkboxes by 4 px in some browsers because the css was nonsensical, and browsers would cope differently.

But you are seeing a different bug, where the checkboxes will be aligned with the top of the text, if they are higher than the space that is available for them (bigger than the line-height, essentially). I seem to be unable to reproduce this, however. I have tried Debian and macOS, each with their respective defaults, but in both cases, the checkboxes are alot tinier. Could you provide me with some information about your particular setup?

fredreichbier:
I see, thanks for the explanation! I remember a similar conversation with @Mipronimo -- apparently I have a weird browser setup :)

To be honest, I have no idea why the checkboxes look this way in Firefox. -- they arent't this big in other gtk-based applications. I'm on Arch Linux and Firefox Quantum 69.0 -- if you have any idea where to look, I'm happy to help.

In any case, as the checkboxes don't look weirder than before for me, I guess this PR is fine. :)

NuvandaPV:
Could you maybe give me the contents of your `$XDG_CONFIG_HOME/gtk-3.0/settings.ini` and `/etc/gtk-3.0/settings.ini`? Also try adding a new string-key called `widget.content.gtk-theme-override` to `about:config`, with the content `Adwaita:light`. If this really is a gtk-theming issue, that should allow you to see the page as Firefox will render it on most Linux-distros.

fredreichbier:
Just for the record: We found out why the checkboxes looks weird only on my system. The reason is that my Arch Linux system was apparently configured to use Nimbus Sans as a replacement for Helvetica:
```
$ fc-match -s helvetica        
NimbusSans-Regular.otf: "Nimbus Sans" "Regular"
LiberationSans-Regular.ttf: "Liberation Sans" "Regular"
...
```
And with Nimbus Sans, the height of the label's <span> one pixel too small, which messes up the checkbox alignment.

Removing Nimbus Sans makes ``fc-match`` select Liberation Sans instead, which fixes the problem.

As I have no idea why Nimbus Sans is the Helvetica replacement on my system, I'll merge this. :)

PR #1864, opened 2019-09-21
1196/random pin

### LINKED ISSUE #1196: Allow helpdesk to reset token pin
We should allow the helpdesk to reset the PIN of a token, without having to manually read the PIN to the user.

The user calls the helpdesk and asks to reset the PIN.
The helpdesk would reset the PIN, but the helpdesk would not see the PIN and the helpdesk would not read the PIN to the user.

The PIN could be sent to the user's email address.

This could be accomplished using the NotificationHandler or a PINHandlingModule.
We also would need a API call like ``token/resetpin``.

> Currently the helpdesk would enroll a registration token for the user, the user would use this registration token to login once and then the user would have to reset the PIN himself.

cornelinux:
This closes #1196 

The endpoint /token/setrandompin can be used manually in the WebUI.

The notification handler uses the tag *{pin}*.

The token handler has an action to set a random PIN.


privacyidea/api/token.py
```diff
@@ -671,6 +671,41 @@ def setpin_api(serial=None):
...
+    :rtype: json object
+    """
+    if not serial:
+        serial = getParam(request.all_data, "serial", required)
+    g.audit_object.log({"serial": serial})
+    user = request.User
+    encrypt_pin = getParam(request.all_data, "encryptpin")
+    pin = getParam(request.all_data, "pin")
+    if not pin:
+        raise TokenAdminError("We have an empty PIN. Please check you policy 'otp_pin_random'.")
```
fredreichbier: *your

privacyidea/static/components/token/views/token.details.html
```diff
@@ -362,6 +362,22 @@ <h3 name="Token Details"><translate>Token details for {{ tokenSerial }}
             </td>
         </tr>
 
+        <tr ng-hide="token.locked || !checkRight('setrandompin')">
```
fredreichbier: I'm not entirely sure about the UI here:
* I think having an extra row that only contains the "set random PIN" and, after a button click, the randomly generated PIN looks a bit weird. What do you think about moving the "set random PIN" button next to the ordinary "set PIN" button, and displaying the randomly generated PIN in a modal dialog with a message "The randomly generated PIN is:" instead? 
* The endpoint ``/token/setrandompin`` does not work without a OTPPINRANDOM policy. But the "set random PIN" button is displayed in any case, which I find it unfortunate. What do you think? We could, for example, use a default length to make the endpont work even without a OTPPINRANDOM policy.

privacyidea/api/token.py
cornelinux: thanks. will do

privacyidea/lib/eventhandler/usernotification.py
```diff
@@ -324,7 +325,7 @@ def do(self, action, options=None):
                                    escape_html=action.lower() == "sendmail" and
                                                handler_options.get("mimetype", "").lower() == "html")
 
-            body = body.format(googleurl_img=googleurl_img,
+            body = body.format(googleurl_img=googleurl_img, pin=pin,
```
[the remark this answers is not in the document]
cornelinux: The html escape sounds sensible. 
I think I did not add it, since the pin was only used at one place. I will check again and take according actions.

privacyidea/static/components/token/views/token.details.html
cornelinux: I am also not completely happy with that. But I was not able to come up with better ideas. Here are my thoughts:

I personally do not like modal dialogs! ;-)
I think we will often use the "set random pin" with an event handler, that actually removes the PIN from the response. So we will get an empty modal dialog. This is wired. We get the blue infogox that the pin is set. I think this is fine.

I think an additional line is not the problem. Because either policies will be configured this way, that a helpdesk user is allowed to manually set a pin or to set a random pin.

In regards to the default length: I will give it some thinking...
This would also avoid raising the error above. But currently I can not remember if there was a reason for that. 

In near future I am sure, that we not only would want to define the length of the random PIN but also the charset (like only set digits)

privacyidea/api/lib/prepolicy.py
```diff
@@ -151,6 +152,9 @@ def init_random_pin(request=None, action=None):
                                                unique=True,
                                                audit_data=g.audit_object.audit_data)
 
+    if action=="setrandompin" and len(pin_pols) == 0:
```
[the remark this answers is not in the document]
cornelinux: I think we might have a logic problem here: the otp_pin_random policy current generates a random PIN during the enrollment of a token. 

This might be problematic: If we want to set a random pin only manually, the configuration/policy will also set a random pin during /token/init, which might not be the desired behaviour.

What I think woud actually be the right way would be to remove the decorator "init_random_pin" from /token/init and only have the policy otppinrandom for the /token/setrandompin endpoint *and* for the token handler.

The /token/init random PIN could than be handled by the tokenhandler.

I think it is important to split the /token/init and the /token/setrandompin in regards to actually create a random pin. Since as mentioned above.

What do *you* think?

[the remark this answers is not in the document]
cornelinux: Or we should use a 2nd decorator and a 2nd policy.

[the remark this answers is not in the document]
cornelinux: I did this in 7d23bd01

privacyidea/api/token.py
cornelinux: done in 499ed665

privacyidea/lib/eventhandler/usernotification.py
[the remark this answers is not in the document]
cornelinux: done in 6d96ca1a

doc/policies/admin.rst
```diff
@@ -201,6 +209,20 @@ and it must not contain any other characters.
...
+otp_pin_set_random
+~~~~~~~~~~~~~~~~~~
+
+type: integer
+
+range: 1-31
+
+The administrator can set a random pin for a token
+with the endpoint ``token/setrandompin``.
+This policy is needed to define how log the PIN will be.
```
fredreichbier: *long

privacyidea/api/lib/prepolicy.py
```diff
@@ -131,9 +131,44 @@ def policy_wrapper(*args, **kwds):
...
+    PIN during Token enrollment
+    """
+    params = request.all_data
+    policy_object = g.policy_object
+    # Fixme: we need to allow passing the user
+    #user_object = get_user_from_param(params)
+    # get the length of the random PIN from the policies
+    adminname = g.logged_in_user.get("username")
+    adminrealm = g.logged_in_user.get("realm")
+    pin_pols = policy_object.get_action_values(action=ACTION.OTPPINSETRANDOM,
```
fredreichbier: Correct me if I'm wrong, but wouldn't this give wrong results if the currently logged-in user is a user, not an admin? In this case, we would match the policy attribute ``adminrealm`` against the user's realm.

privacyidea/lib/eventhandler/usernotification.py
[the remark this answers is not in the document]
fredreichbier: thanks!

privacyidea/api/lib/prepolicy.py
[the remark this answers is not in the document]
fredreichbier: Thanks, I like the solution with a second policy.

privacyidea/api/lib/prepolicy.py
fredreichbier: I just noticed that the docstring says ``OTPPINSETRANDOM`` is in scope=ENROLL instead of scope=ADMIN -- I think this would make more sense?

privacyidea/static/components/token/views/token.details.html
fredreichbier: Hm, I understand your hesitation regarding the modal dialog -- But in the default configuration (without any policies), the token view now contains an additional row only with a "Set random PIN", which in my opinion looks a bit weird :) And once a corresponding ``otp_pin_set_random`` is defined, I find it kind of counter-intuitive that a clicking "Set random PIN" makes the PIN appear in the second cell of the row, while the first cell is still empty. :-/ Maybe we could add a label like
```
<span ng-show="randomPin">Randomly generated PIN:</span>
```
to the first cell of the row?

cornelinux: I make a translateable "random generated PIN".

I am not sure if randomly generated means it might be generated or not.
But it means if it is generated, it is a random PIN ;-)

cornelinux: done in 8107fb3d

privacyidea/api/lib/prepolicy.py
cornelinux: A user can not set a random pin.

cornelinux: done in a2785dfe

doc/policies/admin.rst
cornelinux: done in a2785dfe

privacyidea/api/lib/prepolicy.py
fredreichbier: Hm, I'm not sure I understand: Doesn't ``user.rst`` say that 

> If the ``setrandompin`` action is defined, the user
> is allowed to call the endpoint, that sets a random PIN on his
> specified token.

?

cornelinux: Yes it says. and lib/policy.py also allows the user to do so.

cornelinux: Easy, so we need to distinguish the loggedin role here.

cornelinux: done in 83a95f73
(I think I will not write an additional test ;-)

privacyidea/static/components/token/views/token.details.html
fredreichbier: thx :)

privacyidea/api/lib/prepolicy.py
fredreichbier: Sorry, I'm still not sure this is correct: Now, if a user is logged in and uses the ``/token/setrandompin`` endpoint, the decorator checks for a scope=USER, ``otp_pin_set_random`` policy action. But it is impossible to define such a policy in the web UI. :/

Do we even want a policy action ``otp_pin_set_random`` in scope=ADMIN as well as in scope=USER?

cornelinux: otp_pin_set_random defines the length of the OTP PIN, right?

In the scope user we have a policy "setrandompin". So I think we allow the user for whatever reasons to set a random pin on his tokens.

But: Then we are simply missing the policy ``otp_pin_set_random`` in the policy definition. I would simply add it.

fredreichbier: OK, I guess that makes sense. So this would even allow to configure different lengths for random PINs for users and admins.

cornelinux: done in 83a95f73

cornelinux:
@fredreichbier hm, squasching would be nice - I think :-)

privacyidea/api/lib/prepolicy.py
fredreichbier: thanks! I think that's nicer for consistency.
Right now, we still have the behavior that without any user policies, the "set random PIN" button is displayed, but does not work without a "otp_pin_set_random" policy. Should we keep it like that, or should we set a default length?

privacyidea/api/lib/prepolicy.py
```diff
@@ -131,9 +131,52 @@ def policy_wrapper(*args, **kwds):
...
+                                               user=username,
+                                               realm=realm,
+                                               #user_object=user_object,
+                                               client=g.client_ip,
+                                               unique=True,
+                                               audit_data=g.audit_object.audit_data)
+
+    if len(pin_pols) == 0:
+        # We do this to avoid that an admin sets a random PIN manually!
+        raise TokenAdminError("You need to specify an admin policy 'otp_pin_set_random'.")
```
fredreichbier: For users, we need a user policy, not an admin policy. Should we just remove the word "admin" from this error message?

PR #1865, opened 2019-09-24
Add documentation on out-of-band tokens and the /validate/polltransaction endpoint

### LINKED ISSUE #1838: Add an endpoint for polling the transaction status which automatically deletes answered challenges

fredreichbier:
Closes #1838

### DIFF
--- a/doc/configuration/authentication_modes.rst
+++ b/doc/configuration/authentication_modes.rst
@@ -1,0 +1,133 @@
+.. _authentication_modes:
+
+Authentication Modes
+====================
+
+privacyIDEA supports a variety of tokens that implement different
+authentication flows. We call these flows *authentication modes*. Currently,
+tokens may implement three authentication modes, namely ``authenticate``,
+``challenge`` and ```outofband``.
+
+Application plugins need to implement the three authentication modes
+separately, as the modes differ in their user experience. For example:
+
+* The HOTP token type implements the ``authenticate`` mode, which is a
+  single-shot authentication flow. For each authentication request, the user
+  uses their token to generate a new HOTP value and enters it along with their
+  OTP PIN. The plugin sends both values to privacyIDEA, which decides whether
+  the authentication is valid or not.
+* The E-Mail token type implements the ``challenge`` mode. With an E-Mail token,
+  the authentication flow consists of two steps: In a
+  first step, the plugin triggers a challenge. privacyIDEA sends the challenge
+  response --- a fresh OTP value --- to the user via E-Mail.
+  In a second step, the user responds to the challenge by entering the
+  respective OTP value in the plugin's login form. The plugin sends the
+  challenge response to privacyIDEA, which decides whether the authentication
+  is valid or not.
+* The PUSH token type implements the ``outofband`` mode. With a PUSH token,
+  the authentication step also consists of two steps:
+  In a first step, the user triggers a challenge. privacyIDEA pushes the
+  challenge to the user's smartphone app. In a second step, the user approves
+  the challenge on their phone, and the app responds to the challenge on behalf
+  of the user. The plugin periodically queries privacyIDEA to check if
+  the challenge has been answered correctly and the authentication is valid.
+
+The following describes the authentication flows of the three authentication
+modes in more detail.
+
+.. _authentication_mode_authenticate:
+
+``authenticate`` mode
+---------------------
+
+ * The user enters a OTP PIN along with an OTP value.
+ * The plugin sends a request to the ``/validate/check`` endpoint:
+
+   .. code-block:: text
+
+     POST /validate/check
+
+     user=<user>&pass=<PIN+OTP>
+
+  and privacyIDEA returns whether the authentication request has succeeded
+  or not.
+
+.. _authentication_mode_challenge:
+
+``challenge`` mode
+------------------
+
+ * The plugin triggers a challenge, for example via the
+   ``/validate/triggerchallenge`` endpoint:
+
+   .. code-block:: text
+
+     POST /validate/triggerchallenge
+
+     user=<user>
+
+   The plugin receives a transaction ID which we call ``transaction_id`` and
+   asks the user for the challenge response.
+ * The user enters the challenge response, which we call ``OTP``.
+   The plugin forwards the response to privacyIDEA along with the
+   transaction ID:
+
+   .. code-block:: text
+
+     POST /validate/check
+
+     user=<user>&transaction_id=<transaction_id>&pass=<OTP>
+
+  and privacyIDEA returns whether the authentication request succeeded or not.
+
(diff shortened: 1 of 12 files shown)


doc/configuration/authentication_modes.rst
```diff
@@ -0,0 +1,133 @@
...
+``challenge`` and ```outofband``.
+
+Application plugins need to implement the three authentication modes
+separately, as the modes differ in their user experience. For example:
+
+* The HOTP token type implements the ``authenticate`` mode, which is a
+  single-shot authentication flow. For each authentication request, the user
+  uses their token to generate a new HOTP value and enters it along with their
+  OTP PIN. The plugin sends both values to privacyIDEA, which decides whether
+  the authentication is valid or not.
```
cornelinux: There is only one authentication request

doc/configuration/authentication_modes.rst
```diff
@@ -0,0 +1,133 @@
...
+* The E-Mail token type implements the ``challenge`` mode. With an E-Mail token,
```
cornelinux: "The E-Mail token or SMS-Token type impement"

doc/configuration/authentication_modes.rst
```diff
@@ -0,0 +1,133 @@
...
+  the authentication flow consists of two steps: In a
+  first step, the plugin triggers a challenge. privacyIDEA sends the challenge
+  response --- a fresh OTP value --- to the user via E-Mail.
+  In a second step, the user responds to the challenge by entering the
+  respective OTP value in the plugin's login form. The plugin sends the
+  challenge response to privacyIDEA, which decides whether the authentication
+  is valid or not.
+* The PUSH token type implements the ``outofband`` mode. With a PUSH token,
```
cornelinux: "or TiQR token type"

doc/configuration/authentication_modes.rst
```diff
@@ -0,0 +1,133 @@
...
+  the authentication step also consists of two steps:
+  In a first step, the user triggers a challenge. privacyIDEA pushes the
+  challenge to the user's smartphone app. In a second step, the user approves
+  the challenge on their phone, and the app responds to the challenge on behalf
```
cornelinux: it responds to the challenge by communicating to privacyIDEA server on behalf of the user.

doc/configuration/authentication_modes.rst
```diff
@@ -0,0 +1,133 @@
...
+  and privacyIDEA returns whether the authentication request has succeeded
+  or not.
+
+.. _authentication_mode_challenge:
+
+``challenge`` mode
+------------------
+
+ * The plugin triggers a challenge, for example via the
+   ``/validate/triggerchallenge`` endpoint:
```
cornelinux: we should either add a reference to challenge response or we should note, that the challenge is also triggered by sending the PIN via /validate/check

doc/configuration/authentication_modes.rst
```diff
@@ -0,0 +1,133 @@
...
+
+   .. code-block:: text
+
+     GET /validate/polltransaction
+
+     transaction_id=<transaction_id>
+
+   If this endpoint returns ``false``, the challenge has not been answered yet.
+ * The user approves the challenge on a separate device, e.g. their
+   smartphone app. The app communicates with a tokentype-specific endpoint
```
cornelinux: tokentype-specific endpoint of privacyIDEA!

doc/configuration/authentication_modes.rst
```diff
@@ -0,0 +1,133 @@
...
+   finalize the authentication via the ``/validate/check`` endpoint, sending
+   an empty password:
+
+   .. code-block:: text
+
+     POST /validate/check
+
+     user=<user>&transaction_id=<transaction_id>&pass=
+
+   This step is crucial because the ``/validate/check`` endpoint takes defined
```
cornelinux: note the empty pass.

doc/configuration/authentication_modes.rst
fredreichbier: thanks! fixed in 076624be5af5d799ff9e8cdf6e6d362f1781e495

doc/configuration/authentication_modes.rst
fredreichbier: fixed in 076624be5af5d799ff9e8cdf6e6d362f1781e495

doc/configuration/authentication_modes.rst
fredreichbier: fixed in 076624be5af5d799ff9e8cdf6e6d362f1781e495

doc/configuration/authentication_modes.rst
fredreichbier: fixed in 076624be5af5d799ff9e8cdf6e6d362f1781e495

doc/configuration/authentication_modes.rst
fredreichbier: fixed in 076624be5af5d799ff9e8cdf6e6d362f1781e495

doc/configuration/authentication_modes.rst
fredreichbier: made this more explicit in 076624be5af5d799ff9e8cdf6e6d362f1781e495

Issue #1444, opened 2019-02-12
Optimize SQLAlchemy relationship loading techniques


fredreichbier:
In our models, we use several different lazy and eager [loading techniques](https://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html). We should investigate at which places it makes sense to use lazy or eager loading techniques. This might improve performance, e.g. if we get rid of an eager JOIN whose result is never read, or if we get rid of an extra lazy SELECT if we always need its result.

fredreichbier:
[Some thoughts](https://github.com/privacyidea/privacyidea/pull/1368/files/4fefec62f7611ede769eb5df970fd633cb94aff6#r255291624) by @cornelinux:
> I think Lazy loading (lazy='select' https://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html) is good if
> 
>     1. you usually do not need the relation or need it later or
> 
>     2. if you related object is very big.
> 
> 
> The lazy=select issues a 2nd select statement for the relation. It many cases it is probably better to jave _one_ joines select statement than 2 select statements.
> 
> So we need to look at, where we need the relationship immediately or where we need the relationship later - we have no _big_ object (So this reason for lazy loading does not count).
> # class Token
> 
> _info_ and _owner_. I think we _mostly_ need the additional tokeninfo and we _always_ need the owner of the token, if we are handling with tokens.
> So my guess would be to change these two
> https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L180
> https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L184
> actually to 'joined'.
> # TokenInfo
> 
> oups, we have a double realtion here,
> https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L642
> although we only have an 1:m relation (one token, many attributes)
> But also in this case, I would prefer a 'joined' laziness.
> I will check, if we can drop the info in the Token.
> # Resolver / ResolverConfig
> 
> https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L921
> https://github.com/privacyidea/privacyidea/blob/1288/more-users-per-token/privacyidea/models.py#L960token/privacyidea/models.py#L921
> 
> Here we also have a probably unnessary double relation.
> But this needs to be joined, since a resolver is never used without its config.
> # ResolverRealm
> 
[...]

PR #1866, opened 2019-09-24
Models: Remove obsolete relationships

fredreichbier:
This only removes some cases of duplicated relationships: e.g. there was a relationship ``info`` from ``Token`` to ``TokenInfo`` with a backref called ``info``, *and* a relationship ``token`` from ``TokenInfo`` to ``Token`` with a backref called ``info_list``: Hence, ``Token`` had both ``info`` and ``info_list``, and ``TokenInfo`` had both ``info`` and ``token``. With this PR, ``Token`` has only ``info_list`` and ``TokenInfo`` has only ``token``.

In the unit tests, this PR also replaces all occurrences of ``Token.owners.first()`` with ``Token.first_owner`` in order to make it easier to change the relationship loading technique of ``owners`` from ``dynamic`` to something else in the future.

This commit does not alter the relationship loading techniques and should not have any performance impact.

Working on #1444

### DIFF
--- a/privacyidea/models.py
+++ b/privacyidea/models.py
@@ -190,9 +190,7 @@ class Token(MethodsMixin, db.Model):
                             default=1000)
     rollout_state = db.Column(db.Unicode(10),
                               default=u'')
-    info = db.relationship('TokenInfo',
-                           lazy='dynamic',
-                           backref='info')
+    info_list = db.relationship('TokenInfo', lazy='select', backref='token')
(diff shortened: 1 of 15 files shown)


cornelinux:
Did we have an issue for that? It would be great, if we could reference one.
I am a bit cautious to not break things, and if so, we have a ref point to an issue.

fredreichbier:
Yup, this is a first step of #1444, and this issue should actually be referenced in the commit mesage already :)

PR #1869, opened 2019-09-26
Enhance OverrideAuthorizationClients setting to allow multiple proxies

### LINKED ISSUE #1868: Allow more sophisticated proxy settings in OverrideAuthorizationClients
Right now, ``OverrideAuthorizationClients`` allows to specify proxies that may map the client IP to addresses in specified subnets, [see here](https://privacyidea.readthedocs.io/en/latest/configuration/system_config.html?highlight=override#override-authorization-client).

However, this setting is not flexible enough in setups with multiple (reverse) proxies.

Consider the following setup:
```
1.2.3.4 -------> 10.0.0.1 -------> 192.168.1.1 -----> privacyIDEA
client           proxy1              proxy2
```

It would be nice if ``OverrideAuthorizationClients`` allowed arbitrarily deeply nested of mapping settings. This way, we could define the setting ``192.168.1.1>10.0.0.1>1.2.0.0/16``: In the above setup, this would allow the two trusted proxies ``proxy1`` and ``proxy2`` to map the client IP to 1.2.3.4.

fredreichbier:
Fixes #1868

### DIFF
--- a/privacyidea/lib/utils/__init__.py
+++ b/privacyidea/lib/utils/__init__.py
@@ -661,7 +661,8 @@ def get_client_ip(request, proxy_settings):
     # From the X-Forwarded-For header, we determine the path to the actual client, i.e. the list of proxy servers
     # that the HTTP response will pass through, including the final client IP:
     # If a client C talks to a proxy P1, which in turn talks to a proxy P2, which talks to privacyIDEA,
-    # X-Forwarded-For will be "P1, P2", and path_to_client will be [P2, P1, C].
+    # X-Forwarded-For will be "C, P1", the HTTP client IP will be P2, and path_to_client
+    # consequently will be [P2, P1, C].
     # However, if we get such a request, we cannot be sure if the X-Forwarded-For header is correct,
     # or if it was sent by a rogue client in order to spoof its IP address.
     # To prevent IP spoofing, privacyIDEA allows to configure a list of proxies that are allowed to override the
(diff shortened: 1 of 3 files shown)


fredreichbier:
Hm, in this implementation, ``check_proxy`` allows to take parts of proxy paths into account, even if the whole path does not match, e.g. in the following setup:

```
1.2.3.4 --------> 127.0.0.1 --------> privacyIDEA
client            proxy
```

with the proxy setting ``"127.0.0.1>1.2.3.4>2.3.4.5"``, it would map the IP to ``1.2.3.4``, even though the setup does not match ``"127.0.0.1>1.2.3.4>2.3.4.5"``:
```
 >>> check_proxy([IPAddress("127.0.0.1"), IPAddress("1.2.3.4")], "127.0.0.1>1.2.3.4>2.3.4.5")
IPAddress('1.2.3.4')
```

The question is whether this is OK, or if we should rather fall back to ``127.0.0.1``.

fredreichbier:
With 96cfacb, we should now fall back to 127.0.0.1 in the above example.

privacyidea/lib/utils/__init__.py
```diff
@@ -545,52 +546,66 @@ def parse_date(date_string):
 
 def parse_proxy(proxy_settings):
     """
-    This parses the string of the system settings
-    OverrideAuthorizationClient.
+    This parses the string of the system settings OverrideAuthorizationClient into a set of "proxy paths",
```
cornelinux: Can you please add some log output ``log.debug`` in this code, so that we can know afterwards, why a certain client IP was identitfied.

fredreichbier: good idea! I pushed some debug printouts

PR #1870, opened 2019-09-26
[DRAFT] Use new API for policy matching at more places

fredreichbier:
OK, I know I said this PR wouldn't be so big. As it turns out, this was a lie. :)
But in isolation, these commits aren't so bad:

* a2dfdce adds ``Match.generic``, a "legacy" interface for policy matching with the same parameters as ``match_policies``/``get_action_values``.
* It also moves the ``PolicyClass.ui_*`` methods from the ``PolicyClass`` object to to a new module ``privacyidea.api.lib.ui`` and, more importantly, updates them to use the new ``Match`` API.
* 8974623 converts some more policy matching operations to the new API
 * da57564 changes the signature of ``TokenClass.get_default_settings`` such that the classmethod receives the ``g`` object, and then uses the new policy matching API there
*  da57564 adds a new method ``Match.allowed()`` that implements the ``action``/``action_at_all`` pattern that we often use: This function, for example, checks if a certain scope=ADMIN action is allowed, which is the case if there are *no* scope=ADMIN policies or if there is a policy that allows it.
* 3502456 moves a remaining usage of ``get_action_values`` to the new API.

I realize some of these changes are a bit questionable, so it might even make sense to only cherry-pick some changes (like a2dfdce), or so. I just thought it would be good to push them anyway so they don't get lost.


cornelinux:
Honestly I am not that happy with this Pull Request.
We really should talk about major changes and code rewrite **before** doing them, having a pull request and then pop up and tell "look what I have, lets talk about that".
This does not only hold for this PR but for a lot of other code that currently happens here. 

Why having a generic method? When the goal would be to remove the old way anyways?

Why moving ui policies at all? They are still policies.

I am closing this. I do not want to cherry pick.

cornelinux:
So since I incorporated all important changes in other pull requests, I am closing this one.

PR #1876, opened 2019-09-30
Allow RSA private key to sign JWT

### LINKED ISSUE #1773: Allow authentication with a predefined JWT

The API should be usable in a user context without the need for the user to authenticate with a password or with an OTP token.

The system calling the API could have a JWT for this very user. privacyIDEA could trust certain public keys. The JWT is sent to the ``/auth`` endpoint. With the trusted public key we could check the signature in the JWT and authenticate the user, so that we can issue further API requests in the user context.

**Additional context**

There can be bascially two different approaches to configure this.
1. We add a new loginmode "JWT" to "disable, privacyIDEA, userstore" in the policy WebUI.
    Then we can simply check this in the lib/auth.py::check_webui_user
    However, this would mean that from a certain IP address a user could only **either** authenticate with a password/OTP **or** with a JWT.

2. Similar to the **REMOTE_USER** functionality we could have an additional policy and
   check for a valid JWT *before* we check for the user's password/OTP.
   This probably creates more unpleasant code than the first approach. This would roughly be here:
   https://github.com/privacyidea/privacyidea/blob/master/privacyidea/api/auth.py#L242
   ~~~~python
   elif is_jwt_user_allowed(request):
       # check the jwt
   ~~~~

3. We could trust the JWT to directly be used as JWT on the API end points. The check with other
    trusted public keys would be directly done in [check_auth_token](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/api/auth.py#L369) or ``verify_auth_token``.

    The public keys could be trusted like this in PI_CFG:

    ~~~~python
    PI_TRUSTED_JWT = [ {"public_key":b'-----BEGIN PUBLIC KEY-----\nMHYwEAYHKoZIzj0CAQYFK4EEAC...' ,
                                           "JWT_algorithm":  "RS256",
                                           "restricted_IPs": "...from where",
                                           "restricted_roles": ["User"],
                                           "restricted_realms": ["realm1", "realm2"],
                                           "restricted_ users": ["userA", "userB"]},
                                          { }]
   ~~~~
   This way we can very well define, that a certain IP would be allowed to impersante only Users (not admins) in a given realm. Note: In additional of course all policies are checked.

I prefer version 3, since it keeps all other mechanisms intact.

# Todos

* load the trusted JWT from ``pi.cfg``. 
* in ``verify_auth_token`` iterate through the different possible JWTs. 
* check, if the given JWT confirms with the defined restrictions.

cornelinux:
Closes #1773


doc/installation/system/inifile.rst
```diff
@@ -155,4 +155,37 @@ can give these different nodes, dedicated names.
...
+
+   PI_TRUSTED_JWT = [{"public_key": "-----BEGIN PUBLIC KEY-----\nMIIBIjANBgkqhkiG9w0BAQEF...",
+                       "algorithm": "RS256",
+                       "role": "user",
+                       "realm": "realm1",
+                       "user": "userA",
+                       "resolver": "resolverX"}]
+
+
+This entry means, that the private key, that corresponds the the given
```
plettich: ... corresponds _to_ the given ...

doc/installation/system/inifile.rst
```diff
@@ -155,4 +155,37 @@ can give these different nodes, dedicated names.
...
+public key can sign a JWT, that can impersonate as the *userA* in resolver
+*resolverX* in *realmA*.
+
+A JWT can be signed like this::
```
plettich: created?

cornelinux: I think I really ment *signed*, since the ``dict`` is somehow signed with the key. 
But you are right, *created* sounds more sensible.

Will do.

cornelinux: done in 5ce20321

doc/installation/system/inifile.rst
cornelinux: done in 5ce20321

privacyidea/api/lib/utils.py
```diff
@@ -247,19 +247,44 @@ def verify_auth_token(auth_token, required_role=None):
...
+    headers = jwt.get_unverified_header(auth_token)
+    if headers.get("alg") == "RS256":
+        # The trusted JWTs are RSA signed
+        trusted_jwts = current_app.config.get("PI_TRUSTED_JWT", [])
+        for trusted_jwt in trusted_jwts:
+            try:
+                if trusted_jwt.get("algorithm") == "RS256":
+                    j = jwt.decode(auth_token,
+                                   trusted_jwt.get("public_key"),
+                                   algorithms=["RS256"])
```
plettich: I think we should not limit the algorithms to 'RS256'.
With ``jwt.decode(auth_token, public_key, algorithms=list(jwt.algorithms.requires_cryptography))`` we can enforce to use only public-key crypto algorithms for the token verification.
Or we could use a list of allowed algorithms (to also include HMAC with a shared secret).

plettich: Also we should at least use RSA-PSS (algorithms ``PS256``, ``PS384`` and ``PS512``) since PKCS1v15 is deprecated.

doc/installation/system/inifile.rst
cornelinux: @plettich if you like this fix, please resolve the conversation

privacyidea/api/lib/utils.py
cornelinux: You are right. Will do.

privacyidea/api/lib/utils.py
```diff
@@ -247,19 +247,44 @@ def verify_auth_token(auth_token, required_role=None):
...
+                    if j == dict((k, trusted_jwt.get(k)) for k in ("role", "user", "resolver", "realm")):
```
plettich: I think we'll run into the unordered dictionary issue here.
Also this would mean, that the payload must not contain more elements than specified in ``PI_TRUSTED_JWT``.
We could use python sets to solve both: ``if {"role", "user", "resolver", "realm"} <= six.viewkeys(j)``

cornelinux: Hm, but in python2 and python3 two dicts compare to true:

    {"H": "b", 1:3} == {1:3, "H": "b"}

And this is what we want here.

privacyidea/api/lib/utils.py
cornelinux: done in  #a3e20f13

privacyidea/api/lib/utils.py
cornelinux: done in 41a29da1
This works for a lot of dicts:

~~~~python
>>> d1 = {"hallo": "du", 1:3, 17:2}
>>> d2 = {1:3, "hallo": "du"}
>>> d1 == d2
False
>>> dict((k, d1.get(k)) for k in ("hallo", 1)) == dict((k, d2.get(k)) for k in ("hallo", 1))
True
~~~~

privacyidea/api/lib/utils.py
```diff
@@ -247,19 +250,46 @@ def verify_auth_token(auth_token, required_role=None):
...
-    except jwt.DecodeError as err:
-        raise AuthError(_("Authentication failure. Error during decoding your token: {0!s}").format(err),
-                        id=ERROR.AUTHENTICATE_DECODING_ERROR)
-    except jwt.ExpiredSignature as err:
-        raise AuthError(_("Authentication failure. Your token has expired: {0!s}").format(err),
-                        id=ERROR.AUTHENTICATE_TOKEN_EXPIRED)
+
+    headers = jwt.get_unverified_header(auth_token)
+    algorithm = headers.get("alg")
+    if algorithm in TRUSTED_JWT_ALGOS:
```
plettich: This will be checked by `jwt.decode()` implicitely.

privacyidea/api/lib/utils.py
```diff
@@ -247,19 +250,46 @@ def verify_auth_token(auth_token, required_role=None):
...
+        # The trusted JWTs are RSA, PSS or eliptic curve signed
+        trusted_jwts = current_app.config.get("PI_TRUSTED_JWT", [])
+        for trusted_jwt in trusted_jwts:
+            try:
+                if trusted_jwt.get("algorithm") in TRUSTED_JWT_ALGOS:
```
plettich: Do we still want to restrict the required algorithm in the trusted JWT description? Technically we just need the public key and the list of allowed algorithms in `jwt.decode()`. That way, the *creator* of the JWT can change the algorithm (as long as it is in the trusted algorithms list).

privacyidea/config.py
```diff
@@ -4,6 +4,27 @@
 basedir = "/".join(basedir.split("/")[:-1]) + "/"
 
 
+pubtest_key = b"""-----BEGIN PUBLIC KEY-----
```
plettich: Since this data is only used in the JWT testcase, i think we should put it in the corresponding test file. I'll fix that.

cornelinux:
Yes. Merge it I explain on Monday.
null

privacyidea/api/lib/utils.py
cornelinux: ￼	Yes. Merge it I explain on Monday.

PR #1877, opened 2019-10-01
Added option to pi-manage to skip header

basvandervlies:
The option is -q/--quiet so we can use it in our configuration
management software.

### DIFF
--- a/pi-manage
+++ b/pi-manage
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/opt/privacyidea/3.1/.venv/bin/python3.7
 # -*- coding: utf-8 -*-
 #
 # 2018-08-07 Cornelius Kölbel <cornelius.koelbel@netknights.it>
@@ -1153,14 +1153,6 @@ def create(name, scope, action, filename=None):
 
             params = ast.literal_eval(contents)
 
-            if params.get("name") and params.get("name") != name:
-                print("Found name '{0!s}' in file, will use that instead of "
-                      "'{1!s}'.".format(params.get("name"), name))
-            else:
-                print("name not defined in file, will use the cli value "
-                      "{0!s}.".format(name))
-                params["scope"] = name
-
             if params.get("scope") and params.get("scope") != scope:
                 print("Found scope '{0!s}' in file, will use that instead of "
                       "'{1!s}'.".format(params.get("scope"), scope))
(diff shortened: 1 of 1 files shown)


pi-manage
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/opt/privacyidea/3.1/.venv/bin/python3.7
```
cornelinux: You can not do this!
Please revert this.

pi-manage
```diff
@@ -1153,14 +1153,6 @@ def create(name, scope, action, filename=None):
 
             params = ast.literal_eval(contents)
 
-            if params.get("name") and params.get("name") != name:
-                print("Found name '{0!s}' in file, will use that instead of "
-                      "'{1!s}'.".format(params.get("name"), name))
-            else:
-                print("name not defined in file, will use the cli value "
-                      "{0!s}.".format(name))
-                params["scope"] = name
```
cornelinux: This deletion is out of scope of the mentioed issue in the topic.

pi-manage
```diff
@@ -1177,8 +1169,7 @@ def create(name, scope, action, filename=None):
                       "{0!s}.".format(action))
                 params["action"] = action
 
-            r = set_policy(params.get("name"),
-                           scope=params.get("scope"),
+            r = set_policy(name, scope=params.get("scope"),
```
cornelinux: s.o.

pi-manage
```diff
@@ -1243,11 +1234,20 @@ def createtoken(role, days, realm, username):
 
 if __name__ == '__main__':
     # We add one blank line, to separate the messages from the initialization
-    print("""
+    print_header = True
+    try:
+        if sys.argv[1] in [ '-q', '--quiet' ]:
+            print_header = False
+            sys.argv = sys.argv[1:]
+    except IndexError:
```
cornelinux: It turns out there is a method in flask_script ``Manager.add_option`` which would do the job. However, it is tricky to use and might not easily work in our code.
However, if you add the ``add_option`` anyways, we at least get a help info:

You can do this at the beginning of the file after

    manager = Manager(app)

there you can call 

    manager.add_option("-q", "--quiet", help="Do not show the privacyIDEA banner.", action="store_true")

basvandervlies:
I will close this pull request as it is totally messed up ;-(.  I will make a better one

PR #1881, opened 2019-10-10
Enable autofocus directive in angular

### LINKED ISSUE #1880: enable autofocus attribute for all form elements
The autofocus attribute does not work correctly in forms loaded via angular.
In #205 this was fixed only for the login screen.
we could add the `focus="True"` attribute to the other input elements but since there are already a few (currently ineffective) `autofocus` attributes in the views, it would be better to overwrite the `autofocus` directive in angular.

plettich:
Make the autofocus attribute work when pages were loaded from angular.
Added and fixed some uses of the `autofocus` attribute.

Fixes #1880

### DIFF
--- a/privacyidea/static/components/login/views/login.html
+++ b/privacyidea/static/components/login/views/login.html
@@ -56,7 +56,6 @@
                        class="form-control round-top"
                        ng-model="login.username" required
                        placeholder="{{ 'Username'|translate }}"
-                       focus="true"
                        autofocus/>
                 <label for="realm" class="sr-only" translate>Realm</label>
                 <select class="form-control round-none"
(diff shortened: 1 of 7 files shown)


plettich:
based on https://gist.github.com/kirkstrobeck/599664399dbc23968741

privacyidea/static/components/login/views/login.html
```diff
@@ -56,7 +56,8 @@ <h2 ng-show="piLoginText" class="form-signin-heading">{{ piLoginText }}</h2>
                        class="form-control round-top"
                        ng-model="login.username" required
                        placeholder="{{ 'Username'|translate }}"
-                       focus="true"/>
+                       focus="true"
```
NuvandaPV: Does the `[focus]` here still do anything? If I read correctly, the focus-directive is removed, so this should probably be removed as well.

plettich: Fixed in 17c361fc

NuvandaPV:
Fixes #1880

Issue #1618, opened 2019-05-11
Confirmation question before certain actions


cornelinux:
When deleting or modifying certain data in the WebUI, the user or admin should get a question for confirmation before changing the data.

This is valid for:

* deleting objects like token, resolvers, realms, policies
* enabling or disabling or changing settings like resolvers or policies...
* minor changes, that do not have an impact on the whole system like changing data of a single token.

I personally do not like that confirmation questions, so these questions should be configurable in a webui policy. Maybe in these levels of the action-types as stated above, so you can have configuration questions for deleting objects but not for changing things.

This could be done with a javacsript function before issuing the REST request.

cornelinux:
# Basic concept

## policy

We can have a  policy in the ``scope=webui`` with an action ``confirm_action``. This can take a list of items, *in which case* the user should be asked for confirmation.

## ui

We already have the javascript function [getRightsValue](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/static/components/login/factories/auth.js#L78) which then could be called accordingly:

    getRightsValue("confirm_action")

# Different Levels

We probably should not require to set confirmation questions for every *single* action in the UI.
We should rather define certain classes of severity levels. These UI then needs to know, which action belongs to which severity level.

Those levels could be defined by:

* Does this action has an impact on the whole system, that can not easily be rebuilt?  (deleting policy)
* Does this action only has impact on a single token or single user (deleting token)
* The action can be easily undone (like enable/disable token, disable policy)
* Is the action performed on a hardware or software token
* Does an admin or user has the revert-action-right
* Does the action only set a certain attribute (which could be set anew any time)
* Does it *change* a certain configuration

These levels will be with the following actions and calls:

* Severe (would require a restore or reconfiguration)
* Difficult
* No harm done (easily revertable)

(*maybe it is easier and more clear after all to have defintions for each endpoint*: But in this case the policy-action should definitively allow a multiselect! -- currently action type is 'str', 'bool' and 'int'. Would be 'multi-str', which we currently split with whitespaces)

But even *No Harm* might be interesting, since a helpdesk user might be allowed to disable a token, but he may not be allowed to enable a token. This way he could render a token useless and the user would have to wait, until s.o. else enables the token again. So in this case the confirmation might be sensible.

cornelinux:
Also check #954

cornelinux:
If you do this in a foreign repo we loose the complete connection to the issue.

[...]

Issue #954, opened 2018-02-27
Ask before deleting hardware tokens


cornelinux:
Deleting a hardware token might be much more problematic than deleting a software token.
A hardware token was imported via a seed file. In certain cases you might not find the seed file again.

Recently we added the tokenkind (hardware, software, virtual).
We could use this, to determine if a token can be deleted or not.

# Abstraction

The tokenkind is a tokeninfo field.
We could add tokeninfo field requirements to token policies. 
Certain actions on tokens then would only be allowed on the corresponding value of the tokeninfo.

cornelinux:
This is a special case of #1618

maxbeth:
What is a better way to change the userdialog?
1. Change "delete token" from the userdialog to unbind and only a unused token can be deleted
2. remove the delete button if it a hardwaredevice
3. the delete only disable and unbind the device from a user

or a policy that only disabled hardware can be deleted.

PR #1889, opened 2019-10-15
1618/confirmation question before certain actions

NuvandaPV:
First implementation of confirmation dialogs.


NuvandaPV:
Fixes #1618

NuvandaPV:
Fixes #954

NuvandaPV:
Note: This implementation does not fetch the verbosity level through `getRightsValue()` as initially outlined and instead adds `$scope.confirm_action` and `$scope.confirm_action_levels`. This is done to remain consistent with how the other webui-policies are implemented and to avoid having to inject an `AuthFactory` everywhere.

privacyidea/static/components/components/controllers/componentControllers.js
```diff
@@ -42,6 +44,16 @@ myApp.controller("componentController", function (ComponentFactory, $scope,
         $location.path("/component/clienttype");
     }
 
+    $scope.uploadAsk = function(files) {
```
cornelinux: Is it necessary to add an extra function f for each call?
I think we might produce less code, when we add the "confirm" function directly into the html template like 

~~~~javascript
        <button
                class="btn btn-primary"
                ng-file-select
                ng-file-change='confirm("difficult", 
                                          {{"Upload Subscription File" | translate}},
                                          {{"Do you want to import the selected file?" | translate}},
                                          {{"Start Upload" | translate}},
                                          $scope.upload($files))' 
                translate>
            Upload Subscription File
        </button>
~~~~

This way

1. we could have much less functions / code
2. have labeling of the button *and* the confirmation texts in one place
3. Also have the confirmation level togeather with the template
4. have all the translation in one place

This might look a bit more complicated, but I think it is really awkward to have to many wrapper functions.
What do you think?

privacyidea/lib/policy.py
```diff
@@ -2083,6 +2102,18 @@ def get_static_policy_definitions(scope=None):
                 'type': 'bool',
                 'desc': _("If this is checked, the seed "
                           "will be displayed as text during enrollment.")
+            },
+            ACTION.CONFIRMACTION: {
+                'type': 'str',
+                'desc': _('When to ask the user to confirm their action with a popup dialog. If set to "difficult", '
+                          'only ask when the change would require some work to revert. Defaults to "severe".'),
```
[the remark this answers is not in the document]
NuvandaPV: Before this PR, the only confirmation dialog is for removing a user. After this PR, the default level (if not set by a policy), is SEVERE (set in `postpolicy.py`). This will still ask before deleting a user, but also before changing settings for a token type, or a resolver,  as well as before deleting a policy, resolver, realm, CA connector, periodic task, server, gateway, or token.

privacyidea/static/components/config/controllers/configControllers.js
```diff
@@ -34,19 +36,52 @@ myApp.controller("policyListController", function($scope, $stateParams,
         });
     };
 
+    $scope.delPolicyAsk = function(policyName) {
+        $scope.confirm(
+            $scope.confirm_action_levels["severe"],
+            "Delete Policy",
```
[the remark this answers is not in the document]
NuvandaPV: Does this mean the `[translate]`-Attributes should be removed for the dynamic fields in the template? Should I add the translation calls to `$scope.confirm()`, to avoid code duplication, or do you this that would reduce clarity?

privacyidea/lib/policy.py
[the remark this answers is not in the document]
cornelinux: OK, I am happy with that.

privacyidea/static/components/config/controllers/configControllers.js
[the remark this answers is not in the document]
NuvandaPV: Fixed.

privacyidea/static/components/config/views/config.periodictasks.list.html
```diff
@@ -20,14 +20,14 @@
...
             <td>
-                <span ng-show="ptask.active"
-                      class="glyphicon glyphicon-ok"
-                      ng-click="disablePeriodicTask(ptask.id)">
-                </span>
-                <span ng-hide="ptask.active"
-                      class="glyphicon glyphicon-minus"
-                      ng-click="enablePeriodicTask(ptask.id)">
-                </span>
+                <span
```
cornelinux: @NuvandaPV Honestly I think this change is not necessary. And I like the previous notation better. It worked well for quite some time.
I am not sure, if we need to define a code style guide or if we need to configure the IDE the same way, so that we beautify in the same direction!

privacyidea/static/components/dialogs/views/dialog.confirm_action.html
```diff
@@ -0,0 +1,14 @@
+<!-- Modal -->
+<div class="modal-header">
+    <button type="button" class="close" ng-click="$ctrl.cancel()" aria-label='{{ "Close" | translate }}'>
+        <span aria-hidden="true">&times;</span>
```
cornelinux: We are using Glyphicons!

privacyidea/static/components/login/controllers/loginControllers.js
```diff
@@ -25,19 +27,46 @@ String.prototype.mysplit = function(separator) {
...
-                                      $scope, $http, $location,
-                                      authUrl, AuthFactory, $rootScope,
-                                      $state, ConfigFactory, inform,
-                                      PolicyTemplateFactory, gettextCatalog,
-                                      hotkeys, RegisterFactory,
-                                      U2fFactory, instanceUrl,
-                                      PollingAuthFactory,
-                                      resourceNamePatterns) {
-
+angular.module("privacyideaApp").controller("mainController", [
```
cornelinux: Revert this file!

privacyidea/static/components/machine/views/machine.details.html
```diff
@@ -96,15 +96,14 @@ <h3 translate>Tokens and Applications for machine {{ machineid }}</h3>
...
+                            ng-click="detachMachineTokenAsk(token.serial,
                             token.application)" translate>Detach
                     </button>
                 </td>
                 <td ng-show="token.optionsEdit">
                     <button class="btn btn-primary"
-                            ng-click="saveOptions(token.serial,
-                            token.application, token.options);
-                            token.optionsEdit=false"
+                            ng-click="saveOptionsAsk(token.serial,
```
cornelinux: Check if this still works, since the optionsEdit ist not changed!

privacyidea/static/components/token/controllers/tokenDetailController.js
```diff
@@ -117,6 +117,24 @@ myApp.controller("tokenDetailController", function ($scope,
             $rootScope.previousState.params);
     };
 
+    $scope.unassignAskAdmin = function() {
+        $scope.confirm(
+            $scope.confirm_action_levels["easy"],
+            gettextCatalog.getString("Unassign Token"),
+            gettextCatalog.getString("Do you really want to remove this token from this users account?"),
```
cornelinux: I think this is a type -> "user's"

I would simply do: "Do you really want to unassign this token from the user"

privacyidea/static/components/user/views/user.details.html
```diff
@@ -47,10 +47,10 @@ <h2>
                 <span class="glyphicon glyphicon-edit"></span>
                 <span translate>Edit user</span>
             </button>
-            <button ng-click="editUser=false; updateUser()"
+            <button ng-click="updateUserAsk()"
```
cornelinux: We need to check, if this still works!

cornelinux:
Can you please in future create a feature branch in the privacyidea repo, we use to do this, which makes it easier for us to work on the PR togeather.

cornelinux:
See my comment in the original issue.

PR #1890, opened 2019-10-15
Add pattern for resource name check

### LINKED ISSUE #1674: Show disallowed characters in policy name

plettich:
Some characters and complete names are not allowed in some input forms
since they either conflict with internal naming schemes or describe a
different resource.
The different patterns are defined in `static/app.js` and used in the
appropriate controllers to add the pattern attribute to the corresponding
input element.
The input element will not be validated unless the pattern matches.

Fixes #1674


privacyidea/static/app.js
```diff
@@ -76,6 +76,10 @@ myApp.constant("smtpServerUrl", backendUrl + instance + "/smtpserver");
 myApp.constant("radiusServerUrl", backendUrl + instance + "/radiusserver");
 myApp.constant("privacyideaServerUrl", backendUrl + instance + "/privacyideaserver");
 myApp.constant("recoveryUrl", backendUrl + instance + "/recover");
+myApp.constant("resourceNamePattern", "^[a-zA-Z0-9_-]+$");
+/* we have to ignore "test" and "test_request" as a resource name explicitly */
+myApp.constant("resourceNamePatternWoTest", "^(?!test$)([A-Za-z0-9_-]+)$");
+myApp.constant("resourceNamePatternWoTestRequest", "^[a-zA-Z0-9_-]+$");
```
cornelinux: I think the "test_request" is missing here? It looks like the "resourceNamePattern".

privacyidea/static/components/config/controllers/configControllers.js
```diff
@@ -840,6 +850,8 @@ myApp.controller("machineResolverController", function ($scope,
         $location.path("/config/machineresolvers/list");
     }
 
+    /* we have to ignore "test" as a resource name explicitly */
+    $scope.inputPattern = '^(?!test$)([A-Za-z0-9_-]+)$';
```
cornelinux: Why cant we use the "resourceNamePatternWoTest" here?

privacyidea/static/components/config/controllers/configControllers.js
```diff
@@ -79,6 +80,7 @@ myApp.controller("policyDetailsController", function($scope, $stateParams,
     $scope.scopes = [];
     $scope.viewPolicyTemplates = false;
     $scope.action_filter = "";
+    $scope.inputPattern = resourceNamePattern;
```
cornelinux: What about the idea to define the different patterns on the ``rootscope``. This way we could avoid defining a new pattern in each controller and simply use the corresponding pattern in the HTML template?

plettich:
There are some changes coming, so don't review yet.

privacyidea/static/components/config/controllers/configControllers.js
plettich: sorry, missed that You already reviewed this PR, i force-pushed a new version which already fixes some of Your remarks.
Also the constants only get injected into the `mainController` and added to the current scope there. I read somewhere that you should not use the root scope but i am just an angluarjs beginner.

privacyidea/static/app.js
```diff
@@ -76,6 +76,16 @@ myApp.constant("smtpServerUrl", backendUrl + instance + "/smtpserver");
 myApp.constant("radiusServerUrl", backendUrl + instance + "/radiusserver");
 myApp.constant("privacyideaServerUrl", backendUrl + instance + "/privacyideaserver");
 myApp.constant("recoveryUrl", backendUrl + instance + "/recover");
+myApp.constant("resourceNamePatterns", {
+    simple: {pattern: "^[a-zA-Z0-9_.-]+$",
+        title: "The resource name must consist of letters, numbers and '_', '-', '.'"},
```
cornelinux: Please make this translateable using ``getTextCatalog``.

privacyidea/static/app.js
```diff
@@ -76,6 +76,16 @@ myApp.constant("smtpServerUrl", backendUrl + instance + "/smtpserver");
...
+/* we have to ignore "test" and "test_request" as a resource name explicitly */
+    withoutTest: {pattern: "^(?!test$)([A-Za-z0-9_.-]+)$",
+        title: "The resource name must consist of letters, numbers and '_', '-', '.' " +
```
cornelinux: Please make this translateable.

privacyidea/static/app.js
```diff
@@ -76,6 +76,16 @@ myApp.constant("smtpServerUrl", backendUrl + instance + "/smtpserver");
...
+            "and must not be the word 'test'"},
+    withoutTestRequest: {pattern: "^(?!test_request$)([a-zA-Z0-9_.-]+)$",
+        title: "The resource name must consist of letters, numbers and '_', '-', '.' " +
```
cornelinux: also this translateable.

privacyidea/static/components/config/views/config.caconnectors.local.html
```diff
@@ -19,6 +19,8 @@
                        class="form-control"
                        ng-model="connectorname" required
                        placeholder="MyConnector1"
+                       pattern="{{ inputNamePatterns.simple.pattern }}"
+                       title="{{ inputNamePatterns.simple.title }}"
```
cornelinux: This is really cool!

privacyidea/static/components/config/views/config.smtp.edit.html
```diff
@@ -16,6 +16,8 @@
             <input name="identifier" id="identifier" class="form-control"
                    ng-model="params.identifier" required
                    placeholder="mySmtpServer"
+                   pattern="{{ inputNamePatterns.simple.pattern }}"
+                   title="{{ inputNamePatterns.simple.title }}"
```
cornelinux: *argh*! :-/
We have an endpoint ``/send_test_email``.
Should we add a pattern here like with test? Or should we rename the endpoint?

privacyidea/static/app.js
plettich: fixed in 84b8a8a0

privacyidea/static/app.js
plettich: fixed in 84b8a8a0

privacyidea/static/app.js
plettich: fixed in 84b8a8a0

privacyidea/static/components/config/views/config.smtp.edit.html
plettich: Added a pattern in 84b8a8a0

privacyidea/static/components/token/views/token.details.html
```diff
@@ -364,7 +364,7 @@ <h3 name="Token Details"><translate>Token details for {{ tokenSerial }}
 
         <tr ng-hide="token.locked || !checkRight('setrandompin')">
             <td>
-                <span ng-show="randomPin" translate>Random generated PIN</span>
```
cornelinux: Please change this back.

plettich: fixed in eb0000f2

privacyidea/api/lib/prepolicy.py
```diff
@@ -168,8 +168,8 @@ def set_random_pin(request=None, action=None):
 
     if len(pin_pols) == 0:
         # We do this to avoid that an admin sets a random PIN manually!
-        raise TokenAdminError("You need to specify a policy '{1!s}' in scope "
-                              "{0!s}.".format(scope, ACTION.OTPPINSETRANDOM))
+        raise TokenAdminError("You need to specify a policy '{0!s}' in scope "
+                              "{1!s}.".format(ACTION.OTPPINSETRANDOM, scope))
```
cornelinux: Thank you :-)

PR #1894, opened 2019-10-17
Add userinfo to the output of the token-janitor

### LINKED ISSUE #1070: Add userinfo to the output of the token-janitor
We could add the user data (loginname, givenname, surname, email) to the output.

This way the token janitor could also produce a snapshot / statistics of the tokens.

https://github.com/privacyidea/privacyidea/blob/master/tools/privacyidea-token-janitor

chamkad:
Print user info for each token resulting from find operation

Closes #1070

### DIFF
--- a/tools/privacyidea-token-janitor
+++ b/tools/privacyidea-token-janitor
@@ -431,16 +431,44 @@ def find(last_auth, assigned, active, tokeninfo_key, tokeninfo_value,
                         token_obj.token.serial,
                         token_obj.token.tokentype,
                         token_obj.token.description,
-                        token_obj.get_tokeninfo()))
+                        token_obj.get_tokeninfo()
+                        ))
+                    user = token_obj.user
+                    if user:
+                        print("\t\t'{0!s}','{1!s}','{2!s}','{3!s}','{4!s}','{5!s}'".format(
+                            user.info.get("username", ""),
+                            user.info.get("givenname", ""),
+                            user.info.get("surname", ""),
+                            user.uid,
+                            user.resolver,
+                            user.realm
+                        ))
+                    else:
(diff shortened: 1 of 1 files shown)


tools/privacyidea-token-janitor
```diff
@@ -431,16 +431,44 @@ def find(last_auth, assigned, active, tokeninfo_key, tokeninfo_value,
                         token_obj.token.serial,
                         token_obj.token.tokentype,
                         token_obj.token.description,
-                        token_obj.get_tokeninfo()))
+                        token_obj.get_tokeninfo()
+                        ))
+                    user = token_obj.user
```
cornelinux: Thanks a lot for the PR.
With the ``user = token_obj.user`` you are *always*, when the token is assigned to a user, also fetching user from the user directory. This can take a long time! So we should have a parameter to explicitly require the username to be printed.

tools/privacyidea-token-janitor
```diff
@@ -431,16 +431,44 @@ def find(last_auth, assigned, active, tokeninfo_key, tokeninfo_value,
...
             else:
                 for token_obj in tlist:
-                    print(u"'{!s}','{!s}','{!s}','{!s}'".format(
-                        token_obj.token.serial,
-                        token_obj.token.tokentype,
-                        token_obj.token.description,
-                        token_obj.get_tokeninfo()
-                    ))
-
+                    user = token_obj.user
```
cornelinux: S.o.

chamkad:
I've added the `--userinfo` option to the find operation, so we can easily require the userinfo to be printed.

cornelinux:
I am sorry. I totally overlooked that the command

     privacyidea-token-janitor find --action listuser

actually does this already!

PR #1900, opened 2019-10-23
FirebaseProvider Enhancements for Push Tokens

droobah:
Changes to FCM message payload are required for push notifications to work properly on iOS in all app states (foreground, background, terminated).  The apns-* and aps sections are required/recommended per Apple's developer documentation.  Adding the category tag for Apple Push Services enables the 'Allow' button to work at the notification screen on iOS as well as from notifications that are pushed to the Apple Watch.

Added TTL for push notifications so they automatically remove themselves from Firebase storage if they are too old to be useful for both iOS and Android.

Added FCM tags for reporting/analytics on the Firebase Console per Google's recommendation.

### DIFF
--- a/privacyidea/lib/smsprovider/FirebaseProvider.py
+++ b/privacyidea/lib/smsprovider/FirebaseProvider.py
@@ -127,7 +127,31 @@ class FirebaseProvider(ISMSProvider):
         fcm_message = {
             "message": {
                         "data": data,
-                        "token": firebase_token
+                        "token": firebase_token,
+                        "android": {
+                                    "priority": "HIGH",
+                                    "ttl": "120s",
+                                    "fcm_options": {"analytics_label": "AndroidPushToken"}
+                                   },
+                        "apns": {
+                                 "headers": {
+                                             "apns-priority": "10",
+                                             "apns-push-type": "alert",
+                                             "apns-collapse-id": "privacyidea.pushtoken",
+                                             "apns-expiration": str(int(time.time()) + 120)
+                                            },
+                                 "payload": {
+                                             "aps": {
+                                                     "alert": {
+                                                               "title": data.get("title"),
+                                                               "body": data.get("question"),
+                                                              },
+                                                     "sound": "default",
+                                                     "category": "PUSH_AUTHENTICATION"
+                                                    },
+                                            },
+                                 "fcm_options": {"analytics_label": "iOSPushToken"}
+                                }
                        }
             }
 


cornelinux:
Thanks a lot for this PR. Can you please open an issue for this and refer the PR to the issue. We try to do no PR without an issue, so that details things can be discussed in the issue (and do not need to be discussed in some comments of the PR)
Thanks!

privacyidea/lib/smsprovider/FirebaseProvider.py
```diff
@@ -127,7 +127,31 @@ def submit_message(self, firebase_token, data):
...
+                                    "priority": "HIGH",
+                                    "ttl": "120s",
+                                    "fcm_options": {"analytics_label": "AndroidPushToken"}
+                                   },
+                        "apns": {
+                                 "headers": {
+                                             "apns-priority": "10",
+                                             "apns-push-type": "alert",
+                                             "apns-collapse-id": "privacyidea.pushtoken",
+                                             "apns-expiration": str(int(time.time()) + 120)
```
cornelinux: @NuvandaPV Would is make sense to make such values configurable?

NuvandaPV: In theory it would make sense to make this configurable, however I am not sure whether that would be worth the effort to add extra configuration options, just so the administrator can pick what ringtone to use on iOS and whether the notification should persist at the top of the screen, or hide after a couple seconds.

cornelinux: I am not thinking that much about different ring tones than future changes in the API, where new key/values might be necessary for push to run.
So maybe we initialize the complete fcm_message from a config file that ships as default with privacyIDEA and can be changed if necessary. (Maybe this can also be an issue for version 3.3)?

NuvandaPV: You mean changes in Firebase's API?

cornelinux: Yes. Changes. Enhancemets. Things we missed. Just like now.

privacyidea/lib/smsprovider/FirebaseProvider.py
```diff
@@ -127,7 +127,31 @@ def submit_message(self, firebase_token, data):
         fcm_message = {
             "message": {
                         "data": data,
-                        "token": firebase_token
+                        "token": firebase_token,
+                        "android": {
+                                    "priority": "HIGH",
```
nilsbehlen: The android part adds nothing, it could be omitted.

NuvandaPV: In my testing it will actually stop the notification from hanging around in the push network, if the phone is offline. That's a small improvement, but I think we should keep it nevertheless.

NuvandaPV:
Closes #1903

PR #1905, opened 2019-10-25
Make 'splitAtSign' more consistent

### LINKED ISSUE #1808: Make splitAtSign consistent
The behaviour of ``splitAtSign`` is not consistent between the endpoint /validate/check and /auth.
We should adapt the endpoint /auth this way, that it behaves in the same was as the endpoint /validate/check.

plettich:
The `splitAtSign` configuration did not affect the `/auth` endpoint thus
leading to inconsistencies.
This PR tries to mimic the user-resolving behaviour of the
`/validate/check` endpoint considering the `splitAtSign` configuration.

Fixes #1808

### DIFF
--- a/tests/test_api_auth.py
+++ b/tests/test_api_auth.py
@@ -1,0 +1,356 @@
+# -*- coding: utf-8 -*-
+""" Test for the '/auth' API-endpoint """
+
+from .base import MyApiTestCase
+import mock
+import six
+from privacyidea.lib.config import set_privacyidea_config
+
+
+class AuthApiTestCase(MyApiTestCase):
+    def test_01_auth_with_split(self):
+        set_privacyidea_config("splitAtSign", "1")
+        self.setUp_user_realms()
+        with self.app.test_request_context('/auth',
+                                           method='POST',
+                                           data={"username": "cornelius",
+                                                 "password": "test"}):
+            res = self.app.full_dispatch_request()
+            self.assertTrue(res.status_code == 200, res)
+            result = res.json.get("result")
+            self.assertTrue(result.get("status"), result)
+            self.assertIn('token', result.get("value"), result)
+            self.assertEqual('realm1', result['value']['realm'], result)
+
+        # test failed auth
+        with self.app.test_request_context('/auth',
+                                           method='POST',
+                                           data={"username": "cornelius",
+                                                 "password": "false"}):
+            res = self.app.full_dispatch_request()
+            self.assertEqual(401, res.status_code, res)
+            result = res.json.get("result")
+            self.assertFalse(result.get("status"), result)
+            self.assertEqual(4031, result['error']['code'], result)
+            self.assertEqual('Authentication failure. Wrong credentials',
+                             result['error']['message'], result)
+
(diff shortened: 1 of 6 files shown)


tests/test_api_validate.py
```diff
@@ -747,22 +747,6 @@ def test_05_check_serial_with_no_user(self):
             self.assertEqual(result.get("status"), True)
             self.assertEqual(result.get("value"), True)
 
-    def test_05_check_serial_with_no_user(self):
```
cornelinux: Why did you remove this test?

tests/test_api_auth.py
```diff
@@ -0,0 +1,356 @@
+# -*- coding: utf-8 -*-
+""" Test for the '/auth' API-endpoint """
+
+from .base import MyApiTestCase
+import mock
+import six
+from privacyidea.lib.config import set_privacyidea_config
+
+
+class AuthApiTestCase(MyApiTestCase):
```
cornelinux: Could you please add a test case for an internal admin?
We have this one "testadmin"/"testpw", but in theory you could also define an internal admin like "admin@int", even if the "int" realm does not exist. And I think the code will and should work, since we do not raise an error (which is fine) the pseudo realm "int" would be split and the admin is verify with the complete ``username`` "admin@int".

tests/test_lib_user.py
```diff
@@ -199,7 +202,27 @@ def test_09_get_user_from_param(self):
                  "realm": self.realm2}
         user = get_user_from_param(param)
         self.assertEqual("{0!s}".format(user), "<cornelius.resolver1@realm2>")
-        
+
+        # test with splitAtSign set to '0'
```
cornelinux: Why did you remove this one? Did it change or did you move it to the other tests?
I think then we can also delete it - not only comment it.

cornelinux:
I added an issue #1908 for the documentation. So my generic comment is obsolete.
Only see my three comments inline.

tests/test_api_validate.py
plettich: Ah, the github diff doesn't show but this function is a duplicate. The same function name with the same code is used just before this one :-)

tests/test_api_auth.py
plettich: will do

tests/test_api_validate.py
cornelinux: thanks.

tests/test_lib_user.py
plettich: This is a remnant from when i thought we could fix this all in one place...
But actually, the `splitAtSign` configuration isn't tested here (it is used in `get_user_from_param()`, not in `splig_user()`) so i would add some tests for the library functions as well.

tests/test_api_auth.py
plettich: Done in 3bcd19af

tests/test_lib_user.py
plettich: removed the commented code and added tests for `splitAtSign` in 3bcd19af

PR #1906, opened 2019-10-27
Streamline the token-janitor user output

### LINKED ISSUE #1070: Add userinfo to the output of the token-janitor

cornelinux:
Catch exceptions if user resolving fails.
Add realm to the user output

Closes #1070

### DIFF
--- a/tools/privacyidea-token-janitor
+++ b/tools/privacyidea-token-janitor
@@ -313,13 +313,20 @@ def export_token_data(token_list):
     for token_obj in token_list:
         token_data = [u"{0!s}".format(token_obj.token.serial),
                       u"{0!s}".format(token_obj.token.tokentype)]
-        user = token_obj.user
+        try:
+            user = token_obj.user
+        except Exception:
+            sys.stderr.write("Failed to determine user for token {0!s}.\n".format(
+                token_obj.token.serial
+            ))
+            token_data.append("**failed user resolving**")
         if user:
             token_data.append(u"{0!s}".format(user.info.get("username", "")))
             token_data.append(u"{0!s}".format(user.info.get("givenname", "")))
             token_data.append(u"{0!s}".format(user.info.get("surname", "")))
             token_data.append(u"{0!s}".format(user.uid))
             token_data.append(u"{0!s}".format(user.resolver))
+            token_data.append(u"{0!s}".format(user.realm))
         tokens.append(token_data)
     return tokens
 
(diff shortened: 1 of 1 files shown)


tools/privacyidea-token-janitor
```diff
@@ -344,7 +356,7 @@ def export_user_data(token_list):
                 user.realm
             )
         else:
-            uid = u"N/A"
+            uid = u"N/A, N/A, N/A, N/A, N/A, N/A, N/A"
```
plettich: since we do not process this information, do we need to add all the 'N/A's?

tools/privacyidea-token-janitor
```diff
@@ -313,13 +313,20 @@ def export_token_data(token_list):
...
         token_data = [u"{0!s}".format(token_obj.token.serial),
                       u"{0!s}".format(token_obj.token.tokentype)]
-        user = token_obj.user
+        try:
+            user = token_obj.user
+        except Exception:
+            sys.stderr.write("Failed to determine user for token {0!s}.\n".format(
+                token_obj.token.serial
+            ))
+            token_data.append("**failed user resolving**")
```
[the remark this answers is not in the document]
cornelinux: I like "**failed**" to be the first word (so that it can be seen e.g. in an Excel sheet.

I can change this to "failed to resolver user"?

tools/privacyidea-token-janitor
cornelinux: The thing is, that the last column can contain the number of tokens assigned to this user.
N/A will sum up the tokens with no user,
So we need that much commas(?) so that in an excel sheed the sum will be in one column.
We could however do
"N/A, , , , , , "
What do you think?

tools/privacyidea-token-janitor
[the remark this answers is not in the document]
cornelinux: change this in 63ffbebb

tools/privacyidea-token-janitor
cornelinux: change this in 63ffbebb

plettich: well, i don't like excel... :-)

PR #1910, opened 2019-10-28
Add notes to read_before_update regarding Python 3

cornelinux:


cornelinux:
> Looks good but the `PI_CUSTOMIZATION` thing is not nice. Could we check this as well during the update?

You are right about the nice-level. But currently I feel it is a bit hard to determine. Users could either have chosen the simple way and put the changes in the original file system. I am not sure, if this can be identified by a relative path - obviously not.

The relative path could also point to a rewrite rule (if set up more professional) which might point to a central customization folder maybe under /etc/privacyidea. Parsing and determining all this sounds a bit challenging. What do you think?

PR #1913, opened 2019-11-04
Improve error handling of privacyidea cron

chamkad:
Add `--retry` and `--period` to try execution the failed jobs, to write the current timestamp for them in the database to `periodictasklastrun` table.


cornelinux:
I thin retrying is a bad idea. 
This could pile up, if something does not work at all. 
You can end up with endlos jobs, that are retrying.

PR #1923, opened 2019-11-12
Use named parameters for redis setex

wuebbels:
In newer versions (tested on Debian 10) the `setex(name, value, time)` call no longer works.
According to the documentation (https://redis-py.readthedocs.io/en/latest/#redis.Redis.setex) setex is supposed to be called like `setex(name, time, value)` (please notice the change in the order of the parameters).
The use of the named parameters value and time makes sure that parameters are not being confused when calling setex.


cornelinux:
Thanks. I created an issue for that.
Unfortunately the tests fail, since we also need to adapt the redismock.

PR #1930, opened 2019-11-15
Fix disabled phone number input field

### LINKED ISSUE #1929: No Phone input field in SMS Token enrollment
Due to a mixed up element in PR #1791 the input field for the phone number is disabled for users.

plettich:
The phone number input field was disabled when enrolling an SMS token as a
user.
Fixes #1929


cornelinux:
"Working on #1929" - wouldn't it fix it?

PR #1932, opened 2019-11-16
1931/pyrad timeout retries

maxbeth:
Hello,
i hope the handling with the values are correct.

### DIFF
--- a/privacyidea/api/lib/postpolicy.py
+++ b/privacyidea/api/lib/postpolicy.py
@@ -519,6 +519,8 @@ def get_webui_settings(request, response):
                                             realm=realm).action_values(unique=True)
         show_seed = Match.realm(g, scope=SCOPE.WEBUI, action=ACTION.SHOW_SEED,
                                 realm=realm).any()
+        confirm_action = Match.realm(g, scope=SCOPE.WEBUI, action=ACTION.CONFIRM_ACTION,
+                                    realm=realm).action_values(unique=False, allow_white_space_in_action=True)
         token_page_size = DEFAULT_PAGE_SIZE
         user_page_size = DEFAULT_PAGE_SIZE
         default_tokentype = DEFAULT_TOKENTYPE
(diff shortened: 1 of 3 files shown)


privacyidea/api/lib/postpolicy.py
```diff
@@ -519,6 +519,8 @@ def get_webui_settings(request, response):
                                             realm=realm).action_values(unique=True)
         show_seed = Match.realm(g, scope=SCOPE.WEBUI, action=ACTION.SHOW_SEED,
                                 realm=realm).any()
+        confirm_action = Match.realm(g, scope=SCOPE.WEBUI, action=ACTION.CONFIRM_ACTION,
+                                    realm=realm).action_values(unique=False, allow_white_space_in_action=True)
```
cornelinux: these changes in this file should not be here.

privacyidea/lib/policy.py
```diff
@@ -2083,6 +2084,11 @@ def get_static_policy_definitions(scope=None):
                 'type': 'bool',
                 'desc': _("If this is checked, the seed "
                           "will be displayed as text during enrollment.")
+            },
+            ACTION.CONFIRM_ACTION: {
+                'type': 'str',
+                'desc': _("Ask for confirmation before execute certain actions."),
+                'value': ["0 - Default", "1 - Critical", "2 - Sane", "3 - Fearful"]
```
cornelinux: these changes in this file should not be here

privacyidea/lib/tokens/radiustoken.py
```diff
@@ -501,7 +498,9 @@ def _check_radius(self, otpval, options=None, radius_state=None):
 
             req = srv.CreateAuthPacket(code=pyrad.packet.AccessRequest,
                                        User_Name=radius_user.encode('utf-8'),
-                                       NAS_Identifier=nas_identifier.encode('ascii'))
+                                       NAS_Identifier=nas_identifier.encode('ascii'),
+                                       timeout=radius_server_object.config.timeout,
+                                       retries=radius_server_object.config.retries)
```
cornelinux: I tried to look at the source code from the documentation but I did not find the parameters "timeout" and "retries".
Where do you get them from?
Or other radius code add the timeout after creating the client:
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/radiusserver.py#L105


Also, I read, that, if a timeout is hit, the timeout exception is raised. So we probabyl would have to catch that exception.

cornelinux:
@maxbeth thanks a lot for the suggestion. I created a new PR, that worked better.
So I am closing this one. Be prepared to have this in version 3.2!

PR #1937, opened 2019-11-19
Ask before deleting hardware token

### LINKED ISSUE #954: Ask before deleting hardware tokens

debianis:
fixes this issue - if user wants to delete token which is hardwaretype, he will be asked.

Closes #954

### DIFF
--- a/privacyidea/static/components/token/controllers/tokenDetailController.js
+++ b/privacyidea/static/components/token/controllers/tokenDetailController.js
@@ -192,39 +192,18 @@ myApp.controller("tokenDetailController", function ($scope,
 
     $scope.deleteTokenAsk = function() {
             $('#dialogTokenDelete').modal();
-        var tokenType = $scope.token.info.tokenkind;
-        if (tokenType == "hardware"){
-        $('#dialogTokenDelete').modal();
-             } else {
-             TokenFactory.delete($scope.tokenSerial, $scope.return_to,
-                function (data) {
-                    if (data.result.value==true) {
-                        inform.add(gettextCatalog.getString("Hardware Token deleted " +
-                            "successfully."),
-                                    {type: "info"});
-                        $scope._getUsers();
-                        $location.path("/user/list");
-                     } else {
-                inform.add(gettextCatalog.getString("Failed to delete Token."), {type: "danger"});
-            };
-              });
-    };
-    };
+            var tokenType = $scope.token.info.tokenkind;
+                if (tokenType == "hardware"){
+                    $('#dialogTokenDelete').modal();
+                    } else {
+                         $scope.delete();
+                    };
+     };
 
-    $scope.deleteToken = function(){
-        TokenFactory.delete($scope.tokenSerial, $scope.return_to,
-            function (data) {
-                if (data.result.value==true) {
-                            inform.add(gettextCatalog.getString("Hardware Token deleted " +
-                                "successfully."),
-                                        {type: "info"});
-                            $scope._getUsers();
-                            $location.path("/user/list");
-                 } else {
-                          inform.add(gettextCatalog.getString("Failed to delete Token."), {type: "danger"});
-                        };
-                             });
-                                 };
+
+    $scope.delete = function () {
+        TokenFactory.delete($scope.tokenSerial, $scope.return_to);
+    };
 
 
     $scope.setRandomPin = function () {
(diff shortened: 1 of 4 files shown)


privacyidea/static/components/token/controllers/tokenDetailController.js
```diff
@@ -190,9 +190,42 @@ myApp.controller("tokenDetailController", function ($scope,
...
+                        $scope._getUsers();
+                        $location.path("/user/list");
+                     } else {
+                inform.add(gettextCatalog.getString("Failed to delete Token."), {type: "danger"});
+            };
+              });
     };
+    };
+
+    $scope.deleteToken = function(){
```
NuvandaPV: This function is pointless. `$scope.delete()` already does the exact same thing. Please use $scope.delete().

privacyidea/static/components/token/controllers/tokenDetailController.js
```diff
@@ -190,9 +190,42 @@ myApp.controller("tokenDetailController", function ($scope,
...
+        TokenFactory.delete($scope.tokenSerial, $scope.return_to,
+            function (data) {
```
NuvandaPV: Explanation: `TokenFactory.delete` only takes two parameters. The third parameter you have added here is ignored completely. This makes your new function identical to `$scope.delete()`.

privacyidea/static/components/token/controllers/tokenDetailController.js
```diff
@@ -190,9 +190,42 @@ myApp.controller("tokenDetailController", function ($scope,
         }, $scope.get);
     };
 
-    $scope.delete = function () {
-        TokenFactory.delete($scope.tokenSerial, $scope.return_to);
+    $scope.deleteTokenAsk = function() {
+            $('#dialogTokenDelete').modal();
```
NuvandaPV: I think this line needs to be removed. My suspicion is, that is would cause the modal to always show, even if the token is not a hardware token.

privacyidea/static/components/token/controllers/tokenDetailController.js
```diff
@@ -190,9 +190,42 @@ myApp.controller("tokenDetailController", function ($scope,
...
+        var tokenType = $scope.token.info.tokenkind;
+        if (tokenType == "hardware"){
+        $('#dialogTokenDelete').modal();
```
NuvandaPV: This looks good, but please fix the indentation. It is really hard to read otherwise. the code should look something like this:
```
if (condition) {
    doSomething();
} else {
    doSomethingElse();
}
```

privacyidea/static/components/token/controllers/tokenDetailController.js
```diff
@@ -190,9 +190,42 @@ myApp.controller("tokenDetailController", function ($scope,
...
+             } else {
+             TokenFactory.delete($scope.tokenSerial, $scope.return_to,
```
NuvandaPV: The third parameter here does not do anything (see explanation below). What are you trying to do here? We already have a function `$scope.delete()` that calls `TokenFactory.delete()` with the correct parameters. Unless there is a specific reason why you need the extra code, please just call `$scope.delete()`.

privacyidea/static/components/token/views/dialog.ask_token_delete.html
```diff
@@ -0,0 +1,38 @@
...
+                    </p>
+                </div>
+
+                <div class="modal-footer">
+                    <button type="button" class="btn btn-default"
+                            data-dismiss="modal" translate>No</button>
+                    <button type="button"
+                            class="btn btn-danger"
+                            data-dismiss="modal"
+                            ng-click="deleteToken()">
```
NuvandaPV: As stated above, there is not need for a new function `deleteToken()`, as that is what `delete()` already does. Please refactor this to use `delete()`.

privacyidea/static/components/token/controllers/tokenDetailController.js
debianis: i tested it in venv and it does work as intended.

Issue #1566, opened 2019-04-12
Restrict admin policies in OTP-Length, Hash, Timestep


cornelinux:
Policies in scope "user" can be restricted in OTP Length, hash and timestep.
Admin-policies should also have this restriction, so that helpdesk users can get a simple view.

plettich:
TODO:
- [ ] improve tests
- [ ] add documentation

cornelinux:
@plettich  Hm, isn't his resolved by the PR?  - or are we missing some docs?

plettich:
Yes, docs are missing, working on it.

Issue #1839, opened 2019-09-04
Redesign handling of hashlib/keysize parameters in /token/init


fredreichbier:
Currently, it is not very clear how the ``hashlib`` and ``keysize`` parameters are handled and which one takes precedence. We should clarify or redesign the handling of these parameters.

It should probably be as follows:
* If only ``keysize`` is given, we derive the hashlib from the keysize OR what?
* If only ``hashlib`` is given, we derive the keysize from the hashlib
* If both are given, we use both parameters.

fredreichbier:
* [ ] Check and document current behavior of hashlib/keysize parameters
* [ ] Discuss the desired behavior
* [ ] Implement it

PR #1938, opened 2019-11-22
Add admin policy restrictions for token parameter

plettich:
The restriction of TOTP/HOTP token parameters like hashlib, timestep or OTP
length could be enforced for users via policies.
This PR adds the possibility to restrict these parameters for the admin
scope as well.

The system-wide default configuration for the token types is still usable
but will be overwritten by a corresponding policy.
In the Web-UI these parameters are disabled/hidden if a corresponding
policy is set.

If a default system configuration is set, it won't be added to the form
data anymore and thus won't add confusing token info data.

Also simplified the `get_class_info()` functions since the actions are mostly the same for HTOP/TOTP.

Working on #1566

### DIFF
--- a/privacyidea/lib/tokens/hotptoken.py
+++ b/privacyidea/lib/tokens/hotptoken.py
@@ -60,9 +60,7 @@ from privacyidea.lib.error import ParameterError
 from privacyidea.lib.apps import create_oathtoken_url as cr_oath
 from privacyidea.lib.utils import (create_img, is_true, b32encode_and_unicode,
                                    hexlify_and_unicode)
-from privacyidea.lib.policydecorators import challenge_response_allowed
 from privacyidea.lib.decorators import check_token_locked
-from privacyidea.lib.auth import ROLE
 from privacyidea.lib.policy import SCOPE, ACTION
 from privacyidea.lib import _
 import traceback
@@ -120,8 +118,8 @@ class HotpTokenClass(TokenClass):
         desc_self1 = _('Specify the hashlib to be used. '
                        'Can be sha1 (1) or sha2-256 (2).')
         desc_self2 = _('Specify the otplen to be used. Can be 6 or 8 digits.')
-        desc_two_step_user =_('Specify whether users are allowed or forced to use '
-                              'two-step enrollment.')
+        desc_two_step_user = _('Specify whether users are allowed or forced to use '
+                               'two-step enrollment.')
         desc_two_step_admin = _('Specify whether admins are allowed or forced to use '
                                 'two-step enrollment.')
         res = {'type': 'hotp',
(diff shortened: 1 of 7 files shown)


privacyidea/static/components/token/controllers/tokenControllers.js
```diff
@@ -524,29 +525,28 @@ myApp.controller("tokenEnrollController", function ($scope, TokenFactory,
...
+        $scope.form.hashlib = $scope.systemDefault["hotp.hashlib"] || 'sha1';
         // Now add the questions
-        angular.forEach(systemDefault, function(value, key) {
+        angular.forEach($scope.systemDefault, function(value, key) {
             if (key.indexOf("question.question.") === 0) {
                 $scope.questions.push(value);
             }
         });
-        $scope.num_answers = systemDefault["question.num_answers"];
+        $scope.num_answers = $scopen.systemDefault["question.num_answers"];
```
cornelinux: typo ``$scopen`` -> ``$scope``

privacyidea/static/components/token/controllers/tokenControllers.js
```diff
@@ -524,29 +525,28 @@ myApp.controller("tokenEnrollController", function ($scope, TokenFactory,
             radius.server, radius.secret...
            are stored in systemDefault and $scope.form
          */
-        var systemDefault = data.result.value;
+        $scope.systemDefault = data.result.value;
```
cornelinux: It the user is not admin, he will not be able to load systemconfig. (see above).
What happens then? The ``$scope.systemDefault`` might not be initialized and it could break in line 276 or 279?
The ``$scope.form`` did exist for both ROLE.ADMIN and ROLE.USER.
Maybe we have to initialize ``$scope.systemDefault``. Is the "systemDefaul" actualy that, what we want?

privacyidea/lib/tokens/hotptoken.py
```diff
@@ -242,9 +250,9 @@ def get_init_detail(self, params=None, user=None):
                                         tokentype=tok_type.lower(),
                                         serial=self.get_serial(),
                                         tokenlabel=tokenlabel,
-                                        hash_algo=params.get("hashlib", "sha1"),
-                                        digits=params.get("otplen", 6),
-                                        period=params.get("timeStep", 30),
+                                        hash_algo=self.hashlib,
+                                        digits=self.get_otplen(),
+                                        period=self.init_details.get("timeStep", 30),
```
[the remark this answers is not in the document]
cornelinux: After all these these three lines are a code refactoring that are not necessary for the issue.
Please revert them.

privacyidea/static/components/token/controllers/tokenControllers.js
plettich: Well actually the user is able to load the (reduced) system config which contains the HOTP and TOTP parameter.
The `$scope.systemDefault` is initialized when the `tokenEnrollController` is instantiated.
The problem with `$scope.form` is, that everything that is put there will be send to privacyIDEA in the token enrollment call. And if we have i.e. a policy which enforces the HOTP hashlib to be `sha512`, but a system config where `hotp.hashlib = sha1` we would have both hashlibs in the token info (`hashlib: sha512, hotp.hashlib: sha1`) while the actual hashlib is `sha1`.
We can move this into #1839

privacyidea/static/components/token/controllers/tokenControllers.js
plettich: fixed in bf04b672

privacyidea/lib/tokens/hotptoken.py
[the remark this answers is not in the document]
plettich: Done in bf04b672

PR #1946, opened 2019-11-28
1908/add description of split at sign

plettich:


doc/configuration/realms.rst
```diff
@@ -18,6 +18,49 @@ Therefor the users need to authenticate with their username and the realm like t
...
+There are several options to relate a user to a specific realm during
+authentication. Usually, if only a login name is given, the user will be
+searched in the default realm.
+
+If a *realm* parameter is given in a :ref:`/auth<rest_auth>` or
+:ref:`/validate/check<rest_validate>` request, it supersedes a possible
+:ref:`split<splitatsign>` realm.
+
+The following table shows different combinations of a users login and a *realm*
+parameter and where the user will be looked for depending on the
```
cornelinux: Should it be "where a user will be looked up" or "where the user will be searched"?

doc/configuration/realms.rst
```diff
@@ -18,6 +18,49 @@ Therefor the users need to authenticate with their username and the realm like t
...
+:ref:`splitatsign` setting:
+
+=============  =======  =======================  =====
+  Input                 :ref:`splitatsign`
```
cornelinux: I was thinking a bit, what I actually see in this table.
Could you either add some explanation or enhance the headers:

Could you change "Input" to "Input parameter", so that it is clear, that these are *request parameters*.

Hm, the ``login`` is actually the ``user`` parameter (or username). I am not sure, if we should change this or if "login" is fine.

Could you change the header ``:ref:`splitatsign` `` to splitatsign-setting (so that it is clearer, that this is the actual configuration)

PR #1953, opened 2019-12-02
Check if travis needs to run

plettich:
Add a script (`./.travis_check.sh`) which checks if there are changes in
certain directories. If i.e. only the docs changed, the tests won't run.

### DIFF
--- a/.travis_check.sh
+++ b/.travis_check.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+set -ev
+
+# setup patterns for files to ignore
+declare -a ignore_patterns
+ignore_patterns=(
+  "^doc/"
+  "^privacyidea/static/"
+  "^privacyidea/translations/"
+  "^migrations/"
+  "^deploy/"
+  "^contrib/"
+  )
+
+# get the base branch for checking changes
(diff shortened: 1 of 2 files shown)


.travis_check.sh
```diff
@@ -0,0 +1,59 @@
...
+  for pattern in "${ignore_patterns[@]}"; do
+    if [[ ${CHANGED_FILE} =~ ${pattern} ]]; then
+      match_found=True
+      echo "Match found! Changed file: $CHANGED_FILE, pattern: $pattern"
+      break
+    fi
+  done
+  if [[ ${match_found} == False ]]; then
+    IGNORE_BUILD=False
+    echo "Do not ignore build! File: $CHANGED_FILE, IGNORE_BUILD: $IGNORE_BUILD, match_found:
```
cornelinux: Negations are difficult to read. We should change it to something like this:

"Do not ignore build!" -> "Found a relevant changed file. Testing"

OK, I understand that the complete logic and also naming of variables is about finding out, if the tests shoud **not** run. I would have chosen the approach the other way round.

plettich: fixed in ea67081

Issue #1958, opened 2019-12-10
Add readthedocs configuration


NuvandaPV:
readthedocs.io supports configuration through a yaml-file in the project root these days. We should move our configuration to such a file.

PR #1962, opened 2019-12-11
Add .readthedocs.yaml

NuvandaPV:
This adds a config file for readthedocs, that *should* be equivalent to our current settings.


NuvandaPV:
fixes #1958

PR #1964, opened 2019-12-12
Filter tokenrealm in UI

### LINKED ISSUE #545: Filter token realms in token view
The following additional filter attributes in token list would be great:

- user
- realm

Additionally, the REST API documentation for GET /TOKEN should be altered:

- `tokenrealm` parameter is actually used, not `viewrealm` like mentioned in code documentation

cornelinux:
The token table in the UI contains a filter field for the tokenrealm.
It filters for a substring of all realms, that the token
is assigned to.

Closes #545

### DIFF
--- a/privacyidea/lib/token.py
+++ b/privacyidea/lib/token.py
@@ -186,7 +186,7 @@ def _create_token_query(tokentype=None, realm=None, assigned=None, user=None,
     if stripped_realm is not None:
         # filter for the realm
         if "*" in realm:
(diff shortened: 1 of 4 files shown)


cornelinux:
> Looks good to me. Currently a user typing too fast will cause exceptions to be thrown though. Should we do something about that?

How is that?
This would probably happen in all filter components? If so, I would leave this as a generic new issue.

PR #1965, opened 2019-12-12
No link on Logo, if no user is logged in

### LINKED ISSUE #1944: Link in Logo does not work, when no user is logged in.
The top left logo leads to the /token/list tab in the UI.
If a user is not yet logged in, the call to this tab will create an error message, that the user is not logged in.

We could have different links behind this logo, depending on whether the user is logged in.

* logged in state: https://privacyidea/#/token/list
* not logged in: (no link)

cornelinux:
The logo in the WebUI should only have a link,
if a user is logged in.

Closes #1944

### DIFF
--- a/privacyidea/static/templates/menu.html
+++ b/privacyidea/static/templates/menu.html
@@ -11,7 +11,11 @@
 </div>
 <div class="navbar navbar-default navbar-fixed-top">
     <div class="container">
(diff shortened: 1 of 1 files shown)


cornelinux:
There may be more elegant solutions, but I could not figure something out *without* adding any new directives, which seemed like overkill to me.

Issue #1375, opened 2019-01-14
 Limit number of token types per user


blsho:
Admin can limit count of user tokens only by setting `max_token_per_user` (and `max_token_per_realm`). However, we would like to limit user so they can have only up to one token of each token type.

Possible solution from my point of view:
- Add a single option `max_same_tokentype_per_user` to policy. Value of the option will be integer and it will apply to all token types individually.
  - Benefits: 
    - Simple configuration and Web UI
    - generic approach and easier token type management (this functionality will not increase complexity of implementing new token types)
  - Downsides:
    - Reduced customization per token type
- Add multiple options (e.g. `max_hotp_per_user`, `max_totp_per_user`,  `max_sms_per_user`, `max_u2f_per_user`). Each token type will have its own option.
  - Benefits:
    - Highly customizable
  - Downsides:
    - Complex configuration and Web UI
    - Options and their count are variable a defined by supported token types. Count of options will have to reflect changes in supported token types.

Is such a feature reasonable? Which variant? Thanks.

cornelinux:
Thanks for your interest and suggestion. At first this sounds very specific to me. And specific requirements are always a bit difficult since they can clutter the feature list. I always like to keep things very flexible. So the variant A does not sound nice to me at all.

Variant B would - as you already said - add to many options, which would be probably not used in 99,9999% of the scenarios ;-)

However, what do you want to achieve - from a top level view. Why would every user be allowed to have
* one TOTP token
* one HOTP token
* one U2F token
* one Email Token
* one 4eyes token
* ...
This sounds like a very strange requirements to me.
So what is actually your scenarion.

Maybe we can achieve something with the event handler rather then with a rigid, huge policy list.

blsho:
We have multiple reasons:
- We want to promote U2F as a main token (with possibility of TOTP as a backup token) instead of 2 TOTP tokens (one main and second as a backup)
- Our end-users are used to refer to tokens by its type, so it is easier for operation team to have strong relation 1:0..1 in token_type:token for a given user
- From our experience, when end-users lost their token (e.g. change their phone number or change their phone device without backuping its TOTP) they just enroll the same token type again and don't remove the old one.
- Reduce fees for SMS validation (if user has more SMS tokens)

We were thinking about implementing custom event handler script, but it has one big downside - it can not provide instant feedback to end-user over Web UI.

cornelinux:
Thanks for the input.
In the meantime some hints:

[...]

PR #1966, opened 2019-12-12
Add tokentype specifc limits

cornelinux:
The admin can define policies to limit the
number of tokens a user may enroll based on the
specific tokentype.

Working on #1375

### DIFF
--- a/privacyidea/lib/tokens/papertoken.py
+++ b/privacyidea/lib/tokens/papertoken.py
@@ -29,9 +29,8 @@ import logging
 from privacyidea.lib.log import log_with
 from privacyidea.lib.tokenclass import TokenClass
 from privacyidea.lib.tokens.hotptoken import HotpTokenClass
-from privacyidea.lib.policy import SCOPE
+from privacyidea.lib.policy import SCOPE, ACTION, GROUP
 from privacyidea.lib import _
-from privacyidea.lib.policydecorators import libpolicy
 
 log = logging.getLogger(__name__)
 DEFAULT_COUNT = 100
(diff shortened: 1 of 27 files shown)


cornelinux:
Work in progress. Still need to add the tests...

cornelinux:
Most of the changed files are the policy definitions of the token classes.
The real logic is contained in ``api/lib/preprolicy.py``.

privacyidea/lib/tokens/papertoken.py
```diff
@@ -107,7 +106,18 @@ def get_class_info(key=None, ret='all'):
...
+                       },
+                       ACTION.MAXTOKENUSER: {
+                           'type': 'int',
+                           'desc': _("The user may only have this maximum number of paper tokens assigned."),
+                           'group': GROUP.TOKEN
+                       },
+                       ACTION.MAXACTIVETOKENUSER: {
+                           'type': 'int',
+                           'desc': _("The user may only have this maximum number of active paper tokens assigned."),
+                           'group': GROUP.TOKEN
```
plettich: remember my (rejected) changes from #1938? We could define template policies which will be automatically generated in each token class. We could save some code duplication. Should i implement this?

cornelinux: No. I do not remember. The 1938 points to something merged.
Did we have an issue for this idea? Otherwise you can create one and I would like to have a few lines of meta code in the issue so we have a common understanding what this implementation would do.

I also do not like the many lines of code - although they are no lines of **code**.
We would still have specific ``desc`` of the policies, so I am not so sure, how some templates could help, keeping the flexibility and producing *less* lines of code.

cornelinux:
> looks good. Should i look into changing the token policies?

Please put policy templates in another issue and have the discussion *before* the PR.

PR #1968, opened 2019-12-16
Cleanup created files after CA test

### LINKED ISSUE #1960: Update CA testdata
Some certificates in the directory tests/testsdata/ca are about to expire in January 2020.
The CA itself is valid till 2022. 
We need to check and update/renew certificates.

cornelinux:
Closes #1960


tests/test_lib_caconnector.py
```diff
@@ -280,6 +281,25 @@ def test_05_templates(self):
         self.assertTrue(ddiff.days > 740, ddiff.days)
         self.assertTrue(ddiff.days < 760, ddiff.days)
 
+    def test_99_cleanup(self):
```
plettich: Could we also reset the modified files `index.txt` and `serial`?

tests/test_lib_caconnector.py
```diff
@@ -280,6 +281,25 @@ def test_05_templates(self):
...
+        filelist = glob.glob("{0!s}/100*.pem".format(WORKINGDIR))
+        for f in filelist:
+            os.remove(f)
+
+        FILES = ["cornelius_user@localhost.localdomain_realm1.pem",
+                 "cornelius_user@localhost.localdomain_realm1.req",
```
plettich: actually these two are not tracked and don't exist after a current run of the test.
So the `os.remove` call fails and the test breaks on a fresh checkout.

tests/test_lib_caconnector.py
```diff
@@ -280,6 +281,25 @@ def test_05_templates(self):
...
+                 "DE_Hessen_privacyidea_usercert.pem",
+                 "DE_Hessen_privacyidea_usercert.req",
+                 "index.txt.attr.old",
+                 "index.txt.old",
+                 "serial.old",
+                 "crl.pem",
+                 "Steve_Test.der",
+                 "Steve_Test.txt"]
+        for f in FILES:
+            os.remove("{0!s}/{1!s}".format(WORKINGDIR, f))
```
plettich: maybe we should use `try-except` here as well (catching `IOError` should be enough).

tests/test_lib_caconnector.py
```diff
@@ -299,3 +319,12 @@ def test_01_create_ca(self):
...
             # check if the generated files exist
             self.assertTrue(os.path.exists(os.path.join(workdir, 'cacert.pem')))
+
+    def test_02_cleanup(self):
+        filelist = glob.glob("{0!s}2/*".format(WORKINGDIR))
+        for f in filelist:
+            try:
+                os.remove(f)
+            except:
+                print("Error deleting file {0!s}.".format(f))
```
plettich: catching `IOError` here should be enough. Also we could issue a [warning](https://docs.pytest.org/en/latest/warnings.html) which pytest shows in the summary.

tests/test_lib_caconnector.py
```diff
@@ -280,6 +281,25 @@ def test_05_templates(self):
...
         self.assertTrue(ddiff.days < 760, ddiff.days)
 
+    def test_99_cleanup(self):
+        filelist = glob.glob("{0!s}/100*.pem".format(WORKINGDIR))
+        for f in filelist:
+            os.remove(f)
+
+        FILES = ["cornelius_user@localhost.localdomain_realm1.pem",
+                 "cornelius_user@localhost.localdomain_realm1.req",
+                 "DE_Hessen_privacyidea_requester.localdomain.req",
```
plettich: This one is currently tracked and should be removed from git.
Also it's corresponding PEM-File (`DE_Hessen_privacyidea_requester.localdomain.pem`) should be removed here.

tests/test_lib_caconnector.py
cornelinux: will do

tests/test_lib_caconnector.py
cornelinux: will check

tests/test_lib_caconnector.py
cornelinux: Hm, if the test run successfully, the files should exist. Actually I *had* this in the first place but then realized that this would not be necessary and removed it before the commit, since it *looked nicer* ;-)

tests/test_lib_caconnector.py
cornelinux: will do

tests/test_lib_caconnector.py
cornelinux: will check

tests/test_lib_caconnector.py
cornelinux: will do.

tests/test_lib_caconnector.py
cornelinux: I created a backup copy of the files and move them back in the cleanup step.
Move the cleanup step into teardownClass.

tests/test_lib_caconnector.py
cornelinux: done in 151a9d2d

tests/test_lib_caconnector.py
cornelinux: done in 151a9d2d

tests/test_lib_caconnector.py
cornelinux: done in 151a9d2d

tests/test_lib_caconnector.py
cornelinux: done in 151a9d2d

Issue #1871, opened 2019-09-26
Versioning for Angular-Components


NuvandaPV:
It would be nice to have some sort of versioning for the Angular components, to allow for reliable cache invalidation on updates.

NuvandaPV:
Related #1860 #1261

NuvandaPV:
Paging @cornelinux @fredreichbier

NuvandaPV:
Versioning the components will require employing a slightly different approach from the use we use to version our css and js, since the urls to call are generated on the client. If we want to use the same versioning scheme we use for the server-side versioning, we have to send the last modification timestamps for all of Angular's views to the client (possibly as hidden fields in `footer.html`). Alternatively, we could use `privacyideaVersionNumber`, falling back to generating a random string, if the version number is not set, to make sure caches are invalidated every time when a user runs pre-release software.

fredreichbier:
I think some kind of versioning for the Angular templates would be nice, especially for development: I've often had the situation that caches were not properly invalidated.

My 2 cents: I guess using ``privacyideaVersionNumber`` would be pretty easy and this would probably already solve the problem for endusers. But we would still have the same problem during development -- not sure how we should approach that. Maybe we could detect client-side if we're using a ``devX`` version and append a random query string, effectively disabling the cache?

cornelinux:
@NuvandaPV Is this closed by the latest PR?

NuvandaPV:
Yup.

PR #1969, opened 2019-12-16
Versioning for AngularJS components

NuvandaPV:


NuvandaPV:
fixes #1871

privacyidea/static/providers/versioningProvider.js
```diff
@@ -0,0 +1,58 @@
...
+
+/*
+ * The version number of the privacyIDEA server software.
+ */
+versioning.constant('PRIVACYIDEA_VERSION_NUMBER', document.getElementById('PRIVACYIDEA_VERSION_NUMBER').value);
+
+/*
+ * A suffix, that can be appended to an url to invalidate the caches when necessary.
+ *
+ * This factory will provide a provider to provide a suffix for an url that sets a parameter `v` to either the
```
cornelinux: Admit it - you constructed this sentence on purpose! :-)

privacyidea/static/providers/versioningProvider.js
```diff
@@ -0,0 +1,58 @@
...
+ * privacyideaVersionNumber, if known, or a random string otherwise. This has the effect of preserving caches, but not
+ * across updates of the privacyIDEA server software and not if the user is running a development version of
+ * privacyIDEA.
+ */
+versioning.provider('versioningSuffixProvider', [
```
cornelinux: Is there a deeper idea why call this provider "...Provider"?
We end up calling

    versioningSuffixProviderProvider.$get().$get()

which is rather difficult to type when adding a new route.

As the versioing is always a suffix(?) wouldn't it be possible to simply call it ``'versioning'``  to end up with ``versioningProvider``?

PR #1971, opened 2019-12-19
Set line-length to 100

plettich:
Configuration according to https://github.com/OrkoHunter/pep8speaks#configuration


NuvandaPV:
This is covered by #1972

PR #1974, opened 2019-12-19
Fix description of lost token

### LINKED ISSUE #1973: Description and translation of "lost token" is misleading
The description for the lost-token feature as well as its translation are a little bit misleading. It might give the impression that the token is already marked as lost before pushing the button.

plettich:
Also adds missing translation.

Fixes #1973


cornelinux:
Can you please base this PR tag/v3.2 aka branch-3.2

PR #1981, opened 2020-01-10
Add some diagrams to the documentation

### LINKED ISSUE #1914: Add some diagrams to the documentation
It would be nice to have a few diagrams in the documentation. Specifically I think a deployment diagram giving an example for the components involved in a typical use of push tokens, as well as sequence diagrams for the three authentication modes would be nice.

NuvandaPV:
This adds sphinxcontrib-plantuml, along with the necessary configuration that
*should* allow it to work with readthedocs.io (no real way to test that locally).
Four diagrams are included. A deployment diagram for a typical setup of push tokens,
and diagrams for the different authentication modes. For the latter a whitespace
change was necessary due to the way ReST parses the uml-directive, sorry about the
somewhat ugly diff :|

Fixes #1914

### DIFF
--- a/doc/configuration/authentication_modes.rst
+++ b/doc/configuration/authentication_modes.rst
@@ -47,17 +47,17 @@ modes in more detail.
   Service -> PrivacyIDEA: POST /validate/check
   Service <-- PrivacyIDEA
 
- * The user enters a OTP PIN along with an OTP value.
- * The plugin sends a request to the ``/validate/check`` endpoint:
+* The user enters a OTP PIN along with an OTP value.
+* The plugin sends a request to the ``/validate/check`` endpoint:
 
-   .. code-block:: text
+  .. code-block:: text
 
-     POST /validate/check
+    POST /validate/check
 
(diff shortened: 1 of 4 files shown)


cornelinux:
I set a tag on this branch to verify the building on readthedocs.

doc/configuration/authentication_modes.rst
```diff
@@ -41,113 +41,178 @@ modes in more detail.
 ``authenticate`` mode
 ---------------------
 
- * The user enters a OTP PIN along with an OTP value.
- * The plugin sends a request to the ``/validate/check`` endpoint:
+.. uml::
+  :width: 500
 
-   .. code-block:: text
+  Service -> PrivacyIDEA: POST /validate/check
```
cornelinux: You are adding the word *Service* here.
Maybe we should add some notes, what the *Service* is: that *Service* is the application that gets the 2FA protection by talking to privacyIDEA. I currently have no perfect wording myself.

Please write privacyIDEA with a lower "p".

PR #1984, opened 2020-01-13
Update requirements for v3.3

plettich:
Clean-up of setup.py, add `doc`, `test` and `postgres` as extras.


setup.py
```diff
@@ -5,8 +5,7 @@
 import stat
 import sys
 
-#VERSION="2.1dev4"
```
cornelinux: This is an important line to be able to remember how the format for development versions is.
Please leave it in there ;-)

plettich:
Just a wrap-up of the requirements where the major version changes:
- [x] [alembic](https://alembic.sqlalchemy.org/en/latest/changelog.html): No backward-incompatibilities expected
- [ ] [huey](https://github.com/coleifer/huey/releases): Incompatibilities expected! Revert to 1.11 and test properly
- [x] [oauth2client](https://github.com/googleapis/oauth2client/releases): No backward-incompatibilities expected, See also #1990 
- [x] [Pillow](https://github.com/python-pillow/Pillow/blob/master/CHANGES.rst): No backward-incompatibilities expected
- [x] [PyOpenSSL](https://www.pyopenssl.org/en/stable/changelog.html): No backward-incompatibilities expected, See also #1991
- [x] [pytz](https://pypi.org/project/pytz/): No backward-incompatibilities expected
- [x] [redis](https://pypi.org/project/redis/): This is only used by `huey` and the the apache2 auth-module. `huey` (v1.11) requires `redis` (>=v3.0.0) so we should be safe here.

setup.py
plettich: Maybe we should drop hard-coding of version numbers and use git tags entirely for versioning.

plettich: Fixed in 67e3d62

cornelinux:
@plettich  As far as I am concerned, this can be merged.

PR #1989, opened 2020-01-14
Add encoding to Popen

### LINKED ISSUE #1988: Add encoding to Popen calls.
Add encoding to Popen

https://github.com/privacyidea/privacyidea/blob/0989bea33833728f3cdae33d42c71ab9284de276/privacyidea/lib/caconnectors/localca.py#L426

cornelinux:
Closes #1988

### DIFF
--- a/privacyidea/lib/utils/__init__.py
+++ b/privacyidea/lib/utils/__init__.py
@@ -49,6 +49,7 @@ import threading
 import pkg_resources
 import time
 import cgi
+from subprocess import Popen
 
 from privacyidea.lib.error import ParameterError, ResourceNotFoundError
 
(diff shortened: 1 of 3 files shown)


cornelinux:
Code seems not nice, but obviously works for all python versions

privacyidea/lib/utils/__init__.py
```diff
@@ -1336,4 +1337,15 @@ def check_serial_valid(serial):
     """
     if not re.match(ALLOWED_SERIAL, serial):
         raise ParameterError("Invalid serial number. Must comply to {0!s}.".format(ALLOWED_SERIAL))
-    return True
\ No newline at end of file
+    return True
+
+
+def my_Popen(command, stdin=None, stdout=None, stderr=None, cwd=None, shell=False,
+             encoding=None):
```
plettich: Could we pass `'utf8'` as the default encoding? This would save some code since it is the default...

cornelinux: done

PR #1992, opened 2020-01-15
Cherry pick 1988 to branch 32

plettich:


cornelinux:
@plettich: Is there a reason being that a draft?
Should I wait for something?

PR #1995, opened 2020-01-15
Add policy to allow the administrator to read CA connectors

### LINKED ISSUE #1994: Missing policy CACONNECTORREAD
We are checking for the policy "CACONNECTORREAD", but it does not exist.

cornelinux:
Fixes #1994


cornelinux:
Argh. I set a wrong tag and deleted it again. This is why the travis check fails.

PR #2004, opened 2020-01-29
1867/better distinction adminuser username

cornelinux:


privacyidea/lib/policy.py
```diff
@@ -1017,20 +1026,15 @@ def ui_get_enroll_tokentypes(self, client, logged_in_user):
...
-        else:
-            # If the logged-in user is a user, we pass an user object to allow matching for userinfo attributes
-            user_name = None
-            user_object = User(logged_in_user.get("username"),
-                               logged_in_user.get("realm"))
-            admin_realm = None
+        # In this case we do not distinguish the userobject as for whom an administrator would enroll a token
+        # We simply want to know, which tokentypes a user or an admin in generally allowed to enroll. This is
+        # why we pass an empty params.
+        (role, username, userrealm, adminuser, adminrealm) = determine_logged_in_userparams(logged_in_user, {})
```
plettich: sometimes the returned `scope` is called "`scope"` and sometimes "`role`". We should use consistent naming here.

privacyidea/lib/tokens/totptoken.py
```diff
@@ -675,31 +676,39 @@ def get_default_settings(cls, params, logged_in_user=None,
         ret = {}
         if not logged_in_user:
             return ret
+        (scope, username, userrealm, adminuser, adminrealm) = determine_logged_in_userparams(logged_in_user,
```
plettich: Just an idea: if we return a dictionary here, we could pass it to the other functions without explicitly stating the parameter names. This could reduce the code a bit.

tests/test_api_policy.py
```diff
@@ -268,6 +268,65 @@ def test_02_set_policy_conditions(self):
...
+                                           method='GET',
+                                           headers={
+                                               'Authorization': self.at}):
+            res = self.app.full_dispatch_request()
+            self.assertTrue(res.status_code == 200, res)
+            data = res.json
+            result = data.get("result")
+            policies = result.get("value")
+            # We have two polices
+            self.assertEqual(2, len(policies))
```
plettich: Why are there two policies? We should really try to make test-runs more independent. Is the policy from the previous test still needed? Then maybe we should add/remove it in the setUp/tearDown functions.

tests/test_lib_utils.py
```diff
@@ -781,3 +781,26 @@ def test_32_allowed_serial_numbers(self):
...
+                                                                                      "username": "hans",
+                                                                                      "realm": "realm1"},
+                                                                                     {"user": "peter",
+                                                                                      "realm": "domain"})
+
+        self.assertEqual(scope, "admin")
+        self.assertEqual(user, "peter")
+        self.assertEqual(realm, "domain")
+        self.assertEqual(adminuser, "hans")
+        self.assertEqual(adminrealm, "realm1")
```
plettich: Please also check for the PolicyError in case the "role" is unknown.

Changelog
```diff
@@ -1,3 +1,18 @@
...
+
+  Enhancements:
+  * Policies:
+    * Improve destincion between the username of the administrator
+      and the username of the user. Add an admin username to policies. (#1867)
+      Thus allowing:
+      * User attribute conditions in admin policies
+      * default settings for hashlib and otplen for HOTP and TOTP token 
+        and default timestep for TOTP token can now be dependent on
+	admin user and for which user the admin does the enrollment
```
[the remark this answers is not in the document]
cornelinux: Will do.

privacyidea/lib/policy.py
cornelinux: Actually what is returned is the "role" of the user, which is used as "scope" in the policies.
So I will check to change this to role.

privacyidea/lib/tokens/totptoken.py
cornelinux: I understand. But since this is only a part of the paramters taken by the policy  functions I think this could result in less readable code, since there would be some implicit functionality.

tests/test_api_policy.py
cornelinux: OK, will check.

tests/test_lib_utils.py
cornelinux: will check and do.

Changelog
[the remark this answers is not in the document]
cornelinux: done in 548717046b9c11e90c29d8b50b9fb3b4635e7eeb

privacyidea/lib/policy.py
cornelinux: done in 7649246d672eb64ff314a7ec69826f2ea43fc933

tests/test_lib_utils.py
cornelinux: Done in ff3bb82be7f1697a96015bfa486b034c659d3e0f

tests/test_api_policy.py
cornelinux: I actually see this a bit more relaxed. This is why we can have test classes, so that several tests could build on each other. However, done in 0798f7a598ec6a111c0e64955240d01cd6c7780f

privacyidea/lib/tokens/totptoken.py
cornelinux: Will not do it in this case.

privacyidea/lib/utils/__init__.py
```diff
@@ -1345,7 +1345,7 @@ def determine_logged_in_userparams(logged_in_user, params):
     from the params.
 
     If an administrator is acting, the "adminuser" and "adminrealm" are set from the logged_in_user
-    information and the user parameters are taken from the request parameters.
+    information and the user parameters are takdetermine_en from the request parameters.
```
plettich: determined or taken?

cornelinux: I guess this was an unintentional edit. Will revert and merge. THanks

Issue #1226, opened 2018-09-06
Feature Request: Add option to add {googleurl_img} as an email attachement


speedy01:
Outlook does not support inline images encoded in base64. This is the current method used for QR code in email  {googleurl_img} and usernotifcation.  Please consider adding the option to include the QR image as an attachment.  Another option may be to use the CID or link to the img (if linking, then img/link should expire)

quick link that goes over CID tags
https://sendgrid.com/blog/embedding-images-emails-facts/

cornelinux:
We would have to add another parameter like ``attachments`` (being a list) to this function 
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/smtpserver.py#L130

We would need an email config in the notification handler, to define if images should be sent inline or as attachment. https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/eventhandler/usernotification.py#L113

Here (https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/eventhandler/usernotification.py#L281) we would need to read this config and call ``send_email_identifier`` (https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/eventhandler/usernotification.py#L334) accordingly.

cornelinux:
Resources:
https://docs.python.org/2/library/email.mime.html
https://stackoverflow.com/questions/3362600/how-to-send-email-attachments

cornelinux:
Here is the explanation how this can be done for outlook.
https://community.privacyidea.org/t/qr-code-via-sms-or-email/793/9

plettich:
It seems, google does not allow inline data URLs in gmail: https://stackoverflow.com/a/42014708/7036742
Adding the QR-Code as an attachment solves this.

cornelinux:
I am looking forward to your pull request ;-)

plettich:
Documentation is done, so i'll close this.

PR #2005, opened 2020-01-30
Allow sending images as attachments

plettich:
Some email programs and providers have problems with inline images or
data-URLs so this will be fixed with this commit.
The UserNotificationHandler gets an additional option 'attachment' which
triggers the creation of a multipart email (text and image) and pass it to
the smtp-server.

Working on #1226

TODO: Documentation

### DIFF
--- a/privacyidea/lib/eventhandler/usernotification.py
+++ b/privacyidea/lib/eventhandler/usernotification.py
@@ -55,7 +55,7 @@ from email.mime.text import MIMEText
 from email.mime.image import MIMEImage
 try:
     from urllib.request import urlopen
-except ImportError:
+except ImportError:  # pragma: no cover
     from urllib import urlopen
 import logging
 import os
@@ -306,8 +306,7 @@ class UserNotificationEventHandler(BaseEventHandler):
                     with open(filename, "r", encoding="utf-8") as f:
                         body = f.read()
                 except Exception as e:
-                    log.warning(u"Failed to read email template from file "
-                                u"{0!r}: {1!r}".format(filename, e))
+                    log.warning(u"Failed to read email template from file {0!r}: {1!r}".format(filename, e))
                     log.debug(u"{0!s}".format(traceback.format_exc()))
 
             subject = handler_options.get("subject") or \
@@ -356,7 +355,6 @@ class UserNotificationEventHandler(BaseEventHandler):
                 if attachment and googleurl_img:
                     # get the image part of the googleurl
                     googleurl = urlopen(googleurl_img)
-                    assert googleurl.headers.get('Content-Type') == 'image/png'
                     mail_body = MIMEMultipart('related')
                     mail_body.attach(MIMEText(body, 'html'))
                     mail_img = MIMEImage(googleurl.read())
--- a/privacyidea/lib/eventhandler/usernotification.py
+++ b/privacyidea/lib/eventhandler/usernotification.py
@@ -155,19 +155,19 @@ class UserNotificationEventHandler(BaseEventHandler):
                         NOTIFY_TYPE.INTERNAL_ADMIN,
                         NOTIFY_TYPE.ADMIN_REALM,
                         NOTIFY_TYPE.EMAIL]},
-                "To "+NOTIFY_TYPE.ADMIN_REALM: {
+                "To " + NOTIFY_TYPE.ADMIN_REALM: {
                     "type": "str",
                     "value": get_app_config_value("SUPERUSER_REALM", []),
                     "visibleIf": "To",
                     "visibleValue": NOTIFY_TYPE.ADMIN_REALM},
-                "To "+NOTIFY_TYPE.INTERNAL_ADMIN: {
+                "To " + NOTIFY_TYPE.INTERNAL_ADMIN: {
                     "type": "str",
                     "value": [a.username for a in
                               get_db_admins()],
                     "visibleIf": "To",
                     "visibleValue":
                         NOTIFY_TYPE.INTERNAL_ADMIN},
-                "To "+NOTIFY_TYPE.EMAIL: {
+                "To " + NOTIFY_TYPE.EMAIL: {
                     "type": "str",
                     "description": _("Any email address, to which the notification "
                                      "should be sent."),
--- a/privacyidea/lib/eventhandler/usernotification.py
+++ b/privacyidea/lib/eventhandler/usernotification.py
@@ -50,6 +50,13 @@ from privacyidea.lib.user import User, get_user_list
 from privacyidea.lib.utils import create_tag_dict
 from privacyidea.lib.crypto import get_alphanum_str
 from privacyidea.lib import _
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+from email.mime.image import MIMEImage
+try:
+    from urllib.request import urlopen
+except ImportError:
+    from urllib import urlopen
 import logging
 import os
 import traceback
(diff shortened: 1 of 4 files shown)


privacyidea/lib/eventhandler/usernotification.py
```diff
@@ -301,7 +306,8 @@ def do(self, action, options=None):
                     with open(filename, "r", encoding="utf-8") as f:
                         body = f.read()
                 except Exception as e:
-                    log.warning(u"Failed to read email template from file {0!r}: {1!r}".format(filename, e))
+                    log.warning(u"Failed to read email template from file "
```
cornelinux: Are you breaking those lines manually? Or does your IDE break it automatically?
Since the line still is shorter than 120 characters. I think we confirmed on this line length.
If we confirm on a certain line length I would recommend to use this and not add breaks where not necessary

privacyidea/lib/eventhandler/usernotification.py
```diff
@@ -345,7 +351,20 @@ def do(self, action, options=None):
                 mimetype = handler_options.get("mimetype", "plain")
                 useremail = recipient.get("email")
                 reply_to = handler_options.get("reply_to")
-
+                attachment = handler_options.get("attachment", False)
```
cornelinux: If "attachment" would not exist, then ``attachment`` will be None and then below ``if attachment`` will also work.
Is there a reason for defaulting it to False instead of None?

privacyidea/lib/eventhandler/usernotification.py
```diff
@@ -345,7 +351,20 @@ def do(self, action, options=None):
...
+
+                if attachment and googleurl_img:
```
cornelinux: Just a logical question: The "attachment" option will always attach a googleurl_img, if it exists in the response, no matter if the "googleurl_img" is contained in the body.
So the "attachment" implicitly means "attach QR Code from enrolled tokens"?
OK, the description indicates it. But maybe we might have some other "attachment" in the future and then actually the name should be more specific like "qrcode_attachment"? What do you think?

privacyidea/lib/eventhandler/usernotification.py
```diff
@@ -345,7 +351,20 @@ def do(self, action, options=None):
...
+                    # get the image part of the googleurl
+                    googleurl = urlopen(googleurl_img)
+                    assert googleurl.headers.get('Content-Type') == 'image/png'
```
cornelinux: We always create the googleurl_img to be a png.
So what would be the assert for?

tests/test_lib_events.py
```diff
@@ -2147,6 +2173,8 @@ def test_02_sendmail(self):
         un_handler = UserNotificationEventHandler()
         res = un_handler.do("sendmail", options=options)
         self.assertTrue(res)
+        msg = smtpmock.get_sent_message()
+        assert 'To: user@localhost.localdomain' in msg
```
cornelinux: Please use 

    self.assertIn()

in tests.

tests/test_lib_events.py
```diff
@@ -3347,7 +3397,85 @@ def test_20_sendmail_googleurl(self):
...
+                   "handler_def": {
+                       "conditions": {"serial": "123.*"},
+                       "options": {"body": "<img src='cid:token_image' />",
+                                   "mimetype": "html",
+                                   "attachment": True,
+                                   "emailconfig": "myserver"}}}
+
+        un_handler = UserNotificationEventHandler()
+        res = un_handler.do("sendmail", options=options)
+        assert res
```
cornelinux: Please use

    self.assertTrue()

in tests.

tests/test_lib_events.py
```diff
@@ -3347,7 +3397,85 @@ def test_20_sendmail_googleurl(self):
...
+        parsed_email = email.message_from_string(smtpmock.get_sent_message())
+        assert parsed_email.get_content_maintype() == 'multipart'
```
cornelinux: self.assertEqual()

and all the assert below.

tests/test_lib_smtpserver.py
```diff
@@ -103,6 +111,21 @@ def test_05_test_email(self):
                                   "Test Email from privacyIDEA",
                                   "This is a test email from privacyIDEA. "
                                   "The configuration %s is working." % identifier)
+        assert r
```
cornelinux: self.assertTrue

and change all assert below.

tests/test_lib_smtpserver.py
```diff
@@ -103,6 +111,21 @@ def test_05_test_email(self):
...
+        parsed_email = email.message_from_string(smtpmock.get_sent_message())
+        assert parsed_email.get_content_type() == 'text/plain'
```
cornelinux: e.g.

    self.assertEqual()

privacyidea/lib/eventhandler/usernotification.py
plettich: I've added the breaks since i find it more readable. Also i would prefer a 100 chars line limit overall, but this can be discussed later.
I'll fix this to 100 chars for now.

privacyidea/lib/eventhandler/usernotification.py
plettich: Just a precaution and making things more clear. If You like i can remove it.

privacyidea/lib/eventhandler/usernotification.py
plettich: Of course, this specifically targets the `googleurl`. There is also the `oathurl` and `otpkey` as QR-Code images in the response but i  haven't seen them used anywhere else.
So ideally we would allow to select all (QR-Code) images as attachments (and maybe some other stuff as well? PDF with descriptions, ...) but i think handling this here is overkill.
We could add attachments based on [`CID` URLs](https://tools.ietf.org/html/rfc2392) in the body. But this means we would have to parse the body before.
Maybe we can externalize the whole email creation stuff and we only provide the QR-Code images.

privacyidea/lib/eventhandler/usernotification.py
plettich: Was a check for me that the URL parsing worked as expected. I'll remove it.

privacyidea/lib/eventhandler/usernotification.py
cornelinux: I think part of my point was that "attachent" sounds like a very generic keyword. If we ever^(TM) would want to attach other things to a notification, this generic word would be wasted. 

Image we would one day send notifactions to students including a PDFs with certain information and we would attach the PDF, or we would send a notification "you are ready to enroll your smartcard, please use the attached enrollment tool" and we would attach a ZIP or EXE or whatever.
At the moment this might be far fetch but it is a feeling in my guts, with the experience with privacyIDEA, that "attachment" might be something bigger int he future.

So please use a keyword like "attach_qrcode" to be able to be more generic in the future.

privacyidea/lib/eventhandler/usernotification.py
cornelinux: For me it would be clearer the other way round.
But I am not sure, if we need certain styles at this point.

privacyidea/lib/eventhandler/usernotification.py
cornelinux: I think 100 chars is to short. And we added 120 chars in the .pep8speaks.
This line is 108 chars. I think there are a lot of lines that are longer than 100chars. Because an average notification text might be around 70-80 lines. With indentation and command such a line will almost always be longer than 100 lines. So you will break all those lines.
Please stick with 120.

privacyidea/lib/eventhandler/usernotification.py
plettich: will do

privacyidea/lib/eventhandler/usernotification.py
plettich: ok

privacyidea/lib/eventhandler/usernotification.py
plettich: Done in 4dc32558

privacyidea/lib/eventhandler/usernotification.py
plettich: Done in bb386797

tests/test_lib_events.py
plettich: Done in bb386797

privacyidea/lib/eventhandler/usernotification.py
plettich: Done in bb386797

cornelinux:
Thanks! Looks good to me, besides the failing test.

Issue #1967, opened 2019-12-16
Dust off documentation


cornelinux:
* [x] ~~remove references to PPA~~
* [ ] update pictures
* [ ] references to debian packages?
* [ ] Some policies seem to be missing, token specific policies...
* [ ] #1504 
* [x] ~~remove update chapters and~~ add link to READ_BEFORE_UPDATE.md (polish update information)
* [ ] #1787 
* [ ] #1831

cornelinux:
I think it makes sense to have the PPA repo reference.
There are installations running with 2.23 or older.
If they want to update to 3.0, they will probably need to install a *new* 2.23, from the PPA repositories. 

It always bug me, if OSS projects remove old versions or documentations.

cornelinux:
I would still keep an upgrade/update chapter, because it is referenced a lot. And I also would keep this chapter, since it contains the information, how to update.
But a reference to READ_BEFORE_UPDATE makes sense.

PR #2011, opened 2020-02-06
Add documentation for attaching token image

plettich:
Also updated some of the usernotification documentation.

Working on #1967
Working on #1226

### DIFF
--- a/doc/eventhandler/usernotification.rst
+++ b/doc/eventhandler/usernotification.rst
@@ -11,34 +11,103 @@ administrators in case of any event.
 Possible Actions
 ~~~~~~~~~~~~~~~~
 
+.. _sendmail_action:
+
 sendmail
 ........
 
-The *sendmail* action sends an email to the tokenowner user. The email is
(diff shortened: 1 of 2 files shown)


doc/eventhandler/usernotification.rst
```diff
@@ -11,34 +11,103 @@ administrators in case of any event.
...
+
+.. warning:: The KEY URI and the QR Code contain the secret OTP key in plain
+   text. Everyone who receives this data has a detailed copy of this token.
+   Thus we very much recommend to **never** send these data in an unencrypted
+   email!
+
+**attach_qrcode**
+
+Instead of sending the QR-Code as an inline data image (which is not supported
+by some email clients / webmailer), enabling this option sends the email as a
```
cornelinux: Should we add something "... like Outlook", since people usually have problems with that?

plettich: will do

plettich: done in 7b39aa7e

PR #2021, opened 2020-02-11
Fix PrivacyIdea-OTP LDAP_STARLTLS error

darrenche:
Fix the condition for start_tls when using postgres SQL

### DIFF
--- a/privacyidea/lib/resolvers/LDAPIdResolver.py
+++ b/privacyidea/lib/resolvers/LDAPIdResolver.py
@@ -1008,7 +1008,7 @@ class IdResolver (UserIdResolver):
                                       receive_timeout=timeout,
                                       auto_referrals=not param.get(
                                            "NOREFERRALS"),
-                                      start_tls=param.get("START_TLS", False))
+                                      start_tls=is_true(param.get("START_TLS", False)))
             #log.error("LDAP Server Pool States: %s" % server_pool.pool_states)
             if not l.bind():
                 raise Exception("Wrong credentials")


plettich:
Hi @darrenche and thanks for the PR,
could You open an issue for this PR? We try to track the development via GitHub issues.
So if i understand correctly, You want to test the connection and pass the `START_TLS` parameter with a value of `False`? Actually the json-decoder already takes care of translating a json-`true` to a python-`True`.
Could You elaborate when this issue occurs for You?

darrenche:
@plettich So the test done in here utilizes the `is_true` wrapper as shown here: https://github.com/privacyidea/privacyidea/blob/db22f533badbff33ff112ae3ed546b59150c7272/privacyidea/lib/resolvers/LDAPIdResolver.py#L345
and 
https://github.com/privacyidea/privacyidea/blob/db22f533badbff33ff112ae3ed546b59150c7272/privacyidea/lib/resolvers/LDAPIdResolver.py#L760

However, when creating a connection, there is no check that utilizes this same `is_true` wrapper and from my testing, when we pull the parameters from a database, `START_TLS` is `"False"` as a string and not a boolean which caused problems on our side. I understand this can be as simple as fixing the parameters we're sending to this function but I think it might be better to have this check.
https://github.com/privacyidea/privacyidea/blob/db22f533badbff33ff112ae3ed546b59150c7272/privacyidea/lib/resolvers/LDAPIdResolver.py#L1011

cornelinux:
Merging this PR, although it really does not have a nice issue with #2025, which actually describes nothing.
We assume that in certain cases - probably with a different database - the JSON may contain a not well parsed "TRUE".

PR #2022, opened 2020-02-11
Fix questionnaire token message bug

darrenche:
Fix for message bug for questionnaire to display correctly


plettich:
Hi @darrenche and thanks for the PR,
again, could You open an issue for this PR?
Actually `replace()` could also break (for something like `'test.typedef.type'`, better check with `endswith('.type')` and then remove the last chars.

darrenche:
@plettich My apologies for not opening an issue initially again!

As for this issue, you're definitely right if the question as `.type` inside of it, the token doesn't even populate. But also, since the condition checks if the key ends with `.type` and the value is `password`, then it shouldn't be a problem unless the user has `.type` inside their security question.

plettich:
This is fixed in #2033

PR #2023, opened 2020-02-12
Trigger Challenge fix for serial_wildcard

darrenche:
Two different changes:

1) Feature Request to support wildcard serials when doing a trigger challenge (Lines 451, 529, 533-538)

2) Retain the request params when doing a trigger challenge e.g. `application` param (Lines 528-531)


cornelinux:
Hello @darrenche ,
I will close this pull request because it is most probable that we do not merge it this way.
1. as @plettich pointed out, please open an issue for a PR
2. We are not completely sure, what this PR should actually do and why
3. Your changes touches central functionality of privacyIDEA. Such things should not be discussed in a Pull Request but in an open issue. You probably want to achieve something (what). But we might find another design to do this.
Thanks a lot for your understanding.

PR #2034, opened 2020-02-14
added missing "position" argument (e_import)

### LINKED ISSUE #2036: Missing "position" argument importing events through pi-manage
### Top-level intent

Importing events does not load position param.

### Steps to reproduce

1. Go to events panel, and create one event with "pre" position. 
2. Export event: `pi-manager event e_export > /tmp/events.py`
3. Import event: `pi-manager event e_import --cleanup /tmp/events.py`
4. Go to events and see the event with "post" position set.

### Expected outcome

`e_import` command should load "position" argument from file

### Actual outcome

`e_import` command always set events position as "post"

### Configuration

* **privacyIDEA version**: master
* **Installation method**: github
* **Python version**: 3.7
* **Operating system**: Ubuntu 18.04
* **Webserver**: development server
* **Token database**: MSSQL (https://github.com/privacyidea/docker/pull/4)

brunocascio:
Ref #2036

Closes #2036

### DIFF
--- a/pi-manage
+++ b/pi-manage
@@ -1034,7 +1034,7 @@ def e_import(filename, cleanup=False):
                       ordering=event.get("ordering"),
                       options=event.get("options"),
                       active=event.get("active"),
-                      position=event.get("position"))
+                      position=event.get("position", "post"))
         print("Added event '{0!s}' with result {1!s}".format(name, r))
 
 
--- a/pi-manage
+++ b/pi-manage
@@ -1033,7 +1033,8 @@ def e_import(filename, cleanup=False):
                       conditions=event.get("conditions"),
                       ordering=event.get("ordering"),
                       options=event.get("options"),
-                      active=event.get("active"))
+                      active=event.get("active"),
+                      position=event.get("position"))
         print("Added event '{0!s}' with result {1!s}".format(name, r))
 
 


plettich:
Thanks for the PR @brunocascio
Could You open an issue with some description so we can track the development?

brunocascio:
@plettich done :)

pi-manage
```diff
@@ -1033,7 +1033,8 @@ def e_import(filename, cleanup=False):
                       conditions=event.get("conditions"),
                       ordering=event.get("ordering"),
                       options=event.get("options"),
-                      active=event.get("active"))
+                      active=event.get("active"),
+                      position=event.get("position"))
```
cornelinux: If we have an old exported event definition, that does not know post and pre positions, maybe we should default to "post":
~~~~python
   position=event.get("position", "post"))
~~~~

Please consider enhancing this.

brunocascio: Hi @cornelinux!

Should "Post" be the default value? If an exported configuration doesn't know about position, assuming "post" could be wrong. Anyway, if yes, I'll do it.

cornelinux: If it does not know about positions, the export is from a time, when *all* event handlers where post. (No other positions have been available) 
So post would always be right in this case.

brunocascio: Thanks for the explanation!

I have just pushed the change

PR #2035, opened 2020-02-14
Propagate finalize_log() call for ContainerAudit

### LINKED ISSUE #2029: Missing Audit entries when using the ContainerAudit module
When using the container audit module, the `finalize_log()` call isn't properly propagated to the actual audit log modules in all cases.

plettich:
The `ContainerAudit` module did not propagate the `finalize_log()` call to
its sub-modules in all cases.
This is fixed together with a little reorganization of the test cases.

Fixes #2029

### DIFF
--- a/privacyidea/lib/auditmodules/base.py
+++ b/privacyidea/lib/auditmodules/base.py
@@ -89,8 +89,10 @@ class Audit(object):  # pragma: no cover
         """
         self.name = "AuditBase"
         self.audit_data = {}
+        self.config = config or {}
         self.private = ""
         self.public = ""
+        self.have_data = False
 
     @log_with(log)
     def initialize(self):
@@ -160,6 +162,10 @@ class Audit(object):  # pragma: no cover
         """
         return None
 
+    @property
+    def has_data(self):
+        return self.have_data
+
     @log_with(log)
     def log(self, param):
         """
@@ -175,6 +181,7 @@ class Audit(object):  # pragma: no cover
         """
         for k, v in param.items():
             self.audit_data[k] = v
+        self.have_data = True
 
     def add_to_log(self, param, add_with_comma=False):
         """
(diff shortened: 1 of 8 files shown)


privacyidea/lib/auditmodules/base.py
```diff
@@ -89,8 +89,10 @@ def __init__(self, config=None):
         """
         self.name = "AuditBase"
         self.audit_data = {}
+        self.config = config or {}
         self.private = ""
         self.public = ""
+        self.have_data = False
```
cornelinux: I find this irritating, that there is a object variable "have_data" and a property "has_data".

I looks like we usually could call "if f.audit_object.have_data" in before_after anyways.
I would like to have the internal variable renamed, maybe to ``self._has_data``. Thus it is more clear, that this belongs together and future audit modules could still evaluate the fact, if there is data more sophisticated in the property ``has_data``.
What do you think?

privacyidea/lib/auditmodules/containeraudit.py
```diff
@@ -64,6 +63,7 @@ def log(self, param):
         """
         for module in self.write_modules:
             module.log(param)
+        self.have_data = True
```
cornelinux: Assume we have no writeable modules, then it would contain no data - right?
So I think it should be

~~~~python
for module in self.write_modules:
    module.log(param)
    self.have_data = True
~~~~

privacyidea/lib/auditmodules/base.py
plettich: Yeah, i was trying to be smart... I'll clean that up.

privacyidea/lib/auditmodules/containeraudit.py
plettich: I'll look into it

privacyidea/lib/auditmodules/base.py
plettich: Done in 1997585

privacyidea/lib/auditmodules/containeraudit.py
plettich: Done in 1997585

PR #2037, opened 2020-02-14
Avoid password error in sqlresolver

### LINKED ISSUE #2030: Allow SQL Resolver to have no password mapping
The SQL resolver alwas requires a password mapping to be defined.
In certain cases it is not necessary to have a password for a user.
But the password is, when adding a user in a sqlresolver, a required field.

It should be possible to define sqlresolvers without a password field.

What should happen, if the policy "otppin=userstore" is defined. The password from the user store can not be verified, so I think in this case authentication should fail.

cornelinux:
If a editable SQL resolver is defined without a password mapping,
then the line

   del attributes("password")

will cause a Keyerror. Thus it was not possible to define SQL resolvers
without a password. Deleting the password, only if it exists, now
actually allows us to create users in a SQL resolver without a password
mapping.

Fixes #2030

### DIFF
--- a/privacyidea/api/user.py
+++ b/privacyidea/api/user.py
@@ -176,7 +176,8 @@ def create_user_api():
     # Remove the password from the attributes, so that we can hide it in the
     # logs
     password = attributes.get("password")
-    del attributes["password"]
+    if hasattr(attributes, "password"):
+        del attributes["password"]
     r = create_user(resolvername, attributes, password=password)
     g.audit_object.log({"success": True,
                         "info": u"{0!s}: {1!s}/{2!s}".format(r, username,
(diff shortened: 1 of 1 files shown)


privacyidea/api/user.py
```diff
@@ -176,7 +176,8 @@ def create_user_api():
     # Remove the password from the attributes, so that we can hide it in the
     # logs
     password = attributes.get("password")
-    del attributes["password"]
+    if "password" in attributes:
+        del attributes["password"]
```
plettich: is there a reason why we delete the pw from the attributes here and add it again in [`create_user()`](https://github.com/privacyidea/privacyidea/blob/1143f41d7abc9aa6809fd50ae91d86b4eb6a071f/privacyidea/lib/user.py#L511)?

cornelinux: Yes. Otherwise we can not hide the password from the @log wrapper.
It would be contained in the params dictionary and would be visible in the log.debug level.

The create_user is decorated with
~~~~python
@log_with(log, hide_kwargs=["password"])
def create_user(resolvername, attributes, password=None):
~~~~
Which means, never show the arg "password" in the logs :-)

Issue #830, opened 2017-10-28
Update angular.js


cornelinux:
We should update angular.js.
If we update e.g. to 1.6.6 we need at least:

* update angular.js
* update in static/contrib/js/ngmodules/ 
  * angular-animate.js
  * angular-resource.js
  * angular-route.js
  * angular-ui-router.js

The ``$http.get`` and ``.post`` do not support ``.success()`` anymore. Now it is called ``.then()``.

And there might be a lot other changes.

plettich:
TODO:
- [ ] update `READ_BEFORE_UPDATE.md` for possible changes in templates.

plettich:
The update of the angular-ui-router package broke the reload button and the return to the token list after deleting a token.
The [`$stateChangeSuccess`](https://ui-router.github.io/guide/ng1/migrate-to-1_0#state-change-events) event is replaced by transition hooks.

plettich:
I think we covered all the issues here regarding the update of the UI components. I'll close this now and if something doesn't work, we can open a new issue to work on that.

PR #2044, opened 2020-02-19
Update UI components

plettich:
- Update bootstrap to v3.4.1
    - add required css files and fonts
- Update jQuery to v3.4.1

Working on #830


cornelinux:
@plettich Please check if we need to add info to read_before_update. This could be necessary if some templates have changed in a way, that customizations would prevent the usage of new components.

plettich:
A few things like u2f-token and tokenwizzard haven't been tested yet.

plettich:
Updated READ_BEFORE_UPDATE
Also checked u2f enrollment.

PR #2053, opened 2020-02-24
Remove Ubuntu 14.04 package references

plettich:
Ubuntu 14.04 is not supported any more so we should remove it from the
docs.
Also some of the plugin packages aren't available in the community
repository anymore, which lead to misunderstandings (e.g. privacyidea-pam).

Working on #1967

### DIFF
--- a/doc/application_plugins/index.rst
+++ b/doc/application_plugins/index.rst
@@ -28,8 +28,8 @@ communicates with the privacyIDEA server via the API. The PAM module also suppor
 authentication. In this case you need to configure an offline machine application. (See
 :ref:`application_offline`)
 
-You can install the PAM module with a ready made Debian package for Ubuntu or
-just use the source code file. It is a python module, that requires pam-python.
+You can install the PAM module by using the source code file. It is a python module, that
+requires pam-python.
 
 The configuration could look like this::
 
--- a/doc/application_plugins/index.rst
+++ b/doc/application_plugins/index.rst
@@ -23,9 +23,9 @@ Pluggable Authentication Module
 
 .. index:: offline, PAM
 
-The PAM module of privacyIDEA directly communicates with the privacyIDEA
-server via the API. The PAM module also supports offline authentication. In
-this case you need to configure an offline machine application. (See
+The `PAM module of privacyIDEA <https://github.com/privacyidea/pam_python>`_ directly
+communicates with the privacyIDEA server via the API. The PAM module also supports offline
+authentication. In this case you need to configure an offline machine application. (See
 :ref:`application_offline`)
 
 You can install the PAM module with a ready made Debian package for Ubuntu or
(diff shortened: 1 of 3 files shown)


doc/application_plugins/index.rst
```diff
@@ -23,9 +23,9 @@ Pluggable Authentication Module
...
 
-The PAM module of privacyIDEA directly communicates with the privacyIDEA
-server via the API. The PAM module also supports offline authentication. In
-this case you need to configure an offline machine application. (See
+The `PAM module of privacyIDEA <https://github.com/privacyidea/pam_python>`_ directly
+communicates with the privacyIDEA server via the API. The PAM module also supports offline
+authentication. In this case you need to configure an offline machine application. (See
 :ref:`application_offline`)
 
 You can install the PAM module with a ready made Debian package for Ubuntu or
```
cornelinux: So we also should remote "ready made Debian package"?

doc/application_plugins/index.rst
```diff
@@ -165,17 +165,18 @@ With either setup, you can test the RADIUS setup using a command like this::
 Microsoft NPS server
 --------------------
 You can also use the Microsoft Network Protection Server with privacyIDEA.
-A full featured integration guide can be found at the NetKnights webpage
-[#NPS]_.
```
cornelinux: What is wrong with the footnote references?

doc/application_plugins/index.rst
```diff
@@ -266,16 +267,11 @@ Anyway, in a productive environment you probably want to uncheck this feature.
 OTRS
 ----
 
-There are two plugins for OTRS. For OTRS version 4.0 and higher use
-*privacyIDEA-4_0.pm*.
+The OTRS Plugin can be found in its own
+`GitHub Repository <https://github.com/privacyidea/otrs>`__.
```
cornelinux: Are two underscores at the end correct?

doc/installation/ubuntu.rst
```diff
@@ -8,7 +8,7 @@ Ubuntu Packages
 
 There are ready made packages for Ubuntu.
 Up to privacyIDEA 2.23 packages are available for 
-Ubuntu 14.04 LTS and 16.04 LTS [#ubuntu1604]_.
```
cornelinux: 14.04 packages are still available at launchpad.
So in fact they could be installed there. I think we can leave this in here.

doc/application_plugins/index.rst
plettich: huh, missed that, was looking for ubuntu :-) will fix.

doc/application_plugins/index.rst
plettich: Since the link is often an integral part of the text ("... see here.", "... get from"), footnotes are not appropriate (in my opinion).
I was thinking of adding footnotes additionally to the links (for people who don't like to click on links) .

doc/application_plugins/index.rst
plettich: Yes, it's an anonymous reference (https://github.com/sphinx-doc/sphinx/issues/3921#issuecomment-315581557). Otherwise sphinx would complain about to many "Duplicate explicit target name".

doc/installation/ubuntu.rst
plettich: will do

plettich: Done in e4a50d61

doc/application_plugins/index.rst
plettich: Done in e4a50d61

PR #2060, opened 2020-03-03
algorithm has to be upper case for Google Authenticator on iOS

juliansiebert:
The "algorithm=" parameter has to be in upper case to be valid in Google Authenticator App on iOS. Microsoft Authenticator takes it in upper and lower case
I have not tested on any other platforms.

### DIFF
--- a/privacyidea/lib/apps.py
+++ b/privacyidea/lib/apps.py
@@ -148,7 +148,7 @@ def create_google_authenticator_url(key=None, user=None,
     url_issuer = quote(issuer.encode("utf-8"))
 
     if hash_algo.lower() != "sha1":
-        hash_algo = "algorithm={0!s}&".format(hash_algo)
+        hash_algo = "algorithm={0!s}&".format(hash_algo.upper())
     else:
         # If the hash_algo is SHA1, we do not add it to the QR code to keep
         # the QR code simpler


plettich:
Hi and thanks for the PR.
Could You open an issue for this PR? We like to track changes and discussions there.
Also note that while the algorithm is specified in the documentation as upper-case, the google authenticator might not accept anything other than SHA1:
https://github.com/google/google-authenticator/wiki/Key-Uri-Format#algorithm
And the Microsoft authenticator has its issues as well:
https://community.privacyidea.org/t/do-not-use-microsoft-authenticator-it-is-evil/448
Just use the [privacyIDEA Authenticator](https://github.com/privacyidea/privacyidea-authenticator-ios), it works best the privacyIDEA server :-)

cornelinux:
Hello @juliansiebert ,

we investigated some apps in #2079.
And honestly, I recommend you use sha1 with your Google Authenticator. Why? To avoid problems in the future.

Please note that the Google Authenticator on Android does **not** support sha256 and not SHA256. In simply silently assumes a sha1. So if ever one of your users comes around with a android phone, this very user will fail to authenticate with no warnings. forever.
In regards to HOTP/TOTP sha1 has no disadvantage in regards to sha256.
So rather than pushing this uppercase to deal with on crappy app on one platform, I would recommend that you adapt your settings.
Note that all other apps, that actually support sha256 do not care about upper or lower case.

Does this make sense to you?

Regards,
Cornelius

juliansiebert:
Hi,
we opted for the sha1 hashes as we don't want to impose which app to use on our users.

Thanks for all the effort put into this!

Best regards!

PR #2069, opened 2020-03-08
Honor usernames in webui policies.

### LINKED ISSUE #2038: use Match.generic in get_webui_settings
In ``api/lib/postpolicy:get_webui_settings`` we use a lot of ``Match.realm``. This seems to have historic reasons, because at this point *after* ``/auth`` we know the logged in user, so we could also match based on logged_in_user, username and resolver. Maybe we can also use ``Match.admin_or_user``.

We should change this, since currently these webui settings are only evaluated based on the realm.

see: https://github.com/privacyidea/privacyidea/blob/master/privacyidea/api/lib/postpolicy.py#L486

cornelinux:
The webui policies that are called in get_webui_settings are
fetched after the user has logged in. Thus we already know the
username of the logged in user. Thus we can not only match the
realm but also the username of the logged in user to match webui
settings.

Closes #2038

### DIFF
--- a/privacyidea/api/lib/postpolicy.py
+++ b/privacyidea/api/lib/postpolicy.py
@@ -484,20 +484,21 @@ def get_webui_settings(request, response):
     if content.get("result").get("status") is True:
         role = content.get("result").get("value").get("role")
         loginname = content.get("result").get("value").get("username")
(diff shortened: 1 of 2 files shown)


cornelinux:
@laclaro  Added test in e703dea8

PR #2074, opened 2020-03-10
Fix accordion UI component

plettich:
After an update of angular and bootstrap in #830 the accordion component
has changed.

Working on #830

### DIFF
--- a/privacyidea/static/components/config/views/config.policies.details.html
+++ b/privacyidea/static/components/config/views/config.policies.details.html
@@ -273,16 +273,6 @@
                 </label>
 
                 <div class="col-sm-9">
-                    <div isteven-multi-select
-                         ng-hide="isActionValues"
-                         input-model="actions"
-                         output-model="selectedActions"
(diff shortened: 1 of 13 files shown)


privacyidea/static/components/config/views/config.policies.details.html
```diff
@@ -273,16 +273,6 @@ <h3 translate>Action</h3>
                 </label>
 
                 <div class="col-sm-9">
-                    <div isteven-multi-select
```
cornelinux: Why did this have to go?

plettich: As far as i could see, this isn't used and just left an empty drop-down something unless a "scope" was selected above. Then the accordion would show up.
I don't know what the intentions of this part were.
Tested only on Firefox...

cornelinux: Right. This might be a remnant from when I restructured the UI of the policy details with the two tabs.

Issue #2059, opened 2020-03-03
Python suggests using dictConfig for logging configuration instead of files


plettich:
According to the [Python docs](https://docs.python.org/3/library/logging.config.html#configuration-file-format), the `fileConfig()` API is kind of deprecated and won't be updated in the future.
This already impacts logging configuration files with the [`RotatingFileHandler()`](https://docs.python.org/3/library/logging.handlers.html#rotatingfilehandler) which ignores the `maxBytes` and `backupCount` parameter.

cornelinux:
Does this mean we would have to read the config file ourselves, parse it into a dict and then do ``dictConfig``?

plettich:
Yes, this is needed for setting up logging to logstash/splunk (#1580).
With a logging configuration in YAML this will take two lines of code.

plettich:
TODO:
- [ ] update docs

PR #2080, opened 2020-03-12
Read YAML based logging configuration

plettich:
The python docs suggest using `dictConfig()` instead of `fileConfig()` for
complex logging configurations.
This PR adds the possibility to read logging configuration from a config
file as well as a YAML file. The YAML based configuration will be passed to
`dictConfig()` and takes precedence over a config-file based configuration.
If both files don't exist (or could not be read), the default logging
configuration from `privacyidea.lib.log` will be used.

Working on #2059

### DIFF
--- a/privacyidea/app.py
+++ b/privacyidea/app.py
@@ -26,8 +26,9 @@ import logging.config
 import sys
 import yaml
 from flask import Flask, request, Response
+from flask_babel import Babel
+from flask_migrate import Migrate
 from flaskext.versioned import Versioned
-from privacyidea.lib import queue
 
 # we need this import to add the before/after request function to the blueprints
 import privacyidea.api.before_after
@@ -59,12 +60,10 @@ from privacyidea.api.smsgateway import smsgateway_blueprint
 from privacyidea.api.clienttype import client_blueprint
 from privacyidea.api.subscriptions import subscriptions_blueprint
 from privacyidea.api.monitoring import monitoring_blueprint
+from privacyidea.lib import queue
 from privacyidea.lib.log import DEFAULT_LOGGING_CONFIG
 from privacyidea.config import config
 from privacyidea.models import db
-from flask_migrate import Migrate
-from flask_babel import Babel
-
 
 ENV_KEY = "PRIVACYIDEA_CONFIGFILE"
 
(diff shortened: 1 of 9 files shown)


privacyidea/app.py
```diff
@@ -196,37 +172,42 @@ def create_app(config_name="development",
...
             logging.config.fileConfig(log_config_file)
+            have_logging_conf = True
             if not silent:
-                print("Reading Logging settings from {0!s}".format(log_config_file))
-        else:
-            raise Exception("The config file specified in PI_LOGCONFIG does "
-                            "not exist.")
+                print("Read Logging settings from {0!s}".format(log_config_file))
+        # now try to read from yaml config
+        log_conf_yaml = app.config.get("PI_LOGCONFIG_YAML",
```
cornelinux: But this means, if I have both in my pi.cfg
* PI_LOGCONFIG
* PI_LOGCONFIG_YAML
both the cfg file and the yaml files are read. Does this make sense?
Also, If I first successfully read the cfg file but do an exception during yaml read, I would have a valid cfg file config, but still enter the ``except``.  OK. But sind have_logging_conf is true, I would actually use the cfg file.

At first glance this looks a bit weired to me, since several things are tried and if something works out, it is used.
I would rather expect that I **define** to use yaml config and it fails I would not want a fallback on the cfg file.
Hm, in this case I would simply have to remove ``PI_LOGCONFIG`` from the pi.cfg. OK, technically fine.

But this is a strange behaviour, which needs a good documentation.

plettich: Yeah, i was trying to be backwards compatible as much as possible but i think we should just use an either/or approach:
- Have only one PI_LOGCONFIG in the config file
- detect whether it is a conf or yaml file and use the appropriate reader
I'll update this.

cornelinux: detecting the file type would be totally awesome.
Maybe we can try to load the file as a yaml file (new behaviour) and if it thows an exception load it as cfg (old behaviour)

plettich: I've added a detection based on the ending.

privacyidea/app.py
```diff
@@ -196,38 +171,37 @@ def create_app(config_name="development",
...
 
+    # Setup logging
     try:
         # Try to read logging config from file
         log_config_file = app.config.get("PI_LOGCONFIG",
                                          "/etc/privacyidea/logging.cfg")
         if os.path.isfile(log_config_file):
-            logging.config.fileConfig(log_config_file)
+            _, ext = os.path.splitext(log_config_file)
+            if ext.upper() in ['.YML', '.YAML']:
```
cornelinux: Honestly I do not like file endings so much! :-/ Sounds like windows.
I would think it is more pythonic to catch eceptions.
Or to use something like python-magic. On the other hand, I do not want to get it too bloated.
What was your reason to go for file endings?

privacyidea/app.py
```diff
@@ -196,38 +171,37 @@ def create_app(config_name="development",
...
+                log_cnf_dict = yaml.safe_load(open(log_config_file, 'r').read())
+                logging.config.dictConfig(log_cnf_dict)
+            else:
+                logging.config.fileConfig(log_config_file)
             if not silent:
-                print("Reading Logging settings from {0!s}".format(log_config_file))
+                sys.stdout.write("Read Logging settings from {0!s}".format(log_config_file))
```
cornelinux: Did you mean ``stderr`` or what is the reason to change print or write?

privacyidea/app.py
```diff
@@ -196,38 +171,37 @@ def create_app(config_name="development",
...
         else:
-            raise Exception("The config file specified in PI_LOGCONFIG does "
-                            "not exist.")
+            raise OSError('Unable to use logging configuration file {0!s}'.format(log_config_file))
```
cornelinux: Well, if ``os.path.isfile(log_config_file)`` is False, the file does not exist, right?
Or what is the reason to change the text? Wouldn't you usually not use vairables in exceptions?

privacyidea/app.py
cornelinux: I am not that happy with it. See my  new comments.

privacyidea/app.py
plettich: Unfortunately there is no way in detecting the content type of the file (except reading it and catching an exception).
Both the builtin mimetypes and python-magic only detect it as plain text files.

plettich: Fixed in 9136a076

privacyidea/app.py
plettich: Fixed in 9136a076

privacyidea/app.py
plettich: Fixed in 9136a076

privacyidea/app.py
cornelinux: thanks - looks awesome! :-)

plettich: Fixed typo and docu in 42218553

plettich:
Fixed typo and docu in 4221855

Issue #1580, opened 2019-04-17
Send info to logstash et.al. via event handlers


cornelinux:
**Is your feature request related to a problem? Please describe.**

We want to send information about events in the privacyIDEA system to a logging facility like logstash or...

**Describe the solution you'd like**

There is already an issue for this in #1072 to send the audit log to the logging facility. But there might also be another approach by either adding a new event handler or enhancing the notification handler.

This way we could create more specific logging events, since logging events could be related to the set of conditions.

The log-event-handler could read a python logging.conf file. Using the python logging mechanism could help us to have enough flexibility for either writing to a file or to syslog or to email or wherever.

For this we would need to define a new object "logging facility". One class of logging facilities could be the python logging and would require a logging.conf for configuration.
Other logging facilities might be more specific.

The notification handler then would allow notification via

  * email (email server) / already implemented
  * sms (sms gateway) / already implemented
  * logging (logging facility) / implemented with this issue.

There are already a lot of tags for the "message" in the event handler. The message then would simply be the log entry sent to the logging facility.

(To be discussed for 3.1)

fredreichbier:
See #1072

cornelinux:
@plettich please add the splunk HTTP python logging config as an example.

In the file audit module we also used the python logging facility. We should check there how this was done. (As @fredreichbier already linked in #1072 for the FileAudit ;-) )

cornelinux:
We simply create a notification handler that does one call to ``log.info``.
The logging.cfg then can determine, where the information is put.

plettich:
TODO:
- [ ] tests

PR #2082, opened 2020-03-12
Add a logging event handler

plettich:
The event handler sends a configurable log message to the python logging
facility whenever the event is triggered (and the conditions match).
The log message can then be processed with a suitable logging
configuration.

Working on #1580


plettich:
> Please check the PEP8 comments and also add a documentation, which especially should help with:

Done in ba58c86 and e740f12
> 
>     * [x]  Define the logmessage template

I could also add a link to the usernotification handler docs

>     * [ ]  Configure the log config in the YAML file to e.g. pass this specific message to wherever.

I'll add this when this PR is merged because in this branch i don't have the YAML configuration.
This is also the reason why the travis push tests failed, i forgot to add a requirement which is already available in #2080

cornelinux:
@plettich textfixtures is missing and in runs into an import error.

tests/test_lib_eventhandler_logging.py
```diff
@@ -0,0 +1,161 @@
...
+"""
+This file tests:
+
+lib/eventhandler/logging.py
+"""
+from mock import mock
+from datetime import datetime
+from werkzeug.test import EnvironBuilder
+from flask import Request
+from testfixtures import log_capture
```
cornelinux: The testfixtures seems to be missing.

doc/eventhandler/logginghandler.rst
```diff
@@ -0,0 +1,89 @@
...
+  * ``DEBUG``
+
+**message**
+
+  * *default:* ``"event={action} triggered"``
+
+The message to send to the logging facility. This message can be customized
+with the following tags:
+
+  * ``{admin}``
```
cornelinux: Are these always the same tags as in the notification handler?
Then we could in a new PR/issue move the tags to a document section "event handler tags" and simply add a reference here. 
What do you think?

plettich: Yes, basically the same but we do not want to include PINs or QR-Codes in the log message.

tests/test_lib_eventhandler_logging.py
plettich: Yes, they are already in master from the YAML PR :-)

doc/eventhandler/logginghandler.rst
cornelinux: Hm, so the build in the master will succeed.

tests/test_lib_eventhandler_logging.py
cornelinux: Hm, so the build in the master will succeed.

Issue #2081, opened 2020-03-12
HTTP UserId Resolver support


brunocascio:
**Is your feature request related to a problem? Please describe.**
What are you trying to achieve?

Currenlty, pi supports `sql`, `ldap`, `passwd` and `scim` user's resolvers. A useful case for microservices is retrieving users from an external API. For example, `http://domain.com/users/<userId>`

**Describe the solution you'd like**

- What is the purpose of the resolver

Use third party HTTP API for retrieving user data without follows the SCIM specs.

- How it works

Since PI does not store users, it uses resolvers like LDAP, SCIM, SQL, etc. Today, there is no way to resolve user information through an API but SCIM. SCIM uses an authorization server to authenticate the request, HTTP resolver will not. HTTP resolver could authenticate users via `Authorization` headers instead.

- How it is configured

The user would create an HTTP resolver only adding an HTTP endpoint under `Add httpresolver` UI. The endpoint must contain the '%s' symbol inside, symbol where pi will replace with their userId.

**Describe alternatives you've considered**

Add inversion control in order to be the user able to create custom resolvers instead of modifying pi code directly.

**Additional context**
Add any other context or screenshots, that might help us to better understand your idea, your need and your circumstances.

![image](https://user-images.githubusercontent.com/2397875/76535745-1deb1000-645a-11ea-9c9b-d33548e97b15.png)

cornelinux:
Thanks a lot for your suggestion. 
But please note, that a reference to a pull request is NOT a CONSISE description. So I do not actually know, what you are suggesting!

SCIM actually IS an HTTP API to retrieve users. Please describe in more detail, what you want to do, otherwise we will need to close this issue.
Thanks a lot for your understanding.

brunocascio:
SCIM is for auth servers. My PR is for a simple HTTP API. For example, an external REST user API. `http://domain.com/users/<userId>`

cornelinux:
Sorry. Missing spec or description. We do not do any development without spec.

brunocascio:
The user sources is an http API. How do accomplish that today? I've made a pull request with that if you take a look..

cornelinux:
Didn't I make myself clear?

We do not accept Pull Requests without knowing what the pull request does.

So you need to tell
* What is the purpose of the resolver
* How it works
* How it is configured.

This is also necessary so that conceptual discussions can take place *before* code is generated.
Nice, right! This is why privacyIDEA is such a cool piece of software.
[...]

PR #2083, opened 2020-03-12
Added HTTP Resolver

brunocascio:
Fixes: #2081


privacyidea/static/components/config/views/config.resolvers.httpresolver.html
```diff
@@ -0,0 +1,99 @@
...
+        <div class="text-center" ng-show="showResult">
+            <label class="label" ng-class="{'label-success':result.result,
+                       'label-danger':!result.result}">
+                {{ result.description }}<span class="glyphicon glyphicon-remove" ng-click="showResult=false"></span>
+            </label>
+        </div>
+        <div class="container-fluid">
+            <div class="text-right">
+                <div class="form-group row form-inline">
+                    <input name="testEmail" id="testEmail" class="form-control" ng-model="params.testEmail"
```
plettich, line 82: i think we should not use *email* here, we query the *userid* (whether it is an email or something else).

privacyidea/lib/resolvers/HTTPResolver.py
```diff
@@ -0,0 +1,229 @@
...
+#
+# You should have received a copy of the GNU Affero General Public
+# License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+from .UserIdResolver import UserIdResolver
+import requests
+import logging
+import json
+from urllib.parse import urlencode
```
plettich, line 35: Could You use `from six.moves.urllib.parse import urlencode` here to get this to work with Python 2?

doc/configuration/useridresolvers.rst
```diff
@@ -354,6 +355,43 @@ The available attributes for the ``Attribute mapping`` are:
  * mobile,
  * email.
 
+HTTP resolver
+.............
+
+.. index:: HTTP resolver
+
+HTTP resolver is useful when you need to retrieve user information from an 
+external HTTP API.
```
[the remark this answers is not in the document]
brunocascio: Let me show you a simple diagram:

image

A user login into keycloak with their username and password (e.g. pepe)  as usual. Keycloak knows that the user needs 2fa authentication before proceeding. Now, keycloak calls to privacyidea (through trigger_challenge endpoint) with `user = "pepe"` and other params. Privacyidea does not have the user at the moment, so it uses the HTTP resolver to request the user information needed for sending for example an email with the OTP code.

Is it more clear? :)

doc/configuration/useridresolvers.rst
```diff
@@ -354,6 +355,43 @@ The available attributes for the ``Attribute mapping`` are:
...
+The available attributes for the ``Response mapping`` are:
+
+ * username *(mandatory)*,
+ * givenname,
+ * surname,
+ * phone,
+ * mobile,
+ * email.
+
+Often, APIs are not RESTful (i.e. always returns 200 OK even if it fails). For this you can use ``Special error handling`` input:
```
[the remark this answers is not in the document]
brunocascio: Exaclty! I'll do it! Thanks

privacyidea/lib/resolvers/HTTPResolver.py
```diff
@@ -0,0 +1,229 @@
...
+
+    @classmethod
+    def testconnection(cls, param):
+        """
+        This function lets you test if the parameters can be used to create a
+        working resolver.
+        The implementation should try to connect to the user store and verify
+        if users can be retrieved.
+        In case of success it should return a text like
+        "Resolver config seems OK. 123 Users found."
```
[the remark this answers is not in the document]
brunocascio: Ups! Yes, you're right. Thanks!

privacyidea/lib/resolvers/HTTPResolver.py
```diff
@@ -0,0 +1,229 @@
...
+        :param param: The parameters that should be saved as the resolver
+        :type param: dict
+        :return: returns True in case of success and a descriptive text
+        :rtype: tuple
+        """
+        desc = ""
+        success = False
+        try:
+            response = cls._getUser(param, param.get('testEmail'))
+            desc = response
```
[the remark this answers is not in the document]
brunocascio: Yes, Could be any string so `testEmail` isn't a good name. Got it!

privacyidea/lib/resolvers/HTTPResolver.py
```diff
@@ -0,0 +1,229 @@
...
+            success = True
+        except Exception as e:
+            success = False
+            desc = "failed: {0!s}".format(e)
+        return success, desc
+
+    #
+    #   Private methods
+    #
+    @classmethod
```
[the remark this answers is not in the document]
brunocascio: This is because I used it inside of another class method (testconnection) but I figure out that python can do it anyway. Thanks

privacyidea/lib/resolvers/HTTPResolver.py
```diff
@@ -0,0 +1,229 @@
...
+    def _getUser(cls, param, userid):
+        method = param.get('method').lower()
```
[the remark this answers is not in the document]
brunocascio: I didn't find a clean way to do this based on other resolvers. But I can add some validations, no problem.

privacyidea/lib/resolvers/HTTPResolver.py
```diff
@@ -0,0 +1,229 @@
...
+
+        # Raises HTTPError, if one occurred.
+        httpResponse.raise_for_status()
+
+        jsonHTTPResponse = httpResponse.json()
+
+        if hasSpecialErrorHandler:
+            # verify if error response mapping is a subset of the json http response
+            if errorResponseMapping.items() <= jsonHTTPResponse.items():
+                raise Exception(jsonHTTPResponse)
```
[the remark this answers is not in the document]
brunocascio: Something like this?

```python
logger.error(jsonHTTPResponse)
raise Exception( 'Received an error while searching for user "pepe")')
```

privacyidea/lib/resolvers/HTTPResolver.py
```diff
@@ -0,0 +1,229 @@
...
+    #
+    @classmethod
+    def _getUser(cls, param, userid):
+        method = param.get('method').lower()
+        endpoint = param.get('endpoint')
+        requestMappingJSON = json.loads(param.get('requestMapping').replace("{userid}", userid))
+        responseMapping = json.loads(param.get('responseMapping'))
+        headers = json.loads(param.get('headers', '{}'))
+        hasSpecialErrorHandler = bool(param.get('hasSpecialErrorHandler'))
+        errorResponseMapping = json.loads(param.get('errorResponseMapping', '{}'))
```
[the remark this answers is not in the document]
brunocascio: Good Catch!

privacyidea/lib/resolvers/HTTPResolver.py
```diff
@@ -0,0 +1,229 @@
...
+# You should have received a copy of the GNU Affero General Public
+# License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+from .UserIdResolver import UserIdResolver
+import requests
+import logging
+import json
+from urllib.parse import urlencode
+from pydash import get
```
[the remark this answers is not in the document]
brunocascio: `pydash` really helps when you work with nested objects. When I should document this?

doc/configuration/useridresolvers.rst
```diff
@@ -354,6 +355,75 @@ The available attributes for the ``Attribute mapping`` are:
...
+
+   { "success": true, "message": "Oops, error!" }
+
+You can able to configure something like this:
+
+.. figure:: images/http_resolver_2.png
+   :width: 500
+
+So, now the HTTP Resolver checks if the response contains your input and then throws and error if it's match.
+In this case, An rrror will be thrown due ``{ "success": true }`` is part of the response.
```
plettich, line 423: ... error ...

privacyidea/lib/resolvers/HTTPResolver.py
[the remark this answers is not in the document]
plettich: Just add it to the resolver documentation. Just a note that nested object can be used as well (maybe with a little inline example).

privacyidea/lib/resolvers/HTTPResolver.py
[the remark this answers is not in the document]
plettich: we have the [`getParam()`](https://github.com/privacyidea/privacyidea/blob/5fe428af3c75cd1c0013b244950fd2abd98b950d/privacyidea/api/lib/utils.py#L55) function which can throw an error if a required parameter isn't given.

privacyidea/lib/resolvers/HTTPResolver.py
```diff
@@ -0,0 +1,248 @@
...
+    #
+    def _getUser(self, userid):
+        param = self.config
+        method = param.get('method').lower()
+        endpoint = param.get('endpoint')
+        requestMappingJSON = json.loads(param.get('requestMapping').replace("{userid}", userid))
+        responseMapping = json.loads(param.get('responseMapping'))
+        headers = json.loads(param.get('headers', '{}'))
+        hasSpecialErrorHandler = bool(param.get('hasSpecialErrorHandler'))
+        errorResponse = json.loads(param.get('errorResponse'))
```
plettich, line 222: Now this would throw a `JSONDecodeError` if the parameter is empty. I think we should leave it as before (with a default value of `'{}'`= and check below if `errorResponse` is `True`  (an empty dict evaluates to `False`).

doc/configuration/useridresolvers.rst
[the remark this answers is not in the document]
brunocascio: [This](https://github.com/privacyidea/privacyidea/issues/2081#issuecomment-623025402) could be helpful

privacyidea/lib/resolvers/HTTPResolver.py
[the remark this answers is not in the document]
brunocascio: Ok, I'm going to use it. Also, let me show you something:

image

If I set `allowed_values=(get, post)` and optional as False, `getParam()` should throw an error if I send for example "put". Now, this method puts the default value instead of throwing an exception. Is it expected?

privacyidea/lib/resolvers/HTTPResolver.py
```diff
@@ -0,0 +1,247 @@
...
+            httpResponse = requests.get(endpoint, urlencode(requestMappingJSON), headers=headers)
+
+        # Raises HTTPError, if one occurred.
+        httpResponse.raise_for_status()
+
+        jsonHTTPResponse = httpResponse.json()
+
+        if hasSpecialErrorHandler:
+            # verify if error response mapping is a subset of the json http response
+            if errorResponse.items() <= jsonHTTPResponse.items():
```
plettich, line 235: i think we should use a `set` here for comparison:
`if errorResponse and set(errorResponse.items()) <= set(jsonHTTPResponse.items()):`

plettich: Ok, so this doesn't work if there are embedded dictionaries. I'll look for something else.

privacyidea/lib/resolvers/HTTPResolver.py
```diff
@@ -0,0 +1,247 @@
...
+        "method": 1,
+        "requestMapping": 1,
+        "headers": 1,
+        "responseMapping": 1,
+        "hasSpecialErrorHandler": 0,
+        "errorResponse": 0
+    }
+
+    def __init__(self):
+        super().__init__()
```
plettich, line 58: Could You use `super(HTTPResolver, self).__init__()` here to get this working with Python 2.7?

tests/test_lib_resolver.py
```diff
@@ -2452,3 +2455,353 @@ def test_15_try_to_delete_used_resolver(self):
...
+        r_type = instance.getResolverDescriptor()
+        self.assertTrue("httpresolver" in r_type)
+        r_type = instance.getResolverType()
+        self.assertEqual("httpresolver", r_type)
+
+        # Test with invalid method
+        instance = HTTPResolver()
+        invalidMethodDict = params.copy()
+        invalidMethodDict.update({'method': ''})
+        self.assertRaisesRegex(ParameterError,
```
plettich, line 2532: Please use `assertRaisesRegexp()` to get this working with Python 2.7.

privacyidea/lib/resolvers/HTTPResolver.py
plettich: the following works:
`all([x in jsonHTTPResponse.items() for x in errorResponse.items()])`
but we still need to check if `errorResponse` is empty.

brunocascio: I've changed with your implementation and looks great!

> but we still need to check if errorResponse is empty.

Here's https://github.com/privacyidea/privacyidea/pull/2083/files#diff-7d7a63c6aac770da073091e1bb88a464R177 :)

privacyidea/lib/resolvers/HTTPResolver.py
```diff
@@ -0,0 +1,247 @@
...
+            config,
+            'hasSpecialErrorHandler',
+            optional=False,
+            default=False,
+            allow_empty=False
+        )
+
+        # Validate special error handler
+        if hasSpecialErrorHandler and not config.get('errorResponse'):
+            getParam(config, 'errorResponse', optional=False, allow_empty=False)
```
plettich, line 177: I should have read the code more closely... the checking is all unnecessary since we receive the data from the database and assume that this data is correct.
So a simple `self.config = config` should suffice here.
Sorry for the trouble.

privacyidea/lib/resolvers/HTTPResolver.py
plettich: Since this value can still be empty i would use
`errorResponse = json.loads(param.get('errorResponse', '{}'))`
to avoid the `JSONDecodeError`.

privacyidea/lib/resolvers/HTTPResolver.py
brunocascio: Ok... So validations are not necessary for this resolver or Should I move them to another part?

privacyidea/lib/resolvers/HTTPResolver.py
brunocascio: I didn't add the default value because it's validated at the ` getConfig() step. But if I remove validations according to [this](https://github.com/privacyidea/privacyidea/pull/2083#discussion_r469853298) I definitely have to add it.

privacyidea/lib/resolvers/HTTPResolver.py
plettich: The validations occur when the resolver is created so when we load the configuration from the db we can assume that everything is correct. Simply copying the configuration should be enough here.

privacyidea/lib/resolvers/HTTPResolver.py
plettich: Yes, that would be the idea since this entry is optional, it can be empty in the db (or we check it together with the `hasSpecialErrorHandler` option, but i think, just using the default string here should be enough).

brunocascio:
Thanks, @plettich @cornelinux for accepting my PR :rocket:

cornelinux:
Thanks @brunocascio for your effort and your patience!

PR #2084, opened 2020-03-12
Enckey as argument support

### LINKED ISSUE #2075: "pi-manage create_enckey" command should accept static key by arg
Creates an enckey with a password string set via CLI

`pi-manage create_enckey --enckey=<key>`

PR:  #2084

**Additional context**

I'm trying to deploy pi with docker swarm. In a multi-host environment, there's no easy way to distribute volumes data, so I'm trying to avoid file persistence. I'm thinking about creating the same enckey for all the instances instead of random bytes (default behavior). Is it possible?

brunocascio:
Closes #2075

### DIFF
--- a/pi-manage
+++ b/pi-manage
@@ -93,6 +93,7 @@ from privacyidea.lib.utils import parse_timedelta
 from privacyidea.lib.crypto import create_hsm_object
 import jwt
 import ast
+import base64
 
 SILENT = True
 MYSQL_DIALECTS = ["mysql", "pymysql", "mysql+pymysql"]
@@ -446,9 +447,13 @@ def encrypt_enckey(encfile):
 
 
 @manager.command
-def create_enckey():
+def create_enckey(enckey_b64=None):
     """
-    If the key of the given configuration does not exist, it will be created
+    If the key of the given configuration does not exist, it will be created.
+
+
+    :param enckey_b64: (Optional) base64 encoded plain text key
+    :return:
     """
     print()
     filename = app.config.get("PI_ENCFILE")
@@ -457,8 +462,15 @@ def create_enckey():
               filename)
         sys.exit(1)
     with open(filename, "wb") as f:
-        f.write(DefaultSecurityModule.random(96))
-
+        if enckey_b64 is None:
+            f.write(DefaultSecurityModule.random(96))
+        else:
+            print("Warning: Passing enckey via cli input is considered harmful.")
+            bin_enckey = base64.b64decode(enckey_b64)
+            if len(bin_enckey) != 96:
+                print("Error: enckey must be 96 bytes length")
+                sys.exit(1)
+            f.write(bin_enckey)
     print("Encryption key written to %s" % filename)
     os.chmod(filename, 0o400)
     print("The file permission of %s was set to 400!" % filename)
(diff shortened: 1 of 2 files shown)


pi-manage
```diff
@@ -457,8 +457,14 @@ def create_enckey():
               filename)
         sys.exit(1)
     with open(filename, "wb") as f:
-        f.write(DefaultSecurityModule.random(96))
-
+        if enckey is None:
+            f.write(DefaultSecurityModule.random(96))
+        else:
+            print("Warning: Passing enckey via cli input is considered harmful.")
+            if not len(enckey.encode()) is 96:
```
cornelinux: In which encoding is the user supposed to pass the encryption key?
The encryption key is 96 bytes, so it could be a string of 192 hex values, but it is not clear.

You also should enhance the docstring or use the decorator ``@manager.option`` to give the user a better clue, how the encryption key is to be passed.
Or add some words to the documentation: https://github.com/privacyidea/privacyidea/blob/4929d4e0a7a11cbc564a88d034f7903665ba8024/doc/installation/system/pimanage/index.rst

brunocascio: Hi @cornelinux,

> In which encoding is the user supposed to pass the encryption key?
The encryption key is 96 bytes, so it could be a string of 192 hex values, but it is not clear.

Yes, you're right. Encoding could be [ASCII](https://es.wikipedia.org/wiki/ASCII) in order to avoid ambiguous chars or chars like `ñáéíú....`.

Example `XeH5xCz2RIB54pd4xwLO3XTJf69umi2z6k8mHRoXuOp5Vdzn6raFEEHR02jzTFXs0r0mbeIZCI9FVDdaBAAiWoRyS8!#&%+,`

What do you think?

brunocascio: I've just updated the PR. Any change or recommendation just let me know and I'll do it :+1:

cornelinux: If you allow those characters like "&" or "!" you will get challenges at the command line.
I would probably use 
a) Hex encoding, which is proabably too long:
~~~~    
'c55b31c1c39eeab4d4084378e80e4f28217497c47eb39dd9eb22804a87d29b4d0b6715b036da2fb20c76c18f88bf80479a61b73607d03be8dea22cae2918b55cfb859ec5bcead307fff84014b91237548b9c95bde70837bad0fc175ec9e694a1'
~~~~
b) base32
~~~~
'2Y7Y5VY7XKDSO3WECLVMOBV6WTLIYOMTMZRMPRGF663TBCQCB5VUZZABT3C4C7Z62HCUTQNVCVEMSI2N6WCO4N3GKYNG6UV3KE6JFDTD4XNY6LT3A5QYBCE5JV35EXZZ2DHPIYTDVKLGUV2PHOKIPKPZLE======'
~~~~
WHich is still safe in regards to characters, but always needs some padding
c) base64 might still have problems with the + (no it does not)
~~~~
'QMVSDPSL6d/K4J1xl0mpPY10Oi+YBdnBlm+oTykQimdehe2xFdjW8FDl8KR1ZPr7xfNp2VxvR+/ODC4VEYyFwnGH6iJYzixswA9QB4F5oFtRDb2gUKZDcWcYcHIlCtzE'
~~~~
but it is simpler to escape with upper ticks.

I would go for base64!

cornelinux: @brunocascio Please change this to base64 encoding. I do not see any sense in ascii - only problems. We do not do such encoding at any other place in privacyIDEA.

brunocascio: Done!

PR #2085, opened 2020-03-12
Privacyidea Dockerized

brunocascio:


cornelinux:
We started to move all aspects of packaging privacyIDEA to other repositories.
This way we have solely the code in this repository and dedicated repositories for different way of packaging (deb for Ubuntu, rpm for CentOS)...
I would see any docker specific stuff also in another repository - unless I am getting something completely wrong, here.

As a matter of fact we already have an (extremely old) repository for providing privacyIDEA via docker. 
https://github.com/privacyidea/docker
Please take a look at that and see if you can open an issue and issue a pull request there.

Thank you.

PR #2088, opened 2020-03-12
MSSQL support

brunocascio:
Use example:

```py
SQLALCHEMY_DATABASE_URI = 'mssql+pymssql://user:pass@host/dbname'
```


cornelinux:
We use the requirements.txt to build our common installation setup. 
It is also used to build debian packages and red hat packages.
Of course you could use MSSQL, but the requirements.txt is not ment for containing all possible setups. (You see, it also does not contain e.g. Oracle drivers)

So if you need to connect to MSSQL servers, you need to install the driver manually afterwards.

PR #2112, opened 2020-03-19
Fix the sorting of token table

### LINKED ISSUE #2111: Sorting by column in token table does not work
In version 3.2 sorting in the token table is broken.
The table is sorted up and down but not by the correct column.

The GET /token/ API expects the lowecase parameter "sortby", while for some reason the WebUI tries to pass the parameters sortBy, while the sorting directive also returns the parameter sortby. (lowercase)

cornelinux:
The API expects a lowercase "sortby".
The directive returns a lowercase "sortby".
So lowercasing the parameters.

Fixes #2111

### DIFF
--- a/privacyidea/static/components/token/controllers/tokenControllers.js
+++ b/privacyidea/static/components/token/controllers/tokenControllers.js
@@ -46,7 +46,7 @@ myApp.controller("tokenController", function (TokenFactory, ConfigFactory,
             $scope.params.userid = "*" + ($scope.userIdFilter || "") + "*";
             $scope.params.resolver = "*" + ($scope.resolverFilter || "") + "*";
             $scope.params.pagesize = $scope.token_page_size;
-            $scope.params.sortBy = $scope.sortBy;
+            $scope.params.sortby = $scope.sortby;
             if ($scope.reverse) {
                 $scope.params.sortdir = "desc";
             } else {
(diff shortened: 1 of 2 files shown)


privacyidea/static/components/token/controllers/tokenControllers.js
```diff
@@ -46,7 +46,7 @@ myApp.controller("tokenController", function (TokenFactory, ConfigFactory,
             $scope.params.userid = "*" + ($scope.userIdFilter || "") + "*";
             $scope.params.resolver = "*" + ($scope.resolverFilter || "") + "*";
             $scope.params.pagesize = $scope.token_page_size;
-            $scope.params.sortBy = $scope.sortBy;
```
plettich: What about the ``sortBy`` in
https://github.com/privacyidea/privacyidea/blob/5f6aa31a467ebf01cd25efad397c15ebf4b4c982/privacyidea/static/components/token/controllers/tokenChallengesController.js#L20

cornelinux: This is the endpoint GET /token/challenges

This should also be "sortby" lowercase.
Will do.

cornelinux: Fixed it in 15244ca8
although the token challenge list is probably not that long.
Thanks for noting it!

PR #2126, opened 2020-03-22
Replace stateChangeSuccess by transitions.onBefore and onSuccess

### LINKED ISSUE #2122: After deleting token does not return to user
on the current master branch when deleting a token of a user, the dialog stays on the deleted token and does not return to the user.

Reproduce:

* go on user details
* select one of the users tokens -> token detail
* delete token
(The view should return to the user, but it stays on the deleted token)

### LINKED ISSUE #2125: Reload button not visible.
on the current master branch, the reload button in the upper right corner is not visible.
Although, if some action is done, the spinner at the same location is spinning.

cornelinux:
In the new angular-ui-router the event stateChangeSuccess has
been removed. The changing of a state is now handled by
hooks to transitions
onBefore and onSuccess.
In the onBefore hook we now need to save the *current* state
and parameters, so that we can return to this state, when needed.

Fixes #2122

In the onSuccess hook we if all Listeners have reloaed.

Fixes #2125

### DIFF
--- a/privacyidea/static/components/login/controllers/loginControllers.js
+++ b/privacyidea/static/components/login/controllers/loginControllers.js
@@ -106,10 +106,8 @@ angular.module("privacyideaApp")
         function () {
             // The stateParams or $state.params are always changed as a reference.
             // So we need to do a deep copy, to preserve it between state transitions
-            var oldParams = new Array();
-            for (let [key, value] of Object.entries($state.params)) {
-                oldParams[key] = value;
-            }
+            var oldParams = {};
+            angular.copy($state.params, oldParams);
             $rootScope.previousState = {
                 state: $state.current.name,
                 params: oldParams
--- a/privacyidea/static/components/login/controllers/loginControllers.js
+++ b/privacyidea/static/components/login/controllers/loginControllers.js
@@ -35,7 +35,7 @@ angular.module("privacyideaApp")
                                       PolicyTemplateFactory, gettextCatalog,
                                       hotkeys, RegisterFactory,
                                       U2fFactory, webAuthnToken, instanceUrl,
-                                      PollingAuthFactory,
+                                      PollingAuthFactory, $transitions,
                                       resourceNamePatterns) {
 
     $scope.instanceUrl = instanceUrl;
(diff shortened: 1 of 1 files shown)


privacyidea/static/components/login/controllers/loginControllers.js
```diff
@@ -102,19 +102,24 @@ angular.module("privacyideaApp")
     $scope.myCountdown = "";
     // We save the previous State in the $rootScope, so that we
     // can return there
-    $rootScope.$on('$stateChangeSuccess',
-        function (ev, to, toParams, from, fromParams) {
-            //debug: console.log("we changed the state from " + from + " to " + to);
-            //debug: console.log(from);
-            //debug: console.log(fromParams);
-            //debug: console.log(to);
+    $transitions.onBefore({},
```
plettich: We could limit the call to transitions with target 'token.details' so this doesn't get called on every transition.

privacyidea/static/components/login/controllers/loginControllers.js
```diff
@@ -102,19 +102,24 @@ angular.module("privacyideaApp")
...
+        function () {
+            // The stateParams or $state.params are always changed as a reference.
+            // So we need to do a deep copy, to preserve it between state transitions
+            var oldParams = new Array();
```
plettich: We can use [`angular.copy`](https://docs.angularjs.org/api/ng/function/angular.copy) here

privacyidea/static/components/login/controllers/loginControllers.js
```diff
@@ -102,19 +102,24 @@ angular.module("privacyideaApp")
...
             $rootScope.previousState = {
-                state: from.name,
-                params: fromParams
+                state: $state.current.name,
+                params: oldParams
             };
-
+        });
+    $transitions.onSuccess({},
+        function() {
```
plettich: Since the `checkReloadListeners()` function currently doesn't really do anything, maybe we can get rid of it.

privacyidea/static/components/login/controllers/loginControllers.js
cornelinux: The ``previousState`` is used in one other location. We could also use it in other cases in the future. I want to keep it with no limits, to have the same functionality as before.

privacyidea/static/components/login/controllers/loginControllers.js
cornelinux: This would be nice. Have you checked, if this actually works?
The source is a stateParams object. I am not sure, if it would behave in the same way as the array.

privacyidea/static/components/login/controllers/loginControllers.js
cornelinux: You are probably right. I would like to put this in an extra issue and commit, to make this change easier trackable.
Will be done in #2127

privacyidea/static/components/login/controllers/loginControllers.js
cornelinux: done in 66620dda

privacyidea/static/components/login/controllers/loginControllers.js
plettich: This could be triggered just for `'token.details'` and `'token.challenges'` but i agree, there is probably not that much overhead calling this function on every transition.

cornelinux: So this can be resolved, right?

PR #2131, opened 2020-03-24
Flexible jwt

cornelinux:


privacyidea/api/lib/utils.py
```diff
@@ -322,6 +326,8 @@ def verify_auth_token(auth_token, required_role=None):
         except jwt.ExpiredSignature as err:
             raise AuthError(_("Authentication failure. Your token has expired: {0!s}").format(err),
                             id=ERROR.AUTHENTICATE_TOKEN_EXPIRED)
+    if r == WRONG_USERNAME:
+        raise AuthError(_("Authentication failure. Your username is not allowed."))
```
plettich: Maybe this error is misleading: the user is not allowed to use this jwt
Or am i misunderstanding something?

cornelinux: Look at my improved error message :-)

PR #2145, opened 2020-04-03
Add example build/usage configs for docker

eritikass:
Needed to setup privacyidea lately and did not find any instantly ready to use docker setup for it, so needed to make one myself. This PR is sharing this setup with hope that someone else can also have some use for it.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -32,3 +32,7 @@ privacyidea
 ===========
 This directory contains some config examples to be deployed from setup.py
 into the python package.
+
+docker
+===========
+Contains everything you need to build working privacyidea docker image, with example docker-compose based deploy using it.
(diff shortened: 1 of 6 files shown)


cornelinux:
Thank you for your pull request.
We try to separate the privacyidea code from installation mechanism or configurations.
So we split e.g. Ubuntu build config and CentOS. And in fact we have an old separated docker repo:
https://github.com/privacyidea/docker
Would you mind taking a look a that and trying to add your PR there?
Thanks!

eritikass:
i would have one suggestion about the docker image - instead of having separate repo where you have configs how to build one yourself - would you not want to have instead pre-build one ( in CI for each release), so getting started would be more easy - as more and more things are made now-days with docker so missing docker image in docker-hub i'n my option is BIG minus

cornelinux:
It is not about having a docker image or not.
It is about where to develop the code and the image. Thanks.

Issue #2073, opened 2020-03-10
UI: provide reasonable error messages on token import


laclaro:
Hi,

I came across an "Internal Server Error" message when accessing the "Import Tokens" in the WebUI. The log file contains a ValueError exception since the directory /etc/privacyidea/gpg was not found.
https://gist.github.com/laclaro/4720ecf2a84b7eaa1a141acd0d428bc2

It concerns this line
https://github.com/privacyidea/privacyidea/blob/2dd920e90ab42ed35fc717e29cec9c1e4c6b797a/privacyidea/lib/importotp.py#L550

We should catch this case providing the user with more information.

How to resolve?
* Check if directory is there, if not, print error message
* If directory is there, check if keys are available, if not, print error message.

Best,

Henning

cornelinux:
I don't think so. Actually this part to my knowledge is used by...
...noone.
So currently I see no need in adding more code here.

laclaro:
Most users have a packaged installation of PI, I suspect. If the directory gets created by the package, no reason to take actions at the moment, agreed.

cornelinux:
We could so a simple existance check, only to get this issue closed ;-)

like

~~~~python
import os
if os.path.isdir(self.gnupg_home):
    self.gpg = gnupg.GPG(gnupghome=self.gnupg_home) 
else:
   # bail out
~~~~

~~~~python
[2020-03-10 12:10:24,391][11480][140689530718016][ERROR][privacyidea.app:1891] Exception on /system/gpgkeys [GET]
Traceback (most recent call last):
  File "/home/henning/Dokumente/Dev/privacyidea-server/venv/lib/python3.6/site-packages/flask/app.py", line 2446, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/henning/Dokumente/Dev/privacyidea-server/venv/lib/python3.6/site-packages/flask/app.py", line 1951, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/henning/Dokumente/Dev/privacyidea-server/venv/lib/python3.6/site-packages/flask/app.py", line 1820, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/henning/Dokumente/Dev/privacyidea-server/venv/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
    raise value
[...]

PR #2149, opened 2020-04-07
catch nonexisting gnupg directory

laclaro:
working on #2073

Now the message is not displayed anymore, but the user would assume that the feature would work.

Better would be to additionally disable the import button and display a message instead. But I don't know where to do this.


cornelinux (approved):
@laclaro To me this looks fine.

But I thinks this PR would actually close #2073 and not only work on?

You thoughts at the beginngin of the PR do not belong into a PR but into the issue. 
And: you **must** not disable the import button, since the import button is independent of gpg. 100% of the imports are done without gpg. Disable the button would be a bad idea, breaking an important feature. (This is why discussions belong into the issue, not the PR ;-)

PR #2156, opened 2020-04-14
Update requirements to current versions

plettich:


tests/test_lib_tokens_webauthn.py
```diff
@@ -62,7 +62,7 @@
 import unittest
 from copy import copy
 
-from sqlalchemy.testing.pickleable import User
+from privacyidea.lib.user import User
```
cornelinux, line 65: Huh, what was that?
OK, ``User`` was only used at one place in self.user, so it seemed somehow to run?

plettich: Yeah, i don't know. Maybe some code completion glitch. It broke on the update of sqlalchemy.

PR #2168, opened 2020-04-22
Fix creation of PGP keys

### LINKED ISSUE #2165: pi-manage create_pgp_keys does not create keys
Hi,

we encountered a problem with `pi-manage create_pgp_keys`. It actually does not create the keys.

The gpg home directory was populated correctly during package install but the above command does not add a key pair. Therefore not key is listed in the "Import Tokens" section in the WebUI.
Also the command
`sudo gpg --homedir /etc/privacyidea/gpg --list-keys`
does not list any keys.

The keys can be created manually using the gpg tool. If you use sudo for this, it is necessary to change the ownership of the tty temporarily (for input of the EMPTY password of the key :smile: ).
- sudo chown <user_before_sudo> /dev/pts/0
- sudo gpg --homedir /etc/privacyidea/gpg --gen-key --verbose
- sudo chown root /dev/pts/0
- sudo gpg --homedir /etc/privacyidea/gpg --list-keys

Result: key ist displayed in the WebUI. Exporting the created key with gpg, encrypting a the token-csv file and importing it in the WebUI works flawlessly.

Best,

Henning

cornelinux:
Newer versions of GnuPG always ask for passphrase.
Suppress passphrase using %no-protection.

Fix #2165

### DIFF
--- a/pi-manage
+++ b/pi-manage
@@ -493,6 +493,14 @@ def create_pgp_keys(keysize=2048, force=False):
     input_data = gpg.gen_key_input(key_type="RSA", key_length=keysize,
                                    name_real="privacyIDEA Server",
                                    name_comment="Import")
+    inputs = input_data.split("\n")
+    if inputs[-2] == "%commit":
+        del(inputs[-1])
+        del(inputs[-1])
+        inputs.append("%no-protection")
+        inputs.append("%commit")
+        inputs.append("")
+        input_data = "\n".join(inputs)
     gpg.gen_key(input_data)
 
 


plettich:
Does this work with GnuPG in ubuntu 16.04 as well? Or did they upgrade GnuPG there as well?

cornelinux:
I dont know if it works on 16.04. But I hope no one will install it on 16.04.
And it is only relevant for creating keys. 
So I personally do not care about 16.04 in this regard.

PR #2172, opened 2020-04-27
catch missing "priority"

### LINKED ISSUE #2171: Catch realm "priority" error in WebUI
Hi!

A missing priority value is not caught by the WebUI, when editing an existing realm (e.g. to add another resolver).

Best,

Henning

laclaro:
When no priority is given this change will result in a multi-resolver realm with no dedicated priorities.

If this causes no further problem, this PR closes #2171

### DIFF
--- a/privacyidea/api/lib/utils.py
+++ b/privacyidea/api/lib/utils.py
@@ -267,11 +267,8 @@ def get_priority_from_param(param):
     """
     priority = {}
     for k, v in param.items():
-        if k.startswith("priority."):
-            try:
-                priority[k[len("priority."):]] = int(v)
-            except TypeError:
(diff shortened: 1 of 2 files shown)


privacyidea/api/lib/utils.py
```diff
@@ -268,7 +268,10 @@ def get_priority_from_param(param):
     priority = {}
     for k, v in param.items():
         if k.startswith("priority."):
-            priority[k[len("priority."):]] = int(v)
+            try:
```
cornelinux, line 271: you could simply use ``isinstance()`` in the if statement and check, if "v" is an int.

Please also add a short test in test_api_lib_utils.py

PR #2180, opened 2020-04-29
Add script to sync users from ownCloud

cornelinux:
This script syncs users from **two** tables oc_users
and oc_accounts from an ownCloud installation to on table.
This way this *one* table can be used in privacyIDEA to
be read in an SQLResolver.


cornelinux:
@plettich I'd like to show you the idea of the script and how it works.

PR #2184, opened 2020-04-30
2181/fix helpdesk audit

cornelinux:
Added an automatic unit test, which failed previously to this fix.

Please also do some manuel tests.
...also for the normal user (which worked for me)


cornelinux:
@plettich @laclaro I also pushed a version 3.3.2dev1 to the community repo for 18.04 to test it manually.

privacyidea/api/lib/prepolicy.py
```diff
@@ -969,7 +969,9 @@ def check_base_action(request=None, action=None, anonymous=False):
     (role, username, realm, adminuser, adminrealm) = determine_logged_in_userparams(g.logged_in_user, params)
 
     # In certain cases we can not resolve the user by the serial!
-    if action not in [ACTION.AUDIT]:
+    if action is ACTION.AUDIT:
```
plettich, line 972: So if i understand this correctly we have to exclude the Audit action here since the user/realm/resolver parameter in the request represent the query for the audit log and not the actual logged in user?
We should explain this somehow for future code reviews :-)

privacyidea/api/lib/prepolicy.py
```diff
@@ -969,7 +969,9 @@ def check_base_action(request=None, action=None, anonymous=False):
...
+        realm = username = resolver = None
+    else:
```
laclaro, line 974: okay, determine_logged_in_userparams function does the checking.

privacyidea/api/lib/prepolicy.py
cornelinux: So you mean we simply should add a bit more comments in the code?

cornelinux: @plettich added more comments

PR #2186, opened 2020-05-01
Add counter condition

### LINKED ISSUE #2147: Event condition counter
The Counter Event handler can count certain occurrances of events and write the counter to the database table ``Eventcounter``.

We could have a new condition for all event handlers called ``eventcounter``.

This condition would check, if the event counter has reached a certain value, is bigger or lower than the given value. This could be done like this:

    countername > 12

Where *countername* would be the keyname of a defined counter. The equal, bigger, smaller comparison would be done like with ``count_auth``...

This condition sounds a bit wired but could open up unexpected possibilities, since we can count any event.
Wiered Example: If the admin hat viewed the audit log for 100 times, then we could define a RequestMangler, that another used would not be able to login anymore.

cornelinux:
In the event handlers the value of a countername
can be compared with "==", ">", "<".

To do so, we add a new function "compare_generic_condition",
since the same mechanism is used for countername and for tokeninfo.

Closes #2147


privacyidea/lib/eventhandler/base.py
```diff
@@ -447,6 +456,15 @@ def check_condition(self, options):
             if not bool(m):
                 return False
 
+        if CONDITION.COUNTERNAME in conditions:
+            # Can be countername==1000
+            if not compare_generic_condition(conditions.get(CONDITION.COUNTERNAME),
+                                             counter_read,
```
plettich, line 462: We should check if the counter exists, otherwise we will get an error when comparing `None` with some value.

cornelinux: If the counter does not exist, the condition would not be true.
However, the request should not fail with an exception.

cornelinux: According to the new commit 8ce52d54 it works as expected.

cornelinux: Darn. I am wrong in regards to Python 3.

tests/test_lib_utils.py
```diff
@@ -849,4 +850,35 @@ def test_33_determine_logged_in_user(self):
...
+        self.assertTrue(compare_generic_condition("a<100",
+                                                  mock_attribute,
+                                                  "Error {0!s}"))
+
+        self.assertTrue(compare_generic_condition("b==100",
+                                                  mock_attribute,
+                                                  "Error {0!s}"))
+
+        # Unknown comparator
+        self.assertFalse(compare_generic_condition("b>=100",
```
plettich, line 872: This does not work since it will be split at the '>' and thus tries to evaluate **`100`** `>` **`=100`** 
In this case we won't even get a warning since the `else` block is never reached.
Using **`!=`** works as a test but still leaves us with the missing warning and unexpected behavior in case of `>=`, `<=`, `=>` or `=<` (This will then compare strings like '1000' > '=100').

privacyidea/lib/utils/__init__.py
```diff
@@ -888,6 +888,34 @@ def compare_value_value(value1, comparator, value2):
...
+
+    If the value does not match, it returns False.
+
+    :param cond: A condition containing a comparator like "==", ">", "<"
+    :param key_method: A function call, that get the value from the key
+    :param warning: A warnnig message to be written to the log file.
+    :return: True of False
+    """
+    key = value = None
+    for comparator in ["==", ">", "<", None]:
```
plettich, line 904: Using the comparator `None` will lead to an error `compare_value_value()` when the condition is something like "tokeninfoattribute value" (Who knows what users put in the inputs).

tests/test_lib_utils.py
cornelinux: You are right. The test is wrong. Also the documentation states, that the values "==", ">" and "<" can be used.
I will fix the text.

privacyidea/lib/utils/__init__.py
cornelinux: I needed "None" as breakpoint. If you have a better suggestion, shoot! :-)

tests/test_lib_utils.py
cornelinux: So the test accidentially checks the correct result. I will change the comparator to a real unknown comparator.

Hm, you migh also want to compare the string-value of attribute "b" and check, if it smaller than the string "=100", would you? ;-)

privacyidea/lib/utils/__init__.py
cornelinux: I actually think, this might be a quite interesting result.
The problem is, what would s.o. expect if he enters a wrong condition. Should it be false or true?
So raising an exception might even be nicer, since the admin actually realizes the problem.
I can totally live with that.

tests/test_lib_utils.py
cornelinux: done in d7ace787

privacyidea/lib/utils/__init__.py
cornelinux: done in 848e1a32

laclaro: I would agree, an exception hints at a wrong condition.
I thought first the support of <= or >= would be nice, but in principle nobody would actually need those here. Just sharing the thought.

cornelinux:
> I would rename the condition "countername" to "countervalue" since the actual condition checks the value, not the name.

Interesting point. Countervalue isn't the right naming either. Compared to the long existing "tokeninfo" condition, the correct naming for this condition should be "counter".

doc/eventhandler/index.rst
```diff
@@ -339,7 +339,16 @@ This can be used to e.g. automatically enroll a token for the user if the
...
+**counter**
 
+The counter condition can compare the value of any arbitrary event counter against a fixed
+value. Valid compares are:
+
+    myCounter == 1000
+    myCounter > 1000
+    myCounter < 1000
+
+"myCounter" being any event counter set with the :ref:`counterhandler`.
```
plettich, line 351: Could You add some documentation about setting the counter value to `0` if the counter does not (yet) exists?

cornelinux: Done in 4685a9f0

PR #2187, opened 2020-05-04
2121/define characters set random pin

### LINKED ISSUE #2121: define characters that can be used in set_random_pin
The policy ``otp_set_random_pin`` can be used to set a random PIN of a specific length.
But the contents, the used characters, of this random PIN can not be configured.

We could also use the policy ``otp_pin_contents``.

laclaro:
fixes #2121

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -80,7 +80,8 @@ from privacyidea.lib.token import (get_tokens, get_realms_of_token, get_token_ty
 from privacyidea.lib.utils import (get_client_ip,
                                    parse_timedelta, is_true, check_pin_policy, get_module_class,
                                    determine_logged_in_userparams)
-from privacyidea.lib.crypto import generate_password
+from privacyidea.lib.crypto import (generate_password,
+                                    generate_password_with_requirements)
 from privacyidea.lib.auth import ROLE
 from privacyidea.api.lib.utils import getParam, attestation_certificate_allowed, is_fqdn
 from privacyidea.lib.clientapplication import save_clientapplication
@@ -153,6 +154,55 @@ class prepolicy(object):
         return policy_wrapper
 
 
+def generate_pin_from_policy(policy,size=6):
+    """
+    This function creates a string of allowed characters from the value of a pincontents policy
+    The policy to check a PIN can contain of "c", "n" and "s".
+    "cn" means, that the PIN should contain a character and a number.
+    "+cn" means, that the PIN should contain elements from the group of characters and numbers
+    "-ns" means, that the PIN must not contain numbers or special characters
+    "[12345]" means, that the PIN may only consist of the characters 1,2,3,4 and 5.
+
+    :param policy: The policy that describes the allowed contents of the PIN.
+    :param size: The desired length of the generated pin
+    :return: Tuple of the generated PIN and a description
+    """
+
+    chars = {"c": string.ascii_letters,         # characters
+             "n": string.digits,                # numbers
+             "s": "\.:,;_<>+*!/()=?$§%&#~^-"}   # special
+
+    # default: full character list
+    default_characters = "".join(chars.values())
+
+    not_allowed = []
+    required = []
+    if policy[0] in ["+", "-"] or policy[0] is not "[":
+        for char in policy[1:]:
+            if char not in chars.keys():
+                raise PolicyError("Unknown character specifier in PIN policy.")
+
+    if policy[0] == "+":
+        # grouping
+        required = []
+        for char in policy[1:]:
+            required.append(chars.get(char))
+
+    elif policy[0] == "-":
+        # exclusion
+        for char in policy[1:]:
+            not_allowed.append(chars.get(char))
+
+    elif policy[0] == "[" and policy[-1] == "]":
+        # only allowed characters
+        default_characters = policy[1:-1]
+
+    ret = generate_password_with_requirements(size=size,
+                                        characters=default_characters,
+                                        exclude="".join(not_allowed),requirements=required)
+    return ret
+
+
 def set_random_pin(request=None, action=None):
     """
     This policy function is to be used as a decorator in the API setrandompin function
@@ -182,8 +232,23 @@ def set_random_pin(request=None, action=None):
         raise TokenAdminError("You need to specify a policy '{0!s}' in scope "
                               "{1!s}.".format(ACTION.OTPPINSETRANDOM, role))
     elif len(pin_pols) == 1:
-        log.debug("Creating random OTP PIN with length {0!s}".format(list(pin_pols)[0]))
-        request.all_data["pin"] = generate_password(size=int(list(pin_pols)[0]))
+        # check pin contents policy per token type, otherwise fall back
+        tokentype = request.all_data.get("type", "hotp")
+        pol_contents = Match.admin_or_user(g, action="{0!s}_{1!s}".format(tokentype, ACTION.OTPPINCONTENTS),
+                                           user_obj=request.User).action_values(unique=True)
+        if not pol_contents:
+            pol_contents = Match.admin_or_user(g, action=ACTION.OTPPINCONTENTS,
+                                               user_obj=request.User).action_values(unique=True)
+
+        if len(pol_contents) == 1:
+            log.info("Creating random OTP PIN with length {0!s} "
+                      "matching the contents policy {1!s}".format(list(pin_pols)[0],list(pol_contents)[0]))
+            # generate a pin which matches the contents requirement
+            r = generate_pin_from_policy(list(pol_contents)[0],int(list(pin_pols)[0]))
+            request.all_data["pin"] = r
+        else:
+            log.debug("Creating random OTP PIN with length {0!s}".format(list(pin_pols)[0]))
+            request.all_data["pin"] = generate_password(size=int(list(pin_pols)[0]))
 
     return True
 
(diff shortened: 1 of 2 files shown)


privacyidea/api/lib/prepolicy.py
```diff
@@ -182,8 +230,23 @@ def set_random_pin(request=None, action=None):
         raise TokenAdminError("You need to specify a policy '{0!s}' in scope "
                               "{1!s}.".format(ACTION.OTPPINSETRANDOM, role))
     elif len(pin_pols) == 1:
-        log.debug("Creating random OTP PIN with length {0!s}".format(list(pin_pols)[0]))
-        request.all_data["pin"] = generate_password(size=int(list(pin_pols)[0]))
+        # check pin contents policy per token type, otherwise fall back
+        tokentype = request.all_data.get("type", "hotp")
```
cornelinux, line 234: The "set_random_pin" request can be called by a button in the UI as a special request for one token. I think there is token tokentype in this request. So we can get it this way:

~~~~python
tokentype = get_token_type(request.all_data.get("serial"))
~~~~

get_token_type is imported from privacyidea.lib.token

privacyidea/api/lib/prepolicy.py
```diff
@@ -153,6 +153,54 @@ def policy_wrapper(*args, **kwds):
         return policy_wrapper
 
 
+def generate_pin_from_policy(policy, size=6):
```
cornelinux, line 156: As far as I understand, this is a module specific function which is not used as a decorater, but only within this module.
So please rename it to ``_generate_pin_from_policy``.

privacyidea/lib/crypto.py
```diff
@@ -811,16 +811,27 @@ def generate_otpkey(key_size=20):
...
     :param size: The length of the password
     :param characters: The characters the password may consist of
+    :param requirements: A list of strings from which the password must contain at least one character
+    :param exclude: A string of characters to exclude
     :return: password
     :rtype: basestring
     """
-    return ''.join(urandom.choice(characters) for _x in range(size))
+    # add one character from each string in the requirements list
+    passwd = [urandom.choice(str) for str in requirements]
```
cornelinux, line 827: Please add a few tests for the method ``generate_password`` in test_lib_crypto.py, since this is rather unintuitive to me.

privacyidea/api/lib/prepolicy.py
cornelinux: Please add some tests for this method, to see, if pins are generated in the correct way.

privacyidea/api/lib/prepolicy.py
laclaro: done

privacyidea/lib/crypto.py
laclaro: added some tests

privacyidea/api/lib/prepolicy.py
laclaro: tests added

privacyidea/lib/utils/__init__.py
```diff
@@ -1123,13 +1126,17 @@ def check_pin_policy(pin, policy):
...
              "n": r"[0-9]",
-             "s": r"[\[\].:,;_<>+*!/()=?$§%&#~^-]"}
+             "s": SPECIAL_CHARS_REGXEP}
     ret = True
     comment = []
 
     if not policy:
         return False, "No policy given."
 
+    # this is neccesary since a policy "cns" equals "+cns"
```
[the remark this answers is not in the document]
laclaro: done

privacyidea/api/lib/prepolicy.py
```diff
@@ -153,6 +153,58 @@ def policy_wrapper(*args, **kwds):
...
+    chars = {"c": string.ascii_letters,         # characters
+             "n": string.digits,                # numbers
+             "s": SPECIAL_CHARS_LIST}   # special
+
+    # default: full character list
+    default_characters = "".join(chars.values())
+
+    not_allowed = []
+    required = []
+    # this is neccesary since a policy "cns" equals "+cns"
```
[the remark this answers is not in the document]
laclaro: done

cornelinux:
@laclaro can you please check the failing tests?

privacyidea/lib/utils/__init__.py
```diff
@@ -58,6 +58,8 @@
 
 ALLOWED_SERIAL = "^[0-9a-zA-Z\-_]+$"
 
+SPECIAL_CHARS_LIST = string.punctuation
+SPECIAL_CHARS_REGXEP = r"[" + SPECIAL_CHARS_LIST + "]"
```
laclaro, line 62: one reason for the failing tests was a mismatch in generation and checking of the pin.

privacyidea/lib/utils/__init__.py
```diff
@@ -1137,13 +1145,10 @@ def check_pin_policy(pin, policy):
 
     if policy[0] == "+":
         # grouping
-        necessary = []
         for char in policy[1:]:
-            necessary.append(chars.get(char))
-        necessary = "|".join(necessary)
```
laclaro, line 1143: the OR connection was not correct here, since we want to have all the string-lists included in the password.

tests/test_api_lib_policy.py
```diff
@@ -676,13 +682,21 @@ def test_07_set_random_pin(self):
         # request, that matches the policy
         req.all_data = {
                         "user": "cornelius",
-                        "realm": "home"}
+                        "realm": "realm1"}
+
         init_random_pin(req)
+        pin = req.all_data.get("pin")
```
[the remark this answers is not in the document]
laclaro: done

privacyidea/lib/utils/__init__.py
cornelinux: The SPECICAL_CHARS_LIST is not the same anymore, E.g. "§" is missing.

Also I think SPECIAL_CHARS_REGEXP is now wrong. The brackets [ and ] need to be escaped. Otherwise they will work as a group.

privacyidea/lib/utils/__init__.py
```diff
@@ -1123,27 +1125,30 @@ def check_pin_policy(pin, policy):
...
     ret = True
     comment = []
 
     if not policy:
         return False, "No policy given."
+    if not pin:
+        return False, "No pin given."
 
-    if policy[0] in ["+", "-"] or policy[0] is not "[":
+    # this is necessary since a policy "cns" equals "+cns"
```
cornelinux, line 1137: This is not true.

cns means it needs to contain a character, a number and a special character.

+cns means it needs to contain something from the group of characters, numbers  and special characters. 

Why was it necessary to touch the check_pin_policy in the first place.

privacyidea/lib/utils/__init__.py
laclaro: According to the documentation, this change is wrong. I'll revert it.

privacyidea/lib/utils/__init__.py
laclaro: Maybe not. I misread the documentation in the sense of grouping (the imnportant example comes after the line break at https://privacyidea.readthedocs.io/en/latest/policies/admin.html#otp-pin-contents). I'll check.

privacyidea/lib/utils/__init__.py
laclaro: okay, I reverted it. However, I removed the square brackets, since they seem not to behave the same for string and regexp. We may discuss this.
It is crucially important that SPECIAL_CHARS_LIST and SPECIAL_CHARS_REGXEP are the same, since one is for generation and one for checking of the generated pins.

privacyidea/lib/utils/__init__.py
laclaro: done

privacyidea/lib/utils/__init__.py
laclaro: https://github.com/privacyidea/privacyidea/pull/2187/commits/40740a7eb1610400b44c7e0b811ee9944101ab67 should resolve the errors.

privacyidea/lib/utils/__init__.py
laclaro: because of this, test test_20_pin_policy atm fails at https://github.com/privacyidea/privacyidea/blob/2446d9a6eaf0f68bff483791912802a0018c463e/tests/test_lib_utils.py#L520

laclaro: added a base string identifier r"[[[". Now those work again. Python interprets two square brackets as one string and the third one is left, causing trouble.

privacyidea/lib/utils/__init__.py
```diff
@@ -1135,6 +1139,51 @@ def truncate_comma_list(data, max_len):
     return ",".join(data)
 
 
+def generate_charlists_from_pin_policy(policy):
+    """
+    This function uses the pin content policy string to create the character lists
```
cornelinux, line 1144: Please add an example, so that it is clear, that this is the policy action.

   ....policy string (like "+cns") ...

privacyidea/lib/utils/__init__.py
```diff
@@ -1135,6 +1139,51 @@ def truncate_comma_list(data, max_len):
...
+    for password generation.
+
+    :param policy: The policy that describes the allowed contents of the PIN (see check_pin_policy)
+    :return: Dictionary with keys "base" for the base set of allowed characters and "requirements"
+     which denotes a list of characters from each of which at least one must be contained in the pin.
+    """
+
+    valid_policy_regexp = re.compile(r'^[+-]*[cns]+$|^\[.*\]+$')
```
cornelinux, line 1152: As this is a fixed value, please capitalize

privacyidea/lib/utils/__init__.py
```diff
@@ -1147,52 +1196,29 @@ def check_pin_policy(pin, policy):
...
-    chars = {"c": r"[a-zA-Z]",
-             "n": r"[0-9]",
-             "s": r"[\[\].:,;_<>+*!/()=?$§%&#~^-]"}
+
     ret = True
     comment = []
 
     if not policy:
         return False, "No policy given."
+    if not pin:
```
cornelinux, line 1205: Please remove this again, since it changes the logic.

privacyidea/lib/utils/__init__.py
laclaro: done

privacyidea/lib/utils/__init__.py
laclaro: done

privacyidea/lib/utils/__init__.py
laclaro: done

PR #2189, opened 2020-05-04
Update CentOS docs

plettich:
Update docs for CentOS installation to include the firewall configuration.

### DIFF
--- a/doc/installation/centos.rst
+++ b/doc/installation/centos.rst
@@ -76,13 +76,16 @@ and install/update some prerequisites::
 
 If this should be a pinned installation (that is the environment we use to build and test),
 we need to install some pinned dependencies first. They should match the version of the targeted
-privacyIDEA. You can get the latest version tag from github::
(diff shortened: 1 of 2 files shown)


doc/installation/centos.rst
```diff
@@ -66,19 +66,23 @@ Create the virtual environment::
...
 
     $ . /opt/privacyidea/bin/activate
 
+and install/update some prerequisites::
+
+    (privacyidea)$ pip install -U pip setuptools
+
 If this should be a pinned installation (that is the environment we use to build and test),
 we need to install some pinned dependencies first. They should match the version of the targeted
 privacyIDEA::
```
cornelinux, line 79: We should add some always valid text like, "please check for the latest version of privacyIDEA" (where ot check!)
If it is "3.3" you need to issue the following commands. (Or use a tag $version).

plettich: Done in 1354eaf

PR #2194, opened 2020-05-07
Fix failing chal-resp login to webUI

### LINKED ISSUE #2192: WebUI does not send username in totp challenge response
### Top-level intent

I would like to login to privacyidea with totp token over WebUI in challenge-response way.

### Steps to reproduce

1. Enable totp `challenge_response` in an authentication policy
2. Entroll a totp for a regular user
3. Open a web-broser with PI and enter username/password of the user
4. Enter totp value

### Expected outcome

WebUI should send a POST message to API after 4th step in a following fashion:
```
curl 'https://privacyidea/auth' \
  -H 'Accept: application/json, text/plain, */*' \
  -H 'Content-Type: application/json;charset=utf-8' \
  --data '{"username":"user","password":"123456","transaction_id":"01234567890123456789"}'
```
### Actual outcome

It actually sends request with blank username and therefore API returns 401:
```
curl 'https://privacyidea/auth' \
  -H 'Accept: application/json, text/plain, */*' \
  -H 'Content-Type: application/json;charset=utf-8' \
  --data '{"username":"","password":"123456","transaction_id":"01234567890123456789"}'
```

### Configuration

* **privacyIDEA version**:  3.3.2

### Log file

N/A

plettich:
While working on #830 the AngularJS promise-API was changed.
The `$http.post` request in the authenticate function had a promise to delete
the user data (username and password) from the scope.
While updating to the new promise-API, this was changed to a `finally`
construct which was called for every request (regardless if there was an
error or success). This way the username was not available in the following
view.
The `finally` function is changed to only remove the password from the
login data in the scope.

Fixes #2192

### DIFF
--- a/privacyidea/static/components/login/controllers/loginControllers.js
+++ b/privacyidea/static/components/login/controllers/loginControllers.js
@@ -198,8 +198,12 @@ angular.module("privacyideaApp")
         }, {
             withCredentials: true
         }).then(function (response) {
+            // successful authentication
             $scope.do_login_stuff(response.data);
+            // login data is not needed anymore, remove from scope
+            $scope.login = {username: "", password: ""};
         }, function (response) {
+            // failed auth request (may be challenge-response)
             //debug: console.log("challenge response");
             //debug: console.log(error);
             let error = response.data;
@@ -290,7 +294,6 @@ angular.module("privacyideaApp")
                     inform.add(gettextCatalog.getString("Challenge Response " +
                             "Authentication. Your response was not valid!"),
                         {type: "warning", ttl: 5000});
-                    $scope.login.password = "";
                     // in case of U2F we try for a 2nd signature
                     // In case of u2f we do:
                     if ($scope.u2f_first_error) {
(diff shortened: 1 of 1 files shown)


privacyidea/static/components/login/controllers/loginControllers.js
```diff
@@ -322,9 +322,9 @@ angular.module("privacyideaApp")
                 }
             }
         }).finally(function () {
-            // We delete the login object, so that the password is not
+            // We delete the password from the login object, so that it is not
             // contained in the scope
-            $scope.login = {username: "", password: ""};
+            $scope.login.password = "";
```
cornelinux, line 327: Please check, if we can do:

~~~~javascript
if ($scope.transactionid) {
   $scope.login.password = "";
} else {
   $scope.login = {username: "", password: ""};
}
~~~~

This way we would drop the username for invalid auth and keep the username only in case we do a challenge response.

plettich: The old behavior was, that only if the authentication succeeded, the login-data was removed.
If the auth request failed (wrong pw or challenge-response) the password was removed immediately (and if it wasn't challenge-response the transaction id was removed later) and only the username was kept.
So my idea is, we keep the error handler as is and just remove the login data in the success-handler.

plettich: Done in f7671b66
Tested with several chal-resp tokens (U2F, WebAuthN, Email, HOTP).

PR #2216, opened 2020-05-22
Allow to set an empty PIN when enrolling a new token

FedeOmoto:
Allow to set an empty PIN when enrolling a new token.


cornelinux:
Thanks a lot for the PR. We require an issue for each Pull Request, so that we know what we need to review against.
Please specify in which case you experience what kind of problem.

FedeOmoto:
Fixes #2218

cornelinux:
I am sorry for being so blind. Totally missed that.

cornelinux:
@plettich to me this makes sense (see issue) and is totally fine.

PR #2221, opened 2020-05-25
Fix IdResolver.getUserInfo

### LINKED ISSUE #2219: Error "Could not get the userinformation" in log file
### Top-level intent

What did you try to achieve?
WEBUI is not displaying **user** column information in `Tokens -> All tokens`.
Also, the WEBUI is not displaying the **Username** in `Token details for OATH**** -> Assigned User`.

### Steps to reproduce

1. In the WEBUI go to `Tokens -> All tokens`, no information is displayed at the **user** column.
2. In the WEBUI go to `Token details for OATH**** -> Assigned User`, no information is displayed in **Username**.

### Expected outcome

WEBUI should display the user information.

### Actual outcome

What did actually happen?
An error is logged in the log file:
```
[2020-05-25 18:43:31,475][74911][140538455636608][ERROR][privacyidea.lib.resolvers.SQLIdResolver:353] Could not get the userinformation: DatabaseError('(cx_Oracle.DatabaseError) ORA-00906: missing left parenthesis',)
```

### Configuration

* **privacyIDEA version**: 3.3.1
* **Installation method**: github, PyPI, RHEL package dependencies
* **Python version**: 3.6.9
* **Webserver**: NGINX Unit 1.17.0
* **Token database**: Oracle 19c (using SQLAlchemy cx_oracle dialect)

### Log file

Log files attached.
[all_tokens.log](https://github.com/privacyidea/privacyidea/files/4679035/all_tokens.log)
[token_details.log](https://github.com/privacyidea/privacyidea/files/4679036/token_details.log)

FedeOmoto:
When using Oracle, there's a bug in the getUserInfo method (_get_userid_filter helper) as it tries to cast the userid column (VARCHAR2 in our case) to VARCHAR2.

Fixes #2219

### DIFF
--- a/privacyidea/lib/resolvers/SQLIdResolver.py
+++ b/privacyidea/lib/resolvers/SQLIdResolver.py
@@ -356,12 +356,14 @@ class IdResolver (UserIdResolver):
     
     def _get_userid_filter(self, userId):
         column = getattr(self.TABLE, self.map.get("userid"))
-        if isinstance(column.type, Integer):
+        if isinstance(column.type, String):
+            return column == str(userId)
+        elif isinstance(column.type, Integer):
             # since our user ID is usually a string we need to cast
             return column == int(userId)
-        else:
-            # otherwise we cast the column to string (in case of postgres UUIDs)
-            return cast(column, String).like(userId)
+
+        # otherwise we cast the column to string (in case of postgres UUIDs)
+        return cast(column, String).like(userId)
 
     def getUsername(self, userId):
         """


privacyidea/lib/resolvers/SQLIdResolver.py
```diff
@@ -356,12 +356,14 @@ def getUserInfo(self, userId):
...
+            return column == str(userId)
+        elif isinstance(column.type, Integer):
             # since our user ID is usually a string we need to cast
             return column == int(userId)
-        else:
-            # otherwise we cast the column to string (in case of postgres UUIDs)
-            return cast(column, String).like(userId)
+
+        # otherwise we cast the column to string (in case of postgres UUIDs)
+        return cast(column, String).like(userId)
```
plettich, line 366: Basically this `cast` does the same as the added `if` part, but if i understand this correctly, Oracle expects a size limit of the String when casting (Something like `String(40)`).
Because this is not given, the Oracle db complains about missing parentheses.
We could add something like this here:
```
            try:
                return cast(column, String(column.type.length)).like(userId)
            except AttributeError as _e:
                return cast(column, String(40)).like(userId)
```
which would mitigate the error and we don't have the double String type handling here.

plettich: So this is only relevant when the id type in the database is a string.
Basically it looks fine and works with postgres as well. Casting with a given string size might get us into different trouble...

Issue #2047, opened 2020-02-19
tokentype in /validate/check and triggerchallenge


cornelinux:
Sometimes an application wants to authenticate the user by a certain tokentype.
Or the application wants the user to choose, which tokentype he uses for authentication. (Please use "Push"-Token and not "SMS"-Token).

While ``/validate/check`` and ``/validate/triggerchallenge`` allow an authentication either based on the user (use *all* his tokens) or on a specific serial number (use only *one* of the user's tokens) we need the authentication to also work on a specific token type. 
This can currently be configured in policies. But then the policy would always require a specific tokentype from a certain IP (have the server choose).

This time, the application should be able to choose the tokentype. And it could be tokentype A in one auth request and tokentype B in another auth request.

# Specification

## Parameters

Add the parameter ``type`` to the validate endpoints ``check``, ``samlcheck`` and ``triggerchallenge``.

Then only tokens of this very type will be searched and included in the tokenlist to be checked.
This happens here:

Either add "tokentype" to the method or add "type" in the options.https://github.com/privacyidea/privacyidea/blob/d536b52b5e25ef2e4225b2bca59b150f367300bc/privacyidea/api/validate.py#L337

Then the "type" is passed to the method ``get_tokens``.
https://github.com/privacyidea/privacyidea/blob/d536b52b5e25ef2e4225b2bca59b150f367300bc/privacyidea/lib/token.py#L1988

## Policies

The problem with this use case arises like this:

A user could use his HOTP token to try to authenticate. User enters HOTP value. 
But the application decides to only search for TOTP tokens. The authentication will fail.
Since the ``get_tokens`` only compiles a list of TOTP tokens, the HOTP otp value is not invalidated!
An attacker could use the HOTP value at another place!

However, since this is also supposed to work for challenge response tokens, we can not get the list of all users tokens. Since then the challenges might actually be triggered (see above - only use "Push" tokens and no "SMS" tokens would still result in an SMS being sent.)

Thus the use of a tokentype parameter introduces a possible attack vector we should make this configurable via an authentication policy:

* scope=authz
* action=application_tokentype:bool
* desc=Allow application to choose the valid tokentype.

cornelinux:
Closes #2028

h3ge:
> This can currently be configured in policies. But then the policy would always require a specific tokentype from a certain IP (have the server choose).

[...]

PR #2229, opened 2020-05-30
Filter used tokens based on authorization tokentype policy

h3ge:
Could save you some time for #2047


cornelinux:
Thanks you for your pull request. But we will most probably close it.
We do not do policy stuff in the API level. Policies are done in the decorators of the API endpoints like here:
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/api/validate.py#L200
I have not looked into #2047, yet. I would like to keep discussions there in the issue.

cornelinux:
The pull request #2230 covers the requirements in #2047.
I added a new *idea* #2231 to be possible to define the tokentypes for challenge response tokens in a policy.

PR #2230, opened 2020-05-30
Add policy to allow application to choose token type

### LINKED ISSUE #2047: tokentype in /validate/check and triggerchallenge

cornelinux:
With the given policy an application is allowed to
define, which tokentype should be used for authentication.
To do this, the application can send the parameter
"type" in the requests /validate/check, ./samlcheck
or ./triggerchallenge - given that the policy allows to
do so.

Closes #2047

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -369,6 +369,35 @@ def check_otp_pin(request=None, action=None):
     return True
 
 
+def check_application_tokentype(request=None, action=None):
+    """
+    This pre policy checks if the request is allowed to specify the tokentype.
+    If the policy is not set, a possibly set parameter "type" is removed
+    from the request.
+
+    Check ACTION.APPLICATION_TOKENTYPE
+
+    This decorator should wrap
+        /validate/check, /validate/samlcheck and /validate/triggerchallenge.
+
+    :param request: The request that is intercepted during the API call
+    :type request: Request Object
+    :param action: An optional Action
+    :type action: basestring
+    :returns: Always true. Modified the parameter request
+    """
+    application_allowed = Match.generic(g, scope=SCOPE.AUTHZ,
+                                        action=ACTION.APPLICATION_TOKENTYPE,
+                                        user_object=request.User,
+                                        active=True).any()
+
+    # if the application is not allowed, we remove the tokentype
+    if not application_allowed:
+        request.all_data["type"] = None
+
+    return True
+
+
 def sms_identifiers(request=None, action=None):
     """
     This is a token specific wrapper for sms tokens
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -391,9 +391,11 @@ def check_application_tokentype(request=None, action=None):
                                         user_object=request.User,
                                         active=True).any()
 
-    # if applications are not allowed to request a certain token type, we remove it
-    if not application_allowed:
-        request.all_data["type"] = None
+    # if the application is not allowed, we remove the tokentype
+    if not application_allowed and "type" in request.all_data:
+        log.info("Removing parameter 'type' from request, "
+                 "since application is not allowed to authenticate by token type.")
+        del request.all_data["type"]
 
     return True
 
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -391,7 +391,7 @@ def check_application_tokentype(request=None, action=None):
                                         user_object=request.User,
                                         active=True).any()
 
-    # if the application is not allowed, we remove the tokentype
+    # if applications are not allowed to request a certain token type, we remove it
     if not application_allowed:
         request.all_data["type"] = None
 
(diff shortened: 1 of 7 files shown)


privacyidea/api/lib/prepolicy.py
```diff
@@ -369,6 +369,35 @@ def check_otp_pin(request=None, action=None):
...
+    :param action: An optional Action
+    :type action: basestring
+    :returns: Always true. Modified the parameter request
+    """
+    application_allowed = Match.generic(g, scope=SCOPE.AUTHZ,
+                                        action=ACTION.APPLICATION_TOKENTYPE,
+                                        user_object=request.User,
+                                        active=True).any()
+
+    # if the application is not allowed, we remove the tokentype
```
laclaro, line 394: this is a general policy. Better write "if applications are not allowed, we remove the tokentype"

tests/test_api_lib_policy.py
```diff
@@ -2852,6 +2852,37 @@ class RequestMock(object):
...
+        set_policy(name="pol1",
+                   scope=SCOPE.AUTHZ,
+                   action=ACTION.APPLICATION_TOKENTYPE)
+        g.policy_object = PolicyClass()
+
+        # check for
+        req.all_data = {"type": "tokentype"}
+        req.User = User("cornelius", self.realm1)
+        check_application_tokentype(req)
+        # Check, that the realm was not set, since there was no user in the request
```
laclaro, line 2875: Please check comment lines.

tests/test_api_validate.py
```diff
@@ -2660,6 +2660,48 @@ def test_34_validate_user_and_serial(self):
...
+        # Trigger challenge for HOTP
+
+        # check that only HOTP was triggered
+
+        # Delete tokens and policies
+        remove_token("tok1")
+        remove_token("tok2")
+        delete_policy("pol_chalresp")
+
+
```
laclaro, line 2704: We may add some further tests of the new functionality.

tests/test_api_lib_policy.py
laclaro: done

tests/test_api_validate.py
laclaro: done

privacyidea/api/lib/prepolicy.py
laclaro: done

privacyidea/api/lib/prepolicy.py
```diff
@@ -369,6 +369,35 @@ def check_otp_pin(request=None, action=None):
...
+    :returns: Always true. Modified the parameter request
+    """
+    application_allowed = Match.generic(g, scope=SCOPE.AUTHZ,
+                                        action=ACTION.APPLICATION_TOKENTYPE,
+                                        user_object=request.User,
+                                        active=True).any()
+
+    # if applications are not allowed to request a certain token type, we remove it
+    if not application_allowed:
+        request.all_data["type"] = None
```
laclaro, line 396: Please add some warning to hint at the required policy if an application tries to use type and is not allowed to.

cornelinux: Done in b3a37804

cornelinux:
Rebase on master, to avoid random failing of tests.

Issue #2204, opened 2020-05-14
Menu overlaps error messages


cornelinux:
In the current master branch the UI Menu overlaps error messages.

This can be reproduced by simply failing to login.

![image](https://user-images.githubusercontent.com/1908620/81923357-d6117200-95dd-11ea-8a99-db32f2eafda0.png)

PR #2232, opened 2020-06-01
2177/dashboard

cornelinux:
Implement a basic dashboard.

A policy defines, if the admin is allowed to see the dashbord or not.


laclaro:
I observe again #2204 if trying to enroll token for nonexisting user.
I did not provoke any other messages to check.
"ERR905: Cannot pass user_object as well as user, resolver, realm in policy (None, 'admin', 'enrollHOTP')"

cornelinux:
> I observe again #2204 if trying to enroll token for nonexisting user.
> I did not provoke any other messages to check.
> "ERR905: Cannot pass user_object as well as user, resolver, realm in policy (None, 'admin', 'enrollHOTP')"

This is due to the fact, that this PR is based on an older version of the master branch, where this is not fixed, yet.

PR #2237, opened 2020-06-04
fix failing test for special characters []\ in pin

laclaro:
related to #2121 

string.punctuation should not plainly be used as character list for generated pins.

Backslashes let the pin policy checking fail. Removing \ [ ] from the lists of strings remove the problem.


privacyidea/lib/utils/__init__.py
```diff
@@ -61,7 +61,7 @@
 # character lists for the identifiers in the pin content policy
 CHARLIST_CONTENTPOLICY = {"c": string.ascii_letters, # characters
                           "n": string.digits,        # numbers
-                          "s": string.punctuation}   # special
```
cornelinux, line 64: Can we leave the ``string.punctuation`` and somehow fix the tests/functionality?

laclaro: I tried to, but I can give it another shot.

laclaro: see #2238

PR #2238, opened 2020-06-04
fix policy checking keeping string.punctuation

laclaro:
related to #2121 

escaping the matching string in check_pin_policy solves the issue. Adding iterations to ensure the validity of the randomly generated pins.

### DIFF
--- a/privacyidea/lib/utils/__init__.py
+++ b/privacyidea/lib/utils/__init__.py
@@ -1216,7 +1216,7 @@ def check_pin_policy(pin, policy):
 
     # check requirements
     for str in charlists_dict["requirements"]:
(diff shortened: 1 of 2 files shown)


privacyidea/lib/utils/__init__.py
```diff
@@ -1216,7 +1216,7 @@ def check_pin_policy(pin, policy):
 
     # check requirements
     for str in charlists_dict["requirements"]:
-        if not re.search(re.compile('[' + str + ']'), pin):
+        if not re.search(re.compile('[' + re.escape(str) + ']'), pin):
```
cornelinux, line 1219: cool. I like this.

PR #2240, opened 2020-06-07
1963/headers smsgateway new column

laclaro:
This pull request provides the initial headers support for SMS gateways.

1. It requires a change in the database scheme of the smsgatewayoption table to relax the "Key" uniqueness constraint to "per Type".

CREATE TABLE smsgatewayoption (
id INTEGER NOT NULL,
"Key" VARCHAR(255) NOT NULL,
"Value" TEXT, "Type" VARCHAR(100), 
gateway_id INTEGER,PRIMARY KEY (id), 
CONSTRAINT sgix_1 UNIQUE (gateway_id, "Key", **"Type"**),
FOREIGN KEY(gateway_id) REFERENCES smsgateway_new (id)
);

2. We must add `type = "option"` to all rows in smsgatewayoption to prevent orphan (unchangeable) entries in the table.

closes #2213

### DIFF
--- a/migrations/versions/e360c56bcf8c_.py
+++ b/migrations/versions/e360c56bcf8c_.py
@@ -1,4 +1,4 @@
-"""empty message
+"""Add constraint for smsoptions and set default type to "option"
 
 Revision ID: e360c56bcf8c
 Revises: a7e91b18a460
--- a/migrations/versions/e360c56bcf8c_.py
+++ b/migrations/versions/e360c56bcf8c_.py
@@ -1,0 +1,83 @@
+"""empty message
+
+Revision ID: e360c56bcf8c
+Revises: a7e91b18a460
+Create Date: 2020-06-15 09:18:43.855589
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = 'e360c56bcf8c'
+down_revision = 'a7e91b18a460'
+
+from alembic import op, context
+import sqlalchemy as sa
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.schema import Sequence, CreateSequence
+from sqlalchemy import orm
+
+Base = declarative_base()
+
+class SMSGateway(Base):
+    __tablename__ = 'smsgateway'
+    __table_args__ = {'mysql_row_format': 'DYNAMIC'}
+    id = sa.Column(sa.Integer, Sequence("smsgateway_seq"), primary_key=True)
+    identifier = sa.Column(sa.Unicode(255), nullable=False, unique=True)
+    description = sa.Column(sa.Unicode(1024), default=u"")
+    providermodule = sa.Column(sa.Unicode(1024), nullable=False)
+    options = orm.relationship('SMSGatewayOption',
+                              lazy='dynamic',
+                              backref='smsgw')
+
+class SMSGatewayOption(Base):
+    __tablename__ = 'smsgatewayoption'
+    id = sa.Column(sa.Integer, Sequence("smsgwoption_seq"), primary_key=True)
+    Key = sa.Column(sa.Unicode(255), nullable=False)
+    Value = sa.Column(sa.UnicodeText(), default=u'')
+    Type = sa.Column(sa.Unicode(100), default=u'option')
+    gateway_id = sa.Column(sa.Integer(),
+                           sa.ForeignKey('smsgateway.id'), index=True)
+    __table_args__ = (sa.UniqueConstraint('gateway_id',
+                                          'Key', 'Type',
+                                          name='sgix_1'),
+                      {'mysql_row_format': 'DYNAMIC'})
+
+# Check if the SQL dialect uses sequences
+# (from https://stackoverflow.com/a/17196812/7036742)
+def dialect_supports_sequences():
+    migration_context = context.get_context()
+    return migration_context.dialect.supports_sequences
+
+def create_seq(seq):
+    if dialect_supports_sequences():
+        op.execute(CreateSequence(seq))
+
+def upgrade():
+    try:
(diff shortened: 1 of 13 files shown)


privacyidea/api/smsgateway.py
```diff
@@ -124,21 +129,23 @@ def delete_gateway(identifier=None):
...
+@smsgateway_blueprint.route('/<type>/<gwid>/<key>', methods=['DELETE'])
 @log_with(log)
 @prepolicy(check_base_action, request, ACTION.SMSGATEWAYWRITE)
-def delete_gateway_option(gwid=None, option=None):
+def delete_gateway_option(gwid=None, key=None, type=None):
     """
     this function deletes an option of a gateway definition
 
     :param gwid: The id of the sms gateway definition
     :return: json with success or fail
```
cornelinux, line 140: Please add in the docstring, that <type> is either "option" or "header", since the documentation for the endpoint

 DEL /option/

will disappear.
So you might also check again, if two bluprints (one for /option and one for /header) work.

privacyidea/lib/smsprovider/HttpSMSProvider.py
```diff
@@ -145,10 +147,11 @@ def submit_message(self, phone, message):
             params = {}
             data = parameter
 
-        log.debug("issuing request with parameters %s and method %s and "
-                  "authentication %s to url %s." % (parameter, method,
+        log.debug("issuing request with parameters %s, headers %s and method %s and "
+                  "authentication %s to url %s." % (params, headers, method,
```
cornelinux, line 151: Please use the {0!s} notation.

   u"issuing request with parameters {0!s}, headers {1!s}and method {2!s} and "
                  "authentication {3!s} to url {4!s}." % (params, headers, method,

privacyidea/models.py
```diff
@@ -2064,11 +2064,15 @@ class SMSGateway(MethodsMixin, db.Model):
     options = db.relationship('SMSGatewayOption',
                               lazy='dynamic',
                               backref='smsgw')
+    headers = db.relationship('SMSGatewayOption',
```
cornelinux, line 2067: I think this is superfluous.
Please check and remove.

privacyidea/models.py
```diff
@@ -2127,7 +2135,21 @@ def option_dict(self):
...
+
+    @property
+    def header_dict(self):
+        """
+        Return all connected options as a dictionary
+
+        :return: dict
+        """
+        res = {}
+        for header in self.headers:
```
cornelinux, line 2150: You do not need the second relation. 
You can do

~~~~python
for header in self.options:
    if header.Type == "header":
        res[header.key] = header.Value
~~~~

privacyidea/static/components/config/controllers/smsgatewayController.js
```diff
@@ -104,6 +123,16 @@ myApp.controller("smsgatewayController", function($scope, $stateParams,
...
+    $scope.addHeader = function() {
+        $scope.headers[$scope.newheader] = $scope.newheadervalue;
+        $scope.newheader = "";
+        $scope.newheadervalue = "";
+    };
+
     // listen to the reload broadcast
     $scope.$on("piReload", $scope.getSMSGateways);
-});
+});
```
cornelinux, line 138: Wait for PEP8! ;-)

privacyidea/models.py
```diff
@@ -2127,7 +2132,21 @@ def option_dict(self):
         """
         res = {}
         for option in self.options:
-            res[option.Key] = option.Value
+            if option.Type == "option" or option.Type is None:
```
cornelinux, line 2135: I tested this with SQLlite. An old gateway config does not get read, if the option does not contain a *type* in the database. Different database dialects can return different "empty" values, could be empty strings, or Nones or Nulls.
This breaks with SQLite and probably other databases.

It might make more sense to change this to 
~~~~python
if options.Type == "option" or not option.Type:
~~~~


~~~~

privacyidea/api/smsgateway.py
```diff
@@ -129,20 +129,18 @@ def delete_gateway(identifier=None):
...
     """
-    if key.startswith("option."):
-        key = key[7:]
-    elif key.startswith("header."):
-        key = key[7:]
+    type = key.split('.')[0]
+    key = ".".join(key.split('.')[1:])
```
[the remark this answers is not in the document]
laclaro: This makes sense, thanks

privacyidea/models.py
laclaro: ok

privacyidea/models.py
```diff
@@ -2174,7 +2174,7 @@ class SMSGatewayOption(MethodsMixin, db.Model):
     id = db.Column(db.Integer, Sequence("smsgwoption_seq"), primary_key=True)
     Key = db.Column(db.Unicode(255), nullable=False)
     Value = db.Column(db.UnicodeText(), default=u'')
-    Type = db.Column(db.Unicode(100), default=u'')
+    Type = db.Column(db.Unicode(100), default=u'option')
```
cornelinux, line 2177: Wouldn't adding a "default" in the models.py create a need for a new migrations script? Or is the default not reflected in the schema?

migrations/versions/e360c56bcf8c_.py
```diff
@@ -0,0 +1,83 @@
+"""empty message
```
cornelinux, line 1: Please add a nice message, that would be displayed in 

    pi-manage db history

like: *Add constraint for smsoptions and set default type to "option"*

laclaro: ok

PR #2244, opened 2020-06-09
Update jQuery to fix broken hamburger-menu

### LINKED ISSUE #2233: collapsed menu won't open
Hi,

I just realized that the collapsed menu won't open (e.g. on small displays on smart phones).

Java Script throws an error here:
https://github.com/privacyidea/privacyidea/blob/cc04f6695338923c22bc2edaae33dec939bda481/privacyidea/static/contrib/js/bootstrap.js#L723

"TypeError: can't convert Object to string"

Best,

Henning

plettich:
- update angularjs to latest 1.8.0
- remove angular-route since we use angular-ui-router
- re-enable menu texts when viewport smaller than 768px
- use `ui-sref-active` to mark active menu entries

Fixes #2233

### DIFF
--- a/privacyidea/static/templates/menu.html
+++ b/privacyidea/static/templates/menu.html
@@ -9,7 +9,7 @@
     log file. You can change the debug level in
     pi.cfg. Afterwards be sure to remove the logfile!
 </div>
-<div class="navbar navbar-default navbar-fixed-top">
+<div class="navbar navbar-default">
     <div class="container">
         <a class="navbar-brand" ui-sref="token.list" ng-show="loggedInUser.role">
             <img ng-src="{{ instanceUrl }}/static/css/{{ piLogo }}"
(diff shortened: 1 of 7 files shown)


plettich:
There is still an issue with the main body. I am on it.

privacyidea/static/templates/menu.html
```diff
@@ -31,70 +31,63 @@
         <div class="navbar-collapse collapse myMainNavbar">
             <ul class="nav navbar-nav" ng-hide="token_wizard">
 
-                <li ng-class="{active: $state.includes('dashboard')}"
+                <li ui-sref-active="active"
```
cornelinux, line 34: Why is that removed? The dashboard should not always be active?
We use the "$state.includes" to mark the top menu as active or not.

PR #2245, opened 2020-06-10
Add some comments about installation with ubuntu

cornelinux:
Please check if the comment makes sense to you and helps to understand the repo thingy.


cornelinux:
Wait - we check if we make the community repo browsable to a certain extent.

cornelinux:
Actually you might also get a 404 in your browser, since the browser does an automatic redirect to https. An on https the directory is really not available, only on http! So maybe we should add this note anyways.

PR #2246, opened 2020-06-10
Allow to switch off or on the translation warning

### LINKED ISSUE #2223: Less prominent translation-missing marker
When there are untranslated strings, they are marked with "[MISSING]: ".

The marking should either be less prominent, or we should completely remove it.

This can be done here:
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/static/app.js#L103

~~~javascript
 gettextCatalog.debug = false;
~~~~

or

~~~~javascript
  gettextCatalog.debug = false;
  gettextCatalog.debugPrefix = "(en) ";
~~~~

cornelinux:
The WebUI prepends all non-translated strings with
the warning [MISSING]. This is often annoying and
irritating. The prefix now needs to be explicitly
turned on using PI_TRANSLATION_WARNING.

Closes #2223

### DIFF
--- a/doc/installation/system/inifile.rst
+++ b/doc/installation/system/inifile.rst
@@ -81,6 +81,9 @@ own UI and you do not want to present the UI to the user or the outside world.
 .. note:: The API calls are all still accessible, i.e. privacyIDEA is
    technically fully functional.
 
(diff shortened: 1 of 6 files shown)


cornelinux:
Done in 86e9cfb3

PR #2255, opened 2020-06-16
INitial writeup

### LINKED ISSUE #2253: linotp migration script: failsafe for tokeninfos
The linotp-magration script fails to migrate token infos, if a phone number contains special characters, which can happen if a user entered faulty characters there.

Als long as #1874 is not completely done, we could add some failsafe for migrating tokeninfo fields.

cornelinux:
Add a failsafe (which only works with python2)
to remove non-ascii characters from configured tokeninfo keys.

Closes #2253

### DIFF
--- a/tools/privacyidea-migrate-linotp.py
+++ b/tools/privacyidea-migrate-linotp.py
@@ -439,7 +439,7 @@ def migrate(config_obj):
                 # Add tokeninfo for this token
                 for (k, v) in ti.items():
                     if k in config_obj.MIGRATE.get("tokeninfo_unicode_keys", []):
-                        v = unicode(v, erros="ignore")
+                        v = unicode(v, errors="ignore")
                     tokeninfo_values.append(dict(
                         serial=r["LinOtpTokenSerialnumber"],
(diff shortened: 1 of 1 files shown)


tools/privacyidea-migrate-linotp.py
```diff
@@ -435,6 +438,8 @@ def insert_chunks(conn, table, values, chunk_size=100000):
             if config_obj.MIGRATE.get("tokeninfo") and ti:
                 # Add tokeninfo for this token
                 for (k, v) in ti.items():
+                    if k in config_obj.MIGRATE.get("tokeninfo_unicode_keys", []):
+                        v = unicode(v, erros="ignore")
```
plettich: Assuming that the value is a unicode string (We read from the table as unicode) we can use
```v = v.encode('ascii', 'ignore').decode('utf8')```
This works in Python 2 and Python 3.

cornelinux: Will do

cornelinux: done in 97f4c86

Issue #1158, opened 2018-08-05
Allow setting phonenumber format


cornelinux:
See #1141 

During enrollment (especially self enrollment of the user), the format of the entered phone number should be restricted, to avoid the user having enter invalid phone numbers.

The phone number should be entered in E.123 international	 (aka DIN 5008 interenation) 

~~~~
   + <countrycode> <area code - without leading zero> <local number>
~~~~

plettich:
Using multiple issues in a `Fixes `-line does not seem to close all issues...

PR #2256, opened 2020-06-16
Add input pattern for phone numbers

### LINKED ISSUE #1141: Optimize Config notes during enrollment
2.22.22

When users go to enroll a new token, they see the following message for email and sms token types.  This should be an admin only message, and I feel shouldn't even show up when they actually have been configured properly.
> For Email tokens to work properly you need to set the Email Token Config!
> For SMS tokens to work properly you need to set the SMS Token Config!

I'd also like the ability to turn off the checkbox below that message:
> Read email address dynamically from user source on each request.
> Read number dynamically from user source on each request.

And for SMS, it would be nice to have the ability to specify a format for the phone number.  For example, Twilio highly recommends using the E.164 format for phone numbers.

plettich:
When enrolling an SMS Token the phone number input accepts any characters.
With the new pattern only numbers and an optional '+' at the beginning are
allowed.
Also the input field for phone and email are required now unless the 'read
from userstore' option is enabled.

Fixes #1141 #1158

### DIFF
--- a/privacyidea/static/components/token/views/token.enroll.sms.html
+++ b/privacyidea/static/components/token/views/token.enroll.sms.html
@@ -18,7 +18,10 @@
     <label for="phone" translate>Phone number</label>
     <input type="text" id="phone"
            autofocus
+           ng-required="! form.dynamic_phone"
            class="form-control"
+           placeholder="+1234567890"
+           pattern="(\+[1-9])?\d{1,14}"
            ng-model="form.phone" name="phone" />
     <div>
         <button class="btn btn-default btn-xs" ng-repeat="phone in phone_list"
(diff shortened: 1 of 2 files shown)


privacyidea/static/components/token/views/token.enroll.sms.html
```diff
@@ -18,7 +18,10 @@ <h4 translate>Token data</h4>
     <label for="phone" translate>Phone number</label>
     <input type="text" id="phone"
            autofocus
+           ng-required="! form.dynamic_phone"
            class="form-control"
+           placeholder="+1234567890"
+           pattern="(\+[1-9])?\d{1,14}"
```
cornelinux, line 24: I am not sure what valid phone numbers actually are or should be.

Some use blanks, dots and minus?  555.123.123
I think it would be no harm to add those?

plettich: We could add those but we might need to think about how we save the numbers internally. The [ITU E.164](https://en.wikipedia.org/wiki/E.164) does not mention any blanks, dots or dashes. We can remove them before saving the number in the DB.
This pattern is also not ITU compliant since it allows zeroes in the number (except after a leading `+`).

plettich: Done in 42cf88a6

PR #2259, opened 2020-06-18
fix check_max_active_token on enabling token

### LINKED ISSUE #2215: Disabled token can be enabled despite Active Tokens Count exceeding the value set in Policy
### Top-level intent

What did you try to achieve?
Token policy is set to have only 01 Active TOTP Token for the user. Checking validation if system prevents users to have multiple Active TOTP tokens despite policy set to allow only 01 active token.

### Steps to reproduce

1. Login as admin, Create policy to allow max 01 Active TOTP token and Max 10 tokens.
2. Enroll TOTP token for a user, disable it.
3. Enroll another TOTP token for same user and its a active/enabled state.
4. Try to enroll another TOTP token for the same user and system will prevent the enrollment.
5. Now try to Enable the disabled token for the same user and system doesn't prevent enabling.

### Expected outcome

System should have checked the policy and prevented user/admin from enabling the token, if the user already has max active tokens for token type.

### Actual outcome
**Bug** - Now user has 2 active TOTP tokens despite policy only allowing max 1 Active TOTP token.

What did actually happen?

### Configuration

* **privacyIDEA version**:
* **Installation method**: (from Ubuntu packages, github, PyPI, ...)
* **Python version**:
* **Operating system**:
* **Webserver**:
* **Token database**: (MySQL, PostgreSQL, ...)

### Log file

laclaro:
closes #2215 

On checking the max_token_user policy, we not get the tokenowner from the serial if no user_obj is present.

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -694,8 +694,11 @@ def check_max_token_user(request=None, action=None):
     params = request.all_data
     serial = getParam(params, "serial")
     user_object = get_user_from_param(params)
-    if not user_object.login and serial:
-        user_object = get_token_owner(serial)
+    if user_object.is_empty() and serial:
+        try:
+            user_object = get_token_owner(serial) or User()
+        except:
+            user_object = User()
     if user_object.login:
         tokentype = getParam(params, "type")
         if not tokentype:
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -699,7 +699,7 @@ def check_max_token_user(request=None, action=None):
         try:
             user_object = get_token_owner(serial) or User()
         except ResourceNotFoundError:
-            user_object = User()
+            pass
     if user_object.login:
         tokentype = getParam(params, "type")
         if not tokentype:
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -76,7 +76,7 @@ from privacyidea.lib.policy import SCOPE, ACTION, PolicyClass
 from privacyidea.lib.policy import Match
 from privacyidea.lib.user import (get_user_from_param, get_default_realm,
                                   split_user, User)
-from privacyidea.lib.token import (get_tokens, get_realms_of_token, get_token_type)
+from privacyidea.lib.token import (get_tokens, get_realms_of_token, get_token_type, get_token_owner)
 from privacyidea.lib.utils import (get_client_ip,
                                    parse_timedelta, is_true, generate_charlists_from_pin_policy,
                                    check_pin_policy, get_module_class,
@@ -693,6 +693,8 @@ def check_max_token_user(request=None, action=None):
     ERROR_ACTIVE_TYPE = "The number of active tokens of type {0!s} for this user is limited!"
     params = request.all_data
     user_object = get_user_from_param(params)
+    if not user_object.login and params.get("serial", None):
+        user_object = get_token_owner(params["serial"])
     if user_object.login:
         serial = getParam(params, "serial")
         tokentype = getParam(params, "type")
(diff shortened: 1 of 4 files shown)


privacyidea/api/lib/prepolicy.py
```diff
@@ -693,6 +693,8 @@ def check_max_token_user(request=None, action=None):
     ERROR_ACTIVE_TYPE = "The number of active tokens of type {0!s} for this user is limited!"
     params = request.all_data
     user_object = get_user_from_param(params)
+    if not user_object.login and params.get("serial", None):
+        user_object = get_token_owner(params["serial"])
     if user_object.login:
```
plettich, line 698: In case a serial is given but there is no user assigned to the token, this fails since `user_object` will be `None`. We could add `if user_object and user_object.login:` to mitigate this.

laclaro: If the token is not assigned, the max_token policy should not block anything...?

I added a test for this, but in my opinion, it works as it should

plettich: The problem is that this pre-policy function will always be executed regardless of an existing max_token policy (the policy will be queried later).
So in case we only have a serial and the token does not have a user assigned, the `get_token_owner(serial)` function returns 'None' and the check breaks.

tests/test_api_lib_policy.py
```diff
@@ -370,9 +370,31 @@ def test_04a_check_max_active_token_user(self):
...
+        enable_token("NEW001", enable=False)
+        # we enroll a new HOTP token, this would now succeed
+        init_token({"serial": "NEW002", "type": "hotp",
+                    "otpkey": "1234567890123456"},
+                   user=User(login="cornelius",
+                             realm=self.realm1))
+        tokenobject_list = get_tokens(user=User(login="cornelius",
+                                                realm=self.realm1))
+        self.assertTrue(len(tokenobject_list) == 2)
+        # now we enable the first hotp token again, which fails due to the policy
```
cornelinux, line 383: This is not a valid test for enabling the token. 
Enabling the token is simply:
  
~~~~http
POST /token/enable
serial=<serial>
~~~~

So there is no user object in the request. However, you have a user in lines 384 and 385. Your line 698 in lib/prepolicy.py will not be covered.

privacyidea/api/lib/prepolicy.py
laclaro: if the token with serial does not exist yet (as in /token/init), get_token_owner(serial) does throw an exception instead of returning None, since the token does not exists yet.

privacyidea/api/lib/prepolicy.py
```diff
@@ -694,8 +694,11 @@ def check_max_token_user(request=None, action=None):
     params = request.all_data
     serial = getParam(params, "serial")
     user_object = get_user_from_param(params)
-    if not user_object.login and serial:
-        user_object = get_token_owner(serial)
+    if user_object.is_empty() and serial:
+        try:
+            user_object = get_token_owner(serial) or User()
+        except:
+            user_object = User()
```
cornelinux, line 702: I think user_object is already an empty User() if an exception occurs (see line 697). So imho there is no need to set the user_object ot an empty user again. You could simply do a ``pass``.

Is it possible to determine, which Exception we would catch, so that we only would do ``except ExeptionType``?

tests/test_api_lib_policy.py
laclaro: I removed the user data to match the enabling request.

cornelinux: Well, I would have thought it would be great, if we actually had a real API test.
I manually checked the API and it was oK:

privacyidea/api/lib/prepolicy.py
laclaro: That's correct. Thank you. The exeption Type is ResourceNotFoundError.

PR #2261, opened 2020-06-18
2236/script sms provider

cornelinux:
Add the ScriptSMSProvider, that can call arbitrary scripts, that take the phone number from the command line and the text from stdin.

### DIFF
--- a/privacyidea/lib/smsprovider/ScriptSMSProvider.py
+++ b/privacyidea/lib/smsprovider/ScriptSMSProvider.py
@@ -34,7 +34,7 @@ SCRIPT_WAIT = "wait"
 
 class ScriptSMSProvider(ISMSProvider):
 
-    def __init__(self, db_smsprovider_object=None, smsgateway=None):
+    def __init__(self, db_smsprovider_object=None, smsgateway=None, directory=None):
         """
         Create a new SMS Provider object fom a DB SMS provider object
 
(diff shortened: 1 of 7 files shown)


cornelinux:
Oups, no tests here.

privacyidea/lib/smsprovider/ScriptSMSProvider.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def __init__(self, db_smsprovider_object=None, smsgateway=None, directory=None):
+        """
+        Create a new SMS Provider object fom a DB SMS provider object
+
+        :param db_smsprovider_object: The database object
+        :param smsgateway: The SMS gateway object from the database table
+            SMS gateway. The options can be accessed via
+            self.smsgateway.option_dict
+        :return: An SMS provider object
```
plettich, line 45: Please describe all parameters

privacyidea/lib/smsprovider/ScriptSMSProvider.py
```diff
@@ -0,0 +1,120 @@
...
+        """
+        self.config = db_smsprovider_object or {}
+        self.smsgateway = smsgateway
+        self.script_directory = directory or get_app_config_value("PI_SCRIPT_SMSPROVIDER_DIRECTORY",
+                                                                  "/etc/privacyidea/scripts")
+
+    def submit_message(self, phone, message):
+        """
+        send a message to a phone via an smpp protocol to smsc
```
plettich, line 54: We just call the script, we have no idea about the protocol, or have we?

privacyidea/lib/smsprovider/ScriptSMSProvider.py
cornelinux: done in 8d67564e

privacyidea/lib/smsprovider/ScriptSMSProvider.py
cornelinux: done in 8d67564e

PR #2269, opened 2020-06-26
Add closed issues to changelog

cornelinux:
This creates a tag for 3.4dev1 so that we
can do a devel snapshot.

This way we can keep track of the changelog and build a debian package snapshot.

I also reverted ldap3 to 2.6.1 so that we will not run into issues.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,42 @@
+Version 3.4dev1, 2020-06-26
+
+  Features:
+   * Add ScriptSMSProvider, that can send SMS through external
+     Gateways using arbitrary scripts (#2236)
+   * Add a basic dashboard as start screen in the WebUI (#2177)
+
+  Enhancements:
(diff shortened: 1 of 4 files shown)


cornelinux:
I just realized, we have multiple heads in our DB migration. So we do not want to build from this.
I will add another commit!

Changelog
```diff
@@ -1,3 +1,42 @@
...
+
+  Features:
+   * Add ScriptSMSProvider, that can send SMS through external
+     Gateways using arbitrary scripts (#2236)
+   * Add a basic dashboard as start screen in the WebUI (#2177)
+
+  Enhancements:
+   * Allow deletion of validity period via UI (#2263)
+   * Remove missing translation marker and allow to set a
+     customer marker (#2223)
```
plettich, line 11: custom?

cornelinux: done in 6fb1028

PR #2277, opened 2020-07-01
Remove the extra code for /validate/samlcheck

cornelinux:
samlcheck is basically the same code as /validate/check.
So we remove samlecheck and incorporate it into /validate/check.


cornelinux:
@plettich I would like to know, what you think.

privacyidea/api/validate.py
```diff
@@ -426,44 +356,61 @@ def samlcheck():
...
     password = getParam(request.all_data, "pass", required)
+    otp_only = getParam(request.all_data, "otponly")
     token_type = getParam(request.all_data, "type")
     options = {"g": g,
-               "clientip": g.client_ip,
-               "token_type": token_type}
+               "clientip": g.client_ip}
     # Add all params to the options
     for key, value in request.all_data.items():
             if value and key not in ["g", "clientip"]:
```
plettich, line 368: i am not entirely sure what the intention of this construct is.
If `value` and `key` exist, the result would be to check if `key` is not in the list.

cornelinux: This is an old construct and not topic of this PR.
However, if the value is not empty, then the key and value are copied to the options.
To my knowledge the intention is, to pass all parameters except "g" and "clientip" to the lower levels.

cornelinux: ...and also empty values were not supposed to be passed.
If we change this, empty values would be passed to the lower level and we would change the behaviour unexpectedly.

cornelinux:
tests for /validate/samlcheck already existed.

privacyidea/api/validate.py
```diff
@@ -426,44 +356,61 @@ def samlcheck():
     mapping.
     """
     user = request.User
+    result = success = False
```
plettich, line 359: We should differentiate here, the success is a bool but result can also be a dict. The usage is a little confusing in this part.

cornelinux: Hm, it looks as if the variables are set later anyways, always.
So what about simply removing them at this point?

cornelinux: I remove this in https://github.com/privacyidea/privacyidea/pull/2277/commits/d3cf61aded400cdf5c228b9cd963af4b1f57544d

Issue #2278, opened 2020-07-01
Proposal: Add new test group *Integration*


plettich:
I think it would be beneficial if we add another test group to *api* and *lib* tests:
Some testcases are over-complicated because we try to test several things at once. We should keep the *api* and *lib* tests clean and test only the basic functionality there.
All other tests with a more complex setup should be put in a different test group.
Especially testing workflows with several policies could fit into such a category.

PR #2279, opened 2020-07-01
Fix missing options parameter

### LINKED ISSUE #2276: RADIUS and Remote Tokens do not pass the options parameter to check_pin
In some token classes the `options` parameter (which contains the `g` object, policies and audit) doesn't get passed when calling further functions. This leads to broken policy decorators (e.g. [`check_pin()`](https://github.com/privacyidea/privacyidea/blob/09a857a27f1558b2c9e6a9703de33797017c1a5d/privacyidea/lib/tokens/remotetoken.py#L206))

plettich:
Some token classes did not pass the options parameter to the `check_pin()`
function which results in not checking the otppin policy.

Fixes #2276

### DIFF
--- a/privacyidea/lib/tokens/radiustoken.py
+++ b/privacyidea/lib/tokens/radiustoken.py
@@ -191,7 +191,7 @@ class RadiusTokenClass(RemoteTokenClass):
         # should we check the pin locally?
         if self.check_pin_local:
             # With a local PIN the challenge response is always a privacyIDEA challenge response!
-            res = self.check_pin(passw)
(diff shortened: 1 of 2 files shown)


plettich:
> Would some kind of tests make sense, here?

Yes definitely. Unfortunately this requires tokens assigned to users from a userstore which provides password checking and corresponding policies. I wouldn't want to clutter out API-testcases with this that's why i created #2278

cornelinux:
OK. Sounds sensible to me.

PR #2281, opened 2020-07-01
Use webauthn-client

### LINKED ISSUE #2273: Use privacyidea/webauthn-client
We should switch to using [privacyidea/webauthn-client](https://github.com/privacyidea/webauthn-client) as a submodule, instead of duplicating the library in this project.

NuvandaPV:
fixes #2273


.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "privacyidea/static/contrib/js/webauthn-client"]
+	path = privacyidea/static/contrib/js/webauthn-client
+	url = https://github.com/privacyidea/webauthn-client
```
cornelinux, line 3: It looks like if there was the possibility to actually use the url of the *tag* to link the submodule to a tag.
But it looks like as if the URL to the tag "v0.1.1" actually ends up on the branch, since it has the same name?

NuvandaPV: Hm… are you sure about this? I can't find it mentioned anywhere in the documentation, and the concept of having a submodule tracking a tag seems pretty counterintuitive to me. As I understand it, tags don't usually move, so I am not sure what it would mean to be tracking a tag (would the tag move? or would it just mean that trying to update the submodule does nothing?). Either way, I do not know what you mean about using the url of the tag. I assume I would edit this file manually (this current state was generated), removing the brach specification and updating the url how?

I am also not sure what we are trying to achieve here. Do you want to avoid naming braches and tags the same way? In that case we could just rename the branch “v0.1”, or even “v0.1.x”, which would probably have been the better name anyways, as it is intended to track the patch-level updates. Or would you like to avoid having release-branches altogether? In this case, we could just remove the branch and have the submodule tracking the master. In that case, updating the submodule to the next patch-release would become more difficult, but updating the submodule to the absolute newest commit would become a little bit quicker. The downside here is losing the explicit documentation, which version that submodule is currently on. But I don't think it would make that much of a difference either way, considering how rarely this submodule will (hopefully) ever need to be updated.

NuvandaPV:
(see my reply in-line)

NuvandaPV:
@cornelinux could you take another look at this?

NuvandaPV:
@plettich do you have some time to look at this?

.gitmodules
plettich: We could drop the branch altogether and reference a specific tag-commit here: https://stackoverflow.com/a/1778247/7036742
We just need to remember to commit a new tag here when the `webauthn-client` repository advances.

Issue #1586, opened 2019-04-22
Challenge timestamp and expiration is saved in database as localtime but incorrectly marked as GMT in challenge lists


droobah:
# What did you try to do?

Trigger challenge for token and then use /token/challenges/<serial> to list challenges for the token.   Initial timestamp for challenge trigger was Mon, 22 Apr 2019 14:01:54 CDT

# What outcome did you expect?

A list of challenges for the token with a timestamp of Mon, 22 Apr 2019 14:01:54 CDT and with an expiration date 120 seconds in the future.  I also would've accepted a timestamp of Mon, 22 Apr 2019 19:01:54 GMT

# What outcome did you experience?

A challenge with a timestamp  Mon, 22 Apr 2019 14:01:54 GMT and expiration of Mon, 22 Apr 2019 14:03:54 GMT.  I was able to have the correct functionality by changing the local timezone of the server from CDT to UTC.

# Configuration

* privacyIDEA Version: 3.0
* Installation method: ubuntu packages

## more details: 

* OS: Ubuntu 16.04LTS
* Webserver: Apache2
* Tokendatabase: Mysql

plettich:
Thanks for the bug-report, i can confirm this behavior. It seems we incorrectly use a different timezone for the challenge timestamps.

plettich:
Actually we already write the incorrect (local) time to the db. We should use only UTC on the server side and let the client handle the conversion.
BTW: The added Timezeone label comes from the `jsonify()` function in Flask.

plettich:
The following code parts write times into the db:
- [ ] [`Audit`](https://github.com/privacyidea/privacyidea/blob/e6bb0a131c5ae4f32a0f6d379571e932466df941/privacyidea/models.py#L2603)

cornelinux:
We need to add tests that write data in timezoneA and read data i timezoneB.
=> Check how timezones can be set in tests.

cornelinux:
In 3.4 we will save challenges in UTC. So we are on the safe side, if a database should not support timezone aware fields.
This issue will only handle challenges.

plettich:
> We need to add tests that write data in timezoneA and read data i timezoneB.
> => Check how timezones can be set in tests.

here is a nice approach: https://blog.xelnor.net/python-mocking-datetime/
though we would need to replace all datetime imports.

PR #2282, opened 2020-07-02
Save challenge timestamps in UTC

plettich:
The challenge timestamps were saved in local time in the database. This
leads to false results when running servers in different timezones.
Saving the timestamps in UTC solves this (assuming the Server-Time is set
correctly).
This also fixes if a server checking the challenge is running fast.

Working on #1586

### DIFF
--- a/tests/test_api_validate.py
+++ b/tests/test_api_validate.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import six
 from six.moves.urllib.parse import urlencode
 import json
 from .base import MyApiTestCase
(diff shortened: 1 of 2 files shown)


tests/test_api_validate.py
```diff
@@ -1269,6 +1294,75 @@ def test_11b_challenge_response_multiple_hotp_failcounters(self):
...
+                                   tzinfo=timezone(datetime.timedelta(hours=+1)))):
+            with self.app.test_request_context('/validate/check',
+                                               method='POST',
+                                               data={"user": "cornelius",
+                                                     "transaction_id": transaction_id,
+                                                     "pass": "755224"}):
+                res = self.app.full_dispatch_request()
+                self.assertTrue(res.status_code == 200, res)
+                result = res.json.get("result")
+                self.assertFalse(result.get("value"))
```
laclaro, line 1360: As discussed, please check that the transaction was really removed by the janitor.

plettich: Done in e260bda1

PR #2283, opened 2020-07-02
2275/authorized policy

### LINKED ISSUE #2275: authorization policy is_authorized

It is currently not possible to deny the access for a user, if he has a valid OTP token.
We can restrict access by token token but not complete access for the user.

Maybe such **authorization** would usually be done with an application, but in certain cases it can be helpful, if privacyIDEA could also deny access.

Add a policy

scope: authorization
action: is_authorized= True|False or  "Allow access" | "Deny access"

This way using normal policy priority we could e.g. deny access on a certain IP for *all* users in a realm but allow access for only a limited user group.

Working with FreeRADIUS ulang.

# Implementation

Here https://github.com/privacyidea/privacyidea/blob/master/privacyidea/api/validate.py#L215 we add a **postpolicy** as decorator to deny access, if the access is not granted by the authorization policy.
Using a postpolicy will also rendet the used OTP value useless.

* [x] define policy in lib/policy.py
* [x] define postpolicy decorator
* [x] decorate /validate/check and /validate/triggerchallenge

cornelinux:
closes #2275

### DIFF
--- a/doc/eventhandler/index.rst
+++ b/doc/eventhandler/index.rst
@@ -36,6 +36,8 @@ all further API calls.
 
    *An action is bound to the event* token_init.
 
+.. _evenhandler_pre_and_post:
+
 Pre and Post Handling
 ---------------------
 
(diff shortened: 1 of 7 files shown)


cornelinux:
@laclaro done in 90243bde

PR #2287, opened 2020-07-04
Add counter decryption

cornelinux:
The SAM Export tool for migration SafeNet
Authentication Manager exports the serial,
the seed and the counter.

Here we also decrypt the counter so that it
can be imported to privacyIDEA.

### DIFF
--- a/tools/privacyidea-decrypt-safeword.py
+++ b/tools/privacyidea-decrypt-safeword.py
@@ -30,7 +30,7 @@ with open(args.file, 'r') as f:
             current_token["serial"] = serial
 
         if re.match("sccTokenData", line):
-            m = re.search("sccKey=\((.*?)\).*?sccSeq=\((.*?)\).*", line)
+            m = re.search(r"sccKey=\((.*?)\).*?sccSeq=\((.*?)\).*", line)
             if m:
                 seed = m.group(1)
                 algo, data = seed.split(":")
--- a/tools/privacyidea-decrypt-safeword.py
+++ b/tools/privacyidea-decrypt-safeword.py
@@ -30,7 +30,7 @@ with open(args.file, 'r') as f:
             current_token["serial"] = serial
 
         if re.match("sccTokenData", line):
-            m = re.search("sccKey=\((.*?)\).*", line)
+            m = re.search("sccKey=\((.*?)\).*?sccSeq=\((.*?)\).*", line)
             if m:
                 seed = m.group(1)
                 algo, data = seed.split(":")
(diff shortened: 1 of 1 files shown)


tools/privacyidea-decrypt-safeword.py
```diff
@@ -30,7 +30,7 @@
             current_token["serial"] = serial
 
         if re.match("sccTokenData", line):
-            m = re.search("sccKey=\((.*?)\).*", line)
+            m = re.search("sccKey=\((.*?)\).*?sccSeq=\((.*?)\).*", line)
```
plettich, line 33: adding a `raw` string specifier here removes the PEP8 warning

tools/privacyidea-decrypt-safeword.py
```diff
@@ -46,6 +46,13 @@
                     des = DES.new(des_key)
                     d = des.decrypt(binascii.unhexlify(data))
                     current_token["seed"] = d.strip("\x00")
+
+                counter = m.group(1)
+                algo, data = counter.split(":")
+                if algo.lower() == "des":
+                    des = DES.new(des_key)
```
plettich, line 53: This requires the unsupported (since 2014) and undocumented PyCrypto library. Even the drop-in replacement PyCryptoDome requires a mode: https://www.pycryptodome.org/en/latest/src/cipher/des.html
According to https://stackoverflow.com/questions/54237266/des-cipher-with-pyca-cryptography-pbewithmd5anddes we can use the `cryptography` library, we just need to check which mode is used.

tools/privacyidea-decrypt-safeword.py
cornelinux: done in b302d64d

tools/privacyidea-decrypt-safeword.py
cornelinux: I think this is a generic problem. We would also have to remove the AES decryption. Well, I think the AES decryption is not used anymore. The DES is used for SAM export ldif files. 
The AES will probably not be used anymore in the future, since new tokens are not delivered in ldif.
So this script is deprecated in itself anyways.

I will look into DES from cryptography, if this can *easily* help.

plettich: Yes, without test data this is cumbersome.
But we need to be clear that this script does not work with the current environment.
We can add a `try...except ImportError` around the import of `Crypto` to catch this.

cornelinux: Maybe we should move the script from the privacyidea-repo to the the scripts repo. Thus it would get a more "unofficial" character.

cornelinux: I moved it here:
https://github.com/privacyidea/scripts/blob/master/toolbox/privacyidea-decrypt-safeword.py

PR #2290, opened 2020-07-09
Make remote_user aware of Split@Sign option

### LINKED ISSUE #1954: Make remote_user policy aware of split@Sign parameter
The [`is_remote_user_allowed()`](https://github.com/privacyidea/privacyidea/blob/eb0000f2c54b37c75b715e536eb23749ac487543/privacyidea/api/lib/prepolicy.py#L1153) policy does not respect the [`split@Sign`](https://privacyidea.readthedocs.io/en/latest/configuration/system_config.html#splitatsign)-configuration. See also #1808

plettich:
Move the check for Split@Sign into the `split_user()` function. This way it
applies to all calls for (not) splitting the username into user and realm.

Also the remote user policy wasn't checked correctly. If set to `disable`,
the log-in with a remote user was still possible. This is fixed now.

Fixes #1954

### DIFF
--- a/privacyidea/api/auth.py
+++ b/privacyidea/api/auth.py
@@ -207,12 +207,9 @@ def get_auth_token():
         raise AuthError(_("Authentication failure. Missing Username"),
                         id=ERROR.AUTHENTICATE_MISSING_USERNAME)
 
-    loginname = username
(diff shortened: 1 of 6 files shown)


plettich:
I'll open a new one without conflicts...

Issue #1874, opened 2019-09-30
Python 3 support for tools


plettich:
Currently not all tools work with python 3. We should fix them and add some basic testing:
- [x] ``privacyidea-standalone``
- [x] ``privacyidea-diag``
- [x] ``privacyidea-migrate-linotp.py``
- [ ] ~~`privacyidea-decrypt-safeword.py`~~
- [ ] ~~`privacyidea-find-oath.py`~~
- [x] `privacyidea-queue-huey`

cornelinux:
Very good idea and very important to keep track. (However, there are some more important tools like 

* [ ] privacyidea-diag
* [ ] privacyidea-token-janitor

as these are used in bigger installations more often - or do these already work?

cornelinux:
The privacyidea-token-janitor works well.

@plettich today I realized, that the migrate-linotp script does not work, if you do a reencrypt.
It happens here:
https://github.com/privacyidea/privacyidea/blob/master/tools/privacyidea-migrate-linotp.py#L142

Probably the output of aes_cbc_decrypt is not allowed for unhexlify.

cornelinux:
I do not know when the script `privacyidea-find-oath` was used at all. I think we can drop it.

cornelinux:
@plettich I just checked and I used privacyidea-migration-linotp successfully on Python 3.8.
I think we should remove privacyidea-find-oath so we could actually cloase this issue. What do you think?

PR #2296, opened 2020-07-17
1874/enable python3 for tools

plettich:
Working on #1874


plettich:
This should fix
- `privacyidea-diag`
- `privacyidea-queue-huey`
- `privacyidea-standalone`

PR #2298, opened 2020-07-17
Fix deprecation warnings in tests

plettich:
- passlib uses the `hash()` function to calculate the pbkdf2 password
- the pbkdf2 digest moved from `passlib.utils` to `passlib.crypto.digest`
- the `inspect.getargspec()` function is deprecated
- add 'r' specifier on regex strings
- some huey queue parameter are deprecated
- refactor: use `get_data_from_params()` in `save_resolver()`
- escape some backslashes in strings
- fix some byte encoding issues
- refactor: the `get_wrapped()` function for all mocks is moved to `smtpmock`
- `assertEquals()` is deprecated and changed to `assertEqual()`
- `assertRaisesRegexp()` is deprecated in favour of `assertRaisesRegex()`
- `assertDictContainsSubset()` is deprecated
- `sqlalchemy.testing` emits a `BytesWarning`
- remove some unused code and fix some PEP8

For functions that are not available in either Python2 or Python3 we can
- add a context manager to catch the `DeprecationWarning` or
- add a `try...except AttributeError` and have both function calls

### DIFF
--- a/privacyidea/lib/crypto.py
+++ b/privacyidea/lib/crypto.py
@@ -74,12 +74,7 @@ from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.asymmetric import padding as asym_padding
 
 import passlib.hash
-if hasattr(passlib.hash.pbkdf2_sha512, "encrypt"):
-    hash_admin_pw = passlib.hash.pbkdf2_sha512.encrypt
-elif hasattr(passlib.hash.pbkdf2_sha512, "hash"):
-    hash_admin_pw = passlib.hash.pbkdf2_sha512.hash
-else:
-    raise Exception("No password hashing method available")
+hash_admin_pw = passlib.hash.pbkdf2_sha512.hash
 
 if not PY2:
     long = int
--- a/privacyidea/lib/utils/compare.py
+++ b/privacyidea/lib/utils/compare.py
@@ -97,7 +97,13 @@ def _compare_matches(left, comparator, right):
     :return: True or False
     """
     try:
-        return re.match("^" + right + "$", left) is not None
+        # check for regex modes
+        m = re.match(r'^(\(\?[a-zA-Z]+\))(.+)$', right)
+        if m and len(m.groups()) == 2:
+            regex = m.group(1) + r'^' + m.group(2) + r'$'
+        else:
+            regex = r"^" + right + r"$"
+        return re.match(regex, left) is not None
     except re.error as e:
         raise CompareError(u"Error during matching: {!r}".format(e))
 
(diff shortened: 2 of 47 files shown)


privacyidea/lib/crypto.py
```diff
@@ -74,12 +74,7 @@
 from cryptography.hazmat.primitives.asymmetric import padding as asym_padding
 
 import passlib.hash
-if hasattr(passlib.hash.pbkdf2_sha512, "encrypt"):
-    hash_admin_pw = passlib.hash.pbkdf2_sha512.encrypt
-elif hasattr(passlib.hash.pbkdf2_sha512, "hash"):
-    hash_admin_pw = passlib.hash.pbkdf2_sha512.hash
-else:
-    raise Exception("No password hashing method available")
+hash_admin_pw = passlib.hash.pbkdf2_sha512.hash
```
cornelinux, line 77: I think we had this, since the interface had changed. Is this obsolete for us, now?

privacyidea/lib/utils/compare.py
```diff
@@ -97,7 +97,13 @@ def _compare_matches(left, comparator, right):
     :return: True or False
     """
     try:
-        return re.match("^" + right + "$", left) is not None
+        # check for regex modes
+        m = re.match(r'^(\(\?[a-zA-Z]+\))(.+)$', right)
+        if m and len(m.groups()) == 2:
```
cornelinux, line 102: Why is this special case? Would we get different results?

tests/test_api_lib_utils.py
```diff
@@ -75,8 +76,12 @@ def test_03_verify_auth_token(self):
                                          "resolver": "resolverX"},
                                 key=key,
                                 algorithm="RS256")
-        self.assertRaisesRegexp(AuthError, "The username hanswurst is not allowed to impersonate via JWT.",
-                                verify_auth_token, auth_token=auth_token, required_role="user")
+        with warnings.catch_warnings():
```
cornelinux, line 79: Tell me more...

privacyidea/lib/utils/compare.py
plettich: We support [Regex Modifiers](https://regex.programmingpedia.net/en/tutorial/5138/regex-modifiers-flags-) which need to be put at the beginning of the pattern (before the `"^"`).

tests/test_api_lib_utils.py
plettich: Well, it catches warnings :-) ... and ignores them, in this case a `DeprecationWarning` emitted from `pytest`. In the end i did not wrap all emitters of deprecation warnings.

privacyidea/lib/crypto.py
cornelinux: OK. Thanks.

PR #2300, opened 2020-07-21
token-janitor can set tokenrealms

### LINKED ISSUE #2299: Allow token janitor to set tokenrealms
The tokenjanitor should be capable of setting tokenrealms.

This could be done like this:

    privacyidea-token-janitor find --asigned FALSE --action tokenrealms --tokenrealms realm1,realm2

This would set the tokenrealms realm1 and realm2. Other realms would be removed!

To completely remove all tokenrealms from a token, the admin could call

    privacyidea-token-janitor find --asigned FALSE --action tokenrealms --tokenrealms ,

cornelinux:
The token-janitor can now set the tokenrealms,
either to a list of given realms or it can completely
clean the tokenrealms of the found tokens.

Closes #2299

### DIFF
--- a/doc/workflows_and_tools/tools/index.rst
+++ b/doc/workflows_and_tools/tools/index.rst
@@ -24,6 +24,8 @@ type, description or token info.
 It can unassign, delete or disable those tokens and it can set additional
 tokeninfo or descriptions.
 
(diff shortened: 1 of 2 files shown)


cornelinux:
I needed this other functionality and I thought it was well located in the tokenjanitor.

PR #2306, opened 2020-07-22
Allow to set email address for Email Tokens

### LINKED ISSUE #2305: Can't set the email address of an Email Token using the /token/set/(serial) endpoint
### Top-level intent

What did you try to achieve?
Set the email address of Email Tokens using then **/token/set/(serial)** endpoint.

### Expected outcome

Being able to set the email address of Email Tokens using then **/token/set/(serial)** endpoint.

### Actual outcome

It's currently not possible to set the email address of an Email Token through the **/token/set/(serial)** endpoint.

### Configuration

* **privacyIDEA version**: 3.3.1
* **Installation method**: github, PyPI, RHEL package dependencies
* **Python version**: 3.6.9
* **Webserver**: NGINX Unit 1.17.0
* **Token database**: Oracle 19c (using SQLAlchemy cx_oracle dialect)

FedeOmoto:
It's currently not possible to set the email address of an Email Token through the /token/set/(serial) endpoint.

Fixes #2305

### DIFF
--- a/privacyidea/lib/token.py
+++ b/privacyidea/lib/token.py
@@ -1687,6 +1687,30 @@ def set_description(serial, description, user=None):
 
 @log_with(log)
 @check_user_or_serial
+def set_email(serial, email, user=None):
+    """
+    Set the email address of a token
+
+    :param serial: The serial number of the token (exact)
+    :type serial: basestring
+    :param email: The email address for the token
+    :type description: str
+    :param user: The owner of the tokens, which should be modified
+    :type user: User object
+    :return: number of modified tokens
+    :rtype: int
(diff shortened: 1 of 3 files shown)


privacyidea/lib/token.py
```diff
@@ -1685,6 +1685,30 @@ def set_description(serial, description, user=None):
     return len(tokenobject_list)
 
 
+@log_with(log)
+@check_user_or_serial
+def set_email(serial, email, user=None):
```
cornelinux, line 1690: This is a tokentype specific field. Such functionality should not be added here.

PR #2307, opened 2020-07-23
Use SQLALchemy's URL class in censor_conn_string

### LINKED ISSUE #2304: Connection strings starting with "oracle+cx_oracle://" are not sanitized in the logs
### Top-level intent

What did you try to achieve?

I was checking log sanitization and found the database password in the logs.

### Steps to reproduce

1. Add a sql resolver with driver "oracle+cx_oracle"
2. Go to the users list
3. Check the logs and find the line matching "using the connect string"

### Expected outcome

Expected the sanitized connect string "oracle+cx_oracle://user:xxxx@dsnname"

### Actual outcome

What did actually happen?

Found the original connect string "oracle+cx_oracle://user:password@dsnname"

### Configuration

* **privacyIDEA version**: 3.3.3
* **Installation method**: pip
* **Python version**: 3.6.8
* **Webserver**: Nginx Unit 1.17
* **Token database**: Oracle 19

### Log file

Redacted extract

```
[INFO][privacyidea.lib.resolvers.SQLIdResolver:559] using the connect string oracle+cx_oracle://<user>:<password>@<dsnname>
```

gallolp:
Also fix tests to match the new censores pass ***
(Fixes #2304)

This is my proposed solution to #2304 using sqlalchemy's own URL class. Seems like a nice way to support present and future sqlalchemy dialects that may have a non rfc compliant url such as "oracle+cx_oracle://".

Tests passing:

```
tests/test_lib_utils.py::UtilsTestCase::test_18_censor_connect_string PASSED                              [ 52%]
```

Hope it helps. Thanks!

### DIFF
--- a/privacyidea/lib/utils/__init__.py
+++ b/privacyidea/lib/utils/__init__.py
@@ -1077,20 +1077,12 @@ def censor_connect_string(connect_string):
     """
     Take a SQLAlchemy connect string and return a sanitized version
     that can be written to the log without disclosing the password.
-    The password is replaced with "xxxx".
+    The password is replaced with "***".
     In case any error occurs, return "<error when censoring connect string>"
     """
     try:
-        parsed = urlparse(connect_string)
-        if parsed.password is not None:
-            # We need to censor the ``netloc`` attribute: user:pass@host
-            _, host = parsed.netloc.rsplit("@", 1)
-            new_netloc = u'{}:{}@{}'.format(parsed.username, 'xxxx', host)
-            # Convert the URL to six components. netloc is component #1.
-            splitted = list(parsed)
-            splitted[1] = new_netloc
-            return urlunparse(splitted)
-        return connect_string
+        parsed = sqlalchemy.engine.url.make_url(connect_string)
+        return repr(parsed)
     except Exception:
         return "<error when censoring connect string>"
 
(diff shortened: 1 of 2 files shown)


privacyidea/lib/utils/__init__.py
```diff
@@ -1077,20 +1077,12 @@ def censor_connect_string(connect_string):
...
-            # We need to censor the ``netloc`` attribute: user:pass@host
-            _, host = parsed.netloc.rsplit("@", 1)
-            new_netloc = u'{}:{}@{}'.format(parsed.username, 'xxxx', host)
-            # Convert the URL to six components. netloc is component #1.
-            splitted = list(parsed)
-            splitted[1] = new_netloc
-            return urlunparse(splitted)
-        return connect_string
+        parsed = sqlalchemy.engine.url.make_url(connect_string)
+        return repr(parsed)
```
plettich, line 1085: it seems that in python2 `repr()` does not only call the `__repr__()` function of the object. In this case it breaks the tests if the username does contain non-ASCII characters.
Could You change this to `return parsed.__repr__()` which does work in Python 2.7.

gallolp: Hi @plettich thanks for reviewing this PR. Of course I will change it. I'm sorry I only tested Python 3. Good catch.

gallolp:
Will resubmit with proposed changes.

PR #2321, opened 2020-08-08
Dynamic token types

cornelinux:


doc/faq/customization.rst
```diff
@@ -183,4 +183,30 @@ Some tokens allow a special customization.
...
+paper token :ref:`paper_token_customize`.
+
+.. _customize_3rd_party_tokens:
+
+New token classes
+~~~~~~~~~~~~~~~~~
+
+You can write your own Python token class, that should look similar to
+the Python modules located in the directory *privacyidea/lib/tokens/*.
+You should only add on token class per Python module.
```
plettich, line 195: ... one token ...

doc/faq/customization.rst
```diff
@@ -183,4 +183,30 @@ Some tokens allow a special customization.
...
\ No newline at end of file
+paper token :ref:`paper_token_customize`.
+
+.. _customize_3rd_party_tokens:
+
+New token classes
+~~~~~~~~~~~~~~~~~
+
+You can write your own Python token class, that should look similar to
+the Python modules located in the directory *privacyidea/lib/tokens/*.
```
plettich, line 194: it should also inherit from/implement `privacyidea.lib.tokenclass.TokenClass` (which functions need to be implemented?)

doc/installation/system/inifile.rst
```diff
@@ -202,3 +202,16 @@ A JWT can be created like this::
...
 
+.. _picfg_3rd_party_tokens:
+
+3rd party token types
+---------------------
+
+You can add 3rd party token types to privacyIDEA. Read more about this
+at :ref:`customize_3rd_party_tokens`.
+
+You then can add a list of your 3rd party token class modules in ``pi.cfg``
```
plettich, line 213: -then

doc/faq/customization.rst
cornelinux: So you mean we should add some more meat here like, 
* inheritance from which classes, 
* slight description of which method is probably necessary to have.

However, I did not want to write a complete development guide ;-) 
But I will add some.

cornelinux: I do not expect anyone to write a new 3rd party tokenclass soon.
I basically wrote this for us, to remember how to configure a new tokenclass, if we need to implement a token class, which we do not want to add to the core privacyIDEA.
However, I added some more meat in b7e67443.

doc/faq/customization.rst
cornelinux: done in b7e67443

doc/installation/system/inifile.rst
cornelinux: done in b7e67443

privacyidea/lib/config.py
```diff
@@ -716,24 +715,13 @@ def get_token_list():
...
-        coco = ",".join(lines)
-        for module in coco.split(','):
-            if module.strip() != '\\':
-                module_list.add(module.strip())
+    # Dynamic token modules
+    dynamic_token_modules = get_app_config_value("PI_TOKEN_MODULES")
+    if dynamic_token_modules:
+        # In the pi.cfg you can specify a list or set of 3rd party token modules like
+        #    PI_TOKEN_MODULES = [ "myproj.tokens.tok1", "myproj.tokens.tok2" ]
+        module_list.update(dynamic_token_modules)
```
plettich, line 724: I know it documented several times but people tend not to read documentation (at least i ran into this...).
We could use
`module_list.update(dynamic_token_modules if isinstance(dynamic_token_modules, list) else [dynamic_token_modules])`
to allow a single string in the configuration. What do You think?

cornelinux: Hm, the documentation seems not quite correct.
``module_list`` is actually a **set**. And the ``update`` method of sets works for lists/sets and strings.

    module_list.update("mytoken")

and 
    
    module_list.update(["mytoken"])

work the same.

plettich: actually not, in the first case every character of the string will be added to the set

cornelinux: :-( I should have user a string longer than one character.

Again. We have a similar issue with other configurations like the ``SUPERUSER_REALM`` which also expects a list and would end up in wired results, if you would give a string.
So I am hesitant to do a failsafe here and not have a failsafe somewhere else, which is used much more often.

cornelinux: I will do a helper function, which we might use at other places in the future.

cornelinux: done in d1f5619c

PR #2328, opened 2020-08-18
Do not raise on missing Origin.

### LINKED ISSUE #2327: Do not raise on missing Origin in check_otp() for WebAuthn
Currently a missing Origin HTTP header in a request to authenticate with a WebAuthn token will cause a ValueError to be raised and a 500 response to be returned. As this is a client-side error in the response to the challenge, we should instead return a 200 response with a message indicating the response to the challenge was not valid.

NuvandaPV:
fixes #2327


plettich (approved):
looks good

PR #2335, opened 2020-08-20
Remove lightning from webauthn default allowed transports

### LINKED ISSUE #2333: Remove lightning from default WebAuthn transports
Apple has in the meantime finished their implementation of WebAuthn and they consider lightning tokens as having usb transport. Thus we should remove lightning from the default value for WEBAUTHN_ALLOWED_TRANSPORTS.

NuvandaPV:
fixes #2333


NuvandaPV:
@plettich Should I update the translation files as well, or do we do that in bulk with the release?

plettich (approved):
looks good, we'll add the translation when doing the release.

PR #2339, opened 2020-08-20
Treat unsupported attribution as none attribution

### LINKED ISSUE #2336: Treat unsupported attestation as none attestation
Currently we do not allow the user to roll out a token, if we do not understand the attestation statement format, even if the administrator has explicitly allowed rolling out a token with an attestation statement format of NONE. Instead we should treat unsupported attestation statement formats equivalent to NONE attestation.

NuvandaPV:
fixes #2336


NuvandaPV:
This also fixes a bug, where tokens with no attestation couldn't be rolled out, if no description was provided, because the default description was not being set

PR #2346, opened 2020-08-22
2303/user obj in auth

cornelinux:
@laclaro, please review this pull request, since you orginally posted the issue.
I added exactly the test case of a user with now token, who gets a token enrolled by a pre-event handler when logging in against /auth.

### DIFF
--- a/privacyidea/api/auth.py
+++ b/privacyidea/api/auth.py
@@ -267,6 +267,7 @@ def get_auth_token():
     elif verify_db_admin(username, password):
         role = ROLE.ADMIN
         admin_auth = True
+        log.info("Local admin '{0!s}' successfully logged in.".format(username))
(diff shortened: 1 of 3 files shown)


privacyidea/api/auth.py
```diff
@@ -289,6 +300,11 @@ def get_auth_token():
                                                     superuser_realms=
                                                     superuser_realms)
         details = details or {}
+        if db_admin_exist(loginname) and realm == get_default_realm():
+            # If there is a local admin with the same login name as the user
+            # in the default realm, we inform about this in the log file.
+            log.warning("A user '{0!s}' exists as local admin and as user in "
+                        "your default realm!".format(loginname))
```
laclaro, line 307: I like this solution.

laclaro (approved):

PR #2352, opened 2020-08-26
Update changelog

cornelinux:
with user_obj in auth, webauthn

I updated the changelog once more with the necessary, final fixes and enhancements.


plettich (approved):
looks good

PR #2357, opened 2020-08-27
RequestMangler Handler Module Doc - Change user to username

thisisnotacompanyaccount:
The attribute user does not exist in the authentication request, it's called username.


laclaro (approved):
This is true. Thank you for this hint. However, we usually create first an issue and link the pull request accordingly to know what we are reviewing against, at least for larger changes.

PR #2360, opened 2020-08-27
2345/fix url decode

### LINKED ISSUE #2345: Fix url_decode()
The length-calculation for the padding of b64-encoded strings in u2f- and WebAuthn-Tokens fails to account for whitespace in the encoded string.

NuvandaPV:
fixes #2345


privacyidea/lib/tokens/u2f.py
```diff
@@ -49,7 +51,8 @@ def url_decode(url):
     :return: the decoded string
     :rtype: bytes
     """
-    pad_len = -len(url) % 4
+    pad_len = -len(re.sub('[^A-Za-z0-9-_+/]+', '', to_unicode(url))) % 4
```
plettich, line 54: could you write a short comment why we need this?

plettich (changes requested):

privacyidea/lib/tokens/u2f.py
plettich: fixed in b30af4d3

plettich (approved):

PR #2367, opened 2020-09-01
Fix check for yubikey AES-Mode OTP key length

### LINKED ISSUE #2366: length check for Yubikey AES mode enrollment doesn't work
When enrolling a YubiKey in AES mode, there is a "Test YubiKey" input field which can be used to extract the OTP length and possible public UID length from a given OTP value.
This currently does not work correctly, the corresponding lengths are not detected.

plettich:
Fixes #2366


privacyidea/static/components/token/controllers/tokenControllers.js
```diff
@@ -294,6 +296,11 @@ myApp.controller("tokenEnrollController", ["$scope", "TokenFactory", "$timeout",
         } else {
             $scope.form.genkey = true;
         }
+        if ($scope.form.type === "yubikey") {
+            delete $scope.form.otplen;
+        } else {
+            $scope.form.otplen = 6;
```
cornelinux, line 302: Are there side effects like?

Currently if I would change from HOTP with otplen=8 to TOTP, I would keep the otplen=8. Very convinient!

In this case, we would **always reset** to 6. This is a change in behaviour and inconvenient for 99% of the users, who are not using tokentype yubikey anyways. Is there a way to avoid this?

cornelinux (changes requested):

privacyidea/static/components/token/controllers/tokenControllers.js
plettich: The general `else`-part makes not sense, correct. But we should think about what we want to achieve here:
If the user switches from `HOTP` with `otplen=8` to `Yubikey AES` (which only supports `otplen=[32, 44, 48]`) the selected `otplen`-dropdown would be empty (and no `otplen` parameter would send in the `/token/init` request).
If the user then switches back to `TOTP` after changing the `otplen` to e.g. `44`, the `otplen`-dropdown would be empty again.

plettich: Fixed in 46885a7

cornelinux (approved):

PR #2376, opened 2020-09-07
add commands to set and clear default realm to pi-manage

laclaro:
Extends the realm command section of pi-manage.
- create supports multiple resolvers
- set_default
- clear_default
- delete

### DIFF
--- a/pi-manage
+++ b/pi-manage
@@ -979,7 +979,7 @@ def set_default(realm):
 @realm_manager.command
 def clear_default():
     """
-    Set the given realm to default
+    Unset the default realm
     """
     from privacyidea.lib.realm import set_default_realm
     set_default_realm(None)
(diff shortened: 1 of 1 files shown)


pi-manage
```diff
@@ -106,7 +106,7 @@ manager = Manager(app)
 admin_manager = Manager(usage='Create new administrators or modify existing '
                               'ones.')
 backup_manager = Manager(usage='Create database backup and restore')
-realm_manager = Manager(usage='Create new realms')
+realm_manager = Manager(usage='Create new realms, set the default realm')
```
cornelinux, line 109: In other usage descriptions we use 'and' or 'or'. I would like better to 'Create new realms or set the default realm.'

cornelinux (changes requested):
Would we also need a "delete realm"?

pi-manage
laclaro: done. I also added the realm deletion function. I also added simple multiple resolver support for the create command.

cornelinux (approved):

PR #2378, opened 2020-09-08
Fix deletion of registraion token

### LINKED ISSUE #2356: registration token is not deleted?
In the current development master branch for 3.4 it looks like as if the registration token is not deleted anymore after a successful authentication.

It could be, that the multi challenge implementation changed s.th.

Anyways, we should have a test with /validate/check of the registration token which then actually checks if the token is removed.
(I thought we had one)

Please verify.

cornelinux:
If the no_auth_counter is set, we now correctly delete
the registration token by adding a new token method
"post_success".

Fixes #2356


cornelinux:
@laclaro I can explain this to you.

laclaro (approved):

Issue #2094, opened 2020-03-16
Implement a mock authenticator for WebAuthn


jp-hoehmann:
In order to more thoroughly test the WebAuthn implementation we need a mock of an authenticator able to sign arbitrary statements.

PR #2382, opened 2020-09-11
Support webAuthn token without sign_count

### LINKED ISSUE #2361: Support webauthn token without sign_count
Some webauthn tokens (f.e. Apple devices) do not implement a sign_count. We do not currently support this.

NuvandaPV:
fixes #2361


NuvandaPV:
pep8ify

NuvandaPV:
@pep8speaks pep8ify

NuvandaPV:
@plettich In theory there should be another test case here, that asserts that the AuthenticationRejectedException is not raised if sign_count is unsupported (i.e. both the stored sign_count and the sign_count in the response are zero), but I don't want to add even more binary blobs to test_lib_tokens_webauthn.py. The real solution would be implementing #2094, but that is of course a bigger undertaking.

plettich (approved):
Looks good.

PR #2383, opened 2020-09-12
Display challenge_text_header and pin_change message

### LINKED ISSUE #2364: problem with challenge_text_header and multi challenge
The policy challenge_test_header causes the challenge text "Please enter a new PIN" in a pin change via validate (#2163) not to be sent to the client.

The policy for the challenge header comes into play as a post policy
https://github.com/privacyidea/privacyidea/blob/ef8f6bdb5541f02238db215cc65e8ecaa595a5e7/privacyidea/api/validate.py#L195

If the policy is set, the message is expected in content -> detail -> messages, but the multichallenge response text happens to be in a field called content -> detail -> message (singular). content is the json-response of the validate/check request.

https://github.com/privacyidea/privacyidea/blob/7ee038a30fad36021ab0bbacbfd05e611ba42680/privacyidea/api/lib/postpolicy.py#L723

Maybe this should be set somewhere in the [challengeresponsedecorators.py](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/challengeresponsedecorators.py).

_If_ there is always only one item in the messages list, we could simply add a line after the message is set in the reply_dict

```python
def _create_pin_reset_challenge(token_obj, message, challenge_data=None):
   ...
    reply_dict["message"] = message
    reply_dict["messages"] = [message]
```

cornelinux:
In case of a challenge_text_header the pin_change message would
not have been displayed. This assures, that the pin_change
message is contained in reply_dict["messages"] (this list of
reply messages) and thus is correctly handled with the
challenge_text_header.

closes #2364

### DIFF
--- a/privacyidea/lib/challengeresponsedecorators.py
+++ b/privacyidea/lib/challengeresponsedecorators.py
@@ -62,7 +62,9 @@ def _create_pin_reset_challenge(token_obj, message, challenge_data=None):
                                       "serial": token_obj.token.serial,
                                       "type": token_obj.token.tokentype}]
     reply_dict["message"] = message
+    reply_dict["messages"] = [message]
     reply_dict["transaction_id"] = db_challenge.transaction_id
+    # TODO: This line is deprecated: Add the information for the old administrative triggerchallenge
     reply_dict["transaction_ids"] = [db_challenge.transaction_id]
 
     return reply_dict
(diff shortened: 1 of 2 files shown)


laclaro (approved):
This looks good. Thank's for checking the possible cases for the messages field.

PR #2389, opened 2020-09-15
catch FileNotFoundError in get_templates

### LINKED ISSUE #2374: CA connector not accessible if template file not found
The /caconnector/ [GET] endpoint tries to read the templates file. The request fails if the file is not there or not readable.

The exception looks as follows:

```python
[2020-09-07 07:02:07,480][7805][140558478960384][ERROR][privacyidea.app:1892] Exception on /caconnector/ [GET]
  File "/opt/privacyidea/lib/python3.6/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/opt/privacyidea/lib/python3.6/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/opt/privacyidea/lib/python3.6/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/opt/privacyidea/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/opt/privacyidea/lib/python3.6/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/opt/privacyidea/lib/python3.6/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/opt/privacyidea/lib/python3.6/site-packages/privacyidea/lib/log.py", line 155, in log_wrapper
    return func(*args, **kwds)
  File "/opt/privacyidea/lib/python3.6/site-packages/privacyidea/api/lib/prepolicy.py", line 154, in policy_wrapper
  File "/opt/privacyidea/lib/python3.6/site-packages/privacyidea/api/caconnector.py", line 54, in get_caconnector_api
    return_config=True)  # the endpoint is only accessed by admins
  File "/opt/privacyidea/lib/python3.6/site-packages/privacyidea/lib/caconnector.py", line 169, in get_caconnector_list
    c_obj = c_obj_class(conn.name, data)
  File "/opt/privacyidea/lib/python3.6/site-packages/privacyidea/lib/caconnectors/localca.py", line 285, in __init__
    self.set_config(config)
  File "/opt/privacyidea/lib/python3.6/site-packages/privacyidea/lib/caconnectors/localca.py", line 327, in set_config
    self.templates = self.get_templates()
  File "/opt/privacyidea/lib/python3.6/site-packages/privacyidea/lib/caconnectors/localca.py", line 471, in get_templates
    with open(self.template_file, 'r') as content_file:
FileNotFoundError: [Errno 2] No such file or directory: '/nonexisting/path/to/templates.conf'
```

The error states the problem, but in case of one broken file entry, no connectors are listed. We should introduce a softer handling for this. Maybe in [localca.py](https://github.com/privacyidea/privacyidea/blob/4929d4e0a7a11cbc564a88d034f7903665ba8024/privacyidea/lib/caconnectors/localca.py#L465) like this. But I'm not sure where the CAError goes to...
```python
    def get_templates(self):
        """
        Return the dict of available templates, which are read from the
        template YAML file.

        :return: dict
        """
        content = {}
        if self.template_file:
            if os.access(self.template_file, os.R_OK):
                with open(self.template_file, 'r') as content_file:
                    file_content = content_file.read()
                    content = yaml.safe_load(file_content)
            else:
                CAError("The Certificates template file {0!s} is not readable.",
                        self.template_file)
        return content
```

laclaro:
We catch the error if the file cannot be found or privacyidea is not permitted to read it.

closes #2374

### DIFF
--- a/privacyidea/lib/caconnectors/localca.py
+++ b/privacyidea/lib/caconnectors/localca.py
@@ -40,6 +40,7 @@ import logging
 import os
 from six import string_types
 from six.moves import input
+import traceback
 
 log = logging.getLogger(__name__)
 
@@ -472,9 +473,10 @@ class LocalCAConnector(BaseCAConnector):
                 with open(self.template_file, 'r') as content_file:
                     file_content = content_file.read()
                     content = yaml.safe_load(file_content)
-            except (FileNotFoundError, PermissionError):
+            except EnvironmentError:
                 log.warning("Template file {0!s} for {1!s} not found or "
                             "not permitted.".format(self.template_file, self.name))
+                log.debug('{0!s}'.format(traceback.format_exc()))
         return content
 
     def publish_cert(self):
--- a/privacyidea/lib/caconnectors/localca.py
+++ b/privacyidea/lib/caconnectors/localca.py
@@ -472,9 +472,9 @@ class LocalCAConnector(BaseCAConnector):
                 with open(self.template_file, 'r') as content_file:
                     file_content = content_file.read()
                     content = yaml.safe_load(file_content)
-            except FileNotFoundError:
-                log.warning("Template file {0!s} for {1!s} not found.".format(
-                    self.template_file, self.name))
+            except (FileNotFoundError, PermissionError):
+                log.warning("Template file {0!s} for {1!s} not found or "
+                            "not permitted.".format(self.template_file, self.name))
         return content
 
     def publish_cert(self):
--- a/privacyidea/lib/caconnectors/localca.py
+++ b/privacyidea/lib/caconnectors/localca.py
@@ -468,9 +468,13 @@ class LocalCAConnector(BaseCAConnector):
         """
         content = {}
         if self.template_file:
-            with open(self.template_file, 'r') as content_file:
-                file_content = content_file.read()
-                content = yaml.safe_load(file_content)
+            try:
+                with open(self.template_file, 'r') as content_file:
+                    file_content = content_file.read()
+                    content = yaml.safe_load(file_content)
+            except FileNotFoundError:
+                log.warning("Template file {0!s} for {1!s} not found.".format(
+                    self.template_file, self.name))
         return content
 
     def publish_cert(self):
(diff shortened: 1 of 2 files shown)


privacyidea/lib/caconnectors/localca.py
```diff
@@ -468,9 +468,13 @@ def get_templates(self):
...
         content = {}
         if self.template_file:
-            with open(self.template_file, 'r') as content_file:
-                file_content = content_file.read()
-                content = yaml.safe_load(file_content)
+            try:
+                with open(self.template_file, 'r') as content_file:
+                    file_content = content_file.read()
+                    content = yaml.safe_load(file_content)
+            except (FileNotFoundError, PermissionError):
```
cornelinux, line 475: I think you could use ``EnvironmentError`` instead and it would run on python 2 and 3.

privacyidea/lib/caconnectors/localca.py
```diff
@@ -468,9 +468,13 @@ def get_templates(self):
...
+                log.warning("Template file {0!s} for {1!s} not found or "
+                            "not permitted.".format(self.template_file, self.name))
```
cornelinux, line 477: If the admin is really confused about this, "oh, but this file exists!", then please do as we do in a lot of other places and also add a 

    log.debug('{0!s}'.format(traceback.format_exc()))

Need to import ``traceback`` before.

This way the admin can turn on debug level and knows, why the file can not be read.

cornelinux (changes requested):

privacyidea/lib/caconnectors/localca.py
laclaro: Thank's for the hint. That does the trick. EnvironmentError -> IOError, OSError, ...

privacyidea/lib/caconnectors/localca.py
laclaro: done

cornelinux: Oups, we started to use 

   log.debug(u'{0!s}'...

if the traceback contains a hint to the problem, like the filename, we could have a filename with a non-ascii character and then the log.debug would cause another exception.
I will add the u"".

cornelinux (approved):

cornelinux:
I will merge after the test runs.

PR #2393, opened 2020-09-18
Add transparent documentation for smstext policy

laclaro:
Add small note in the documentation of the smstext policy.

Add an explicit hint to the <otp> tag to emailtext.

### DIFF
--- a/doc/policies/authentication.rst
+++ b/doc/policies/authentication.rst
@@ -155,14 +155,14 @@ type: string
 This is the text that is sent via SMS to the user trying to
 authenticate with an SMS token.
 You can use the tags *<otp>* and *<serial>*.
-Longer texts must be enclosed in single quotes.
+Texts containing whitespaces must be enclosed in single quotes.
 
 Starting with version 2.20 you can use the tag *{challenge}*. This will add
 the challenge data that was passed in the first authentication request in the
 challenge parameter. This could contain banking transaction data.
 
-Include the custom message set by this policy by using the *{otp}* tag
-in the :ref:`sms_gateway_config`.
+In the :ref:`sms_gateway_config` the tag *{otp}* will be replaced by the custom
+message, set with this policy.
 
 Default: *<otp>*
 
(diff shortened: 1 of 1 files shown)


doc/policies/authentication.rst
```diff
@@ -155,11 +155,15 @@ type: string
 This is the text that is sent via SMS to the user trying to
 authenticate with an SMS token.
 You can use the tags *<otp>* and *<serial>*.
+Longer texts must be enclosed in single quotes.
```
cornelinux, line 158: I think *longer* is not quite correct. You must include it in quotes, if the text contains spaces, right?
A text like

    Hello_my_dear_your_OTP_value_is:<otp>

would still work. (SPACE usually splits multiple values in a policy action value)

doc/policies/authentication.rst
```diff
@@ -155,11 +155,15 @@ type: string
...
 
 Starting with version 2.20 you can use the tag *{challenge}*. This will add
 the challenge data that was passed in the first authentication request in the
 challenge parameter. This could contain banking transaction data.
 
+Include the custom message set by this policy by using the *{otp}* tag
+in the :ref:`sms_gateway_config`.
```
cornelinux, line 165: Honestly, I do not understand this :-/

doc/policies/authentication.rst
```diff
@@ -182,9 +186,9 @@ emailtext
 type: string
 
 This is the text that is sent via Email to be used with Email Token. This
-text should contain the OTP value.
+text should contain the OTP value *<otp>*.
```
cornelinux, line 189: "It should contain the OTP value" is basically  right. "the OTP value *<otp>*" is somehow incorrect, since *<otp>* is not the otp value but the tag.
It would be more clear to say

    This text should contain the tag *<otp>* or *{otp}* that is replaced with the actual OTP value.

Or only

    This text should contain the tag *<otp>* or *{otp}*.

Or leave it as is and add the information below at the tags, that the tag *{otp}* will actually fill in the otp value.

cornelinux (changes requested):

doc/policies/authentication.rst
laclaro: You are right, I thought the {otp} tag is not used in the message but only <otp>. That's why I added the notation.

But it is added here:
https://github.com/privacyidea/privacyidea/blob/5c265bba65258a24abde5c918ea4396ba41fca79/privacyidea/lib/tokens/emailtoken.py#L439

I favor adding both tag types to the list.

doc/policies/authentication.rst
laclaro: True. Changed.

doc/policies/authentication.rst
laclaro: Done

doc/policies/authentication.rst
cornelinux: In the long run I would probably would like to get rid of ``<otp>``. Or: If I would design it anew, I would not use ``<otp>`` anymore. So I would rather only add ``{otp}`` to the tag list.

cornelinux (approved):

PR #2400, opened 2020-09-22
use autocomplete='new-password'

### LINKED ISSUE #2292: Browser autocomplete in token assignment username field
### Top-level intent

I want to assign a token to a user.

### Steps to reproduce

1. /#!/token/enroll//
2. Begin typing username in form field

### Expected outcome

I should (only) see the user list generated by privacyidea

### Actual outcome

The username autocomplete list from the browser is shown in front of the list of privacyidea.
Pressing escape leads to both lists being closed...

From skimming through https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion it seems to me that the only proper way to prevent this is to rename the form fields to something other than `username`...

### Configuration

* **privacyIDEA version**: 3.3.3
* **Python version**: 3.7.3
* **Operating system**: Debian 10
* **Webserver**: Nginx/UWSGI
* **Token database**: (MySQL, PostgreSQL, ...) MySQL
* **Browser**: Firefox 77.0

### Screenshot (instead of log file ;-) )

Screenshot from 2020-07-10 11-52-43

laclaro:
autocomplete="new-password" makes the browser not use the autocomplete feature.

This pull request replaces the old autocomplete="no" in the token details, enrollment and several other places.

closes #2292

### DIFF
--- a/privacyidea/static/components/token/views/token.enroll.html
+++ b/privacyidea/static/components/token/views/token.enroll.html
@@ -28,6 +28,7 @@
         <div class="form-group" ng-hide="loggedInUser.role == 'user' && !checkRight('setdescription')">
             <label for="description" translate>Description</label>
             <input type="text" class="form-control" id="description"
+                   autocomplete="new-password"
                    placeholder="{{ 'Some nice words...'|translate }}"
                    ng-model="form.description" />
         </div>
--- a/privacyidea/static/components/directives/views/directive.assigntoken.html
+++ b/privacyidea/static/components/directives/views/directive.assigntoken.html
@@ -26,10 +26,12 @@ This is the directive for assigning tokens to users.
         -->
     <label for="otppin" translate>PIN</label>
     <input name="otppin" ng-model="newTokenObject.pin"
+            autocomplete="new-password"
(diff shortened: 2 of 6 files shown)


privacyidea/static/components/token/views/token.enroll.html
```diff
@@ -28,6 +28,7 @@ <h3 translate>Enroll a new token</h3>
         <div class="form-group" ng-hide="loggedInUser.role == 'user' && !checkRight('setdescription')">
             <label for="description" translate>Description</label>
             <input type="text" class="form-control" id="description"
+                   autocomplete="new-password"
```
cornelinux, line 31: I think it was not necessary here, since it is a description field?

cornelinux (changes requested):

privacyidea/static/components/token/views/token.enroll.html
laclaro: I double-checked. Apparently, it is needed, but I also cannot see why.

cornelinux (approved):

PR #2411, opened 2020-09-24
Several hardenings

### LINKED ISSUE #2407: Only use trusted algorithm with TRUSTED_JWT
While validating trusted JWTs in ``verify_auth_token()``, the same public key can
be used with more than one externally specified encryption algorithm as long as
it is included in the `TRUSTED_JWT_ALGOS` list.

~~~~python
  for trusted_jwt in trusted_jwts:
    try:
    if trusted_jwt.get("algorithm") in TRUSTED_JWT_ALGOS:
        j = jwt.decode(auth_token,
            trusted_jwt.get("public_key"),
            algorithms=TRUSTED_JWT_ALGOS)
~~~~

We should only do:

     algorithms=trusted_jwt.get("algorithm")

### LINKED ISSUE #2409: Remove pyCrypto even from tools
The tool ``privacyidea-decrypt-safeword.py`` still uses unmaintained ``pyCrypto`` library or the drop-in-replace-ment pycryptodome.
However, this tool is now mainted in the repo privacyidea/scripts, so we should remove it here.

Also check, if there are dependencies, that can be removed.

cornelinux:
Closes #2407 
Closes #2409


laclaro (approved):
Looks good to me. Thank you.

PR #2416, opened 2020-09-24
2412/authcache argon2

### LINKED ISSUE #2412: Use Argon2 as hash algo for the authcache
The authcache currently uses a week hashing algorithm. Although authcache is used seldomly and the cached data is not stored for a long period if time, we should move to argon2 from ``passlib.hash``.

### LINKED ISSUE #2413: Use Argon2 instead of PBKDF2 for local admins
Instead of using pbkdf2 we should use argon2 for local admins.

Note, that we still need to keep the backward compatibility and be able to verify pbkdf2 hashed passwords.

cornelinux:
Closes #2412 
Closes #2413


privacyidea/config.py
```diff
@@ -26,6 +28,14 @@
 """
 
 
+def random_password(size):
```
laclaro, line 31: Why not use generate_password from crypto.py?

laclaro: creates flask dependencies. Ok.

privacyidea/config.py
```diff
@@ -26,6 +28,14 @@
...
+    passwd = [random.choice(string.ascii_lowercase +
+                            string.ascii_uppercase + string.digits) for _x in range(size)]
+    # return shuffled password
+    random.shuffle(passwd)
```
laclaro, line 35: should we write log.warning?

privacyidea/config.py
```diff
@@ -106,7 +116,7 @@ class ProductionConfig(Config):
         'sqlite:///' + os.path.join(basedir, 'data.sqlite')
     #SQLALCHEMY_DATABASE_URI = "mysql://pi2:pi2@localhost/pi2"
     # This is used to encrypt the auth_token
-    SECRET_KEY = os.environ.get('SECRET_KEY') or 't0p s3cr3t'
+    SECRET_KEY = os.environ.get('SECRET_KEY') or random_password(24)
     # This is used to encrypt the admin passwords
     PI_PEPPER = "Never know..."
```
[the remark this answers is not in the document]
laclaro: Better not randomize, since the hashing would be different on every restart.

privacyidea/lib/crypto.py
```diff
@@ -193,7 +195,13 @@ def verify_with_pepper(passwordhash, password):
     # get the password pepper
     password = password or ""
     key = get_app_config_value("PI_PEPPER", "missing")
-    success = passlib.hash.pbkdf2_sha512.verify(key + password, passwordhash)
+
+    success = False
+    if passwordhash.startswith("$argon2"):
```
laclaro, line 220: We should add a not to the READ_BEFORE_UPDATE that passwords should be updated to use the new hash

laclaro (changes requested):

privacyidea/config.py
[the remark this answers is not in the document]
cornelinux: Right, this would require admins to set the password anew after each restart ;-)

privacyidea/config.py
cornelinux: Will do

privacyidea/lib/crypto.py
cornelinux: Will do.

Will add to Changelog, so we can remeber to note this in our email communication.

privacyidea/config.py
cornelinux: Done in 3b3b8a9c

privacyidea/lib/crypto.py
cornelinux: Done in 4156eea9

laclaro (approved):

PR #2422, opened 2020-09-29
Use secure way to compare secrets

### LINKED ISSUE #2415: Use compare_digest for secrets
To reduce possible risks of timing attacks we could in addition use ``hmac.compare_digest()`` (> python 2.7) or  ``secrets.compare_digest()`` (> python 3.6) in `lib/crypto.py`  in `SecretObjects.compare()` or similar situations.

cornelinux:
In several cases we compare secrets or OTP values.
Trying to do this in a more secure way.

Closes #2415

### DIFF
--- a/privacyidea/lib/tokens/passwordtoken.py
+++ b/privacyidea/lib/tokens/passwordtoken.py
@@ -64,7 +64,7 @@ class PasswordTokenClass(TokenClass):
             # getKey() returns bytes and since we can not assume, that the
             # password only contains printable characters, we need to compare
             # bytes strings here. This also avoids making another copy of 'key'.
-            if safe_compare(key, to_bytes(password)):
+            if safe_compare(key, password):
                 res = 0
 
             zerome(key)
--- a/privacyidea/lib/tokens/passwordtoken.py
+++ b/privacyidea/lib/tokens/passwordtoken.py
@@ -20,7 +20,7 @@ This file contains the definition of the password token class
 
 import logging
 
-from privacyidea.lib.crypto import zerome
+from privacyidea.lib.crypto import zerome, safe_compare
 from privacyidea.lib.tokenclass import TokenClass
 from privacyidea.lib.log import log_with
 from privacyidea.lib.decorators import check_token_locked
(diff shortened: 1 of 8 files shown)


cornelinux:
@plettich  Open for discussion.

privacyidea/lib/tokens/passwordtoken.py
```diff
@@ -64,7 +64,7 @@ def check_password(self, password):
             # getKey() returns bytes and since we can not assume, that the
             # password only contains printable characters, we need to compare
             # bytes strings here. This also avoids making another copy of 'key'.
-            if key == to_bytes(password):
+            if safe_compare(key, to_bytes(password)):
```
plettich, line 67: 1. `safe_compare` already does the conversion to bytes so this can be omitted here
2. We can call `self.secretObject.compare(password)` directly here, no need to decrypt the secret.

plettich (changes requested):
Looks good

privacyidea/lib/tokens/passwordtoken.py
cornelinux: I will change this an run the test.

cornelinux: It looks like unfortunately we can not do this, since secretObject.compare assumes to compare a hexlified secret. Which the password is not. So we can not take the shortcut!

~~~~python
            res = -1
            if self.secretObject.compare(password):
                res = 0

            return res
~~~~

It would however not be an issue of the PR and we should not rewrite the password token on a bigger scale in a patch release.
So I ask you to merge this. Thx.

plettich (approved):

cornelinux:
@plettich You asked me to and I dared a force-push after my local rebase. Hope everything stays smooth.

plettich:
Unfortunately then `to_bytes()` issue was introduced again... we should test these rebases/merges more often :-)

cornelinux:
I think I didn't fix it in the first place. Done in 76b272ab

plettich (approved):

PR #2425, opened 2020-09-29
Add "active" query param to GET /token

### LINKED ISSUE #2424: Allow GET /token list to filter by active tokens
It would be helpful to be able to filter the token list by active/inactive tokens.

Example: `GET https://piserver/token/?active=True`

This would reduce the list size for users that have several disabled or otherwise inactive tokens. Also it would allow better token management using the api (similar to what we can do with token janitor on the cli).

The implementation is fairly easy as the current [get_tokens_paginate](https://github.com/privacyidea/privacyidea/blob/3d82e9c8060fe45a25103d0b0240f5888d39afa2/privacyidea/lib/token.py#L426) implementation in the token lib already has an "active" parameter. It is just a matter of exposing it as a query param on GET /token api.

I'm willing to send a PR for your evaluation.

Thank you.

gallolp:
Add docstring
Add tests
(Fixes #2424)

Added que "active" query param to GET /token. Also added some tests to test with and without the parameter.

### DIFF
--- a/privacyidea/api/token.py
+++ b/privacyidea/api/token.py
@@ -358,6 +358,7 @@ def list_api():
     :query sortdir: asc/desc
     :query page: request a certain page
     :query assigned: Only return assigned (True) or not assigned (False) tokens
+    :query active: Only return active (True) or inactive (False) tokens
     :query pagesize: limit the number of returned tokens
     :query user_fields: additional user fields from the userid resolver of
         the owner (user)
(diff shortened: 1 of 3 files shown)


cornelinux (approved):
Looks good to me.

cornelinux:
@plettich would you take a 2nd look? Thx.

plettich (approved):
looks good

PR #2427, opened 2020-10-01
Fix error when logging in with REMOTE_USER

### LINKED ISSUE #2423: Audit object not initialized on login route
The Audit object for the current request is not initialized for the login route:
https://github.com/privacyidea/privacyidea/blob/88c36350fb9bded7ea97255e22cea6fac477dee2/privacyidea/webui/login.py#L57

When checking whether remote users are allowed:
https://github.com/privacyidea/privacyidea/blob/4d59eab2941308f742927851e3e4a45eee7ae739/privacyidea/api/lib/prepolicy.py#L1213
the `action_values()` function fails when trying to write to the audit log:
https://github.com/privacyidea/privacyidea/blob/d6fd3f83de5354c762fc1b8110179e1821302041/privacyidea/lib/policy.py#L2366

plettich:
In order to avoid cluttering the audit log for each login-page load, the
`audit_object` is not initialized for the login request.
To avoid calling functions on an uninitialized object the
`write_to_audit_log` parameter is added to the `is_remote_user_allowed()`
function which passes it on to the policy matching.

Fixes #2423

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -1197,13 +1197,15 @@ def mock_fail(req, action):
     raise Exception("This is an Exception in an external check function")
 
 
-def is_remote_user_allowed(req):
+def is_remote_user_allowed(req, write_to_audit_log=True):
     """
     Checks if the REMOTE_USER server variable is allowed to be used.
 
     .. note:: This is not used as a decorator!
 
     :param req: The flask request, containing the remote user and the client IP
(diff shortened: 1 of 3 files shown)


cornelinux (approved):
lgtm

plettich:
> lgtm, but some tests fail?

Huh, i didn't even request a review :-) Just wanted to see if the tests work.

cornelinux:
@plettich Please "repull" this to branch-3.4. THx

PR #2430, opened 2020-10-02
Use passlibs CryptContext for password hashes

plettich:
The
[`passlib.context.CryptContext`](https://passlib.readthedocs.io/en/stable/lib/passlib.context.html#the-cryptcontext-class) provides an easy way to handle several password hash algorithms.
Since changing from pbkdf2 to argon2 we still need to validate hashes with
the former method.
Unfortunately passlib does not support plain SHA256 hashes so we still need to check them separately.

Working on #2413

### DIFF
--- a/privacyidea/lib/crypto.py
+++ b/privacyidea/lib/crypto.py
@@ -78,11 +78,14 @@ from cryptography.hazmat.primitives.asymmetric import padding as asym_padding
 if not PY2:
     long = int
 
+# Number of hash rounds for the argon2 algorithm
+ROUNDS = 9
+
 # The CryptContext makes it easier to work with multiple password hash algorithms:
 # The first algorithm in the list is the default algorithm used for hashing.
 # When verifying a password hash, all algorithms in the context are checked
 # until one succeeds (or all fail).
-pass_ctx = CryptContext(['argon2', 'pbkdf2_sha512'], argon2__rounds=9)
+pass_ctx = CryptContext(['argon2', 'pbkdf2_sha512'], argon2__rounds=ROUNDS)
 
 FAILED_TO_DECRYPT_PASSWORD = "FAILED TO DECRYPT PASSWORD!"
 
--- a/privacyidea/lib/crypto.py
+++ b/privacyidea/lib/crypto.py
@@ -57,6 +57,7 @@ import ctypes
 import base64
 import traceback
 from six import PY2
+from passlib.context import CryptContext
 
 from privacyidea.lib.log import log_with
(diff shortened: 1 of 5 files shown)


privacyidea/lib/crypto.py
```diff
@@ -73,15 +74,17 @@
...
-
 
 if not PY2:
     long = int
 
+# The CryptContext makes it easier to work with multiple password hash algorithms:
+# The first algorithm in the list is the default algorithm used for hashing.
+# When verifying a password hash, all algorithms in the context are checked
+# until one succeeds (or all fail).
+pass_ctx = CryptContext(['argon2', 'pbkdf2_sha512'], argon2__rounds=9)
```
cornelinux, line 85: A bit nitpicking: Could you put the 

    ROUNDS=9

at the top of the file and the add ``argon2__rounds=ROUNDS``.
This might be ease the change of the number of rounds later.

privacyidea/lib/crypto.py
```diff
@@ -165,24 +168,29 @@ def hash(val, seed, algo=None):
...
-    Hash values with argon2
-    :param val:
-    :return:
+    Hash password with crypt context
+    :param password: The password to hash
+    :type password: str
+    :return: The hash string of the password
     """
-    pw_dig = argon2.using(rounds=ROUNDS).hash(password)
+    pw_dig = pass_ctx.hash(password)
```
cornelinux, line 178: How would we **assure** that argon2 is used!?

What happens if argon2_cffi is not installed. It will fallback to pbkdf2? 
I would rather like to enforce argon2. Can we do this?
With a cryptcontext, that only contains argon2?

Maybe we can have two contexts:

~~~~python
verify_ctx = CryptContext(['argon2', 'pbkdf2_sha512'], argon2__rounds=ROUNDS)
hash_ctx =  CryptContext(['argon2'], argon2__rounds=ROUNDS)
~~~~

To assure hashing with argon2

cornelinux (changes requested):

privacyidea/lib/crypto.py
plettich: According to the [passlib docs](https://passlib.readthedocs.io/en/stable/narr/context-tutorial.html#basic-usage), it always picks the first entry in the list for hashing unless told otherwise.
If no argon2 backend is installed, hashing fails completely.

privacyidea/lib/crypto.py
plettich: Done in 5b0334d7

cornelinux (approved):

PR #2431, opened 2020-10-02
add groups in scope.authz

### LINKED ISSUE #2428: group the actions in scope authorization
We can easily add some groups in all the actions in scope authorization.

There are some actions, that only add some things to the response and some actions that also modify the outcome of the authentication request.

To do so, we need to adapt the file ``lib/policy.py`` and adapt the definitions in 

~~~~python
SCOPE.AUTHZ: {
~~~~

and add certain groups like:

~~~~python
SCOPE.AUTHZ: {
            ACTION.AUTHORIZED: {
                'type': 'str',
                'desc': _("Allow the user to authenticate (default). If set to '{0!s}', "
                          "the authentication of the user will be denied.").format(AUTHORIZED.DENY),
                'value': [AUTHORIZED.ALLOW, AUTHORIZED.DENY]
                'group': 'modifying'
            },
~~~~

NNLemling:
add some groups for all the actions in scope authorization and adapt the definitions in lib/policy.py

closes #2428

### DIFF
--- a/privacyidea/lib/policy.py
+++ b/privacyidea/lib/policy.py
@@ -377,7 +377,9 @@ class GROUP(object):
     MACHINE = "machine"
     USER = "user"
     PIN = "pin"
-
+    MODIFYING_RESPONSE = "modifying response"
+    CONDITIONS = "conditions"
+    SETTING_ACTIONS = "setting actions"
 
 class MAIN_MENU(object):
     __doc__ = """These are the allowed top level menu items. These are used
@@ -389,6 +391,7 @@ class MAIN_MENU(object):
     CONFIG = "config"
     AUDIT = "audit"
     COMPONENTS = "components"
+
 
 
 class LOGINMODE(object):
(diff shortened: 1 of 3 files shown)


privacyidea/lib/policy.py
```diff
@@ -2021,77 +2024,100 @@ def get_static_policy_definitions(scope=None):
...
+                'group': GROUP.MODIFYING_RESPONSE,
             },
             ACTION.APPLICATION_TOKENTYPE: {
                 'type': 'bool',
                 'desc': _("Allow the application to choose which token types should be used "
                           "for authentication. Application may set the parameter 'type' in "
                           "the request. Works with validate/check, validate/samlcheck and "
-                          "validate/triggerchallenge.")
+                          "validate/triggerchallenge."),
+                'group': GROUP.CONDITIONS,
```
cornelinux, line 2036: I am sorry. This is actually a policy that allows the client to choose a tokentype. 
So this is rather not a condition.
I think it is...
...nothing. So I would leave the at misecellaneous. Please remove the group.

cornelinux (changes requested):
Please remove the one group

privacyidea/lib/policy.py
NNLemling: done in 8ce2e724

cornelinux (approved):

PR #2433, opened 2020-10-03
Allow to truncate longer audit columns

### LINKED ISSUE #1756: Allow to modify the truncated length of columns of pidea_audit
The lengths of the columns of the ``pidea_audit`` table are hardcoded in the ``audit_column_length`` dictionary in ``models.py``:

https://github.com/privacyidea/privacyidea/blob/96b3e18c63814b771242c1e42cd3ff7109e5b9db/privacyidea/models.py#L2555

If users want to increase the column lengths, they can manually alter the SQL table (see #525). However, if ``PI_AUDIT_SQL_TRUNCATE`` is True (as it is in new installations now), the audit class will truncate fields to the lengths specified in ``audit_column_length`` even if the actual column size might be greater. It would be nice if one could increase the truncated size without having to edit ``models.py``.

I think this could be fixed in two ways:
* We could allow users to overwrite parts of ``audit_column_length`` in ``pi.cfg`` and use these settings in ``models.py`` too. Then, if a user writes
  ```
  PI_AUDIT_COLUMN_LENGTHS = {"user": 40}
  ```
  in ``pi.cfg``, a ``pi-manage createdb`` would create an audit table with an enlarged ``user`` column, and also truncate values accordingly. This way might be tricky, because when we create the models, we do not have access to the Flask ``app.config`` because the Flask app doesn't exist yet. But there might be a way to circumvent that :-)
* Alternatively, we could require users to manually alter the column sizes, and only allow them to overwrite the truncated lengths, e.g.
    ```
    PI_AUDIT_SQL_TRUNCATE = True
    PI_AUDIT_SQL_TRUNCATED_LENGTHS = {"user": 40}
    ```
   which would cause the SQL audit module to truncate the ``user`` field to 40 characters, not 20.

cornelinux:
The admin can raise the length of audit log columns
in the database. To still use a valid truncation, the
admin can define in pi.cfg a setting
PI_AUDIT_SQL_COLUMN_LENGTH as a dictionary with the
length of the columns that were altered.

Closes #1756

### DIFF
--- a/doc/audit/index.rst
+++ b/doc/audit/index.rst
@@ -136,13 +136,24 @@ This will read the configuration (only the database uri) from the config file
 Table size
 ~~~~~~~~~~
 
-Sometimes the entires to be written to the database may be longer than the
+Sometimes the entries to be written to the database may be longer than the
 column in the database. You can either enlarge the columns in the database or
 you can set
 
    PI_AUDIT_SQL_TRUNCATE = True
 
 in ``pi.cfg``. This will truncate each entry to the defined column length.
+
+If the default column length of the audit log table are to shore, you can
+make the columns longer in the database. You should then also increase this
+information for truncation with a setting in the config file ``pi.cfg`` using
+the setting:
+
+    PI_AUDIT_SQL_COLUMN_LENGTH = {"user": 100,
+                                  "policies": 1000}
+
+which will increase truncation of the user column to 100 and the policies
+column to 1000.
 
 .. _logger_audit:
 
--- a/doc/audit/index.rst
+++ b/doc/audit/index.rst
@@ -137,23 +137,25 @@ Table size
 ~~~~~~~~~~
 
 Sometimes the entries to be written to the database may be longer than the
-column in the database. You can either enlarge the columns in the database or
-you can set
+column in the database. You should set
 
    PI_AUDIT_SQL_TRUNCATE = True
 
 in ``pi.cfg``. This will truncate each entry to the defined column length.
 
-If the default column length of the audit log table are to shore, you can
-make the columns longer in the database. You should then also increase this
-information for truncation with a setting in the config file ``pi.cfg`` using
-the setting:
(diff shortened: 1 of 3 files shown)


doc/audit/index.rst
```diff
@@ -136,14 +136,25 @@ This will read the configuration (only the database uri) from the config file
...
-Sometimes the entires to be written to the database may be longer than the
+Sometimes the entries to be written to the database may be longer than the
 column in the database. You can either enlarge the columns in the database or
 you can set
 
    PI_AUDIT_SQL_TRUNCATE = True
 
 in ``pi.cfg``. This will truncate each entry to the defined column length.
 
+If the default column length of the audit log table are to shore, you can
```
plettich, line 147: ... short ...

doc/audit/index.rst
```diff
@@ -136,14 +136,25 @@ This will read the configuration (only the database uri) from the config file
...
+make the columns longer in the database. You should then also increase this
+information for truncation with a setting in the config file ``pi.cfg`` using
```
plettich, line 149: This sentence is kind of weird. How about:
--------------------------8<------------------------------------
Then you should tell privacyIDEA about the changes. In :ref:`cfgfile` ``pi.cfg`` add the setting:

    PI_AUDIT_SQL_COLUMN_LENGTH = {"user": 100,
                                  "policies": 1000}

which will increase truncation of the user column to 100 and the policies
column to 1000.
--------------------------8<------------------------------------

plettich (changes requested):
I'll add a short test for this setting.

doc/audit/index.rst
cornelinux: done in 222a103b

doc/audit/index.rst
cornelinux: done in 222a103b

plettich (approved):
Looks good, waiting for travis...

PR #2434, opened 2020-10-03
Fix log message for userinfo policy

cornelinux:
If a user_object does not exist in the
request a wrong log message is written.
This fixes the log message, so that it
states correctly, that no user_object is
available.


plettich (approved):
Looks good.
Fixed the failing test, waiting for travis to finish

PR #2435, opened 2020-10-03
Update CircleCI configuration

plettich:
Now using CircleCIs Orbs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,14 @@
 jobs:
   build:
-    executor: python/default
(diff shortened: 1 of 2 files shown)


cornelinux:
So this would be running the unittests against mysql, in contrast to running the unittests against sqlite?

plettich:
> So this would be running the unittests against mysql, in contrast to running the unittests against sqlite?

Yes, that is the idea. But it is still work in progress.

plettich:
superseded by  #2482

PR #2442, opened 2020-10-10
2254/startdate audit log

### LINKED ISSUE #2254: Auditlog with start time and end time
Sometime it would come in handy if the audit log contained the start time and the endtime of a request, thus knowing how long it took the request to be processed.

This way it will be easier to identify problems with slow components like slow LDAP connections.

The audit log could also contain the time-to-process (the difference between end and start)
This way we could even filter the audit log for long-running requests, which would also come in handy to track down installation issues.

cornelinux:
Closes #2254


privacyidea/lib/auditmodules/loggeraudit.py
```diff
@@ -71,6 +71,8 @@ class Audit(AuditBase):
 
     def __init__(self, config=None):
         super(Audit, self).__init__(config)
+        # overwrite the startdate from the superclass with the isoformat
+        self.audit_data["startdate"] = datetime.utcnow().isoformat()
```
plettich, line 75: why don't You use `isoformat()` in the base class as well?

And shouldn't SQLALchemy be able to handle datetime objects as well?

privacyidea/lib/auditmodules/sqlaudit.py
```diff
@@ -235,6 +235,10 @@ def finalize_log(self):
                             "Error occurs in action: {0!r}.".format(self.audit_data.get("action")))
                 if not "token_type" in self.audit_data:
                     self.audit_data["token_type"] = self.audit_data.get("tokentype")
+            if self.audit_data.get("startdate"):
+                duration = datetime.datetime.now() - self.audit_data.get("startdate")
```
plettich, line 239: if there is a (isoformat)-string in "startdate", does this calculation still work?

privacyidea/lib/auditmodules/sqlaudit.py
```diff
@@ -249,7 +253,9 @@ def finalize_log(self):
                           client=self.audit_data.get("client", ""),
                           loglevel=self.audit_data.get("log_level"),
                           clearance_level=self.audit_data.get("clearance_level"),
-                          policies=self.audit_data.get("policies")
+                          policies=self.audit_data.get("policies"),
+                          startdate=self.audit_data.get("startdate"),
+                          duration=duration
```
plettich, line 258: Just a note: these entries won't be included in the audit-log signature.

plettich (commented):
Please see my comments, we can discuss this further in person.

migrations/versions/3236a1abf1c6_.py
```diff
@@ -0,0 +1,28 @@
...
+down_revision = '140ba0ca4f07'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    try:
+        op.add_column('pidea_audit', sa.Column('duration', sa.Interval(), nullable=True))
+        op.add_column('pidea_audit', sa.Column('startdate', sa.DateTime(), nullable=True))
```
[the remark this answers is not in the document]
cornelinux: I want to have all three! ``startdate``, ``duration`` and ``date`` (being currently the enddate).
Only then you can sort and filter for all of these!

migrations/versions/3236a1abf1c6_.py
```diff
@@ -0,0 +1,28 @@
...
+    except Exception as exx:
+        print("Could not add retries and timeout to radiusserver")
```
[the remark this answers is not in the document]
cornelinux: Done in 65e95d12

privacyidea/lib/auditmodules/base.py
```diff
@@ -88,32 +89,11 @@ def __init__(self, config=None):
         :return:
         """
         self.name = "AuditBase"
-        self.audit_data = {}
+        self.audit_data = {'startdate': datetime.datetime.now()}
```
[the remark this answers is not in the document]
cornelinux: Will look into this and lets discuss.

privacyidea/lib/auditmodules/loggeraudit.py
cornelinux: There was a reason, I can not remeber. Need to check this.

privacyidea/lib/auditmodules/sqlaudit.py
cornelinux: Yes, indeed it does!

privacyidea/lib/auditmodules/sqlaudit.py
cornelinux: Hm, we have a ``sig_ver`` like "rsa_sha256_pss" but unfortunately we do not now, which data were input for the signature. We might check in ``_log_to_string()``, which is also used during [verify](https://github.com/privacyidea/privacyidea/blob/8a076b7b56b35c298c25b89d1fb044b3eaedba63/privacyidea/lib/auditmodules/sqlaudit.py#L487), if ``startdate`` and ``duration`` is set and only then add this to the string, that is signed and verified. This way we could verify old entries without these values set. 
Do you have a better approach?

cornelinux: Done in b97f3d9b

privacyidea/lib/auditmodules/loggeraudit.py
cornelinux: OK, here is the ting!
The SQLAudit actually uses the baseclass, where "startedate" is set to datetime object using

      datetime.datetime.now()

So just like the "enddate" the "startdate" is stored as a datetime object.

See the old "enddate"

https://github.com/privacyidea/privacyidea/blob/master/privacyidea/models.py#L2627

See the startdate in the baseclass.

The isoformat is simply used here, since it is the loggeraudit, which actually writes a *string* to a file. So we can not store a datetime object.

privacyidea/lib/auditmodules/sqlaudit.py
cornelinux: Correction. In the SQLAudit it actually containes a datetime object. (see my comment above).

So there is no need to change this.

privacyidea/lib/auditmodules/base.py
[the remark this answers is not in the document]
cornelinux: I think there are not that many steps *Before* creating the audit object, as the audit object gathers all data during the request. If we would wanted to do this that early (we also would actually have the same minimal offset without any real interaction, since no configuration is yet read, no communication to the database...) we would have to save the time in ``g`` and then pass it to the audit. I am not sure if we would overcomplicate things or if in ``before_app_request`` we have all necessary information.

However, we can move the creation of the audit object [here](https://github.com/privacyidea/privacyidea/blob/8a076b7b56b35c298c25b89d1fb044b3eaedba63/privacyidea/api/before_after.py#L160) to the beginning of the function ``def before_request``.

cornelinux:
@plettich please review my changes (also a331d411) and see my comments on the reviews I do not plan to do ;-) and kindly "resolve" the conversation.
However, there is the discussion left about when to register the startdate. Thanks.

privacyidea/lib/auditmodules/loggeraudit.py
plettich: We could simply add a json decoder for datetime: https://pynative.com/python-serialize-datetime-into-json/
That way we don't have to take extra care for the logger audit module.

privacyidea/lib/auditmodules/sqlaudit.py
plettich: Yes, i got mixed up with the classes.

privacyidea/lib/auditmodules/loggeraudit.py
cornelinux: The logger currently also writes the end date in isoformat. This is why I simply wanted to keep it that way like it currently works.
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/auditmodules/loggeraudit.py#L84
I do not see any reason to change the functionality of the logger audit here. 
And I do not see any reason why one date in the loggeraudit should be isoformat and the other one json.

privacyidea/lib/auditmodules/sqlaudit.py
cornelinux: Me too

privacyidea/lib/auditmodules/sqlaudit.py
plettich: Some time ago i did some research on versioning the audit log signature based on the signed values but didn't come to a conclusion. So this works for now.

privacyidea/lib/auditmodules/loggeraudit.py
plettich: i am ok with this.

privacyidea/lib/auditmodules/base.py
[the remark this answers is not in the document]
cornelinux: done in 9381eac1

cornelinux:
@plettich ready to re-review.

privacyidea/api/before_after.py
```diff
@@ -82,10 +83,19 @@
...
     log.debug(u"Begin handling of request {!r}".format(request.full_path))
+    g.startdate = datetime.datetime.now()
 
 
 @token_blueprint.teardown_app_request
 def teardown_request(exc):
+    try:
+        if g.audit_object.has_data:
+            g.audit_object.finalize_log()
+    except AttributeError:
```
plettich, line 94: is there a reason for the try...except? wouldn't `if 'audit_object' in g and g.audit_object.has_data:` suffice?

privacyidea/lib/auditmodules/loggeraudit.py
```diff
@@ -69,8 +69,10 @@ class Audit(AuditBase):
     .. note:: This audit module does not provide a *Read* capability.
     """
 
-    def __init__(self, config=None):
-        super(Audit, self).__init__(config)
+    def __init__(self, config=None, startdate=None):
+        super(Audit, self).__init__(config, startdate)
+        # overwrite the startdate from the superclass with the isoformat
+        self.audit_data["startdate"] = startdate or datetime.utcnow().isoformat()
```
plettich, line 75: We would need `startdate.isoformat()` here otherwise the `json.dumps()` fails.

plettich, line 75: And i just realized that the duration isn't determined in the logger audit. It should be calculated in `finalize_log()`.

plettich (changes requested):

privacyidea/api/before_after.py
cornelinux: In certain cases there is an audit_object, that is ``None``.
So it would have to be

~~~~python
    if "audit_object" in g and g.audit_object and g.audit_object.has_data:
~~~~
I thought this was a bit long, so I chose the pythonic way.

privacyidea/lib/auditmodules/loggeraudit.py
cornelinux: Will do.

cornelinux: Yes, I saw this before and actually I forgot about it again.
Do we want it?
OK, I will add it.

cornelinux: Done in 1da2c0a5

cornelinux: Done in 1da2c0a5

PR #2444, opened 2020-10-13
disable enroll button if not allowed

### LINKED ISSUE #2426: disable "Enroll New Token" button
Hi,
In case of policy which NOT allowing enroll any type of token, the user can see an "Enroll New Token" button in the users deatails.
image

laclaro:
fixes #2426 by disabling the button (in line with the close-by assignment button)

### DIFF
--- a/privacyidea/static/components/user/views/user.details.html
+++ b/privacyidea/static/components/user/views/user.details.html
@@ -120,6 +120,7 @@
     </div>
     <div class="text-center">
         <button ng-click="enrollToken()" class="btn btn-primary"
+                ng-disabled="checkEnroll()"
                 id="enrollButton"
                 translate>Enroll New Token
         </button>
--- a/privacyidea/static/components/user/views/user.details.html
+++ b/privacyidea/static/components/user/views/user.details.html
@@ -74,7 +74,7 @@
          max-size="5"
          boundary-links="true" ng-change="pageChanged()"></div>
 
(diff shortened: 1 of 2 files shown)


privacyidea/static/components/user/views/user.details.html
```diff
@@ -120,6 +120,7 @@ <h3 translate>Tokens for user {{ username }}</h3>
     </div>
     <div class="text-center">
         <button ng-click="enrollToken()" class="btn btn-primary"
+                ng-disabled="checkEnroll()"
```
cornelinux, line 123: You have an inverted, wrong logic here!

laclaro: fixed in https://github.com/privacyidea/privacyidea/pull/2444/commits/efd10231c5b77d1752c10e224452daef7bf319e8

privacyidea/static/components/user/views/user.details.html
```diff
@@ -74,7 +74,7 @@ <h3 translate>Tokens for user {{ username }}</h3>
          max-size="5"
          boundary-links="true" ng-change="pageChanged()"></div>
 
-    <div class="table-responsive">
+    <div class="table-responsive form-group">
```
laclaro, line 77: The form group class adds the correct spacing to the button below.

cornelinux (approved):

PR #2460, opened 2020-10-21
add resolver import export to pi-manage

### LINKED ISSUE #1329: "pi-manage resolver" should be able to dump resolver configuration
We should have a function like

    pi-manage resolver r_export <resolvername>

to dump the complete resolver configuration of a resolver including passwords.
The root user of the privacyIDEA system has access to the encryption key and could decrypt the ldap password anyways. This would be a simple but still secure way to check resolvers. (See #1271)

laclaro:
closes #1329


laclaro:
I'll generalize the import_export a bit further, before this should be ready for review

laclaro:
This is included in PR #2468

Issue #2452, opened 2020-10-16
Ability to choose smtp identifier in email enrollment event 


brunocascio:
**Is your feature request related to a problem? Please describe.**

We need to set SMTP config per realm due we use multiple brands and we don't want to use the same for example, SMTP sender.

**Describe the solution you'd like**

Add a field in the event framework to override the global configuration from privacyidea config 'email.identifier'

![image](https://user-images.githubusercontent.com/2397875/96319565-6bc68880-0fe6-11eb-8bbf-5e790c0417ce.png)

**Alternative solution**

Use `additional_params` field?

**Additional context**
We're using keycloak as a centralized solution for authentication. We have 2 clients (foo.com and bar.com)
Each client want to send email with their smtp configuration.

cornelinux:
Gahh! You want to use email tokens? Evil! Very specific.
And you think you need to autoenroll these tokens to the users?

The options of the tokenhandler are defined here:
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/eventhandler/tokenhandler.py#L169

So you could add something like this:
~~~~python
"smtp_server": {
                            "type": "bool",
                            "visibleIf": "tokentype",
                            "visibleValue": "sms",
                            "description": _("The server which will be used to send email-tokens."),
                            "value": smtp_server_list
                        },
~~~~

You could get the "smtp_server_list" from ``lib/smtpserver.py:get_smtpservers()``.

Jameru:
I just made a fairly naive implementation for this feature. If there is any feedback it'd be appreciated :)

brunocascio:
Hey folks @cornelinux @Jameru,

 I think @Jameru's pull request works pretty well for the feature requested.

I'm glad to collaborate if needed :raised_hands:

PR #2461, opened 2020-10-21
Ability to choose smtp identifier in email enrollment event

Jameru:
Closes https://github.com/privacyidea/privacyidea/issues/2452

### DIFF
--- a/privacyidea/lib/eventhandler/tokenhandler.py
+++ b/privacyidea/lib/eventhandler/tokenhandler.py
@@ -178,8 +178,8 @@ class TokenEventHandler(BaseEventHandler):
                             "type": "str",
                             "visibleIf": "tokentype",
                             "visibleValue": "email",
-                            "description": _("Use a specific SMTP server for this token."),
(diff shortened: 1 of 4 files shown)


privacyidea/lib/eventhandler/tokenhandler.py
```diff
@@ -405,6 +413,8 @@ def do(self, action, options=None):
                         if not init_param['email']:
                             log.warning("Enrolling EMail token. But the user {0!s}"
                                         "has no email address!".format(user))
+                    if handler_options.get("smtp_identifier"):
+                        init_param["email.identifier"] = handler_options.get("smtp_identifier")
```
cornelinux, line 417: Please add a test for this code line.

You can add the test here:
https://github.com/privacyidea/privacyidea/blob/master/tests/test_lib_events.py#L1802
where an event handler is created. You can add one, that also has the "smtp_identifier" as an option.

cornelinux (changes requested):
Thank you for the PR.
It basically looks all right.

I have one style request and one test.

cornelinux (approved):

Issue #2152, opened 2020-04-12
Add attestation certificate to endpoint /token/init for certificate tokens


cornelinux:
A certificate can be enrolled by sending a CSR to the privacyIDEA server via ``/token/init``.

See: https://privacyidea.readthedocs.io/en/latest/modules/lib/tokentypes/certificate.html

The Yubikey and possbily other PIV smartcards allow an attestation certificate, which assure, that the private key of a CSR was created on a smartcard. For this, the smartcard can generate an attestation certificate, that contains the same public key as the CSR.

# Generating CSR and attestation certificate with the yubikey

## Creating the CSR

The client can prepare the CSR like this:

~~~~bash
ykman piv generate-key --algorithm RSA2048 9a user.pub
ykman piv generate-csr -s "username" 9a user.pub user.req
~~~~

## Creating the attestation certificate

In addition the client can create an attestation certificate:

~~~~bash
ykman piv attest 9a attest-slot-9a.pem
~~~~

# privacyIDEA REST API with attestation certificate

The ``/token/init`` for the tokentype ``certificate`` should allow the parameter ``attestation`` which would expect the attestation certificate in PEM format.

The privacyIDEA server, needs to 

* verify the validity of the attestation certificate using the whole key chain and
* verify that the public key of the attestation certificate and the CSR are the same.

Then it can issue the certificate.

* [x] **Step 1**: Add ``attestation``  as parameter to the certificate enrollment. 

In a first implementation step we use the attestation certificate **if** it is given but not requiring it.
Policies will only be evaluated in a 2nd implementation step.

# Configuration

## Policies

We need some policies:

* [x]  **Step 3**: scope *enrollment* to require an attestation certificate and which certificate chain should be trusted for the attestation certificate. action: ``require_attestation_cert``, type: ``string`` (like "required and verify", "verify if provided", "ignore"). (This is a tokentype (certificate) specific policy, so we define this in the certificate token)
  * [x] documentation
  * [x] define policy  
  * [x] Add policy for required attestation
  * [x] use policy to replace VERIFY_CERTIFICATE_CHAIN
* [x] ~~scope *enrollment* which CA should be  allowed for the enrollment~~ This is out of scope of this issue but should be part of  a generic other issue like #2159 

## Generic configuration

Then we need to store the Root CA and chain for the trusting of the attestation certificates

[...]

PR #2465, opened 2020-10-26
Verify the attestation certificate

cornelinux:
When a /token/init for a certificate tokens comes
with an attestation certificate, the request is
verified against the attestation certificate and the
attestation certificate is verified against the
certificate chain.

Working on #2152


privacyidea/lib/tokens/certificatetoken.py
```diff
@@ -168,6 +325,27 @@ def get_class_info(key=None, ret='all'):
                            'type': 'int',
                            'desc': _("The user may only have this maximum number of active certificates assigned."),
                            'group': GROUP.TOKEN
+                       },
+                       ACTION.REQUIRE_ATTESTATION: {
+                           'type': 'string',
+                           'desc': _("Enrolling a certificate token can require an attestation certificate. "
+                                     "(Default: ignore)"),
+                           'group': GROUP.TOKEN,
+                           'value': [""]
```
plettich, line 334: Please add the `REQUIRE_ACTIONS` here

privacyidea/lib/tokens/certificatetoken.py
```diff
@@ -33,22 +35,176 @@
...
+    cacert = ""
+    with open(chainfile) as f:
+        lines = f.readlines()
+    for line in lines:
+        if line.startswith("-----BEGIN CERTIFICATE-----"):
+            cacert = line
+        elif line.startswith("-----END CERTIFICATE-----"):
+            cacert += line
+            # End of certificate
+            cacerts.append(cacert)
```
plettich, line 170: we could load the certificate here already so we don't have to load them several times in `_verify_cert()`.
We would also have a check if the certificate in the chain are valid.

privacyidea/api/lib/prepolicy.py
```diff
@@ -1941,3 +1941,34 @@ def _attestation_certificate_allowed(attestation_cert, allowed_certs_pols):
...
+
+    :param request:
+    :param action:
+    :return:
+    """
+    from privacyidea.lib.tokens.certificatetoken import ACTION, REQUIRE_ACTIONS
+    ttype = request.all_data.get("type")
+    if ttype and ttype.lower() == "certificate":
+        # Get attestation certificate requirement
+        require_att = Match.user(g, scope=SCOPE.ENROLL, action=ACTION.REQUIRE_ATTESTATION,
```
plettich, line 1964: Also the `action` needs to prepend the token type (since this is a token specific action).

privacyidea/api/lib/prepolicy.py
```diff
@@ -1941,3 +1941,34 @@ def _attestation_certificate_allowed(attestation_cert, allowed_certs_pols):
...
+    action=REQUIRE_ATTESTATION an exception is raised, if no attestation parameter is given.
+
+    It also checks the policy if the attestation should be verified and sets the
+    parameter verify_attestation accordingly.
+
+    :param request:
+    :param action:
+    :return:
+    """
+    from privacyidea.lib.tokens.certificatetoken import ACTION, REQUIRE_ACTIONS
```
plettich, line 1960: Could we do the import in the `if` block? that way we only do the import if the token is really a certificate token.

privacyidea/lib/tokens/certificatetoken.py
```diff
@@ -33,22 +35,176 @@
...
+1d8Cv4O/MA8GA1UdEwQIMAYBAf8CAQEwDgYDVR0PAQH/BAQDAgEGMA0GCSqGSIb3
+DQEBCwUAA4IBAQBc7Ih8Bc1fkC+FyN1fhjWioBCMr3vjneh7MLbA6kSoyWF70N3s
+XhbXvT4eRh0hvxqvMZNjPU/VlRn6gLVtoEikDLrYFXN6Hh6Wmyy1GTnspnOvMvz2
+lLKuym9KYdYLDgnj3BeAvzIhVzzYSeU77/Cupofj093OuAswW0jYvXsGTyix6B3d
+bW5yWvyS9zNXaqGaUmP3U9/b6DlHdDogMLu3VLpBB9bm5bjaKWWJYgWltCVgUbFq
+Fqyi4+JE014cSgR57Jcu3dZiehB6UtAPgad9L5cNvua/IWRmm+ANy3O2LH++Pyl8
+SREzU8onbBsjMg9QDiSf5oJLKvd/Ren+zGY7
+-----END CERTIFICATE-----"""
+
+YUBICO_ATTESTATION_INTERMEDIATE = """-----BEGIN CERTIFICATE-----
```
plettich, line 79: Maybe we should also send the intermediate certificate from the token, i have another two different intermediate certs from my Yubikeys.

privacyidea/lib/tokens/certificatetoken.py
```diff
@@ -200,6 +409,27 @@ def update(self, param):
...
+                attestation = getParam(param, "attestation", optional)
+                verify_attestation = getParam(param, "verify_attestation", optional)
+                if attestation:
+                    request_numbers = request_csr.public_key().public_numbers()
+                    attestation_cert = load_pem_x509_certificate(to_byte_string(attestation), default_backend())
+                    attestation_numbers = attestation_cert.public_key().public_numbers()
+                    if request_numbers != attestation_numbers:
+                        log.warning("certificate request does not match attestation certificate.")
+                        raise privacyIDEAError("certificate request does not match attestation certificate.")
+                    if verify_attestation:
```
plettich, line 427: Maybe we should always verify the certificate (and log a message and continue if it fails) and only fail hard if it is required by policy.

plettich (changes requested):
I'll also need to update the enrollment script.

privacyidea/api/lib/prepolicy.py
```diff
@@ -1941,3 +1941,34 @@ def _attestation_certificate_allowed(attestation_cert, allowed_certs_pols):
         else None
 
     return attestation_certificate_allowed(cert_info, allowed_certs_pols)
+
+
+def required_piv_attestation(request, action=None):
+    """
+    This is a token specific decorator certificate tokens for the endpoint
```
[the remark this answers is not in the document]
cornelinux: Will do

privacyidea/lib/tokens/certificatetoken.py
```diff
@@ -33,22 +35,176 @@
...
-from privacyidea.lib.policy import SCOPE, ACTION, GROUP
+from privacyidea.lib.policy import SCOPE, ACTION as BASE_ACTION, GROUP, Match
+from privacyidea.lib.error import privacyIDEAError
 
 optional = True
 required = False
 
 log = logging.getLogger(__name__)
 
+YUBICO_ATTESTATION_ROOT_CERT = """-----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
cornelinux: I think this has historical reason during development.
I will remove this.
Chains would be added in the ubuntu packages.

privacyidea/lib/tokens/certificatetoken.py
```diff
@@ -168,6 +325,27 @@ def get_class_info(key=None, ret='all'):
                            'type': 'int',
                            'desc': _("The user may only have this maximum number of active certificates assigned."),
                            'group': GROUP.TOKEN
+                       },
+                       ACTION.REQUIRE_ATTESTATION: {
+                           'type': 'string',
```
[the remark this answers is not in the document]
cornelinux: Will do

privacyidea/lib/tokens/certificatetoken.py
cornelinux: This seems to have vanished.

Will do

privacyidea/api/lib/prepolicy.py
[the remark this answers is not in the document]
cornelinux: done in b94a7c2c

privacyidea/lib/tokens/certificatetoken.py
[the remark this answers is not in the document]
cornelinux: done in b94a7c2c

privacyidea/lib/tokens/certificatetoken.py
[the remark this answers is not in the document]
cornelinux: done in b94a7c2c

doc/policies/admin.rst
```diff
@@ -673,4 +673,22 @@ The admin does not know the secret and can not change the secret.
 
 For more details of this token type see :ref:`indexedsecret_token`.
 
-New in version 3.3.
\ No newline at end of file
+New in version 3.3.
+
+.. _admin_trusted_attestation_CA:
+
+certificate_trusted_Attestation_CA_path
```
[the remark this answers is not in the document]
cornelinux: At least in card management systems there is a distinction between the two scenarios

a) if a user can enroll tokens himself or 
b) if an administrator is allowed to enroll tokens for users.

Using different CA path we can achieve, that e.g. users may only enroll a specific type of hardware while admins are allowed to enroll other types of hardware.

So this is why we have two different policies.

privacyidea/api/lib/prepolicy.py
```diff
@@ -1941,3 +1941,34 @@ def _attestation_certificate_allowed(attestation_cert, allowed_certs_pols):
...
+        require_att = Match.user(g, scope=SCOPE.ENROLL, action=ACTION.REQUIRE_ATTESTATION,
+                                     user_object=request.User if request.User else None).action_values(unique=True)
+        if REQUIRE_ACTIONS.REQUIRE_AND_VERIFY in list(require_att):
+            if not request.all_data.get("attestation"):
+                # There is no attestation certificate in the request, although it is required!
+                log.warning("The request is missing an attestation certificate. {0!s}".format(require_att))
+                raise PolicyError("A policy requires that you provide an attestation certificate.")
+
+        # Add parameter verify_attestation
+        request.all_data["verify_attestation"] = bool(REQUIRE_ACTIONS.VERIFY in list(require_att) or
```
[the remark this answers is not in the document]
cornelinux: Looks nice at the first glance.

But no, we can not simply do this. If there was no policy, then we would have an empty list and ``list()[0]`` would raise an exception.

If there is no policy at all, we currently return False correctly.
Of course we could check for the length of the list and so on, but I do not see why this would be necessary.

Will remove the bool.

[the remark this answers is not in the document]
cornelinux: Done in ee43c003

privacyidea/lib/tokens/certificatetoken.py
```diff
@@ -33,22 +35,176 @@
...
+    from os.path import isfile, join, isdir
+    verified = False
+    for capath in trusted_ca_paths:
+        if isdir(capath):
+            chainfiles = [join(capath, f) for f in listdir(capath) if isfile(join(capath, f))]
+            for chainfile in chainfiles:
+                chain = parse_chainfile(chainfile)
+                try:
+                    verify_certificate(to_byte_string(certificate), chain)
+                    verified = True
```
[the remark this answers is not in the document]
cornelinux: I think you are right, this should be safe to do.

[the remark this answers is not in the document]
cornelinux: Done in 6857c6f3

privacyidea/lib/tokens/certificatetoken.py
cornelinux: Done in b94a7c2cf281652d3be7957a85ebb9549b4ad3b3

privacyidea/api/lib/prepolicy.py
cornelinux: Well, the philosophy about python imports.
Usually I would probably have put this at the beginning of the file.
But this actually is not that elegant, since it is a token specific import and overwirtes the ACTION class.

So I think I simply put it at the beginning of s.th.

An import in the middle of the code looks strange. 
Isn't the import no only done once, anyways?

privacyidea/api/lib/prepolicy.py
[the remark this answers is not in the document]
cornelinux: Yes.

The enrollment policy can depend on the user. So we user ``Match.user`` to pass the user if available.
So I see no need to take an action here.

privacyidea/lib/tokens/certificatetoken.py
cornelinux: You are right, if there is a chain of *root*, *intermediate* and *cert*, then we would call the ``load_pem_x509_certificate`` twice for the intermediate, one time for being the signee and the second time for being the signer.
The *root* and the *cert* will be loaded only once either way.

I would be ready to live with that. 
I would not *load* the certificates when parsing the chain file.
I at all, I would do this in [def verify_certificate](https://github.com/privacyidea/privacyidea/blob/2152_attestation_parameter/privacyidea/lib/tokens/certificatetoken.py#L140) just loading the chain from pem strings to cert objects:

~~~~python
def verify_certificate(certificate, chain):
    certificate = load_pem_x509_certificate(to_byte_string(certificate), default_backend())
    chain = [load_pem_x509_certificate(to_byte_string(c), default_backend()) for c in chain]
    
~~~~

This way we could even omit the helper ``_verify_cert``.

Just drop a note, if you think it is necessary.

The new method would look like this:

~~~~python
def verify_certificate(certificate, chain):
    # first reverse the list, since it can be popped better
    chain = list(reversed(chain))
    if not chain:
        raise privacyIDEAError("Can not verify certificate against an empty chain.")
    certificate = load_pem_x509_certificate(to_byte_string(certificate), default_backend())
    chain = [load_pem_x509_certificate(to_byte_string(c), default_backend()) for c in chain]
    # verify chain
    while chain:
        signer = chain.pop()
        if chain:
            # There is another element in the list, so we check the intermediate:
            signee = chain.pop()
            signer.public_key().verify(
                signee.signature,
                signee.tbs_certificate_bytes,
                padding.PKCS1v15(),
                signee.signature_hash_algorithm
            )
            signer = signee

    # This was the last certificate in the chain, so we check the certificate
    signer.public_key().verify(
        certificate.signature,
        certificate.tbs_certificate_bytes,
        padding.PKCS1v15(),
        certificate.signature_hash_algorithm
    )
~~~~

privacyidea/lib/tokens/certificatetoken.py
cornelinux: Lets do this at another in another issue and PR. I have a conceptual request with the PIV card vendor.

privacyidea/lib/tokens/certificatetoken.py
cornelinux: Nice idea, will do this.

privacyidea/api/lib/prepolicy.py
cornelinux: I am not sure, if I understand this.
You mean the action needs to be prepended with the tokentype?

~~~~python
require_att = Match.user(g, scope=SCOPE.ENROLL, action="cert_{0!s}".format(ACTION.REQUIRE_ATTESTATION,
~~~~
??

privacyidea/lib/tokens/certificatetoken.py
cornelinux: Done in 8a02f887

cornelinux:
@plettich please check my one commit and my remaining comments.

privacyidea/lib/tokens/certificatetoken.py
cornelinux: Done this in d6f3d8d5

privacyidea/api/lib/prepolicy.py
plettich: Yes, because it will be set in `get_dynamic_policy_definitions()` here:
https://github.com/privacyidea/privacyidea/blob/269d1ef24c7e1b3ddd204e5c84890bb7669a56a5/privacyidea/lib/token.py#L2462
before it is sent to the UI. So if the policy is set from the UI, the action is `"certificate_require_attestation"`.
And I would use something like:
```
... action="{0!s}_{1!s}".format(ttype, ACTION.REQUIRE_ATTESTATION), ...
```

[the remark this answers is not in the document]
plettich: This is ok for now, i'll open another issue with this problem.

plettich (changes requested):
please check the token-type prefix

privacyidea/api/lib/prepolicy.py
cornelinux: Thanks. Will do.

cornelinux: Done in 47a374d3

plettich (approved):
looks good

PR #2466, opened 2020-10-26
Push cbor2 version to 5.2.0 to avoid segfault

### LINKED ISSUE #2463: pi-manage always ends with segmentation fault
### Top-level intent

I want to install privacyidea in the latest version running in ubuntu 20.04 with latest supported version of python.
pi-manage runs and works correctly, but it always ends with segmentation fault error.  I made some effort to debug the problem. Looks like a double free in the finalizer code.

### Steps to reproduce

1. Install pi like this:

First I do a git clone, fix the issue with the submodule and checkout v3.4.1.
Then I can install like this:
```
virtualenv -p /usr/bin/python3.8 install
cd install/
source bin/activate
sudo apt install python3.8-dev gcc
pip install -r ../privacyidea/requirements.txt 
pip install ../privacyidea
```
I think this is very close to the instructions given in the official documentation. The gcc and python3.8-dev are required to compile the cbor2.

I get the following warning 3 times during  'pip install -r':

WARNING: Installing from a newer Wheel-Version (1.1)

2. (with python3.8-dbg instead of python3.8):

Run pi-manage in the debug python version to show more information:
```
$ python -X tracemalloc pi-manage
pi-manage:1295: DeprecationWarning: invalid escape sequence \/
  print("""
/home/ppe/work/privacyidea/install/lib/python3.8/site-packages/html5lib/_trie/_base.py:3: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
  from collections import Mapping
/home/ppe/work/privacyidea/install/lib/python3.8/site-packages/flask_script/commands.py:118: DeprecationWarning: inspect.getargspec() is deprecated since Python 3.0, use inspect.signature() or inspect.getfullargspec()
  args, varargs, keywords, defaults = inspect.getargspec(func)

             _                    _______  _______
   ___  ____(_)  _____ _______ __/  _/ _ \/ __/ _ |
  / _ \/ __/ / |/ / _ `/ __/ // // // // / _// __ |
 / .__/_/ /_/|___/\_,_/\__/\_, /___/____/___/_/ |_|
/_/                       /___/
   
usage: pi-manage [-?]
                 ...
optional arguments:
  -?, --help            show this help message and exit
../Include/object.h:541: _Py_NegativeRefcount: Assertion failed: object has negative ref count
Memory block allocated at (most recent call first):
  File "<frozen importlib._bootstrap>", line 219

object address  : 0x3194b50
object refcount : -1
object type     : 0x8f9120
object type name: type
object repr     : <refcnt -1 at 0x3194b50>

Fatal Python error: _PyObject_AssertFailed
Python runtime state: finalizing (tstate=0x10679a0)

Current thread 0x00007f636a310740 (most recent call first):
<no Python frame>
Abgebrochen (Speicherabzug geschrieben)
```
It crashes with segmentation fault in the finalizing stage.

**Hint:** When I remove the definition of

```
@policy_manager.command
def p_export(filename=None):
```

from the code, the segmentation fault does not happen! I do not know how to dig deeper. Perhaps p_export is already defined elsewhere and should not be defined again. When I use pi-manage in an install with database initialized and everything including website working, the segmentation fault shows up as well.  Maybe one of the requirements is outdated? Is the problem in pi-manage or in some python module. Should I use an older python version?

### Expected outcome

I think the command should not end with segmentation fault. Users will be confused by this. Also exit code should be zero, when there is not an error. 

### Actual outcome

I do always get an segmentation fault. Debugging shows, this  happens during the finalizing.

### Configuration

* **privacyIDEA version**: v3.4.1
* **Installation method**: using pip install
* **Python version**: 3.8.5
* **Operating system**: Ubuntu 20.04.1 LTS
* **Webserver**: none
* **Token database**: none

### Log file

Logfile is empty!

plettich:
Python 3.8 (ubuntu 20.04) segfaults at exit (double free) when using the
cbor2 package with some compiled libs (< 5.2.0).
Pushing the required version to 5.2.0 fixes this.

Fixes #2463

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ Babel==2.8.0
 backports.functools-lru-cache==1.6.1; python_version < '3.0'
 bcrypt==3.1.7
 beautifulsoup4==4.9.0
-cbor2==5.1.0
+cbor2==5.2.0
 certifi==2020.4.5.1
 cffi==1.14.0
 chardet==3.0.4
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ def get_file_list(file_path):
 
 
 install_requires = ["beautifulsoup4[lxml]>=4.3.2",
-                    "cbor2>=5.0.1",
+                    "cbor2>=5.2.0",
                     "configobj>=5.0.6",
                     "croniter>=0.3.8",
                     "cryptography>=2.4.2",


plettich:
This is already fixed in ``master`` (cbor2 pinned to 5.1.2: https://github.com/agronholm/cbor2/releases/tag/5.1.2) but we won't add a BugFix release in the 3.4 branch. Closing this.

PR #2468, opened 2020-10-27
enhance the import/export functionality of pi-manage

### LINKED ISSUE #1329: "pi-manage resolver" should be able to dump resolver configuration

### LINKED ISSUE #2467: enable configuration export/import with pi-manage
In Issue #2455 we identified that a configuration transfer is a valuable thing, especially if your work in a corporate environment with several different environments from DEV to PROD.

pi-manage is already capable of exporting events, policies and soon also resolvers (#1329).

These are already the key configurations that one might want to export and import on a different machine.

# Complete pi-manage config export/import capability

There should be a new emnu in pi-manage called "config" which has all the features to export and import configurations.
- policy, event, resolver export/import
- full export/import creating a tar.gz archive with timestamp, similar to the backup functionality
- stdout capability for all exports/imports to make this also scriptable
- backwards-compatibility (e.g. p_export and e_import)

laclaro:
This PR realizes an import/export functionality of pi-manage.

The new menu `pi-manage config` holds all the relevant options. 
- `conf_export` and `conf_import` unify the import-export functionality in single functions.
- working with stdout/stdin enables easy shell scripting
- `fullexport` and `fullimport` work with pi-config-backup tar.gz files similar to `pi-manage backup`
- The export format has slightly changed wrapping the previous lists in a dictionary with "policy", "event" and "resolver" as keys (so far). pi-manage is still backwards-compatible to the old configuration export format

closes #2467 
closes #1329

### DIFF
--- a/pi-manage
+++ b/pi-manage
@@ -98,10 +98,11 @@ from privacyidea.lib.token import import_token
 import jwt
 import ast
 import base64
+import tarfile
 
 SILENT = True
 MYSQL_DIALECTS = ["mysql", "pymysql", "mysql+pymysql"]
-DEFAULT_CONFTYPE_LIST = ["policy", "resolver", "event"]
+DEFAULT_CONFTYPE_LIST = ("policy", "resolver", "event")
 
 app = create_app(config_name='production', silent=SILENT)
 manager = Manager(app)
@@ -1318,7 +1319,7 @@ def conf_export(conftype=DEFAULT_CONFTYPE_LIST, filename=None, name=None, stdout
     if filename:
         with open(filename, 'w') as f:
             f.write(ret_str)
-    if stdout or not filename:
+    if not filename:
         print(ret_str)
 
 
@@ -1462,28 +1463,28 @@ def fullexport(directory=None, stdout=False, archive=False, print_passwords=Fals
         BASE_NAME = "privacyidea-config-backup"
         HOSTNAME = gethostname()
         if not directory:
-            directory = os.path.abspath('./')
+            directory = './'
         else:
-            directory = os.path.abspath(directory)
             call(["mkdir", "-p", directory])
         config_backup_file_base = "%s/%s-%s-%s" % (directory, BASE_NAME, HOSTNAME, DATE)
         config_backup_file = config_backup_file_base + ".py"
         conf_export(filename=config_backup_file, stdout=stdout, print_passwords=print_passwords)
         if archive:
             config_backup_archive = config_backup_file_base + ".tar.gz"
-            call(["tar", "-zcf", config_backup_archive, config_backup_file])
+            tar = tarfile.open(config_backup_archive, "w:gz")
+            tar.add(config_backup_file)
+            tar.close()
             # cleanup
-            if os.path.isfile(config_backup_file):
+            if tarfile.is_tarfile(config_backup_archive):
                 os.remove(config_backup_file)
     else:
         conf_export(filename=None, stdout=True, print_passwords=print_passwords)
 
 
 @config_import_manager.option("--file", "-f", dest="file",
-                              help="The file to import.")
-@config_import_manager.option("--archive", "-a", action="store_true",
-                              help="The specified file is a tar.gz archive containing a configuration "
-                                   "backup file with a name containing 'pi-config-backup'.")
+                              help="The file to import. It can be a plain python file or a tar.gz archive "
+                                   "containing a configuration backup file with a name containing "
+                                   "'privacyidea-config-backup'.")
 @config_import_manager.option("--update", "-u", action="store_true",
                               help="Update the existing configuration. New policies, resolvers and events will also "
                                    "be added.")
@@ -1491,26 +1492,27 @@ def fullexport(directory=None, stdout=False, archive=False, print_passwords=Fals
                               help="The configuration on the target machine will be wiped before the import.")
 @config_import_manager.option("--wipe", "-w", action="store_true", dest="cleanup",
                               help="Wipe is an alias for cleanup.")
-def fullimport(file=None, cleanup=False, update=False, archive=False):
+def fullimport(file=None, cleanup=False, update=False):
     """
     This option reads configuration-backups from a plain file a tar.gz archive or from standard input and imports
     the contained resolvers, policies and event handlers.
     """
     if file:
         if os.path.isfile(file):
-            if archive:
-                archive_file = file
-                p = Popen(["tar", "-ztf", archive_file], stdout=PIPE, universal_newlines=True)
-                std_out, err_out = p.communicate()
-                for line in std_out.split("\n"):
-                    if re.search(r"pi-config-backup", line):
-                        file = "{0!s}".format(line.strip())
-                call(["tar", "-zxf", archive_file, "-C", "./"])
-
-            conf_import(filename=file, cleanup=cleanup, update=update)
-            # cleanup
-            if archive:
-                os.remove(file)
+            if tarfile.is_tarfile(file):
+                tarinfo_objects = []
+                tar = tarfile.open(file)
+                for member in tar.members:
+                    if re.search(r"privacyidea-config-backup", member.name):
+                        tarinfo_objects.append(member)
+                tar.extractall(members=tarinfo_objects)
+                tar.close()
+                for tarinfo in tarinfo_objects:
+                    conf_import(filename=tarinfo.name, cleanup=cleanup, update=update)
+                    # cleanup extracted files
+                    os.remove(tarinfo.name)
+            else:
+                conf_import(filename=file, cleanup=cleanup, update=update)
     else:
         conf_import(cleanup=cleanup, update=update)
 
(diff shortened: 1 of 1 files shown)


pi-manage
```diff
@@ -1319,6 +1281,249 @@ def import_tokens(file, tokenrealm=None):
 token_manager.add_command('import', Command(import_tokens))
 
 
+def conf_export(conftype=DEFAULT_CONFTYPE_LIST, filename=None, name=None, stdout=True, print_passwords=False):
```
plettich, line 1284: And the default list is mutable which breaks some python constraints. You can use a tuple instead.

pi-manage
```diff
@@ -1319,6 +1281,249 @@ def import_tokens(file, tokenrealm=None):
...
+                              help="Compress the created config-backup as tar.gz archive")
+@config_export_manager.option("--stdout", "-s", action="store_true",
+                              help="Print the configuration also to stdout.")
+@config_export_manager.option("--directory", "-d", action="store_true",
+                              help="Directory where the backup will be stored.")
+def fullexport(directory=None, stdout=False, archive=False, print_passwords=False):
+    """
+    This option exports resolvers, policies and event handlers to standard output or to a file and optionally
+    compresses them as tar.gz archive
+    """
```
plettich, line 1457: Maybe we should also explicitly note that the main configuration will not be exported.
In the long run we should implement that as well.

plettich (changes requested):

pi-manage
```diff
@@ -1319,6 +1281,249 @@ def import_tokens(file, tokenrealm=None):
...
+                              help="The specified file is a tar.gz archive containing a configuration "
+                                   "backup file with a name containing 'pi-config-backup'.")
+@config_import_manager.option("--update", "-u", action="store_true",
+                              help="Update the existing configuration. New policies, resolvers and events will also "
+                                   "be added.")
+@config_import_manager.option("--cleanup", "-c", action="store_true",
+                              help="The configuration on the target machine will be wiped before the import.")
+@config_import_manager.option("--wipe", "-w", action="store_true", dest="cleanup",
+                              help="Wipe is an alias for cleanup.")
+def fullimport(file=None, cleanup=False, update=False, archive=False):
```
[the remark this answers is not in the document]
laclaro: This is true, but I could not find out how to add a flask command with proper help for the options.

pi-manage
```diff
@@ -1319,6 +1281,249 @@ def import_tokens(file, tokenrealm=None):
...
+    """
+    This option reads configuration-backups from a plain file a tar.gz archive or from standard input and imports
+    the contained resolvers, policies and event handlers.
+    """
+    if file:
+        if os.path.isfile(file):
+            if archive:
+                archive_file = file
```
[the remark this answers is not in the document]
laclaro: you are right, that's beautifil

pi-manage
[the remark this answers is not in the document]
laclaro: the option exists only, since conf_export is called by fullexport and maybe you would like to have a plain output while you also get an archive file. I would leave it.

laclaro: thank you for the hint

pi-manage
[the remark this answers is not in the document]
laclaro: we should also change this in the backup functionality

laclaro:
I close this PR but leave the branch, so we can cherry pick from the changes after we migrated pi-manage to click.

laclaro:
since major changes in the core code concerning impor/export and also the migration to click will take much time, I pledge to merge this feature into master.

pi-manage
```diff
@@ -1319,6 +1281,249 @@ def import_tokens(file, tokenrealm=None):
...
+                std_out, err_out = p.communicate()
+                for line in std_out.split("\n"):
+                    if re.search(r"pi-config-backup", line):
+                        file = "{0!s}".format(line.strip())
+                call(["tar", "-zxf", archive_file, "-C", "./"])
+
+            conf_import(filename=file, cleanup=cleanup, update=update)
+            # cleanup
+            if archive:
+                os.remove(file)
```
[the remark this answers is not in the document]
laclaro: thank's we use tarfile now.

pi-manage
laclaro: we will change things gradually. More sections of the privacyIDEA configuration can be added later. I think an explicit note is not necessary

pi-manage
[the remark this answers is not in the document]
plettich: The docstring of the function is output as the help text.
Also with `manager.add_command()` you can rename the command seperatly from the function name.

pi-manage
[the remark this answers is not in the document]
plettich: I think our configuration is not that big that we need to output on both channels.
You can just call the export without a file and get the current config and if it looks good, just add the file-parameter to write it to file.

pi-manage
```diff
@@ -1319,6 +1282,250 @@ def import_tokens(file, tokenrealm=None):
...
+                events_with_name = [e for e in cls.events if (name in e.get("name"))]
+                if events_with_name:
+                    exists = True
+                    event_id = events_with_name[0].get("id")
+                else:
+                    exists = False
+                    event_id = None
+            else:
+                print("Warning: conftype {0!s} not in policy, event and resolver".format(conftype))
+            if exists:
```
plettich, line 1402: `exists` and `name` might not be initialized here. Please check.

pi-manage
```diff
@@ -1319,6 +1282,250 @@ def import_tokens(file, tokenrealm=None):
...
+def conf_export(conftype=DEFAULT_CONFTYPE_LIST, filename=None, name=None, stdout=True, print_passwords=False):
+    """
+    Export configurations to a file or write them to stdout if no filename is given.
+    """
+    import pprint
+    pp = pprint.PrettyPrinter(indent=4)
+
+    ret_dict = {}
+    conf = None
+    if isinstance(conftype, list):
```
plettich, line 1294: Since you changed the default parameter to `tuple` you should check for it here as well.

plettich (commented):
Please see my comments.

pi-manage
laclaro: I check for both now.

pi-manage
laclaro: In principle this is not possible, but I initialized them properly.

pi-manage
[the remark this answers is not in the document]
laclaro: I removed the extra stoud option.

pi-manage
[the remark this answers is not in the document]
laclaro: Yes, but I want to have a proper description of the parameters.

pi-manage
```diff
@@ -1319,6 +1284,252 @@ def import_tokens(file, tokenrealm=None):
...
+
+    ret_dict = {}
+    conf = None
+    if isinstance(conftype, list) or isinstance(conftype, tuple):
+        conftype_list = conftype
+    else:
+        conftype_list = [conftype]
+
+    for conftype in conftype_list:
+        if conftype == "policy":
```
plettich, line 1302: in order to make this (and `conf_import`) more readable we can get rid of the `if...elif` chains the following way:
- define helper functions like `_export_resolver()` which do the exporting
- in the loop You can do `if '_export_' + conftype in globals(): globals()['_export_' + conftype]()` 
Alternatively we can define a dictionary containing the functions with the appropriate keys.

pi-manage
```diff
@@ -1319,6 +1284,252 @@ def import_tokens(file, tokenrealm=None):
...
+
+@config_export_manager.option("--print_passwords", "-p", action="store_true",
+                              help="Print the passwords used in the resolver configuration. "
+                                   "This will overwrite existing passwords on import.")
+@config_export_manager.option("--archive", "-a", action="store_true",
+                              help="Compress the created config-backup as tar.gz archive instead "
+                                   "of printing to standard out.")
+@config_export_manager.option("--directory", "-d", action="store_true",
+                              help="Directory where the backup will be stored.")
+def fullexport(directory=None, archive=False, print_passwords=False):
```
plettich, line 1458: We could add the command class explicitly (without the decorators):
https://flask-script.readthedocs.io/en/latest/#adding-arguments-to-commands

plettich (changes requested):

pi-manage
laclaro: I'll do that for more transparency, although it creates additional lines of code.

pi-manage
laclaro: Done the refactoring in https://github.com/privacyidea/privacyidea/pull/2468/commits/355c90262aefa5132729c53b1de338265b7e80d9

cornelinux (approved):
looks nice to me.

cornelinux:
I take the liberty in merging this, so that we to not need to freshen this up in one month.
I leave the feature branch intact, in case anyone wants to look later.

PR #2469, opened 2020-10-27
suppress unpermitted requests in dashboard

### LINKED ISSUE #2456: Only display information, which an admin is allowed to see
The dashboard should only display information, that is actually readable to the admin users.

This can be done by checking either the admin rights or suppressing policy error messages.
The different blocks could be hidden using our javascript method ``checkRight`` like e.g. here: https://github.com/privacyidea/privacyidea/blob/4ba0ff03027e06c408010f9fcc07cbb131659188/privacyidea/static/components/config/views/config.html#L22

We would have to suppress the GET request and the displaying of the corresponding block.

laclaro:
This PR hide information on the dashboard based on user.rights, using checkRights

closes #2456


privacyidea/static/components/dashboard/controllers/dashboardControllers.js
```diff
@@ -149,24 +163,40 @@ myApp.controller("dashboardController", function (ConfigFactory, TokenFactory,
...
-    $scope.get_token_software();
-    $scope.get_policies();
-    $scope.get_events();
-    $scope.getSubscriptions();
-    $scope.getAuthentication();
-    $scope.getAdministration();
-
-        // listen to the reload broadcast
-    $scope.$on("piReload", function() {
+    if (AuthFactory.checkRight('tokenlist')) {
```
cornelinux, line 166: If you are doing the check here, you could actually omit it in the functions get_total_token_number, get_token_hardware and get_token_software.

privacyidea/static/components/dashboard/controllers/dashboardControllers.js
```diff
@@ -149,24 +163,40 @@ myApp.controller("dashboardController", function (ConfigFactory, TokenFactory,
...
         $scope.get_total_token_number();
         $scope.get_token_hardware();
         $scope.get_token_software();
+    };
+    if (AuthFactory.checkRight('policyread')) {
```
cornelinux, line 171: see above and all occurrances later.

cornelinux (changes requested):
You are calling ``checkRight`` *before* calling a function and within the function itself.
I think you only need one.

privacyidea/static/components/dashboard/controllers/dashboardControllers.js
laclaro: correct. Done in https://github.com/privacyidea/privacyidea/pull/2469/commits/c69eade84c8530f5f04e982a89907e4eac0a9fbf

privacyidea/static/components/dashboard/controllers/dashboardControllers.js
laclaro: done

cornelinux (approved):

PR #2476, opened 2020-10-30
rework welcome dialogue

### LINKED ISSUE #2397: Rework the note about privacyIDEA running in community mode
Currently it is not clear to admins that privacyIDEA will start showing the welcome info again, if there is no subscription and the number of users with tokens exceeds 50 users.

The text of the policy only claims: "If this checked, the administrator will not see the welcome dialog anymore."
We should be more specific here and discuss other possibilities like user numbers or easier notifications like the one used for debug mode.

laclaro:
closes #2397

PR includes
- [x] make welcome dialogue independent of starting view
- [x] do not display the welcome dialogue any more if there is no subscription
- [x] instead add a banner notification if there is no subscription and we have more than 50 active users

### DIFF
--- a/privacyidea/static/templates/menu.html
+++ b/privacyidea/static/templates/menu.html
@@ -161,8 +161,7 @@
 <div class="alert alert-warning {{ class_subscription_expired }}"
     ng-show="subscription_state === 2">
     <div translate>
-        <p>Your subscription is not valid anymore. Either you exceeded the number of tokens or the
-            subscription expired. Please check, if you need to renew your <a ui-sref="component.subscriptions">subscription</a>.
-        </p>
+        Your subscription is not valid anymore. Either you exceeded the number of tokens or the
+        subscription expired. Please check, if you need to renew your <a ui-sref="component.subscriptions">subscription</a>.
     </div>
 </div>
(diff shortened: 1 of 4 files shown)


privacyidea/static/templates/menu.html
```diff
@@ -158,3 +158,11 @@
 <div class="alert alert-warning" ng-show="logoutWarning" translate>
     You will be logged out in {{ myCountdown }} second(s).
 </div>
+<div class="alert alert-warning {{ class_subscription_expired }}"
```
cornelinux, line 161: What is this css_class good for? Are you randomizing it, so that it can not be hidden?

privacyidea/static/templates/menu.html
```diff
@@ -158,3 +158,11 @@
...
+    ng-show="subscription_state === 2">
+    <div translate>
+        <p>Your subscription is not valid anymore. Either you exceeded the number of tokens or the
```
cornelinux, line 164: The text is not centered in the yellow notification.
Please remove the paragraph <p> and it will be centered.

cornelinux (changes requested):
On minor style thing...

privacyidea/static/templates/menu.html
laclaro: that was the intention. It can be hidden, but pops up again. If you do it correctly, you can also hide it permanently.

privacyidea/static/templates/menu.html
laclaro: done

cornelinux (approved):

PR #2478, opened 2020-10-31
Add test for challenge-response without PIN

cornelinux:
Just add another test to clarify, that it is not
possible to use an HOTP value of an Email-Token
in a challenge response with a challenge/transaction_id
that has expired.


plettich (approved):
looks good

PR #2479, opened 2020-11-01
Allow to reset PIN during WebUI Login

### LINKED ISSUE #2474: Allow PIN change at WebUI
### Top-level intent

Use the pin change during validate when logging in to the webui

### Steps to reproduce

1. configure ``change-pin-on-first-use``
2. configure ``login_mode=privacyIDEA``
3. Login to webui with 2FA
4. The webui will not display any challenge

### Expected outcome

The user should be able to change his PIN

### Actual outcome

Nothing is displayed

I think the problem is, that the input field is "ng-hidden" by some tokentype in the response. The PIN-change-challenge probably is missing some fields in the JSON response, so that dialog is hidden.

cornelinux:
The challenge response was missing an
attribute. That resulted in an error
in loginController.js in line 248
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/static/components/login/controllers/loginControllers.js#L248

Adding an empty attribute to the
challenge response fixes this issue.

Fixes #2474

### DIFF
--- a/privacyidea/lib/challengeresponsedecorators.py
+++ b/privacyidea/lib/challengeresponsedecorators.py
@@ -59,6 +59,7 @@ def _create_pin_reset_challenge(token_obj, message, challenge_data=None):
     reply_dict = {}
     reply_dict["multi_challenge"] = [{"transaction_id": db_challenge.transaction_id,
                                       "message": message,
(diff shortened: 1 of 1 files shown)


plettich (approved):
looks good

PR #2480, opened 2020-11-01
Initialize the OTP PIN in the token details UI

### LINKED ISSUE #2472: Problem with setting an empty PIN in UI
The administrator tries to set an empty PIN for a token in the webui.

1. click of "Set pin" results in an error response.

2. click on "Set pin" successfully sets the PIN.

cornelinux:
Now that the PIN is initiailized the admin can
immediately set an empty pin.

Closes #2472


plettich (approved):
Looks good

PR #2485, opened 2020-11-03
Add db migration script for "authcache" table

### LINKED ISSUE #2446: argon2 hash too long for authcache column authorization
With the new argon2 hashing, introduced by #2413, the hashes may exeed VARCHAR(64) as e.g. in the authcache table in the authorization column.

We have to increase that size.

- [x] Check for the length of the argon2 hashes as used by us - the hash generated by test_06_hash_pepper has 77 characters. We will do VARCHAR(255)
- [x] Check if this affects not only the authcache table
- [ ] Add migration script (-> v3.5)

This bug may cause validate/check requests to fail with exception if the auth cache is enabled.

plettich:
The Argon2 hashes introduced in #2413 are too long for the
"authorization" column in the "authcache" table.
The column is set to 255 characters with this PR.

Fixes #2446


cornelinux (approved):

yangboyd:
how to generate such revision files?

plettich:
> how to generate such revision files?

We use [Flask-Migrate](https://flask-migrate.readthedocs.io/en/latest/) as a front-end to SQLAlchemy/Alembic.
With `flask db revision` an empty migration file can be generated which must then be adapted accordingly.

PR #2490, opened 2020-11-04
2402/increment failcounter

### LINKED ISSUE #2402: Event Handler: Increase or decrease failcounter
Allow Token-Event-Handler to not only **set failcounter** but also to increment or decrease failcounter.

This could be done by entering +1 or -1 do increase or decrease the failcounter.
We could then parse for "+" or "-" and act accordingly.

laclaro:
closes #2402

This enables the token event handler to increment failcounters by specifying +n or -n.

### DIFF
--- a/doc/eventhandler/tokenhandler.rst
+++ b/doc/eventhandler/tokenhandler.rst
@@ -165,9 +165,10 @@ and ``{ua_string}`` for information on the user agent and ``{username}`` and
 set failcounter
 ...............
 
-Using the action ``set failcounter`` you can reset the fail counter by
-setting it to 0 or also "block" the token by setting the fail counter to what
-ever value the "max_fail" is, e.g. 10. Only integer values are allowed.
+Using the action ``set failcounter`` you can set, increase or decrease the fail counter
+of the token by setting the action to ``n``, ``+n`` or ``-n`` respectively.
+You may use this to reset the failcounter by setting it to 0 or to "block" the token by setting
+the fail counter to the value of "max_fail" or greater.
 
 set random pin
 ..............
--- a/doc/eventhandler/tokenhandler.rst
+++ b/doc/eventhandler/tokenhandler.rst
@@ -168,7 +168,7 @@ set failcounter
 Using the action ``set failcounter`` you can set, increase or decrease the fail counter
 of the token by setting the action to ``n``, ``+n`` or ``-n`` respectively.
 You may use this to reset the fail counter by setting it to 0 or to "block" the token by setting
-the fail counter to the value of "max_fail" or greater.
+the fail counter to the value of "max_fail".
 
 set random pin
 ..............
--- a/doc/eventhandler/tokenhandler.rst
+++ b/doc/eventhandler/tokenhandler.rst
@@ -167,7 +167,7 @@ set failcounter
 
 Using the action ``set failcounter`` you can set, increase or decrease the fail counter
 of the token by setting the action to ``n``, ``+n`` or ``-n`` respectively.
-You may use this to reset the failcounter by setting it to 0 or to "block" the token by setting
+You may use this to reset the fail counter by setting it to 0 or to "block" the token by setting
 the fail counter to the value of "max_fail" or greater.
 
 set random pin
--- a/privacyidea/lib/eventhandler/counterhandler.py
+++ b/privacyidea/lib/eventhandler/counterhandler.py
@@ -44,7 +44,7 @@ class CounterEventHandler(BaseEventHandler):
     """
 
     identifier = "Counter"
-    description = "This event handler increases arbitrary counters in the database."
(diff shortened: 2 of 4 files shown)


doc/eventhandler/tokenhandler.rst
```diff
@@ -165,9 +165,10 @@ and ``{ua_string}`` for information on the user agent and ``{username}`` and
 set failcounter
 ...............
 
-Using the action ``set failcounter`` you can reset the fail counter by
-setting it to 0 or also "block" the token by setting the fail counter to what
-ever value the "max_fail" is, e.g. 10. Only integer values are allowed.
+Using the action ``set failcounter`` you can set, increase or decrease the fail counter
+of the token by setting the action to ``n``, ``+n`` or ``-n`` respectively.
+You may use this to reset the fail counter by setting it to 0 or to "block" the token by setting
+the fail counter to the value of "max_fail" or greater.
```
cornelinux, line 171: I would not recommend setting the failcounter to "greater". This basically works but might have side effect for which I would not want to take responsibility since "it is written in the docs".

privacyidea/lib/eventhandler/counterhandler.py
```diff
@@ -44,7 +44,7 @@ class CounterEventHandler(BaseEventHandler):
     """
 
     identifier = "Counter"
-    description = "This event handler increases arbitrary counters in the database."
+    description = "This event handler increases, decreases or resets arbitrary counters in the database."
```
cornelinux, line 47: FYI: This is correct but has nothing to do with the fail counter.

privacyidea/lib/eventhandler/tokenhandler.py
```diff
@@ -236,7 +237,8 @@ def actions(cls):
                                    "type": "str",
                                    "required": True,
                                    "description": _("Set the failcounter of "
-                                                    "the token.")
+                                                    "the token. Any value or increment n, "
```
cornelinux, line 240: I am a bit confused by this sentence. "Any value or increment n" sounds like as if "n" would be an increment. May some more text might be less confusing like

"Increase or decrease or set to a specific value by ... doing/puttin/verb-of-your-choice +n, -n or n with n being an integer..."
..or...  rtfm: read the fantactic manual ;-)

privacyidea/lib/eventhandler/tokenhandler.py
```diff
@@ -372,8 +374,15 @@ def do(self, action, options=None):
                                                 d.strftime(DATE_FORMAT))
                 elif action.lower() == ACTION_TYPE.SET_FAILCOUNTER:
                     try:
-                        set_failcounter(serial,
-                                        int(handler_options.get("fail counter")))
+                        handler_option = handler_options.get("fail counter")
+                        if handler_option.startswith("+"):
+                            token_failcount = get_one_token(serial=serial).token.failcount
```
cornelinux, line 379: I would do:

    token_object = get_one_token(serial=serial)

privacyidea/lib/eventhandler/tokenhandler.py
```diff
@@ -372,8 +374,15 @@ def do(self, action, options=None):
...
+                            set_failcounter(serial, token_failcount + int(handler_option[1:]))
```
cornelinux, line 380: You already have to token_object ``token_object``. So if you are calling ``set_cailcounter`` with the serial number, the code will do a second token lookup.
You can save the 2nd token lookup if you do:

~~~~python
token_object.set_failcount(token_oject.token.failcount + int(handler_option[1:]))
~~~~

...
or we could enhance ``lib/token.py:set_failcounter`` to also directly accept increments or decrements.

cornelinux (changes requested):
Please check my notes about texts and the double token lookup.

Also please add a test to increment or decrement like here
https://github.com/privacyidea/privacyidea/blob/master/tests/test_lib_events.py#L2159

doc/eventhandler/tokenhandler.rst
laclaro: ok

privacyidea/lib/eventhandler/tokenhandler.py
laclaro: After pushing, I also wondered if the word order is misleading. Changed.

privacyidea/lib/eventhandler/tokenhandler.py
laclaro: ah, thank's

privacyidea/lib/eventhandler/counterhandler.py
laclaro: that's true. It is the change in https://github.com/privacyidea/privacyidea/pull/2490/commits/b1e02548ac4586789dc36bb2d07e9fe79f926b11

laclaro:
see #2496

PR #2496, opened 2020-11-05
2402/change failcounter

### LINKED ISSUE #2402: Event Handler: Increase or decrease failcounter

laclaro:
closes #2402

introduces change failcounter action to the token handler.

### DIFF
--- a/doc/eventhandler/tokenhandler.rst
+++ b/doc/eventhandler/tokenhandler.rst
@@ -172,11 +172,14 @@ ever value the "max_fail" is, e.g. 10. Only integer values are allowed.
 change failcounter
 ..................
 
-Using the action ``change failcounter`` you can increase or decrease the failcounter.
-Positive and negative integer values are allowed.
+Using the action ``change failcounter`` you can increase or decrease the fail counter.
+Positive and negative integer values are allowed. Positive values will increase the
+fail counter, negative values will decrease it.
 
-.. note:: To limit a token handler in decreasing the fail counter, you may use the event handler
-   condition **failcounter** (c.f. :ref:`handlerconditions`).
+.. note:: To limit a token handler in decreasing the fail counter, you may use the
+   event handler condition **failcounter** (c.f. :ref:`handlerconditions`) and set
+   it to e.g. ">-5". Once this condition is not met anymore, the event handler will
+   not be triggered.
 
 
 set random pin
--- a/doc/eventhandler/tokenhandler.rst
+++ b/doc/eventhandler/tokenhandler.rst
@@ -169,6 +169,16 @@ Using the action ``set failcounter`` you can reset the fail counter by
 setting it to 0 or also "block" the token by setting the fail counter to what
 ever value the "max_fail" is, e.g. 10. Only integer values are allowed.
 
+change failcounter
+..................
+
+Using the action ``change failcounter`` you can increase or decrease the failcounter.
+Positive and negative integer values are allowed.
+
+.. note:: To limit a token handler in decreasing the fail counter, you may use the event handler
+   condition **failcounter** (c.f. :ref:`handlerconditions`).
+
+
 set random pin
 ..............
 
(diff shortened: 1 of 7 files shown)


doc/eventhandler/tokenhandler.rst
```diff
@@ -169,6 +169,16 @@ Using the action ``set failcounter`` you can reset the fail counter by
 setting it to 0 or also "block" the token by setting the fail counter to what
 ever value the "max_fail" is, e.g. 10. Only integer values are allowed.
 
+change failcounter
```
cornelinux, line 172: honestly, I find this naming a bit difficult.

*set failcounter* is also a *change*.

doc/eventhandler/tokenhandler.rst
```diff
@@ -169,6 +169,16 @@ Using the action ``set failcounter`` you can reset the fail counter by
...
+..................
+
+Using the action ``change failcounter`` you can increase or decrease the failcounter.
+Positive and negative integer values are allowed.
```
cornelinux, line 176: I think they are required. Using negative values will decrease the fail counter, using a positive value will increase the fail counter.

doc/eventhandler/tokenhandler.rst
```diff
@@ -169,6 +169,16 @@ Using the action ``set failcounter`` you can reset the fail counter by
...
+
+.. note:: To limit a token handler in decreasing the fail counter, you may use the event handler
+   condition **failcounter** (c.f. :ref:`handlerconditions`).
```
cornelinux, line 179: I as a user would not understand that.

tests/test_lib_events.py
```diff
@@ -2203,6 +2203,19 @@ def test_10_set_failcounter(self):
...
 
+        # check the change failcount option starting with the set failcount of 7
+        handler_options = {-1: {"change fail counter": "-8"},
+                           1: {"change fail counter": "2"},
+                           2: {"change fail counter": "+1"}}
+        for failcount, handler_option in handler_options.items():
+            options["handler_def"]["options"].update(handler_option)
+            res = t_handler.do(ACTION_TYPE.CHANGE_FAILCOUNTER, options=options)
+            self.assertTrue(res)
+            # Check if the token has the correct sync window
```
cornelinux, line 2214: ...has the "correct failcounter".

cornelinux (changes requested):
looks good to me.
But please add some polishing to the docs.

tests/test_lib_events.py
laclaro: done

doc/eventhandler/tokenhandler.rst
laclaro: changed

doc/eventhandler/tokenhandler.rst
laclaro: ok

doc/eventhandler/tokenhandler.rst
cornelinux: I can not wrap my head around the text "to limit a token handler in decreasing the fail counter". Something like "If you want to avoid that a fail counter decreases indefinitively, you can set a fail counter condition...".

cornelinux (changes requested):
For some reason the increasing and decreasing tests fail in python 2.7 and 3.5.
It looks like as if the last test actually **sets** the fail counter and does not increase it.

laclaro:
> For some reason the increasing and decreasing tests fail in python 2.7 and 3.5.
> It looks like as if the last test actually **sets** the fail counter and does not increase it.

Well, python dicts are not ordered... fixed in https://github.com/privacyidea/privacyidea/pull/2496/commits/831f2dad8498417a330145b69088c04b7c0bb8a1

cornelinux (approved):

PR #2497, opened 2020-11-05
Differentiate in C/R between tokens with/out DB-challenge

### LINKED ISSUE #2491: is_challenge_response should not depent on the presence of a transaction_id in the database
The method ``is_challenge_response`` determines if the request is the response to a given challenge.
In this method it is also checked, if a challenge is available in the database table.
In certain scenarios this can lead to the effect, that privacyIDEA thinks, this request would not be a challenge.

We should remove the dependency on the availability of the challenge.

cornelinux:
A response to a challenge can arrive so late, that the
original challenge does not exist anymore in the database.
It still would be a response to a challenge, so we need to
differentiate between being a legit or non-legit response at all
(it has a "transaction_id")
and a response, for which still a challenge-DB-object exists.

When merging this PR, it will also rendet this PR useless:
Closes #2493

After all it...
Closes #2491

### DIFF
--- a/privacyidea/lib/tokenclass.py
+++ b/privacyidea/lib/tokenclass.py
@@ -453,7 +453,6 @@ class TokenClass(object):
 
         return pin_match, otp_counter, reply
 
-
     @staticmethod
     def decode_otpkey(otpkey, otpkeyformat):
         """
@@ -1387,7 +1386,7 @@ class TokenClass(object):
 
     def is_challenge_response(self, passw, user=None, options=None):
         """
-        This method checks, if this is a request  that is supposed to be
+        This method checks, if this is a request that is supposed to be
         the answer to a previous challenge.
 
         The default behaviour to check if this is the response to a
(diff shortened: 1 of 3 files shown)


cornelinux:
This is another approach than #2493. I like this here more, since it is clearer about categorizing the tokens in the ``token_lists`` in ``lib/token.py:check_token_list``.

privacyidea/lib/tokenclass.py
```diff
@@ -1387,8 +1387,8 @@ def is_challenge_request(self, passw, user=None, options=None):
 
     def is_challenge_response(self, passw, user=None, options=None):
         """
-        This method checks, if this is a request, that is the response to
-        a previously sent challenge.
+        This method checks, if this is a request  that is supposed to be
```
plettich, line 1390: please remove superfluous space

plettich (changes requested):
just a minor typo

privacyidea/lib/tokenclass.py
cornelinux: Hawk eye. Did not see that! Will do!

plettich (approved):
looks good

PR #2500, opened 2020-11-10
2372/limit audit log view api level

### LINKED ISSUE #2372: Limit audit log view for e.g. 1st level helpdesk
Audit log, only show selected columns in audit log and only show selected actions.

This could be helpful for 1st level helpdesk users. They e.g. might only want to take a look at actions like /validate/check and would not be interested in information like "administrator".

Via policy we could limit the view to certain columns and actions.

laclaro:
Reorganize PR #2445 to squash commits

Closes #2445, closes #2372


privacyidea/static/components/audit/views/audit.log.html
```diff
@@ -101,30 +115,32 @@
...
+                    <a ng-if="loggedInUser.role == 'admin'" ui-sref="config.realms.list">{{ audit.realm }}</a>
                     <span ng-if="loggedInUser.role == 'user'">{{ audit.realm }}</span>
                 </td>
-                <td><a ng-if="loggedInUser.role == 'admin'" ui-sref="config.resolvers.list">{{ audit.resolver }}</a>
+                <td ng-if="audit_columns.includes('resolver')">
+                    <a ng-if="loggedInUser.role == 'admin'" ui-sref="config.resolvers.list">{{ audit.resolver }}</a>
                     <span ng-if="loggedInUser.role == 'user'">{{ audit.resolver }}</span>
                 </td>
-                <td ng-if="loggedInUser.role == 'admin'">
+                <td ng-if="audit_columns.includes('policies')" ng-if="loggedInUser.role == 'admin'">
```
cornelinux, line 143: You can connect this via
  
    ng-if="audit_columns.included('policies') && loggedInUser.role == 'admin'

It is not clear what happens with two ng-ifs.

privacyidea/static/components/audit/controllers/auditControllers.js
```diff
@@ -65,6 +65,11 @@ myApp.controller("auditController", function (AuditFactory, $scope, $rootScope,
             $scope.getParams();
             AuditFactory.get($scope.params, function(data) {
                 $scope.auditdata = data.result.value;
+                if ($scope.auditdata.auditdata[0] && (typeof $scope.audit_columns !== 'undefined')) {
```
cornelinux, line 68: I am not sure, how javascript behaves here. If  ``$scope.auditdata.auditdata`` is not defined, we will get a javascript exception. Can this happen?

Why do you test for ``$scope.audit_colums``? I think this is not necessary.

privacyidea/lib/audit.py
```diff
@@ -105,10 +106,20 @@ def search(config, param=None, user=None):
...
+        del param["hidden_columns"]
 
     pagination = audit.search(param, sortorder=sortorder, page=page,
                               page_size=page_size, timelimit=timelimit)
 
+    # delete hidden columns from response
+    if hidden_columns:
+        for audit_row in pagination.auditdata:
+            for col in hidden_columns:
+                if col in audit_row:
```
cornelinux, line 120: This looks strange.
Would there be a more elegant way with using dict comprehension or do we create a new object?

~~~~python
audit_row = { key: value for key, value in audit_row.items() if key not in hidden_columns }
~~~~

cornelinux (changes requested):
Please check my comments.

privacyidea/lib/audit.py
laclaro: that's nice indeed.

privacyidea/static/components/audit/controllers/auditControllers.js
laclaro: done, thank's

privacyidea/static/components/audit/views/audit.log.html
laclaro: done, thank's

privacyidea/static/components/audit/controllers/auditControllers.js
laclaro: an empty data.result.value is not probable in the webUI. It works if the request works properly. If you have no access, this code will not run. If no entries are returned, data.result.value is defined also.

The typeof thing is not necessary. I introduced it to have the if-block only run once in the ng-repeater, since there was a related error. But apparently it stemmed from somewhere else.

Changed.

cornelinux (approved):

cornelinux:
@laclaro You could change the way you define the assertions, so that you do not run into problems with the order of the lists/keys.

~~~~python
self.assertIn(list, "realm")
self.assertIn(list, "user")
~~~~

instead of 

~~~~python
 self.assertEqual(list, ["realm", "user"]
~~~~

laclaro:
> @laclaro You could change the way you define the assertions, so that you do not run into problems with the order of the lists/keys.
> 
> ```python
> self.assertIn(list, "realm")
> self.assertIn(list, "user")
> ```
done

cornelinux:
@laclaro action und serial have the same problem. THere might be others!

Issue #2342, opened 2020-08-21
Update documentation, templates and strings for v3.4


plettich:
Before releasing v3.4 we should fix some documentation issues and translations.
- [x] merge `doc/installation/system/pimanage/` into docs since the separate doc tree is not used (#2503)
- [x] fix some misspellings, broken links and other warnings sphinx (#2525)
- [x] update brute-force docs (#2344 )
- [x] fix formatting of API docs (generated from source files) (#2525)
- [x] add translations
- [x] fix missing numbers in doc index ("nested numbered toctree") (#2502)

plettich:
Should we keep this issue as a fixed issue (not only for v3.4) for documentation related PRs? This could include spelling, formatting, rendering, ...

cornelinux:
I don't know. I feel a bit uncomfortable with issues, which will not be closed.
So I would rather have this as a concrete issue and close this. And open new issues for new typos or docs adjustments.

plettich:
Most of the changes are merged so i close this issue.

PR #2502, opened 2020-11-11
fix missing numbers by removing :numbered: in any sub-toctree (nested)

laclaro:
The number in the primary documentation chapter index was sometimes missing. This PR fixes this behavior by removing any `:numbered:` options in sub-indeces (->"nested numbered toctree").

working on #2342


pablo-knight (approved):
fine

PR #2504, opened 2020-11-11
2312/verify value mac

### LINKED ISSUE #2312: PSKC import does not verify ValueMac of encrypted token secrets
In [RFC6030 (PSKC)](https://tools.ietf.org/html/rfc6030#section-6.1.1) a method is described to verify token key data with MACs. We currently [implement](https://github.com/privacyidea/privacyidea/blob/35b705a372d87ffb8e0ccbd81d1765c32e451509/privacyidea/lib/importotp.py#L627) this for export to PSKC-Files but we don't check it when importing the PSKC-File.

TODOs:
- [x] Implement verification of MAC values
- [x] Fix export of tokens / calculation of MAC values on export
- [x] Add parameter to import `validate_mac`:
    - `check_fail_hard` (default) - no import on invalid mac
    - `no_check` - ignore mac on import
    - `check_fail_soft` - only import valid mac
- [x] Add setting to UI

timosturm:
When importing tokens from PSKC-Files the method of validating the MAC can now be selected. There are three options:
- 'no_check' : MAC values are not validated
- 'check_fail_soft' : Tokens with invalid MAC are skipped, valid tokens are imported
- 'check_fail_hard' : If an invalid MAC is encountered, no token is imported (not even the valid ones)

'check_fail_hard' is the default option.

The user can select one of the aforementioned options in the user interface, additionally more feedback is given: The UI now shows the number of not imported tokens next to the number of imported tokens.

The `privacyidea_token_manager' too supports selecting a validation method.

Both the code and the user interface should be checked when reviewing this pull request. Closes #2312

### DIFF
--- a/privacyidea/api/token.py
+++ b/privacyidea/api/token.py
@@ -911,6 +911,8 @@ def loadtokens_api(filename=None):
                    'Yubikey CSV', 'pskc']
     file_type = getParam(request.all_data, "type", required)
     hashlib = getParam(request.all_data, "aladdin_hashlib")
+    # TODO Get parameter psck_validateMac with default to <?>
+    # TODO Add parameter to description
     aes_psk = getParam(request.all_data, "psk")
     aes_password = getParam(request.all_data, "password")
     if aes_psk and len(aes_psk) != 32:
--- a/privacyidea/api/token.py
+++ b/privacyidea/api/token.py
@@ -925,7 +925,7 @@ def loadtokens_api(filename=None):
     if trealms:
         tokenrealms = trealms.split(",")
 
-    not_imported_names = []
+    not_imported_serials = []
     TOKENS = {}
     token_file = request.files['file']
     file_contents = ""
--- a/privacyidea/api/token.py
+++ b/privacyidea/api/token.py
@@ -992,7 +992,7 @@ def loadtokens_api(filename=None):
                         'success': True})
     # logTokenNum()
 
-    return send_result({'n_imported': len(TOKENS), 'n_not_imported': len(not_imported_names)})
+    return send_result({'n_imported': len(TOKENS), 'n_not_imported': len(not_imported_serials)})
 
 
 @token_blueprint.route('/copypin', methods=['POST'])
--- a/privacyidea/api/token.py
+++ b/privacyidea/api/token.py
@@ -969,7 +969,7 @@ def loadtokens_api(filename=None):
     elif file_type in ["yubikeycsv", "Yubikey CSV"]:
         TOKENS = parseYubicoCSV(file_contents)
     elif file_type in ["pskc"]:
-        TOKENS, not_imported_names = parsePSKCdata(file_contents, preshared_key_hex=aes_psk,
+        TOKENS, not_imported_serials = parsePSKCdata(file_contents, preshared_key_hex=aes_psk,
                                                    password=aes_password, validate_mac=aes_validate_mac)
 
     # Now import the Tokens from the dictionary
--- a/privacyidea/api/token.py
+++ b/privacyidea/api/token.py
@@ -904,7 +904,8 @@ def loadtokens_api(filename=None):
         "oathcsv" or "yubikeycsv".
     :jsonparam tokenrealms: comma separated list of realms.
     :jsonparam psk: Pre Shared Key, when importing PSKC
-    :jsonparam pskcValidateMAC: Determines how invalid MACs should be handled when importing PSKC
+    :jsonparam pskcValidateMAC: Determines how invalid MACs should be handled when importing PSKC.
+               Allowed values are 'no_check', 'check_fail_soft' and 'check_fail_hard'.
     :return: The number of the imported tokens
     :rtype: int
     """
--- a/privacyidea/api/token.py
+++ b/privacyidea/api/token.py
@@ -902,6 +902,7 @@ def loadtokens_api(filename=None):
         "oathcsv" or "yubikeycsv".
     :jsonparam tokenrealms: comma separated list of realms.
     :jsonparam psk: Pre Shared Key, when importing PSKC
+    :jsonparam pskcValidateMAC: Determines how invalid MACs should be handled when importing PSKC
     :return: The number of the imported tokens
     :rtype: int
     """
@@ -911,8 +912,7 @@ def loadtokens_api(filename=None):
                    'Yubikey CSV', 'pskc']
     file_type = getParam(request.all_data, "type", required)
     hashlib = getParam(request.all_data, "aladdin_hashlib")
-    # TODO Get parameter psck_validateMac with default to <?>
-    # TODO Add parameter to description
+    aes_validate_mac = getParam(request.all_data, "pskcValidateMAC", default='check_fail_hard')
     aes_psk = getParam(request.all_data, "psk")
     aes_password = getParam(request.all_data, "password")
     if aes_psk and len(aes_psk) != 32:
@@ -967,7 +967,7 @@ def loadtokens_api(filename=None):
         TOKENS = parseYubicoCSV(file_contents)
     elif file_type in ["pskc"]:
         TOKENS = parsePSKCdata(file_contents, preshared_key_hex=aes_psk,
-                               password=aes_password)
+                               password=aes_password, validate_mac=aes_validate_mac)
 
     # Now import the Tokens from the dictionary
     ret = ""
--- a/privacyidea/api/token.py
+++ b/privacyidea/api/token.py
@@ -3,6 +3,8 @@
 # http://www.privacyidea.org
 # (c) cornelius kölbel, privacyidea.org
 #
+# 2020-11-11 Timo Sturm <timo.sturm@netknights.it>
+#            Select how to validate PSKC imports
 # 2020-01-28 Jean-Pierre Höhmann <jean-pierre.hoehmann@netknights.it>
 #            Add WebAuthn token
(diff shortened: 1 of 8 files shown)


tools/privacyidea-token-janitor
```diff
@@ -509,13 +509,18 @@ def find(last_auth, assigned, active, tokeninfo_key, tokeninfo_value,
...
     from privacyidea.lib.importotp import parsePSKCdata
 
     with open(pskc, 'r') as pskcfile:
         file_contents = pskcfile.read()
 
-    tokens = parsePSKCdata(file_contents, preshared_key_hex)
+    tokens, _ = parsePSKCdata(file_contents, preshared_key_hex, validate_mac)
     success = 0
     failed = 0
     failed_tokens = []
```
timosturm, line 528: Should the tokens that were ignored by the import function itself be counted here too?

privacyidea/static/components/token/views/token.import.html
```diff
@@ -69,12 +69,27 @@ <h3 translate>Import tokens from a file</h3>
...
-               ng-keypress="form.password =''"
+               ng-keypress="form.password = ''"
                ng-model="form.psk" minlength="0" maxlength="32">
         <input class="form-control"
                placeholder="{{ 'Password' | translate }}"
                ng-keypress="form.psk = ''"
                ng-model="form.password">
+        <div></div>
+        <div class="mt-5px">
+            <p class="help help-block" translate>
```
timosturm, line 80: @cornelinux It is unclear how translations are handled, should the developer add e.g. german translations in the push request?

privacyidea/lib/importotp.py
```diff
@@ -469,11 +468,18 @@ def parsePSKCdata(xml_data,
...
     :param do_checkserial: Check if the serial numbers conform to the OATH
         specification (not yet implemented)
+    :param validate_mac: Operation mode of hmac validation. Possible values:
+        - 'check_fail_hard' : If an invalid hmac is encountered no token gets parsed.
+        - 'check_fail_soft' : Skip tokens with invalid MAC.
+        - 'no_check' : Hmac of tokens are not checked, every token is parsed.
 
-    :return: a dictionary of token dictionaries
-        { serial : { otpkey , counter, .... }}
+    :return: a dictionary of token dictionaries and a list of serial of not imported tokens
```
timosturm, line 478:
```suggestion
    :return: tuple of a dictionary of token dictionaries and a list of serial of not imported tokens
```

privacyidea/api/token.py
```diff
@@ -964,8 +967,8 @@ def loadtokens_api(filename=None):
     elif file_type in ["yubikeycsv", "Yubikey CSV"]:
         TOKENS = parseYubicoCSV(file_contents)
     elif file_type in ["pskc"]:
-        TOKENS = parsePSKCdata(file_contents, preshared_key_hex=aes_psk,
-                               password=aes_password)
+        TOKENS, not_imported_names = parsePSKCdata(file_contents, preshared_key_hex=aes_psk,
```
timosturm, line 972:
```suggestion
        TOKENS, not_imported_serials = parsePSKCdata(file_contents, preshared_key_hex=aes_psk,
```

privacyidea/api/token.py
```diff
@@ -987,7 +990,7 @@ def loadtokens_api(filename=None):
                         'success': True})
     # logTokenNum()
 
-    return send_result(len(TOKENS))
+    return send_result({'n_imported': len(TOKENS), 'n_not_imported': len(not_imported_names)})
```
timosturm, line 995:
```suggestion
    return send_result({'n_imported': len(TOKENS), 'n_not_imported': len(not_imported_serials)})
```

privacyidea/api/token.py
```diff
@@ -902,6 +904,7 @@ def loadtokens_api(filename=None):
         "oathcsv" or "yubikeycsv".
     :jsonparam tokenrealms: comma separated list of realms.
     :jsonparam psk: Pre Shared Key, when importing PSKC
+    :jsonparam pskcValidateMAC: Determines how invalid MACs should be handled when importing PSKC
```
cornelinux, line 907: The API documentation is generated from the API docstrings. So it would be helpfull to add s.th. like
"Allowed values are 'no_check', 'check_fail_soft' and 'check_fail_hard'". Since this then will be in a prominent location at privacyidea.readthedocs.io.

privacyidea/lib/importotp.py
```diff
@@ -455,6 +455,7 @@ def derive_key(xml, password):
 def parsePSKCdata(xml_data,
                   preshared_key_hex=None,
                   password=None,
+                  validate_mac='check_fail_hard',
```
cornelinux, line 458: Just one thought, sometimes it might be safer to add a new parameter at the end of the parameter list in case there is any call to this function, that does not specify the keywords, like a call to:

    parsePKSCdata(data, None, "secret", True)

when the caller wants to check the serials. 
(No need to change this, just for information)

privacyidea/lib/importotp.py
```diff
@@ -540,7 +573,12 @@ def parsePSKCdata(xml_data,
                 token["timeShift"] = key.data.timedrift.text.strip()
 
         tokens[serial] = token
-    return tokens
+
+    if abort:
+        not_imported_serials = [serial for serial in tokens]
```
cornelinux, line 578: If you want to get the list of the keys from a dictionary, you can simply do this like this:

    not_imported_serials = list(tokens)

privacyidea/lib/importotp.py
```diff
@@ -519,18 +527,43 @@ def parsePSKCdata(xml_data,
...
+                    enc_data_bin = base64.b64decode(enc_data)
+                    hm = hmac.new(key=mac_key, msg=enc_data_bin, digestmod=hashlib.sha1)
+                    mac_value_calculated = b64encode_and_unicode(hm.digest())
+
+                    mac_value_xml = key.data.find('valuemac').text.strip()
+
+                    is_invalid = not hmac.compare_digest(mac_value_xml, mac_value_calculated)
+
+                    if is_invalid and validate_mac == 'check_fail_hard':
+                        abort = True
```
cornelinux, line 556: If you actually want to abort and leave the for-loop on an error, you can add a

~~~~python
break
~~~~
here, which will leave the for-loop immediately.

privacyidea/static/components/token/views/token.import.html
cornelinux: No :-)
Just mark it, that it should be translated.

privacyidea/static/components/token/views/token.import.html
```diff
@@ -69,12 +69,27 @@ <h3 translate>Import tokens from a file</h3>
...
+                You can choose how the authenticity the stored tokens should be verified.
```
cornelinux, line 81: I would say

    "You can choose how the authenticity of the tokens to be imported should be verified".

I might be otherwise confusing, that the "stored tokens" would be the token in the privacyIDEA database.

privacyidea/static/components/token/views/token.import.html
```diff
@@ -124,4 +139,8 @@ <h3 translate>Import tokens from a file</h3>
     <div translate>
         Uploaded Tokens: {{ uploadedTokens }}
     </div>
+    <div>
+        <!--        How to translate this?-->
+        Skipped Tokens: {{ notImportedTokens }}
```
cornelinux, line 144: You can either do:

~~~~html
<div translate>
    Skipped Tokens: {{ notImportedTokens }}
</div>
~~~~

Or 

~~~~html
<div>
    <translate>Skippted Tokens<translate>: {{ notImportedTokens }}
</div>
~~~~

I think the 2nd is better.

tools/privacyidea-token-janitor
```diff
@@ -509,13 +511,18 @@ def find(last_auth, assigned, active, tokeninfo_key, tokeninfo_value,
...
```
cornelinux, line 528: Yes, we might add s.th. like

~~~~python
not_parsable_tokens = []
~~~~
and add simply the serial numbers to this list.

Furher below we could add

~~~~python
if not_parsable_tokens:  # Only print if tokens could not be parsed
    print("The following tokens could not be read from the PSKC file: {0!s}".format(not_parsable_tokens))
~~~~

cornelinux (changes requested):
Nice. I added some comments and requests.

privacyidea/lib/importotp.py
timosturm: This is done to get the list of all serials at the end, if I break the loop here I would have to parse all serials from the .xml again later.

tools/privacyidea-token-janitor
timosturm: Should this be done instead of `failed_tokens` or additionally?

privacyidea/lib/importotp.py
cornelinux: OK, thx.

tools/privacyidea-token-janitor
cornelinux: ``failed_tokens`` are the tokens, that have been successfully parsed from the xml file but could not be imported to privacyIDEA.
``not_parsable_tokens`` would already fail the parsing.
So yes, I would do this in addition.

privacyidea/lib/importotp.py
cornelinux: OK. thx

cornelinux (approved):

PR #2505, opened 2020-11-11
add privacy statement link to baseline

### LINKED ISSUE #2325: link to privacy statment (GDPR)
If privacyIDEA is publically facing the company running privacyIDEA should be able to provide a link for the privacystatement.

This link should be available in the footer of the website availble without the need to login.

The URL should be configurable via policy.

As long as no URL is configured the privacy statement would not be displayed.

* [x] Define policy
* [x] pass config to UI
* [x] enhance HTML template

laclaro:
This PR adds a policy to add a custom link to the a privacy statement in the baseline.

closes #2325

### DIFF
--- a/privacyidea/static/templates/baseline.html
+++ b/privacyidea/static/templates/baseline.html
@@ -25,5 +25,11 @@
                 target="documentation">
             <translate>Online Documentation</translate>
         </a>
+       <a class="navbar-text pull-right pull-right"
+           ng-show="piGDPRLink"
+           href="{{ piGDPRLink }}"
+           target="_external">
+            <translate>Privacy Statement</translate>
+        </a>
     </div>
 </div>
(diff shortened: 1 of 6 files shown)


privacyidea/static/templates/baseline.html
```diff
@@ -25,5 +25,11 @@
                 target="documentation">
             <translate>Online Documentation</translate>
         </a>
+       <a class="navbar-text pull-right pull-right"
+           ng-show="piGDPRLink"
+           href="{{ piGDPRLink }}"
+           target="_external">
+            <translate>Privacy Statement</translate>
```
plettich, line 32: So if someone wants to change the link text they must change the baseline template?
And what about an imprint link? Should we provide a functionality for that as well?

plettich (commented):
Looks good, please see my comments

privacyidea/static/templates/baseline.html
laclaro: I think we should keep it simple. If people want to change more, they should change the baseline.html themselves.

plettich (approved):

PR #2507, opened 2020-11-13
2284/registration token content length

### LINKED ISSUE #2284: Length and Contents of registration code

The registrationcode contains characters, that are hard to read and differentiate.
In some cases the registrationcode needs to be read and typed.

Add a policy in scope enrollment and action

* registrationcode_length
* registrationcode_contents

The contents would be passed as ``characters`` keyword argument in generate_password:
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/tokens/registrationtoken.py#L167

Enrollment policies should be defined in the tokeclass here:
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/tokens/registrationtoken.py#L133

# Some thoughts and discussions

Should we do a decorator, that modifies the actual input?
So should the ``/token/init?type=registractioncode`` actually come with these additional parameters "length" and "contents"?
Or should it be settings on the lib level?

I think adding these parameters to the API and have the decorator overwrite these, make more sense.

laclaro:
This PR implements the registrationcode_contents and length policies.

It uses the code that is also used to generate otp pins.

closes #2284


privacyidea/lib/tokens/registrationtoken.py
```diff
@@ -164,7 +165,17 @@ def update(self, param):
             # Otherwise genkey and otpkey will raise an exception in
             # PasswordTokenClass
             del param["genkey"]
-        param["otpkey"] = generate_password(size=self.otp_len)
+        if "registrationcode_length" in param:
```
cornelinux, line 168: We tried to to have a token specific parameter naming like ``<type>.<parameter>`` to avoid name clashes.
I recommend changing the parameter ot ``registration.length`` and ``registration.contents`` respectively.

privacyidea/api/token.py
```diff
@@ -161,6 +163,8 @@ def init():
     :jsonparam type: the type of the token
     :jsonparam tokenrealm: additional realms, the token should be put into
     :jsonparam otplen: length of the OTP value
+    :jsonparam registrationcode_length: the length of the generated registration code
```
cornelinux, line 166: Usually we do not have the token specific parameters in the API descritpion.
Please note my comment on changing the name of the parameters.

privacyidea/api/lib/prepolicy.py
```diff
@@ -506,6 +506,31 @@ def init_token_defaults(request=None, action=None):
...
+
+    If there is a valid policy set the action values of REGISTRATIONCODE_LENGTH
+    and REGISTRATIONCODE_CONTENTS are added to request.all_data as
+
+    { 'registrationcode_length': '10', 'registrationcode_contents': 'cn' }
+    """
+    params = request.all_data
+    tokentype = params.get("type")
+    if tokentype == "registration":
+        user_object = get_user_from_param(params)
```
cornelinux, line 521: You should set a default length and contents here:

~~~~
request.all_data["registration.length"] = 24
request.all_data["registration.length"] = please_check_what_the_default_is
~~~~

Otherwise a user or administrator will be able to manuall set a short registration code length if no policy is given. This is what I ment in my comment https://github.com/privacyidea/privacyidea/issues/2284#issuecomment-725384088

privacyidea/lib/tokens/registrationtoken.py
```diff
@@ -164,7 +165,17 @@ def update(self, param):
...
+            size = param["registrationcode_length"]
+            del param["registrationcode_length"]
+        else:
+            size = self.otp_len
+        if "registrationcode_contents" in param:
+            otpkey = _generate_pin_from_policy(param["registrationcode_contents"], size=int(size))
+            del param["registrationcode_contents"]
+        else:
+            otpkey = generate_password(size=int(size))
```
cornelinux, line 177: Can't we always use _generate_pin_from_policy, and provide the size:

    otpkey = _generate_pin_from_policy(param["registrationcode_contents"], size)

cornelinux (changes requested):
The decorator needs to set the default parameters

privacyidea/lib/tokens/registrationtoken.py
laclaro: if the parameter registration.contents has proper defaults, one can, indeed. Done.

privacyidea/api/lib/prepolicy.py
laclaro: I set the defaults in registrationtoken.py and import them where needed. 
I think the default contents were "cns". But since the length is already 24, I now set it to "cn" only.

privacyidea/api/token.py
laclaro: Removed.

privacyidea/lib/tokens/registrationtoken.py
laclaro: Changed.

privacyidea/api/lib/prepolicy.py
```diff
@@ -513,8 +513,11 @@ def init_registrationcode_length_contents(request=None, action=None):
     If there is a valid policy set the action values of REGISTRATIONCODE_LENGTH
     and REGISTRATIONCODE_CONTENTS are added to request.all_data as
 
-    { 'registrationcode_length': '10', 'registrationcode_contents': 'cn' }
+    { 'registration.length': '10', 'registration.contents': 'cn' }
```
cornelinux, line 516: The policyname could have stayed "registration_length" ;-)

privacyidea/api/lib/prepolicy.py
```diff
@@ -523,11 +526,17 @@ def init_registrationcode_length_contents(request=None, action=None):
...
             request.all_data[ACTION.REGISTRATIONCODE_LENGTH] = list(length_pols)[0]
+            no_length_policy = False
         content_pols = Match.user(g, scope=SCOPE.ENROLL, action=ACTION.REGISTRATIONCODE_CONTENTS,
                                 user_object=user_object).action_values(unique=True)
         if len(content_pols) == 1:
             request.all_data[ACTION.REGISTRATIONCODE_CONTENTS] = list(content_pols)[0]
-
+            no_content_policy = False
+    # if there is no policy, set defaults.
+    if no_length_policy:
```
cornelinux, line 536: I think, this is an unneccessary boolean, since you could have set the DEFAULT_LENGTH before checking for the policy directly. But I leave it up to you.

privacyidea/api/lib/prepolicy.py
```diff
@@ -523,11 +526,17 @@ def init_registrationcode_length_contents(request=None, action=None):
...
+        request.all_data[ACTION.REGISTRATIONCODE_LENGTH] = DEFAULT_LENGTH
+    if no_content_policy:
```
cornelinux, line 538: I think, this is an unneccessary boolean, since you could have set the DEFAULT_CONENT before checking for the policy directly. But I leave it up to you.

doc/policies/enrollment.rst
```diff
@@ -213,7 +213,7 @@ type: bool
 If set the OTP PIN of a token will be encrypted. The default
 behaviour is to hash the OTP PIN, which is safer.
 
-registrationcode_length
+registration.length
 ~~~~~~~~~~~~~~~~~~~~~~~
```
cornelinux, line 217: In restructed text, the length of the underline should match the title

doc/policies/enrollment.rst
```diff
@@ -222,7 +222,7 @@ type: int
 
 This is the length of the generated registration codes.
 
-registrationcode_contents
+registration.contents
 ~~~~~~~~~~~~~~~~~~~~~~~
```
cornelinux, line 226: In restructed text, the length of the underline should match the title

cornelinux (commented):
I only have some nitpicking left. You decide, if you want to change things. But for me this is no blocking for merging this. THanks

cornelinux (approved):

PR #2509, opened 2020-11-15
demote "missing HTTP headers" to warning in policy handling

jmozd:
After creating policies that use custom HTTP headers, for me, authentication via push tokens started to fail. I chased this down to reported exceptions within policy handling:

- a policy ("policyX") is defined, relying on a custom header (active condition comparing a header against a given value). Sample purpose of policyX is to inititate a push token request for a specific PAM scenario

- the client sends the "/validate/check" POST request including the header, traverses the policy chain and policyX works as expected

- i.e. the push token reply is coming in ("/ttype/push" GET request) and traverses the policy chain, but does not have the custom header

- when checking "policyX" against that "/ttype/push" request, the missing HTTP header is reported as an error, an exception is thrown and further processing seems to stop

This PR changes the behavior of _policy_matches_request_header_condition() to only log a warning and to return as if the comparison on the header failed.

The comparator parameter of _policy_matches_request_header_condition() was not (yet) included in the decision. Returning "true" for "inverse matches" (!in, !contains,...) would possibly create a different semantic, but of course this can be subject of discussion.

### DIFF
--- a/privacyidea/lib/policy.py
+++ b/privacyidea/lib/policy.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 #
+#  2020-11-15 Jens-U. Mozdzen <jmozdzen@nde.ag>
+#             missing http headers now are only reported as warnings
 #  2020-06-05 Cornelius Kölbel <cornelius.koelbel@netknights.it>
 #             Add privacyIDEA nodes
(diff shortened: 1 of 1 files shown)


cornelinux:
This PR can not be merged.
Please see my comment in the issue.
https://github.com/privacyidea/privacyidea/issues/2514

PR #2511, opened 2020-11-16
Fix missing blank in log messages

jmozd:
Messages for the "applications" LUKS and SSH missed a blank, separating msg stem and parameter:

[INFO][privacyidea.lib.applications.ssh:89] Token 'PIPU00XXXX', type 'push' is not supported bySSH application module


plettich (approved):
Thanks for the PR.
Usually we use issues to track the development but it is cumbersome to add an issue for every white space change :-)

PR #2517, opened 2020-11-16
Policy to force login with remote user

### LINKED ISSUE #2072: Force user to login with remote_user at webui
The webui policy [remote_user](https://privacyidea.readthedocs.io/en/latest/policies/webui.html#remote-user) shows a dialog which asks the user, if he wants to login with this REMOTE_USER.

It would be nice, if we could also configure, that the user is **always** logged in with the remote user.

As the policy setting is a string, maybe we can simply add an additional string value? 
like "allowed", "disable", "**force**"?

cornelinux:
If a remote user exists a policy can now force
the login with this user and not provide
the login-with-credentials button.

Closes #2072

I was thinking about if this policy is necessary at all or if we could simply use remote_user=allow and disallow the webui login.
This would be the same for users, who actually authenticated with BASIC Auth. But if a user does not authenticate with basic auth this user would not be able to login (since credential login would be overall disallowed).
So this way we have a slightly bigger flexibility to have only one button (login with BASIC auth - no login with credential auth) for a user who has a remote_user.

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -1229,7 +1229,7 @@ def is_remote_user_allowed(req, write_to_audit_log=True):
     :param req: The flask request, containing the remote user and the client IP
     :param write_to_audit_log: whether the policy name should be added to the audit log entry
     :type write_to_audit_log: bool
-    :return: 0 if disabled, 1 if allowrd, 2 if forced.
+    :return: 0 if disabled, 1 if allowed, 2 if forced.
     """
     res = 0
     if req.remote_user:
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -1229,9 +1229,9 @@ def is_remote_user_allowed(req, write_to_audit_log=True):
     :param req: The flask request, containing the remote user and the client IP
     :param write_to_audit_log: whether the policy name should be added to the audit log entry
     :type write_to_audit_log: bool
-    :return: 0 if disabled, 1 if allowed, 2 if forced.
+    :return: Return a value or REMOTE_USER, can be "disable", "active" or "force".
+    :rtype: basestring
     """
-    res = 0
     if req.remote_user:
         loginname, realm = split_user(req.remote_user)
         realm = realm or get_default_realm()
@@ -1242,12 +1242,10 @@ def is_remote_user_allowed(req, write_to_audit_log=True):
                                                                 write_to_audit_log=write_to_audit_log)
         # there should be only one action value here
         if ruser_active:
-            if list(ruser_active)[0] == REMOTE_USER.ACTIVE:
-                res = 1
-            if list(ruser_active)[0] == REMOTE_USER.FORCE:
-                res = 2
+            return list(ruser_active)[0]
 
-    return res
+    # Return default "disable"
+    return REMOTE_USER.DISABLE
 
 
 def save_client_application_type(request, action):
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -1229,9 +1229,9 @@ def is_remote_user_allowed(req, write_to_audit_log=True):
     :param req: The flask request, containing the remote user and the client IP
     :param write_to_audit_log: whether the policy name should be added to the audit log entry
     :type write_to_audit_log: bool
-    :return: a bool value
+    :return: 0 if disabled, 1 if allowrd, 2 if forced.
     """
-    res = False
+    res = 0
     if req.remote_user:
         loginname, realm = split_user(req.remote_user)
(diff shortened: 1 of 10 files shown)


privacyidea/api/lib/prepolicy.py
```diff
@@ -1243,7 +1243,9 @@ def is_remote_user_allowed(req, write_to_audit_log=True):
         # there should be only one action value here
         if ruser_active:
             if list(ruser_active)[0] == REMOTE_USER.ACTIVE:
-                res = True
+                res = 1
+            if list(ruser_active)[0] == REMOTE_USER.FORCE:
+                res = 2
```
plettich, line 1248: why not return the action string itself? It makes things more clear in my opinion.
We could rename the function to `remote_user_allowed` and just check in the two places where it is called if the result matches the strings (or use some kind of enum).

privacyidea/lib/policy.py
```diff
@@ -406,6 +405,7 @@ class REMOTE_USER(object):
     __doc__ = """The list of possible values for the remote_user policy."""
     DISABLE = "disable"
     ACTIVE = "allowed"
+    FORCE = "force"
```
plettich, line 408: Currently there is no way of setting the action to `force` in the UI or did i miss something?
I couldn't really get it to work even when changing the value in the database...

plettich (changes requested):

privacyidea/api/lib/prepolicy.py
cornelinux: Sounds good to me. I will look into this.

privacyidea/api/lib/prepolicy.py
```diff
@@ -1229,9 +1229,9 @@ def is_remote_user_allowed(req, write_to_audit_log=True):
     :param req: The flask request, containing the remote user and the client IP
     :param write_to_audit_log: whether the policy name should be added to the audit log entry
     :type write_to_audit_log: bool
-    :return: a bool value
+    :return: 0 if disabled, 1 if allowrd, 2 if forced.
```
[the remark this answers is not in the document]
cornelinux: Will do

privacyidea/lib/policy.py
cornelinux: I was missing it in the policy definition.

cornelinux: Setting UI in 458eb975

privacyidea/api/lib/prepolicy.py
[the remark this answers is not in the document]
cornelinux: Done in 458eb975

privacyidea/api/lib/prepolicy.py
cornelinux: Done in ae99a827

cornelinux:
Check it with remote_user=force:

image


and remote_user=allow:

image

privacyidea/api/lib/prepolicy.py
```diff
@@ -1229,9 +1229,9 @@ def is_remote_user_allowed(req, write_to_audit_log=True):
     :param req: The flask request, containing the remote user and the client IP
     :param write_to_audit_log: whether the policy name should be added to the audit log entry
     :type write_to_audit_log: bool
-    :return: a bool value
+    :return: Return a value or REMOTE_USER, can be "disable", "active" or "force".
+    :rtype: basestring
```
plettich, line 1233: ... for REMOTE_USER...
and please use `str` as type, `basestring` is so Python 2.7...

tests/test_api_lib_policy.py
```diff
@@ -1270,23 +1270,23 @@ def test_13_remote_user(self):
...
         req = Request(env)
-        self.assertTrue(is_remote_user_allowed(req))
+        self.assertEqual(REMOTE_USER.ACTIVE, is_remote_user_allowed(req))
 
         set_privacyidea_config(SYSCONF.SPLITATSIGN, False)
-        self.assertFalse(is_remote_user_allowed(req))
+        self.assertEqual(REMOTE_USER.DISABLE, is_remote_user_allowed(req))
 
         set_privacyidea_config(SYSCONF.SPLITATSIGN, True)
         delete_policy("ruser")
```
plettich, line 1292: Could You also write a simple check for `REMOTE_USER.FORCE`?
And possibly also an API test case (in `test_api_auth.py`)?

plettich (changes requested):

privacyidea/api/lib/prepolicy.py
cornelinux: Done in e8b69ceb

tests/test_api_lib_policy.py
cornelinux: Done in fb23f58d
Done in dba4c4f4

plettich (approved):
looks good

PR #2519, opened 2020-11-17
Always create Push-Challenge if polling is enabled

### LINKED ISSUE #2438: Create challege even if message could not be submitted to firebase
Can you check that a challenge is generated for a push token even if privacyidea was not able to submit the message to the firebase service?
This happens if the firebase token of the app has changed, but using push-poll, the app is still able to respond to the challenge without using firebase at all.

plettich:
Even if submitting the message for the Push-notification to the firebase
service fails a challenge is created if polling is enabled (either
globally or for that token specifically).
The App is then able to poll for open challenges and answer them
correctly.

Fixes #2438

### DIFF
--- a/privacyidea/lib/tokens/pushtoken.py
+++ b/privacyidea/lib/tokens/pushtoken.py
@@ -735,9 +735,9 @@ class PushTokenClass(TokenClass):
             # But only if polling is allowed.
             allow_polling = get_action_values_from_options(
                 SCOPE.AUTH, PUSH_ACTION.ALLOW_POLLING, options=options) or PushAllowPolling.ALLOW
-            if (allow_polling == PushAllowPolling.ALLOW or
-                    (allow_polling == PushAllowPolling.TOKEN and
-                     is_true(self.get_tokeninfo(POLLING_ALLOWED, default='True')))):
+            if ((allow_polling == PushAllowPolling.ALLOW or
+                 (allow_polling == PushAllowPolling.TOKEN and
+                  is_true(self.get_tokeninfo(POLLING_ALLOWED, default='True')))) or res):
                 validity = int(get_from_config('DefaultChallengeValidityTime', 120))
                 tokentype = self.get_tokentype().lower()
                 # Maybe there is a PushChallengeValidityTime...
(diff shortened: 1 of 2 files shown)


privacyidea/lib/tokens/pushtoken.py
```diff
@@ -729,6 +729,32 @@ def create_challenge(self, transactionid=None, options=None):
                                                      challenge, fb_gateway,
                                                      pem_privkey, options)
             res = fb_gateway.submit_message(self.get_tokeninfo("firebase_token"), smartphone_data)
+
+            # Create the challenge in the challenge table regardless of a failure
+            # when sending the Push message to the Firebase API.
+            # But only if polling is allowed.
+            allow_polling = get_action_values_from_options(
+                SCOPE.AUTH, PUSH_ACTION.ALLOW_POLLING, options=options) or PushAllowPolling.ALLOW
+            if (allow_polling == PushAllowPolling.ALLOW or
```
cornelinux, line 738: Maybe I am mislead here. But we also need to create a challenge in the database if the message is successfully submitted to firebase. Even if not polling is allowed. 
(Which would be the default behaviour. We sill need the challenge, in the database for the token to work correctly like before.)

So to my understanding this should be

~~~~python
if (polling-conditions) or (res):
~~~~

cornelinux (changes requested):
I think authentication will always fail, if polling is not allowed!

privacyidea/lib/tokens/pushtoken.py
plettich: You are absolutely right, good thing we have a review process :-)

plettich: Done in cb4c8fb

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -729,6 +729,32 @@ def create_challenge(self, transactionid=None, options=None):
                                                      challenge, fb_gateway,
                                                      pem_privkey, options)
             res = fb_gateway.submit_message(self.get_tokeninfo("firebase_token"), smartphone_data)
+
+            # Create the challenge in the challenge table regardless of a failure
```
cornelinux, line 733: The if-statement actually now means s.th. like

Create the challenge in the challenge table if either the messages was successfully submitted
to the Firebase API or if polling is allowed in general or for this specific token.

cornelinux (changes requested):
Only nitpicking about the comment...

privacyidea/lib/tokens/pushtoken.py
plettich: Done in 012b6af7

cornelinux (approved):

PR #2521, opened 2020-11-17
Fix the auth count in the dashboard

### LINKED ISSUE #2473: Investigate: Not the correct authentication number in the Dashboard
### Top-level intent

View the authentication request in the dashbard

### Actual outcome

In certain cases the dashboard displays a lower number than there are actually /validate/check requests.

The dashboard displays these requests by using the parameter ``timelimit=1d`` (one day).
Could it be that there is an issue with this parameter or is the number simply limited to 15 from the pagination?

cornelinux:
The dashboard used a GET /audit that
was limited to the first 15-entry page of
all validate-entries.

We now issue two requests for successful and
failed requests and use the overall count.

Fixes #2473


laclaro (approved):
Great, thank's

PR #2527, opened 2020-11-21
Add link for the PUSH-Enrollment

### LINKED ISSUE #2522: Enroll Push Token via Link
Enroll a push token without the need of a second device.
An additional Link to the QR Code displayed which opens the Authenticator and enroll it

**Additional context**
Normally our users enroll their push token on the office PCs.
Sometimes it is needed to ship them a company Laptop with VPN Client and enabled two factor before they enrolled their token.
Some users do not have a second device which can open the PrivacyIdea Site to log in with their registration token and scan the QR code with the app.
So it would be great to do it on one device.

cornelinux:
We add a link that contains the contents of
the QR Code, so that the user on a smartphon
can actually click this link to enroll the
PUSH token.

Fixes #2522


laclaro (approved):
Link is shown as expected.

PR #2532, opened 2020-11-21
2529/audit.success

### LINKED ISSUE #2529: The "1" and "0" is missing in the audit.succss
The audit log is missing the success entry.

image

The JSON contains the data, so it only seems a display issue.

### LINKED ISSUE #2530: Audit Columns can not be filtered-
The audit columns can not be filtered anymore.
This flaw was introduced in the merge of hiding the audit columns.

``ng-if`` mangles with the DOM, so that ``piFilter`` does not work anymore. 
We need to replace ``ng-if`` with ``ng-show``. Then the filtering works again.

### LINKED ISSUE #2531: Audit log completely disappears
The audit log vanishes if we filter for s.th. that does not exist.

Try to fillter for "date" and enter "X".
All columns vanish and you can not make them appear again, since you can not remove the X.

This was not the way in 3.4.1 and happens in current master.

cornelinux:
Fixes #2529 
Fixes #2530 
Fixes #2531 

@laclaro You can take  look at each single commit, since each commit fixes another issue.

### DIFF
--- a/privacyidea/lib/audit.py
+++ b/privacyidea/lib/audit.py
@@ -92,7 +92,7 @@ def search(config, param=None, user=None):
     page_size = 15
     page = 1
     timelimit = None
-    hidden_columns = None
+    hidden_columns = []
     # The filtering dictionary
     param = param or {}
     # special treatment for:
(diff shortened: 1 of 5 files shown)


laclaro (approved):
Thank's for these fixes. It works smoothly now.

PR #2533, opened 2020-11-21
Cast the filter condition of success to int

### LINKED ISSUE #751: DB model incorrect for pidea_audit.success, causes auth failure due to underlying DB failure (type incompatibility)
# What did you try to do?

Authenticate via pam_privacyidea

Successful authentication

Login failed.

Error log:

ProgrammingError: (ProgrammingError) operator does not exist: integer = boolean
LINE 4: ...audit."user" = 'darthmdh' AND pidea_audit.success = false AND ...
HINT: No operator matches the given name and argument type(s).  You might need to add explicit type casts.

# Configuration

* privacyIDEA Version: 2.19.1

* Installation method:
Ubuntu packages (privacyidea PPA)

## more details: 

* OS: Linux

* Webserver: nginx

* Tokendatabase: Postgresql

# Other info
Since PostgreSQL is strongly typed, it appears the constant use of boolean types "true" and "false" with the pidea_audit.success value conflicts with its definition as an INTEGER field.  If its meant to be a BOOLEAN type then the field should be defined as such in the DB model.  This should work transparently with MySQL as that server aliases BOOLEAN to TINYINT internally so the definition is compatible.
Core problematic files seem to be:
lib/auditmodules/sqlaudit.py
models.py

cornelinux:
In all cases where we filter for "success" on the
SQL audit table we cast to int.

Fixes #751

### DIFF
--- a/privacyidea/lib/auditmodules/sqlaudit.py
+++ b/privacyidea/lib/auditmodules/sqlaudit.py
@@ -46,7 +46,7 @@ from privacyidea.lib.crypto import Sign
 from privacyidea.lib.pooling import get_engine
 from privacyidea.lib.utils import censor_connect_string
 from privacyidea.lib.lifecycle import register_finalizer
-from privacyidea.lib.utils import truncate_comma_list
+from privacyidea.lib.utils import truncate_comma_list, is_true
 from sqlalchemy import MetaData, cast, String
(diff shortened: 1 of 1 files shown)


plettich (approved):
looks good

PR #2535, opened 2020-11-27
make g.serial available everywhere

### LINKED ISSUE #2534: Add serial number to g object
In many cases we need the serial number of the token. We need this for policies and event handlers.

There are three possiibilities where the serial number originiates from

* The serial number can be contained in a parameter of the request
* it can bin  in the URL of the endpoint
* or it can be determined during the request, e.g. when the user authenticates and privacyIDEA determines, which serial number (token) was used

The first can be done by ``before_after``, 
the second can be done by a decorator for the specific methods and the third is currently not clear.

laclaro:
This PR implements privacyidea.api.lib.decorators with the class `APIDecorator`.

It is used to modify things based on the request and response, similar to the policy decorators.

The method `g_add_serial` decorates (hopefully all) serial-related functions to either use the URL, the request or the response to set g.serial.

closes #2534

### DIFF
--- a/privacyidea/api/before_after.py
+++ b/privacyidea/api/before_after.py
@@ -178,7 +178,7 @@ def before_request():
                          request.host
     # Already get some typical parameters to log
     serial = getParam(request.all_data, "serial")
-    if serial:
+    if serial and not serial == '**':
         g.serial = serial
         tokentype = get_token_type(serial)
     else:
--- a/privacyidea/api/before_after.py
+++ b/privacyidea/api/before_after.py
@@ -179,8 +179,10 @@ def before_request():
     # Already get some typical parameters to log
     serial = getParam(request.all_data, "serial")
     if serial:
+        g.serial = serial
         tokentype = get_token_type(serial)
     else:
+        g.serial = None
         tokentype = None
 
     if request.User:
--- a/privacyidea/api/lib/decorators.py
+++ b/privacyidea/api/lib/decorators.py
@@ -20,12 +20,11 @@
 These are the decorators which run before processing the request and may
 modify API calls e.g. by changing the flask environment.
 
-The functions of this module are tested in tests/test_api_lib_decorators.py
+The postAddSerialToG decorator is tested in the ValidateAPITestCase.
 """
 
 
 import logging
-from privacyidea.api.lib.utils import getParam
 from flask import g
 import functools
 
@@ -35,82 +34,37 @@ optional = True
 required = False
 
 
-class APIDecorator(object):
+class postAddSerialToG(object):
     """
-    This is the decorator to either wrap an API function call or act as response-modifier.
-    The behavior is given by the argument 'position'. With position = "post" it runs
-    after the request and else (position = "pre") wraps the API function itself.
-
+    This decorator checks for the serial in the response and adds it to the
+    flask g object.
     """
-    def __init__(self, function, request, position="pre", options=()):
+    def __init__(self, request):
         """
-        :param function: This is the function the is to be called
-        :type function: function
         :param request: The original request object, that needs to be passed
         :type request: Request Object
-        :param options: Options which are passed to the decorator function
-        :type options: dict
-        :param position: The behavior of the decorator. With "post", it executes
-        after the request. With "pre" (default), the API function itself is wrapped
-        by the decorator.
         """
-        self.function = function
         self.request = request
-        self.position = position
-        self.options = options
 
     def __call__(self, wrapped_function):
         """
-        This decorates the given function according to the given position.
-
         :param wrapped_function: The function, that is decorated.
         :type wrapped_function: API function
         :return: None
         """
         @functools.wraps(wrapped_function)
         def function_wrapper(*args, **kwds):
-            if self.position == "post":
-                response = wrapped_function(*args, **kwds)
-                return self.function(self.request,
-                                     response=response,
-                                     options=self.options,
-                                     *args, **kwds)
-            else:
-                self.function(self.request,
-                              options=self.options)
-                return wrapped_function(*args, **kwds)
+            response = wrapped_function(*args, **kwds)
+            return add_serial_to_g_from_response(response)
 
         return function_wrapper
 
 
-def g_add_serial(request, response=None, options=()):
+def add_serial_to_g_from_response(response):
     """
-    This decorator function fetches the token serial and adds it to the flask
-    g object.
-    With a given response, this decorator checks for the serial in the response.
-    If no response is given, the decorator checks first in the request and in the
-    request path for the serial. By default the serial is expected at the third
-    position of the request path. Set the position of the serial in the URL by
-    specifying
-
-    options = {"serial_position": N}
-
-    N defaults to 3.
+    This function adds the token serial from the response to the flask g object
     """
-    if response is None:
-        if "serial" not in g or not g.serial:
-            # try to get serial from request (also done by before_after.py)
-            g.serial = getParam(request.all_data, "serial")
-            if not g.serial:
-                # try to get serial from request path
-                if request.path:
-                    if "serial_position" not in options:
-                        options = {"serial_position": 3}
-                    path_elements = [x for x in request.path.split('/') if x]
-                    if len(path_elements) > options["serial_position"] + 1:
-                        g.serial = path_elements[options["serial_position"] + 1]
-        return True
-    elif response.is_json:
+    if response.is_json:
         serial = response.json.get("detail", {}).get("serial")
         if serial:
             g.serial = serial
--- a/privacyidea/api/lib/decorators.py
+++ b/privacyidea/api/lib/decorators.py
@@ -70,8 +70,8 @@ class APIDecorator(object):
         @functools.wraps(wrapped_function)
         def function_wrapper(*args, **kwds):
             if self.position == "pre":
-                    self.function(self.request,
-                                  options=self.options)
+                self.function(self.request,
+                              options=self.options)
             elif self.position == "post":
                 response = wrapped_function(*args, **kwds)
                 return self.function(self.request,
--- a/privacyidea/api/lib/decorators.py
+++ b/privacyidea/api/lib/decorators.py
@@ -69,16 +69,16 @@ class APIDecorator(object):
         """
         @functools.wraps(wrapped_function)
         def function_wrapper(*args, **kwds):
-            if self.position == "pre":
-                self.function(self.request,
-                              options=self.options)
-            elif self.position == "post":
+            if self.position == "post":
                 response = wrapped_function(*args, **kwds)
                 return self.function(self.request,
                                      response=response,
                                      options=self.options,
                                      *args, **kwds)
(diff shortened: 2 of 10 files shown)


privacyidea/api/lib/postpolicy.py
```diff
@@ -761,3 +761,13 @@ def is_authorized(request, response):
             raise ValidateError("User is not authorized to authenticate under these conditions.")
 
     return response
+
+
+def token_serial_from_response_g(request, response):
+    """
+    This is decorator function fetches the token serial from the response
```
cornelinux, line 768: I think you should remove the "is".

privacyidea/api/lib/prepolicy.py
```diff
@@ -1201,6 +1201,24 @@ def check_external(request=None, action="init"):
     return True
 
 
+def token_serial_from_path_g(request=None, action=None):
+    """
+    This is decorator function fetches the token serial from the route
```
cornelinux, line 1206: I think you should remove the "is".

privacyidea/api/lib/postpolicy.py
```diff
@@ -761,3 +761,13 @@ def is_authorized(request, response):
...
+    and adds it to the flask g object. It decorates api functions where the serial
+    is only determined during the call, e.g. validate/check.
+    """
+    if g.serial is not None:
+        g.serial = response.json.get("detail", None).get("serial")
```
cornelinux, line 773: This will raise an exception, if the response does not contain a "detail", since it results in ``None.get("serial")``.

You should do 

    response.json.get("detail", {}).get("serial")

privacyidea/api/validate.py
```diff
@@ -201,6 +201,7 @@ def offlinerefill():
 @postpolicy(check_tokeninfo, request=request)
 @postpolicy(check_tokentype, request=request)
 @postpolicy(check_serial, request=request)
+@postpolicy(token_serial_from_response_g, request=request)
```
cornelinux, line 204: Do we also need this at the endpoint ``triggerchallenge``?

privacyidea/api/lib/prepolicy.py
```diff
@@ -1201,6 +1201,24 @@ def check_external(request=None, action="init"):
...
+    and adds it to the flask g object. It decorates all api functions
+    where serial can be part of the url.
+    """
+    if g.serial is not None:
+        if request.path:
+            # request.path is something like '/token/setpin/SERIAL' or '/token/SERIAL'
+            path_list = [x for x in request.path.split('/') if x]
+            if path_list[0] == "token" and len(path_list) == 2:
```
cornelinux, line 1214: I am wondering, if this logic to identify *where* the potential serial can be located in the path will always work as expected or if we could run into trouble.

Please consider adding the position of the serial as parameter to the decorator like this:

~~~~python
def token_serial_from_path_g(request=None, action=None, serial_position=2)
~~~~

Then we do not have to have any knowledge about the structure of the endpoints within the decorator.

We could use the decorator like this:

~~~~python
@token_blueprint.route('/disable', methods=['POST'])
@token_blueprint.route('/disable/<serial>', methods=['POST'])
@prepolicy(token_serial_from_path_g, request, serial_position=2)
def disable_api(serial=None):
~~~~

This way we would put the logical decision right to the place where the endpoint is decorated and keep the logic out of the ``token_serial_from_path_g``.
This would especially help, if one day we would have other endpoints, were e.g. the serial is located at position 4 or 5. (I am not planning on this, but you never know).

cornelinux (changes requested):
Looks cool. Please note may comments.

privacyidea/api/lib/postpolicy.py
laclaro: I already experienced this. Changed.

privacyidea/api/lib/prepolicy.py
laclaro: that's a very good idea indeed.

laclaro:
I could not get a test to work and check flask g afterwards. We may do no specific check here, but only an indirect one for the functionality which uses g.serial.

privacyidea/api/validate.py
laclaro: I first thought no, but I think we need it.

My thought: In the next step, you may have an admin policy "triggerchallenge" with a tokeninfo for the triggeradmin.
If the admin now triggers a challenge with serial, before_after will set g.serial -> no decorator needed.
If you trigger the challenge without serial for a specific user, you would like to have the policy condition tokeninfo act on this action. -> decorator needed.

privacyidea/api/lib/prepolicy.py
laclaro: Ah, we cannot easily pass arbitrary arguments to policy decorators. Lets talk about this.

cornelinux: We do this in the event decorator
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/api/token.py#L150

privacyidea/api/lib/postpolicy.py
laclaro: done

privacyidea/api/lib/prepolicy.py
laclaro: done

privacyidea/api/validate.py
laclaro: Triggerchallenge is now decorated twice as pre and post, since the serial may be in the request or not.

Actually, if the triggerchallenge has a serial in the request, we may end up with NO serial in the response. To not override a given serial by None, the g.serial is only set by the post-decoration if it was NOT set before. I cannot think of any case, where this is a problem. Can you?

privacyidea/api/lib/prepolicy.py
laclaro: The commit https://github.com/privacyidea/privacyidea/pull/2535/commits/e0d18e04fafd70a7510f00d94ebf2fbe11c736ef implements the APIDecorator class which can be used as a pre or a post decorator and accepts options to pass to the decorating function.

https://github.com/privacyidea/privacyidea/pull/2535/commits/e0d18e04fafd70a7510f00d94ebf2fbe11c736ef#diff-2f390b1f7e67d57678da243f6a900cb3ee0fe1a7137732217e0c7d5d87ab9db2

privacyidea/api/lib/decorators.py
```diff
@@ -0,0 +1,119 @@
...
+    specifying
+
+    options = {"serial_position": N}
+
+    N defaults to 3.
+    """
+    if response is None:
+        if "serial" not in g or not g.serial:
+            # try to get serial from request (also done by before_after.py)
+            g.serial = getParam(request.all_data, "serial")
```
cornelinux, line 103: is this necessary, if we already do this in before_after? I do not think so or am I missing s.th?

privacyidea/api/lib/decorators.py
```diff
@@ -0,0 +1,119 @@
...
+            if not g.serial:
+                # try to get serial from request path
+                if request.path:
+                    if "serial_position" not in options:
+                        options = {"serial_position": 3}
+                    path_elements = [x for x in request.path.split('/') if x]
+                    if len(path_elements) > options["serial_position"] + 1:
+                        g.serial = path_elements[options["serial_position"] + 1]
+        return True
```
cornelinux, line 112: I would like this method to get the position directly in the ``option`` parameter.

Then we could do it like this:

~~~~python
if request.path:
    position = option or 2 # 2 is the default?
    path_elements = [x for x in request.path.split('/') if x]
    if len(path_elements) >= position:
        g.serial = path_elements[position -1]

privacyidea/api/lib/decorators.py
```diff
@@ -0,0 +1,119 @@
...
+            # try to get serial from request (also done by before_after.py)
+            g.serial = getParam(request.all_data, "serial")
+            if not g.serial:
+                # try to get serial from request path
+                if request.path:
+                    if "serial_position" not in options:
+                        options = {"serial_position": 3}
+                    path_elements = [x for x in request.path.split('/') if x]
+                    if len(path_elements) > options["serial_position"] + 1:
+                        g.serial = path_elements[options["serial_position"] + 1]
```
cornelinux, line 111: I think this must be -1

privacyidea/api/machine.py
```diff
@@ -181,6 +182,7 @@ def attach_token_api():
 @machine_blueprint.route('/token/<serial>/<machineid>/<resolver>/<application>',
                          methods=['DELETE'])
 @prepolicy(check_base_action, request, ACTION.MACHINETOKENS)
+@APIDecorator(g_add_serial, request=request, options={"serial_position": 2})
```
cornelinux, line 185: Such a call would then by

~~~~python
@APIDecorator(g_add_serial, request=request, option=2)
~~~~

privacyidea/api/token.py
```diff
@@ -572,6 +577,7 @@ def disable_api(serial=None):
 
 
 @token_blueprint.route('/<serial>', methods=['DELETE'])
+@APIDecorator(g_add_serial, request, options={"serial_position": 1})
```
cornelinux, line 580: @APIDecorator(g_add_serial, request, option=1)

cornelinux (changes requested):
I think we should simplify the interface of the decorator!
Lets discuss this and I can explain - or you.

Please add tests!

privacyidea/api/before_after.py
```diff
@@ -178,9 +178,11 @@ def before_request():
                          request.host
     # Already get some typical parameters to log
     serial = getParam(request.all_data, "serial")
-    if serial:
+    if serial and not serial == '**':
```
cornelinux, line 181: It could also be *substr*, so I would do

~~~~python
if serial and not serial.startswith("*"):
~~~~

privacyidea/api/lib/decorators.py
```diff
@@ -0,0 +1,73 @@
...
+optional = True
+required = False
+
+
+class postAddSerialToG(object):
+    """
+    This decorator checks for the serial in the response and adds it to the
+    flask g object.
+    """
+    def __init__(self, request):
```
cornelinux, line 42: We do not need the request.

privacyidea/api/lib/decorators.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        @functools.wraps(wrapped_function)
+        def function_wrapper(*args, **kwds):
+            response = wrapped_function(*args, **kwds)
+            return add_serial_to_g_from_response(response)
+
+        return function_wrapper
+
+
+def add_serial_to_g_from_response(response):
```
cornelinux, line 63: We can omit this function and directly put it into the __call__

And we can change the decorator-class to a decorator-function.
We only need a decorator-class, if the decorator acutally needs parameters.

cornelinux (changes requested):
@laclaro Looks nice and easy. I have some changes locally (according to my comments); which will make it even easier. I will show you.

Issue #2536, opened 2020-11-28
move push token at the end of the token list so that it will not block


cornelinux:
It looks like this is absolutely not possible in the current design, if both push tokens have the same PIN! 
due to the linear code of the ``check_token_list``.

A bit more simpler thing we could do is:

* [x] sort the tokenlist and put the push token at the end

  If we sort the token list this way, that push tokens are processed at the end of this list [here](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/token.py#L2163) we can simply implement the below mentioned 2nd step. Sorting could be done by weighing the tokens, to know which tokens to put at the beginning of the list and which at the end of this list. For now we want to put push tokens at the end of the list, but this might be different in other cases.

  This way we will be able to run through the token list with the push tokens being last, all other tokens sorted in alpabetical order:
  ~~~~python
  def tokentype_weighing(tok):
     if tok.type.upper() == "PUSH":
         return 1000
     else:
         return ord(tok.type[0])
  
  for tokenobject in (sorted(tokenobject_list, tokentype_weighing):
     ....
  ~~~~

* [x] If a token in the list authenticates successfully it is stored in ``valid_token_list`` [here](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/token.py#L2196). The token is ``authenticate``d [here](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/token.py#L2186). 
At this point (``authenticate``) we could also pass the ``validate_token_list`` or the length of the list to the push token. If there are already successful authentication, the push token can decide in case of *push_wait* to not authenticate anymore (since it would block the successful authentication of the previous token). 

  We do not need to enhance the interface of ``tokenclass.authenticate``, we could add the length of the valid_token_list in the ``options``.

  [here](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/tokens/pushtoken.py#L873) we need to add s.th. like

  ~~~~python
  if int(options.get("valid_token_numbers", 0)) > 0:
      # We already have a successful authentication, so we pass
      return False, -1, None
  ~~~~

PR #2539, opened 2020-11-30
Sort token list and put Push at the end

cornelinux:
In the check_token_list in token.py we
sort the token list this way to put all
push tokens at the end of the list, so that
they get checked last.

Working on #2536


plettich (approved):
Looks good

PR #2540, opened 2020-12-03
Add username of failed auth in dashboard

### LINKED ISSUE #2475: Usernames in failed authentications
In the "Authentication" box in the dashboard we can add the usernames to fhe *failed authentication*. 
This way a helpdesk user will immediately see, which user ran into an authentication failure, since the user probably might call the helpdesk...

In a big setup there could be a lot of users who failed to authenticate, which might clutter up the display.
So we might want to limit the view to the 5 (whichever number) last failed users.

But, If we have a *user search* or a *token seach* (see #2458), the admin might simply search the user and then find the user in the audit log in a 2nd step. Hm, but if the failed user is linked, the admin does not need to search the user but can simply click it. I think clicking is a good idea.

cornelinux:
The authentication widget contains the list of the
last 15 failed users.

Note: If a user failed to authenticate more than once he will appear
more than once in the list. We will see, if this helps to identify often
failing users or if we should convert the array to a set (which is not
easily possible in javascript).

Closes #2475

### DIFF
--- a/privacyidea/static/components/dashboard/controllers/dashboardControllers.js
+++ b/privacyidea/static/components/dashboard/controllers/dashboardControllers.js
@@ -120,6 +120,10 @@ myApp.controller("dashboardController", function (ConfigFactory, TokenFactory,
         AuditFactory.get({"timelimit": "1d", "action": "*validate*", "success": "0"},
             function (data) {
                 $scope.authentications.fail = data.result.value.count;
+                $scope.authentications.users = Array();
+                angular.forEach(data.result.value.auditdata, function(auditentry){
+                    $scope.authentications.users.push({"user": auditentry.user, "realm": auditentry.realm});
+                });
             });
      };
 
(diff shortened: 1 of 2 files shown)


privacyidea/static/components/dashboard/controllers/dashboardControllers.js
```diff
@@ -120,6 +120,10 @@ myApp.controller("dashboardController", function (ConfigFactory, TokenFactory,
         AuditFactory.get({"timelimit": "1d", "action": "*validate*", "success": "0"},
             function (data) {
                 $scope.authentications.fail = data.result.value.count;
+                $scope.authentications.users = Array();
+                angular.forEach(data.result.value.auditdata, function(auditentry){
```
laclaro, line 124: There are validate-check requests without user, but with a serial (e.g. token testing).

Those have no user or realm in the auditdata and are displayed with "@" in the dashboard. 

We could instead display and link the serial or exclude those entries. I think they are POST requests.

laclaro (commented):
Please check we want to handle requests without a user.

laclaro (changes requested):
forgot to request.

privacyidea/static/components/dashboard/controllers/dashboardControllers.js
cornelinux: done in c0547b7b

laclaro (approved):

PR #2542, opened 2020-12-04
1961/uis look the same

### LINKED ISSUE #1961: Why do all UIs look the same?
In a setup with redundant machines, internal machines, external machines, or test machines there might be a lot of privacyIDEA UIs.

"Why do all UIs look the same?"

We could simply add the name of the PI-Node in a prominent place in the UI - given, the admin could look at the browser URL line, but who does this? And this line does not scale in contrast to the web site.

# Todos

* [x] Improve documentation of PI_CSS
* [x] Add webui policy with action=show_privacyidea_node" (Note: the policy can have a node-condition!)
* [x] Display the privacyIDEA node in the UI using ``get_privacyidea_node`` from privacyidea.lib.config.py.

cornelinux:
Closes #1961

### DIFF
--- a/doc/faq/customization.rst
+++ b/doc/faq/customization.rst
@@ -92,6 +92,21 @@ The file in the file system is located at *privacyidea/static/contrib/css*.
 You might add a directory *privacyidea/static/custom/css/* and add your CSS
 file there.
 
+The CSS you specify here adds to the already existing styles. Thus a convenient way for
+using this setting is to help you distinguish different privacyIDEA instances like "testing", "acceptances"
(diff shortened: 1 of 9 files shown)


laclaro (approved):
Thank you for implementing this.

@cornelinux I would like the node name more left of the spinner as first element of the navbar-right. It is more prominent in that way and does not change our "corporate menu" :)

Your decision.

cornelinux:
@laclaro Where does it appear on your screen? At the spinner? Actually it appears directly right of the logo.

PR #2543, opened 2020-12-04
2401/change autocomplete on passwords

### LINKED ISSUE #2401: Use autocomplete="new-password" instead of "no"
Initiated by #2292

Since in most cases we do use password-fields not for login, but just to hide the input, the browser should not attempt to autofill. We use autocomplete="no" for this purpose.

However, autocomplete="no" is not honored by modern browsers, since they think storing passwords even without master-password protection is more secure than remembering them all.

https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

We therefore should check the UI code and replace occurrences of `autocomplete="no"` with `autocomplete="new-password"`, wherever useful. 
https://github.com/privacyidea/privacyidea/search?p=2&q=autocomplete

We should not touch the UI login here:
https://github.com/privacyidea/privacyidea/blob/e09465107bf554a99057b3ae6fa6595b6da3db3e/privacyidea/static/components/login/views/login.html#L72

timosturm:
This closes #2401. Most occurrences of `autocomplete="off"` are replaced by `autocomplete="new-password"` to prevent the browser from pre-filling them. `autocomplete="off"` was removed from server logins to prompt the browser to use saved passwords there explicitly.

### DIFF
--- a/privacyidea/static/components/dialogs/views/dialog.lock.html
+++ b/privacyidea/static/components/dialogs/views/dialog.lock.html
@@ -20,7 +20,6 @@
                         <label for="password" class="sr-only"
                                translate>Password</label>
                         <input name="password" type="password"
-                               autocomplete="off"
                                id="password" class="form-control"
                                ng-model="login.password" required
                                placeholder="{{ 'Password'|translate }}"/>
--- a/privacyidea/static/components/config/views/config.mresolvers.ldap.html
+++ b/privacyidea/static/components/config/views/config.mresolvers.ldap.html
@@ -105,7 +105,7 @@
         <div class="col-sm-3">
             <input name="bindpw" class="form-control" id="bindpw"
                    type="password"
-                   autocomplete="off"
+                   autocomplete="new-password"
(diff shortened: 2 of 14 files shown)


privacyidea/static/components/dialogs/views/dialog.lock.html
```diff
@@ -20,7 +20,6 @@ <h4 class="modal-title" id="myModalLock" translate>
                         <label for="password" class="sr-only"
                                translate>Password</label>
                         <input name="password" type="password"
-                               autocomplete="off"
```
timosturm, line 23: For this component I am not quite sure if the browser should pre-fill this, I guess it is a login from to the server?

plettich: This is the password dialog which blocks when the logout time has been reached. I think it is ok to enable autocomplete here.

cornelinux (approved):
lgtm

laclaro (approved):
Seems nice to me.

PR #2544, opened 2020-12-04
Allow to set the max failcount via event handler

### LINKED ISSUE #2541: Add "set maxfail" to the TokenHandler.
Sometimes it might come in handy if an event handler could set the MaxFailCounter of a token, since there is no mass-way to do this.

cornelinux:
The token handler now has an option to set the max failcount
of a token.

Closes #2541


tests/test_lib_events.py
```diff
@@ -2308,3 +2308,53 @@ def test_11_set_random_pin(self):
...
+
+        t_handler = TokenEventHandler()
+        res = t_handler.do(ACTION_TYPE.SET_MAXFAIL, options=options)
+        self.assertTrue(res)
+        # Check if the token has the correct sync window
+        t = get_tokens(serial="SPASS01")
+        fc = t[0].get_max_failcount()
+        self.assertEqual(fc, 123)
+
+        remove_token("SPASS01")
```
laclaro, line 2360: Please add newline.

laclaro (approved):
Works flawlessly. Please fix pep8 and merge.

tests/test_lib_events.py
cornelinux: Done

PR #2547, opened 2020-12-06
Make webauthn tokens work with triggerchallenge

### LINKED ISSUE #2538: triggerchallenge with webauthn token
when using validate/triggerchallenge for a user with a webauthn token, privacyidea returns:
``..."result": {"error": {"code": 905, "message": "ERR905: Missing parameter: 'webauthn_allowed_transports'"}, "status": false}...``

while using validate/check with the token pin returns the WebAuthnSignRequest properly.
privacyidea version is 3.4.1.

cornelinux:
The webauthn decorators pass information into the
request.all_data, which is then moved to the options-dict
to pass this information to the lower layers.

The triggerchallenge actually was missing the code to
move request params to the options dict.

Now all necessary information is passed to the lower layer.

Fixes #2538

### DIFF
--- a/privacyidea/api/validate.py
+++ b/privacyidea/api/validate.py
@@ -367,8 +367,8 @@ def check():
                "user": user}
     # Add all params to the options
     for key, value in request.all_data.items():
-            if value and key not in ["g", "clientip", "user"]:
-                options[key] = value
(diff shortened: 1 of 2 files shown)


plettich (approved):
looks good

PR #2551, opened 2020-12-09
Enable UI login with multiple FIDO tokens

### LINKED ISSUE #2092: Support for one user signing in to the webui with one of multiple Fido-tokens
Using U2F or WebAuthn to sign in to the webui seems to only work for a single token. If the user has multiple U2F tokens, or WebAuthn tokens, respectively, only whichever token happens to be the first one of its type in the multi-challenge will work. Need to investigate further, marking this as possible bug.

plettich:
If a user had multiple FIDO tokens enrolled (U2F, WebAuthn), a unique
challenge was created for each token during authentication. The client
on the other hand expected only one challenge to sign even for multiple
tokens. Since the client library took the challenge of the first
challenge-response token to sign, only the first token was able to
answer the challenge correctly. The second token always failed since it
signed the wrong challenge.
By using the transaction ID which is passed to each `create_challenge()`
call we can detect if a challenge for a WebAuthn or U2F token has been
created and reuse the actual challenge data.

Closes #2092

### DIFF
--- a/privacyidea/lib/tokens/webauthntoken.py
+++ b/privacyidea/lib/tokens/webauthntoken.py
@@ -32,7 +32,7 @@ from privacyidea.lib.config import get_from_config
 from privacyidea.lib.crypto import geturandom
 from privacyidea.lib.decorators import check_token_locked
 from privacyidea.lib.error import ParameterError, RegistrationError, PolicyError
-from privacyidea.lib.token import get_tokens, get_one_token
+from privacyidea.lib.token import get_tokens
 from privacyidea.lib.tokenclass import TokenClass
 from privacyidea.lib.tokens.u2f import x509name_to_string
 from privacyidea.lib.tokens.webauthn import (COSE_ALGORITHM, webauthn_b64_encode, WebAuthnRegistrationResponse,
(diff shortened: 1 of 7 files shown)


privacyidea/lib/tokens/webauthntoken.py
```diff
@@ -1069,13 +1069,30 @@ def create_challenge(self, transactionid=None, options=None):
...
+                    break
+
+        if not challenge:
+            nonce = self._get_nonce()
+            challenge = hexlify_and_unicode(nonce)
+        else:
+            nonce = binascii.unhexlify(challenge)
 
         # Create the challenge in the database
         challenge = Challenge(serial=self.token.serial,
```
cornelinux, line 1093: can you please rename this to "db_challenge", so that it looks the same like in u2ftoken?

Otherwise we have a "challenge", that is the hexlified nonce and here it gets the challenge object in the DB.

cornelinux (changes requested):
Only minor variable naming.

privacyidea/lib/tokens/webauthntoken.py
plettich: will do

plettich: Done in ecf651a6

cornelinux (approved):

Issue #2148, opened 2020-04-05
Organize Event Handler conditions


cornelinux:
The conditions of event handlers are badly organized in the UI.
We could do this in a similar way like the actions in policies.

We could group conditions or we could at least allow filtering conditions and only view enabled conditions.

cornelinux:
@plettich I am assigning this to you, so that you can pass this to/work with sebastian.

NNLemling:
I created a list how the individual conditions could be grouped. Maybe someone else has suggested changes here?

1. **general**: 
                        client_ip
                        detail_error_message
                        detail_message
                        last_auth
                        result_status
                        result_value

2. **token**: 
                        rollout_state
                        serial
                        token_locked
                        token_validity_period
                        tokeninfo
                        tokenrealm
                        tokenresolver
                        tokentype

3. **user**: 
                        logged_in_user
                        realm
                        resolver
                        token_has_owner
                        token_is_orphaned
                        user_token_number

4. **counter**: 
                         count_auth
                         count_auth_fail
                         count_auth_success
                         failcounter
                         otp_counter
                         counter. 

also we should add a second "Create Event Handler Definition" button in the upper part of the GUI just like we did with the policies to avoid that the user always has to scroll all the way down to create an event.

cornelinux:
Great, sounds good. Here a few comments:

* ``last_auth`` is token specific so I would put this to *token*.
*  I would also put ``token_is_orphaned`` and ``token_has_owner`` to *token*, since it is an aspect of a token

It is a cout idea to have an extra group "counter".

NNLemling:
thanks for the comment. I changed everything and updated the list!

PR #2553, opened 2020-12-10
add accordion to event handler conditions (#2148)

NNLemling:
"show selected conditions" and "create event handler" button added
the search function can now also be used for event handler conditions

### DIFF
--- a/privacyidea/static/components/config/views/config.events.details.html
+++ b/privacyidea/static/components/config/views/config.events.details.html
@@ -146,68 +146,67 @@
                     </div>
                 </div>
                 <div class="mt-5px">
-
-                <uib-accordion close-others="false">
-                    <div uib-accordion-group
-                         class="panel-default"
-                         is-open="isOpen"
-                         heading="{{ conditionGroup || 'miscellaneous'}}"
-                         ng-repeat="conditionGroup in conditionGroups">
-                        <table class="table table-hover">
-                            <tr ng-repeat="(conditionname, condition) in handlerConditions"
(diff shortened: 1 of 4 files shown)


privacyidea/static/components/config/views/config.events.details.html
```diff
@@ -107,54 +119,96 @@
...
+                                    <p class="help help-block"
+                                       ng-disabled="!conditionCheckBox[conditionname]">
+                                        {{ condition.desc }}
+                                    </p>
+                                </td>
+                            </tr>
+                        </table>
+                    </div>
+                </uib-accordion>
+                    </div>
```
cornelinux, line 210: Is this the closing ``</div>``  of line 148?
Please fix the indentation and move it to the left.

privacyidea/static/css/content.css
```diff
@@ -79,6 +79,10 @@ input[type="checkbox"] {
     margin-top: 5px;
 }
 
+.pl-0px {
+    padding-left: 0px;
```
cornelinux, line 83: Why do we need a new class here for padding this very widget?
Isn't there a generic rule?

cornelinux (changes requested):
Only two mintor styling comments.

privacyidea/static/components/config/views/config.events.details.html
cornelinux: Checked it and just moved the indentation myself in a5ff3f52

cornelinux (approved):

Issue #2557, opened 2020-12-11
layout adjustment of event handler conditions and policy actions


NNLemling:
the layouts of event handler conditions and policy actions were different.

The idea is to match both.

PR #2558, opened 2020-12-11
layout adjustment of event hanlder conditions and policy actions #2557

NNLemling:
slight change of the layout of the event handler conditions and simultaneous adjustment of the policy actions

### DIFF
--- a/privacyidea/static/components/config/views/config.events.details.html
+++ b/privacyidea/static/components/config/views/config.events.details.html
@@ -119,8 +119,9 @@
     <div class="tab-content">
         <div class="tab-pane active" id="condition">
             <h3 translate>Condition</h3>
-                <div>
-                    <label for="condition" class="col-sm-3 pl-0px">
+            <div class="row">
+                <div class="col-sm-3">
+                    <label for="condition">
                         <button class="btn btn-sm btn-success"
                                 ng-show="!onlySelectedVisible"
                                 ng-click="onlySelectedVisible=true">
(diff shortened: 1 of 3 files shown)


privacyidea/static/components/config/views/config.policies.details.html
```diff
@@ -270,9 +270,7 @@ <h3 translate>Condition</h3>
         <div class="tab-pane action" id="action">
             <h3 translate>Action</h3>
             <div class="form-group">
-                <label for="action" class="col-sm-3 control-label">
-                    <translate>Action</translate>
-                    <br>
+                <label for="action" class="col-sm-3">
```
cornelinux, line 273: THe label in the event handlers has the class ``pl-0px``.

privacyidea/static/components/config/views/config.events.details.html
```diff
@@ -133,7 +133,7 @@ <h3 translate>Condition</h3>
                         </button>
                     </label>
 
-                    <div class="input-group col-sm-9">
+                    <div class="input-group col-sm-12">
```
cornelinux, line 136: The columns are divided into 12 sections. So all ``col`` class should add up to 12.

col-sm (column small)...

So in line 123 you have col-sm-3. This should be accompanied by col-sm-9.
So why change this to col-sm-12. This looks wrong to me.

cornelinux (changes requested):

privacyidea/static/components/config/views/config.policies.details.html
NNLemling: yes for some reason the "show selected conditions only" button is positioned further to the right than it should. the class pl-0px solves this problem. Strangely enough, that's not the case with the policy actions.

privacyidea/static/components/config/views/config.events.details.html
NNLemling: Bildschirmfoto vom 2020-12-11 22-59-05

cornelinux (approved):

PR #2563, opened 2020-12-23
Fix the interface of the container audit module

### LINKED ISSUE #2562: Container audit fails in 3.5
The container audit will fail in version 3.5 due to a missing parameter in the constructor.

https://community.privacyidea.org/t/logging-error-after-update-to-3-5/1811/2

cornelinux:
The container audit module is missing a parameter,
which leads to an exception when called from the
upper layer.

Fixes #2562


laclaro (approved):
Thank you for this fix. I have no further comments.

PR #2566, opened 2021-01-03
Update Requirements to newest version

plettich:
For upcoming privacyIDEA 3.6

### DIFF
--- a/privacyidea/api/auth.py
+++ b/privacyidea/api/auth.py
@@ -365,7 +365,7 @@ def get_auth_token():
                         "authtype": authtype,
                         "exp": datetime.utcnow() + validity,
                         "rights": rights},
-                       secret, algorithm='HS256').decode('utf8')
+                       secret, algorithm='HS256')
 
     # Add the role to the response, so that the WebUI can make decisions
     # based on this (only show selfservice, not the admin part)
--- a/privacyidea/api/auth.py
+++ b/privacyidea/api/auth.py
@@ -69,7 +69,7 @@ from privacyidea.api.lib.prepolicy import (is_remote_user_allowed, prepolicy,
                                            webauthntoken_auth)
 from privacyidea.api.lib.utils import (send_result, get_all_params,
                                        verify_auth_token, getParam)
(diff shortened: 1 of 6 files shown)


plettich:
I think we should squash the commits before merging...

cornelinux:
> I think we should squash the commits before merging...

Oh oh. I read "I think we should smash the commits before merging..." and was a bit puzzled.

plettich:
Working on #2570

cornelinux:
@plettich  Is this still in Draft or ready for review?

privacyidea/api/auth.py
```diff
@@ -365,11 +365,11 @@ def get_auth_token():
                         "authtype": authtype,
                         "exp": datetime.utcnow() + validity,
                         "rights": rights},
-                       secret, algorithm='HS256').decode('utf8')
+                       secret, algorithm='HS256')
```
cornelinux, line 368: Shouldn't we move this hidden fix to an extra issue/PR?
Or is this due to a change in the requirements.

tests/test_lib_usercache.py
```diff
@@ -105,6 +105,8 @@ def test_01_get_username_from_cache(self):
         self.assertEqual(u_name, "")
 
     def test_02_get_resolvers(self):
+        # enable user cache
```
cornelinux, line 108: I think the changes in the test is also part of a fix?
I would also like to move this to another commit/PR.

cornelinux (changes requested):
I think 2 issues/PRs got mixed up here?

privacyidea/api/auth.py
plettich: The `jwt.encode()` returns a unicode string for Python > 3.5 so the `.decode()` fails. I have added the `to_unicode(token)` later to always return a unicode string.
This is due to an update in PyJWT.

tests/test_lib_usercache.py
plettich: This is due to an update of the SQLALchemy package. The `UserCache` entry will be removed after the `self._delete_realm()` call so we can't access the data anymore.

cornelinux (approved):

PR #2573, opened 2021-01-13
rename timer to counter in eventhandler accordion

laclaro:
this adds to #2148 changing the group "timer" to "counter"


plettich (approved):
looks good

PR #2575, opened 2021-01-13
search policy action description in WebUI

### LINKED ISSUE #2574: Search also policy action description texts
We may also search the action description texts.

We have to make sure that the descriptions are searched only after translation to avoid unexpected results.

laclaro:
closes #2574


plettich (approved):
looks good

PR #2576, opened 2021-01-13
Use cryptography for ECDSA operation

### LINKED ISSUE #2410: Replace ecdsa with pyca
In u2f we are using python-ecdsa which is not safe against side channel attacks.

We should replace it with pyca/cryptography curve secp256r1.

plettich:
The `ecdsa` package is not suitable for secure applications
(https://github.com/tlsfuzzer/python-ecdsa#security).
Even though this package is used only for verifying the U2F signature we
can utilize the `cryptography` package for ECDSA operations.
Since it already handles the DER encoding/decoding of the signatures, we
can also remove some wrapper code.

Fixes #2410

### DIFF
--- a/tests/test_api_u2f.py
+++ b/tests/test_api_u2f.py
@@ -1,14 +1,20 @@
+# -*- coding: utf-8 -*-
+
 from .base import MyApiTestCase
-import json
 import binascii
+import struct
+from hashlib import sha256
+from cryptography.hazmat.primitives.asymmetric import ec
+from cryptography.hazmat.primitives import hashes
+
 from privacyidea.lib.token import assign_token, remove_token
 from privacyidea.lib.user import User
 from privacyidea.lib.config import set_privacyidea_config
-from privacyidea.lib.tokens.u2f import (sign_challenge, check_response,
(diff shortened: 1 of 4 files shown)


tests/test_api_u2f.py
```diff
@@ -22,6 +28,44 @@
 OTPVALUES2 = ["551536", "703671", "316522", "413789"]
 
 
+def sign_challenge(user_priv_key, app_id, client_data, counter,
```
cornelinux, line 31: It would have been interesting to use the old sign_challenge for the tests to verify backward compatibility.
Since you changed sign and verify the backward compat is difficult to check.

However, I think we have replay tests in place?

Did you check with a real token?

cornelinux (commented):
lgtm, but see my remark

tests/test_api_u2f.py
plettich: WebAuthn and U2F token work, we also have some "original" signatures in the tests and they seem to work.
So i think this is Ok.

cornelinux (approved):

PR #2586, opened 2021-01-14
Remove border around tables in lists views

### LINKED ISSUE #2585: Remove table borders in list views
Some of the list views in the config section are missing the `<div class="table-responsive">` which leads to a border around each table.

plettich:
By adding the `table-responsive` class to the surrounding `div`, the
borders are removed.

Fixes #2585


laclaro (approved):
Thank's for finding this.

PR #2592, opened 2021-01-17
Allow to configure an alternative Web UI

### LINKED ISSUE #2591: prepare for an alternative or new UI
We can easily prepare the current privacyIDEA to be able to call a new UI.

We "simply" need to replave the index.html. (This is also mentioned in #605)

The [index.html](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/webui/login.py#L186) is the one and only initial html template, that pull all CSS and pulls the javascript framework and thus the complete SPA (single page application).

We can make the index.html configurable in ``pi.cfg`` in ``PI_INDEX_HTML`` and revert to the default one.
This way everyone can work on a new UI with the existing version 3.6 simply by modifying the ``pi.cfg``.

Later we could also provide several different UIs. They could be installed separately and chosen by configuring ``pi.cfg``.

cornelinux:
The pi.cfg can contain alternative index.html file
and alternative static/template path to be able
to easily load an alternative web UI without the
need to mangle with the existing one.

Closes #2591

### DIFF
--- a/doc/installation/system/inifile.rst
+++ b/doc/installation/system/inifile.rst
@@ -225,7 +225,7 @@ The Web UI is a single page application, that is initiated from the file
 ``static/templates/index.html``. This file pulls all CSS, the javascript framework
 and all the javascript business logic.
 
-You can configure privacyIDEA to use a complete different, your own Web UI.
+You can configure privacyIDEA to use a completely different, your own Web UI.
 
 You can do this using the following config values:
 
--- a/doc/installation/system/inifile.rst
+++ b/doc/installation/system/inifile.rst
@@ -216,3 +216,23 @@ in ``pi.cfg`` using the parameter ``PI_TOKEN_MODULES``:
 
     PI_TOKEN_MODULES = [ "myproject.cooltoken", "myproject.lametoken" ]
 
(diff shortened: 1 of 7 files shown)


doc/installation/system/inifile.rst
```diff
@@ -216,3 +216,23 @@ in ``pi.cfg`` using the parameter ``PI_TOKEN_MODULES``:
...
+.. _custom_web_ui:
+
+Custom Web UI
+-------------
+
+The Web UI is a single page application, that is initiated from the file
+``static/templates/index.html``. This file pulls all CSS, the javascript framework
+and all the javascript business logic.
+
+You can configure privacyIDEA to use a complete different, your own Web UI.
```
plettich, line 228: ... completely different Web UI.

plettich (changes requested):
Looks good, just minor spelling.

doc/installation/system/inifile.rst
cornelinux: I needed two steps to make it more clear:
358c881a and 784d97b3

plettich (approved):
Looks good

PR #2597, opened 2021-01-25
Replace deprecated oauth2client with google-auth

### LINKED ISSUE #1990: oauth2client is deprecated, use google-auth
The `oauth2client` package is deprecated.
According to the [deprecation doc](https://google-auth.readthedocs.io/en/latest/oauth2client-deprecation.html) we should switch to `google-auth`.

plettich:
Due to some changes in how the google-auth package is used, the mocks
for the tests needed some rewriting.

Fixes #1990


requirements.txt
```diff
@@ -5,6 +5,8 @@ backports.functools-lru-cache==1.6.1; python_version < '3.0'
 bcrypt==3.1.7; python_version <= '3.5'
 bcrypt==3.2.0; python_version > '3.5'
 beautifulsoup4==4.9.3
+cachetools==3.1.1; python_version < '3.0'
+cachetools==4.2.0; python_version > '3.0'
```
cornelinux, line 9: Are these indirectly needed by google-auth?

cornelinux (commented):
Just see my question.

requirements.txt
cornelinux: OK.thkx.

cornelinux (approved):

PR #2610, opened 2021-02-01
Only display subscription in dashboard

### LINKED ISSUE #2609: Only show subscription, if admin is allowed to see
If the admin does not have the right to manage subscription, the dashboard throws an error.

similar to #2456

cornelinux:
if the admin has the right to manage the subscription.
This way we suppress the error message.

Fix #2609


laclaro (approved):
Thank's

PR #2619, opened 2021-02-09
2569 token-janitor add attributes parameter

NNLemling:
add an attributes parameter to display user attributes.

Closes #2619

### DIFF
--- a/tools/privacyidea-token-janitor
+++ b/tools/privacyidea-token-janitor
@@ -111,7 +111,6 @@ Actions:
 
 """.format("|".join(ALLOWED_ACTIONS))
 
-
 app = create_app(config_name='production', silent=True)
 manager = Manager(app)
 
@@ -135,6 +134,7 @@ def _compare_greater_than(_key, given_value_string):
             return int(value) > given_value
         except ValueError:
             return False
+
     return comparator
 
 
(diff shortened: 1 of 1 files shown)


tools/privacyidea-token-janitor
```diff
@@ -462,7 +470,7 @@ def find(last_auth, assigned, active, tokeninfo_key, tokeninfo_value,
                 for token in tokens:
                     print(u",".join([u"'{0!s}'".format(x) for x in token]))
             else:
-                users = export_user_data(tlist)
+                users = export_user_data(tlist, attributes)
```
cornelinux, line 473: Could please also add the ``--attributes`` to the function ``export_token_data`` in line 469.
This way we can always add attributes, when we do ``--listuser``.

tools/privacyidea-token-janitor
```diff
@@ -400,6 +406,7 @@ def export_user_data(token_list):
                               'the output should only contain the number of tokens owned '
                               'by the user.',
                 dest='sum_tokens', action='store_true')
+@manager.option('--attributes', help='Extends the sum function to display additional user attributes.')
```
cornelinux, line 409: would be great to do

     Extends the "listuser" function to display additional user attributes

(see below.)

cornelinux (changes requested):
Looks good so far.
Would be great to have the attributes without sum, too.

cornelinux (approved):

Issue #2568, opened 2021-01-08
Allow SSL over SMTP and not only STARTTLS


olivier-duval:
Hello

I configured my SMTP server to disable STARTTLS and only allow SSL over SMTP (465).
This kind of configuration is not uncommon I think, however I couldn't find a way to implement it.

In the end, I had to patch the lib/smtpserver.py by replacing (line 107) : 

> mail = smtplib.**SMTP**(config['server'], port=int(config['port']),
>                             timeout=config.get('timeout', TIMEOUT))

with

> mail = smtplib.**SMTP_SSL**(config['server'], port=int(config['port']),
>                             timeout=config.get('timeout', TIMEOUT))

It's working fine for me right now but I'm wondering if it would be a good feature to have another checkbox with "SSL" near the "TLS" checkbox in the SMTP server configuration.

I'm not a python guy so I cant' propose a patch for the whole functionnaly

cornelinux:
To my understanding the configuration would either be TLS **or** SSL, right?
Could there be SMTPoverSSL with an additional StartTLS? (sounds weird)

@maxbeth this should be reflected in the configuration. Currently I do not know how. Maybe we would only do this in the UI using javascript.

olivier-duval:
Hello

Yes, it's EITHER TLS or SSL. Actually the difference between both is:
- STARTTLS is a special protocol, starting plaintext then upgrading the connection using TLS/SSL. 
- SSL is "only" the standard protocol but on a SSL connection instead of a cleartext one

In a way, STARTTLS leak more informatique because the beginning of the conversation is not encrypted.

So STARTTLS on SSL would have no meaning. That's why Python use 2 different methods

laclaro:
in the ldap resolver, we use the URI to decide about STARTTLS or LDAPS. Let's do the same here.

- smtps://myserver -> SSL
- smtp://myserver -> check STARTTLS or use plaintext
- myserver -> check STARTTLS or use plaintext

PR #2623, opened 2021-02-13
Working on #2568/smtp over ssl

maxbeth:
I can't create the db migrate script, because too many dependencies fail


cornelinux:
Hi @maxbeth 
thank you for looking into smtp-over-ssl.

1. Can you please refer to an issue number
2. There are commits, that do not belong here (lile confirm-action and maybe some translation). Can you please remove these or rebase to the latest master?

PR #2625, opened 2021-02-15
2257/fix reading hash algorithm

### LINKED ISSUE #2257: Hash algo during import of seed files
It looks like as if the import of pskc (at least from yubikey personalization tool) does not honor the keylength of the seed.

It should set the hash algo of HOTP or TOTP tokens according to the length of the key or a parameter in the PSKC file (if it exists).

I think the key lenght is honored in a CSV import.

The desired behaviour for "OATH CSV" and "PSCK" should be:

1. If a hashlib is specified in the import file, use this one.
2. Determine the hash algo based on the key length.
3. *completely ignore the token default setting*

This behaviour is explicitly for /token/import endpoint. So it is similar to #1631 but not directly related.

# Todo

* [x] Check and specify the behaviour
* [x] ~~implement ~~
* [x] ~~test~~

timosturm:
May close #2257.
Added that SHA1 is used as the default hash algorithm for imported tokens from PSKC to the documentation:
```
..note:: If the Hash algorithm is not specified for a token, SHA1 is used as the default.
```

Changed / fixed how the hash algorithm specified in the file is read.

### DIFF
--- a/privacyidea/lib/importotp.py
+++ b/privacyidea/lib/importotp.py
@@ -537,9 +537,10 @@ def parsePSKCdata(xml_data,
             # Check if hashlib is explicitly set in file
             if parameters.suite and parameters.suite.string:
                 hash_lib = parameters.suite.string.lower()
-            token["hashlib"] = hash_lib
         except Exception as e:
             log.warning(f"No compatible suite contained, falling back to default {hash_lib}.")
+
+        token["hashlib"] = hash_lib
 
         try:
             if key.data.secret.plainvalue:
--- a/privacyidea/lib/importotp.py
+++ b/privacyidea/lib/importotp.py
@@ -533,10 +533,13 @@ def parsePSKCdata(xml_data,
 
         hash_lib = "sha1"
 
-        # Check if hashlib is explicitly set in file
-        if parameters.suite and parameters.suite.string:
-            hash_lib = parameters.suite.string.lower()
-        token["hashlib"] = hash_lib
+        try:
+            # Check if hashlib is explicitly set in file
+            if parameters.suite and parameters.suite.string:
+                hash_lib = parameters.suite.string.lower()
+            token["hashlib"] = hash_lib
+        except Exception as e:
+            log.warning(f"No compatible suite contained, falling back to default {hash_lib}.")
 
         try:
             if key.data.secret.plainvalue:
--- a/privacyidea/lib/importotp.py
+++ b/privacyidea/lib/importotp.py
@@ -529,10 +529,15 @@ def parsePSKCdata(xml_data,
 
         parameters = key.algorithmparameters
         token["otplen"] = parameters.responseformat["length"] or 6
-        try:
-            token["hashlib"] = parameters.suite["hashalgo"] or "sha1"
-        except Exception as exx:
-            log.warning("No compatible suite contained.")
+        # token["hashlib"] = parameters.suite or "sha1"
+
+        hash_lib = "sha1"
+
+        # Check if hashlib is explicitly set in file
+        if parameters.suite and parameters.suite.string:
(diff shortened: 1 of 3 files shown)


privacyidea/lib/importotp.py
```diff
@@ -529,10 +529,15 @@ def parsePSKCdata(xml_data,
...
-            token["hashlib"] = parameters.suite["hashalgo"] or "sha1"
-        except Exception as exx:
-            log.warning("No compatible suite contained.")
+        # token["hashlib"] = parameters.suite or "sha1"
+
+        hash_lib = "sha1"
+
+        # Check if hashlib is explicitly set in file
+        if parameters.suite and parameters.suite.string:
+            hash_lib = parameters.suite.string.lower()
```
cornelinux, line 538: Is the name of the hash algorithm contained in ``parameters.suite["hashalgo"]`` or in ``paramters.suite.string``?
This looks like a change here. I was looking for quite some time myself, but did not find it.
Also, in regards to the problem, that other vendors are not quite precise in regards to pskc, I would wrap line 537 and 538 still in an try-except block.

cornelinux (changes requested):
Please check my comment about the change, you introduced.
Also consider wrapping it in a try.

privacyidea/lib/importotp.py
timosturm: I have described this change / why I changed this [here](https://github.com/privacyidea/privacyidea/issues/2257#issuecomment-777491824) in more detail. I.e., the only reference I found says that it is contained in the `string` property.

But I will wrap it just to be sure.

privacyidea/lib/importotp.py
```diff
@@ -533,10 +533,13 @@ def parsePSKCdata(xml_data,
...
-        if parameters.suite and parameters.suite.string:
-            hash_lib = parameters.suite.string.lower()
-        token["hashlib"] = hash_lib
+        try:
+            # Check if hashlib is explicitly set in file
+            if parameters.suite and parameters.suite.string:
+                hash_lib = parameters.suite.string.lower()
+            token["hashlib"] = hash_lib
+        except Exception as e:
+            log.warning(f"No compatible suite contained, falling back to default {hash_lib}.")
```
cornelinux, line 542: Just one last comment. If you exit the try block, then ``token["hashlib"]`` will not be set.
And it will default to the server's default (which usually is sha1 ;-)
You might consider, movint the line 540 out of the try-block to line 543 or into a finally (cool):

~~~~python
try:
    if parameters.suite and ...
        hash_lib = ...
except Exception as e:
    log.warning()
finally:
    token["hashlib"] = hash_lib
~~~~

cornelinux (changes requested):
Just one final thought

privacyidea/lib/importotp.py
```diff
@@ -537,10 +537,11 @@ def parsePSKCdata(xml_data,
             # Check if hashlib is explicitly set in file
             if parameters.suite and parameters.suite.string:
                 hash_lib = parameters.suite.string.lower()
-            token["hashlib"] = hash_lib
         except Exception as e:
             log.warning(f"No compatible suite contained, falling back to default {hash_lib}.")
```
cornelinux, line 541: I am so sorry. I only saw this now.

The f-format will not work.

Please use the u-format 

   u"....some test {0!s}".format(hash_lib)

cornelinux (changes requested):

privacyidea/lib/importotp.py
cornelinux: Sounds good to me. Remove the exception-catching and move the log.warning to an else-block.

cornelinux (approved):

PR #2626, opened 2021-02-16
numbers per tokentype in the dashboard

### LINKED ISSUE #2457: Display tokentypes and make this a clickable link
The dashboard could offer links to further specific tokenkinds or tokentypes.
Currently it displays the number of hardware tokens and software tokens.
It would be great if these would be a clickable link. However, it looks like that the token view is difficult to filter for a tokeninfo, so this might not happen.

But we could also display the available tokentypes like SMS, HOTP, TOTP.

   SMS       120 Tokens
   HOTP       55 Tokens
   TOTP        12 Tokens

Often software tokens might be TOTP tokens and hardware tokens HOTP tokens, so there is the differentiation anyways.
Tokentypes with a number of "0", could simply be hidden.
Since most installations only have two or three token types, this might come in handy.

laclaro:
closes #2457

- [x] add counts parameter to `GET /token` to return counts by tokentype
- [x] Add API tests
- [x] Add dashboard queries
- [ ] filter token list via state


laclaro:
too much committed...

privacyidea/lib/token.py
```diff
@@ -538,9 +541,24 @@ def get_tokens_paginate(tokentype=None, realm=None, assigned=None, user=None,
...
 
 
+def get_token_type_counts(token_sql_query_obj):
+    """
+    This function extracts the numbers of tokens per tokentype from the given sql query object
+    """
+    token_list_full = token_sql_query_obj.all()
+    tokentypes_in_list = set([token.tokentype for token in token_list_full])
+    counts = dict.fromkeys(tokentypes_in_list, "")
+    for tokentype in tokentypes_in_list:
```
cornelinux, line 556: Do not do it this way.
You do not want to fetch 50.000 tokens from the database and iterate through them!
This will be getting slow.

This way you are issuing a second sql select statement on all tokens.
This first statement will only fetch the pagination, but this will fetch all tokens.

cornelinux (changes requested):
Do not fetch all tokens.

privacyidea/lib/token.py
laclaro: Sadly, you are right. Let's take a step back an think again.

cornelinux:
Close this, since we will rebuild it

PR #2628, opened 2021-02-16
Make entries available in audit log

### LINKED ISSUE #2627: Make signature and missing line entries visible in audit log
The `Signature` and `Missing Line` fields in the audit log were removed in #2532 since they weren't included in the newly introduced `available_audit_columns` property.

plettich:
The `missing line` and `signature` entries were missing from the audit
log. Adding them to the `available_audit_columns` property fixes this.
Also enhanced tests for audit columns which revealed some minor issues.

Fixes #2627


cornelinux:
@plettich I think this misses also in version 3.5.1.
So I think we should  add this to the branch 3.5 for a 3.5.2

plettich:
PR into branch-3.5 done in #2630

PR #2635, opened 2021-02-19
1059/add limited cache to auth cache

### LINKED ISSUE #1059: Allow auth_cache to cache only for a limited number of authentications
The ``auth_cache`` is policy in scope=auth.
http://privacyidea.readthedocs.io/en/latest/policies/authentication.html#auth-cache

Currently you can define it like 4h/4m, which would mean cache for 4 hours if every 4 minutes there is a request. 

We need to enhance this login to "cache for 5 seconds, but only allow 2 authentications within this 5 seconds".

This could be specified like "5s/2".

To achieve this, we need to

1. allow seconds in the auth_cache
2. count the authentications performed by the auth_cache. Therefore we need to add a column for counting the auth's https://github.com/privacyidea/privacyidea/blob/master/privacyidea/models.py#L2431

timosturm:
Closes #1059.

Limited number of attempts for authentication that are handled by the cache can be specified by `<timeframe>/<#attempts>`, e.g., `10s/2` allows two cached authentications.
This means, after the user authenticated themself with a password, the user can use the same password two additional times for 10s after the initial authentication.

### DIFF
--- a/privacyidea/lib/authcache.py
+++ b/privacyidea/lib/authcache.py
@@ -96,7 +96,7 @@ def cleanup(minutes):
 
 
 def verify_in_cache(username, realm, resolver, password, first_auth=None, last_auth=None,
-                    max_number_of_authentications=0):
+                    max_auths=0):
     """
     Verify if the given credentials are cached and if the time is correct.
     
@@ -108,10 +108,10 @@ def verify_in_cache(username, realm, resolver, password, first_auth=None, last_a
         cache. Only find newer entries 
     :param last_auth: The timestamp when the entry was last successfully 
         verified. Only find newer entries
-    :param max_number_of_authentications: Maximum number of times the authcache entry can be used to skip
+    :param max_auths: Maximum number of times the authcache entry can be used to skip
         authentication, as defined by ACTION.AUTH_CACHE policy. Will return False if the current number of
         authentications + 1 of the cached authentication exceeds this value.
-    :type max_number_of_authentications: int
+    :type max_auths: int
     :return: 
     """
     conditions = []
@@ -135,9 +135,9 @@ def verify_in_cache(username, realm, resolver, password, first_auth=None, last_a
             log.debug("Old authcache entry for user {0!s}@{1!s}.".format(username, realm))
             result = False
 
-        if result and max_number_of_authentications > 0:
+        if result and max_auths > 0:
             # Check if auth_count allows this authentication too
-            result = cached_auth.auth_count < max_number_of_authentications
+            result = cached_auth.auth_count < max_auths
 
         if result:
             # Update the last_auth
--- a/privacyidea/lib/authcache.py
+++ b/privacyidea/lib/authcache.py
@@ -108,9 +108,9 @@ def verify_in_cache(username, realm, resolver, password, first_auth=None, last_a
         cache. Only find newer entries 
     :param last_auth: The timestamp when the entry was last successfully 
         verified. Only find newer entries
-    :param max_number_of_authentications: Maximum number of times the authcache entry can be used to skip authentication,
-        as defined by ACTION.AUTH_CACHE policy. Will return False if the current number of authentications + 1 of the
-        cached authentication exceeds this value.
+    :param max_number_of_authentications: Maximum number of times the authcache entry can be used to skip
+        authentication, as defined by ACTION.AUTH_CACHE policy. Will return False if the current number of
+        authentications + 1 of the cached authentication exceeds this value.
     :return: 
     """
     conditions = []
--- a/privacyidea/lib/authcache.py
+++ b/privacyidea/lib/authcache.py
@@ -45,6 +45,12 @@ def add_to_cache(username, realm, resolver, password):
     return r
 
 
+def increment_auth_count(cache_id):
+    db.session.query(AuthCache).filter(AuthCache.id == cache_id).update(
+        {AuthCache.current_number_of_authentications: AuthCache.current_number_of_authentications + 1})
+    db.session.commit()
+
+
 def update_cache_last_auth(cache_id):
     last_auth = datetime.datetime.utcnow()
     AuthCache.query.filter(
@@ -89,7 +95,8 @@ def cleanup(minutes):
     return r
 
 
-def verify_in_cache(username, realm, resolver, password, first_auth=None, last_auth=None):
+def verify_in_cache(username, realm, resolver, password, first_auth=None, last_auth=None,
+                    max_number_of_authentications=None):
     """
     Verify if the given credentials are cached and if the time is correct.
     
@@ -100,7 +107,10 @@ def verify_in_cache(username, realm, resolver, password, first_auth=None, last_a
     :param first_auth: The timestamp when the entry was first written to the 
         cache. Only find newer entries 
     :param last_auth: The timestamp when the entry was last successfully 
-        verified. Only find newer entries 
+        verified. Only find newer entries
+    :param max_number_of_authentications: Maximum number of times the authcache entry can be used to skip authentication,
+        as defined by ACTION.AUTH_CACHE policy. Will return False if the current number of authentications + 1 of the
+        cached authentication exceeds this value.
     :return: 
     """
     conditions = []
@@ -123,6 +133,12 @@ def verify_in_cache(username, realm, resolver, password, first_auth=None, last_a
         except ValueError:
             log.debug("Old authcache entry for user {0!s}@{1!s}.".format(username, realm))
             result = False
+
+        if result and max_number_of_authentications:
+            result = cached_auth.current_number_of_authentications < max_number_of_authentications
+            increment_auth_count(cached_auth.id)
+            break
+
         if result:
             # Update the last_auth
             update_cache_last_auth(cached_auth.id)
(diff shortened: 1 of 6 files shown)


privacyidea/lib/policydecorators.py
```diff
@@ -156,28 +156,41 @@ def auth_cache(wrapped_function, user_object, passw, options=None):
...
+
             # verify in cache and return an early success
             auth_times = list(auth_cache_dict)[0].split("/")
             # determine first_auth from policy!
             first_offset = parse_timedelta(auth_times[0])
 
-            if len(auth_times) == 2:
-                # Determine last_auth from policy
-                last_offset = parse_timedelta(auth_times[1])
+            import re
```
plettich, line 170: The `import` should be moved to the beginning of the file

privacyidea/lib/authcache.py
```diff
@@ -89,7 +95,8 @@ def cleanup(minutes):
     return r
 
 
-def verify_in_cache(username, realm, resolver, password, first_auth=None, last_auth=None):
+def verify_in_cache(username, realm, resolver, password, first_auth=None, last_auth=None,
+                    max_number_of_authentications=None):
```
plettich, line 99: since this is basically an integer value, the default should be `0`

privacyidea/lib/authcache.py
```diff
@@ -100,7 +107,10 @@ def verify_in_cache(username, realm, resolver, password, first_auth=None, last_a
     :param first_auth: The timestamp when the entry was first written to the 
         cache. Only find newer entries 
     :param last_auth: The timestamp when the entry was last successfully 
-        verified. Only find newer entries 
+        verified. Only find newer entries
+    :param max_number_of_authentications: Maximum number of times the authcache entry can be used to skip
+        authentication, as defined by ACTION.AUTH_CACHE policy. Will return False if the current number of
+        authentications + 1 of the cached authentication exceeds this value.
```
plettich, line 113: Could You add
```
:type max_number_of_authentications: int
```
here to give IDEs a type hint?

privacyidea/lib/policydecorators.py
```diff
@@ -156,28 +156,41 @@ def auth_cache(wrapped_function, user_object, passw, options=None):
...
+            # Use auth cache when number of allowed authentications is defined
+            if len(auth_times) == 2 and re.match(r"^\d+$", auth_times[1]):
```
plettich, line 172: Since we have the same test below, this block should be moved there. Makes the flow a little clearer.

tests/test_lib_policydecorator.py
```diff
@@ -619,15 +619,67 @@ def fake_check_user_pass(user, passw, options=None):
                   first_auth=datetime.datetime.utcnow() - timedelta(hours=2),
                   last_auth=datetime.datetime.utcnow() - timedelta(
                       hours=1)).save()
-        r = auth_cache(fake_check_user_pass, User("cornelius", "myrealm"),
+        r = auth_cache(fake_check_user_pass, User(username, realm),
```
plettich, line 622: please also add  a simple test of `verify_in_cache` in `test_lib_authcache.py`

plettich (changes requested):

cornelinux:
@timosturm Please also enhance the documentation here https://github.com/privacyidea/privacyidea/blob/master/doc/policies/authentication.rst#auth_cache

privacyidea/lib/authcache.py
timosturm: Should values <= 0 be ignored then, or should an exception be thrown? See https://github.com/privacyidea/privacyidea/pull/2635/commits/6a9199e1db3b4f931e7c070f2ae0c305ae782bb7

privacyidea/lib/policydecorators.py
timosturm: Honestly, I am not sure how that could be done. This is the only thing I could come up with:
```
last_offset = None
max_number_of_authentications = None
if len(auth_times) == 2:
    if re.match(r"^\d+$", auth_times[1]):
        # Check if second entry defines max number of authentications
        max_number_of_authentications = int(auth_times[1])
    else:
        last_offset = parse_timedelta(auth_times[1])
else:
    # No second value is defined in auth_times
    last_offset = first_offset

last_auth = None
if last_offset:
    last_auth = datetime.datetime.utcnow() - last_offset

result = verify_in_cache(user_object.login, user_object.realm,
                         user_object.resolver, passw,
                         first_auth=first_auth,
                         last_auth=last_auth,
                         max_number_of_authentications=max_number_of_authentications)
```
and it is not good to read.

privacyidea/lib/authcache.py
plettich: negative values don't make sense here so if someone passes a negative value for the number of auths the cache should not be used (in my opinion). So the test if the value is greater than 0 is enough.

privacyidea/lib/policydecorators.py
```diff
@@ -156,28 +157,39 @@ def auth_cache(wrapped_function, user_object, passw, options=None):
     options = options or {}
     g = options.get("g")
     auth_cache_dict = None
+
     if g:
         auth_cache_dict = Match.user(g, scope=SCOPE.AUTH, action=ACTION.AUTH_CACHE,
                                      user_object=user_object).action_values(unique=True, write_to_audit_log=False)
         if auth_cache_dict:
+
```
timosturm, line 165:
```suggestion
```

timosturm (commented):
Removed the only new empty line I found

privacyidea/lib/policydecorators.py
plettich: You could get rid of the last `if:` when You calculate `last_auth` directly:
```
            # Use auth cache when number of allowed authentications is defined
            if len(auth_times) == 2:
                if re.match(r"^\d+$", auth_times[1]):
                    max_auths = int(auth_times[1])
                else:
                    # Determine last_auth delta from policy
                    last_offset = parse_timedelta(auth_times[1])
                    last_auth = datetime.datetime.utcnow() - last_offset
            else:
                # If there is no last_auth, it is equal to first_auth
                last_auth = datetime.datetime.utcnow() - first_offset

            result = verify_in_cache(...)
```

plettich (changes requested):
Some issues just came up, please have a look.
Also please let me mark the comments as resolved otherwise i do not know what i wrote last week ;-)

privacyidea/lib/policydecorators.py
```diff
@@ -156,28 +157,37 @@ def auth_cache(wrapped_function, user_object, passw, options=None):
...
+            first_auth = datetime.datetime.utcnow() - first_offset
+            last_auth = first_auth  # Default if no last auth exists
+            max_auths = 0  # Default value, 0 has no effect on verification
 
+            # Use auth cache when number of allowed authentications is defined
             if len(auth_times) == 2:
-                # Determine last_auth from policy
-                last_offset = parse_timedelta(auth_times[1])
+                if re.match(r"^\d+$", auth_times[1]):
+                    max_auths = int(auth_times[1])
```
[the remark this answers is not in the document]
timosturm:
```suggestion
                    max_auths = int(auth_times[1])
                    last_auth = None
```

Like this? I am not sure about what difference it actually makes here.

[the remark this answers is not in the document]
plettich: Now this is a tricky one :-)
Actually there are several things going on here:
- You used `models.AuthCache.__init__()` to create entries in the authcache table for the tests, unfortunately this doesn't create a `last_auth` entry (`None` by default), so basically this is an incomplete entry (You need to have a `last_auth` entry, otherwise it shouldn't exist in the authcache table).
- When rewriting this functionality we set `last_auth = first_auth` as a default, this was passed to `verify_in_cache()` which in turn created a db-query from that. But since the `last_auth` entry wasn't set in the db, the query returned nothing and the default authentication took place (this lead to a failed test in `test_lib_policydecorator.py:647`).

1. Without a valid `last_auth` in the authcache table the entry should be treated as invalid.
2. We should be able to pass the `last_auth` in any case to `verify_in_cache()`, if `last_auth = first_auth` this should yield all matching entries as if `last_auth=None` was passed.
3. Please delete the `last_auth = None` here and add a `last_auth` parameter in `test_lib_policydecorator.py:641`

plettich (changes requested):

privacyidea/lib/policydecorators.py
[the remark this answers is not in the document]
timosturm: This should be fixed now too.

privacyidea/models.py
```diff
@@ -2699,8 +2699,16 @@ def __init__(self, username, realm, resolver, authentication,
...
+
+        if first_auth:
+            self.first_auth = first_auth
+        else:
+            self.first_auth = datetime.utcnow()
+
+        if last_auth:
+            self.last_auth = last_auth
+        else:
+            self.last_auth = first_auth
```
plettich, line 2711: You could write this as
```
self.first_auth = first_auth if first_auth else datetime.utcnow()
self.last_auth = last_auth if last_auth else self.first_auth
```
This is more [*Pythonic*](https://blog.startifact.com/posts/older/what-is-pythonic.html)

plettich (approved):
looks good

PR #2645, opened 2021-02-26
Removed aladdin_hashlib and default_hashlib parameters

### LINKED ISSUE #2634: Remove undocumented `aladdin_hashlib` parameter from `token/load/` endpoint
The `token/load/` endpoint currently also accepts a parameter `aladdin_hashlib` which overwrites any given hash algorithm from the parsed data.
We should remove the parameter (which is not documented anyway) and re-evaluate the usage of the `default_hashlib` parameter in [`import_token`](https://github.com/privacyidea/privacyidea/blob/6e2ed66ea6ae01dec0068f705502f5d9786c8cb2/privacyidea/lib/token.py#L943).
This is somehow related to #2257

timosturm:
Closes #2634.

Parameter | Method | Reason for removal
-|-|-
`aladdin_hashlib` | `loadtokens_api` | Undocumented and overrides hash lib specified by the imported file
`default_hashlib` | `import_token` | Has misleading docstring and name, was used to override existing hash lib of token; default hash lib is defined elsewhere in code

### DIFF
--- a/privacyidea/lib/token.py
+++ b/privacyidea/lib/token.py
@@ -1081,8 +1081,7 @@ def init_token(param, user=None, tokenrealms=None,
     if user is not None and user.login != "":
         tokenobject.add_user(user)
 
-    upd_params = param
-    tokenobject.update(upd_params)
+    tokenobject.update(param)
 
     try:
         # Save the token to the database
--- a/privacyidea/lib/token.py
+++ b/privacyidea/lib/token.py
@@ -940,7 +940,7 @@ def gen_serial(tokentype=None, prefix=None):
 
 
 @log_with(log)
-def import_token(serial, token_dict, default_hashlib=None, tokenrealms=None):
+def import_token(serial, token_dict, tokenrealms=None):
     """
     This function is used during the import of a PSKC file.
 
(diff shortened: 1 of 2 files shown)


privacyidea/lib/token.py
```diff
@@ -1081,8 +1076,7 @@ def init_token(param, user=None, tokenrealms=None,
     if user is not None and user.login != "":
         tokenobject.add_user(user)
 
-    upd_params = param
-    tokenobject.update(upd_params)
+    tokenobject.update(param)
```
cornelinux, line 1079: Im am not sure anymore, if there is a reason, why the dict was copied - if it was necessary for a deep copy since it gets modified somewhere.

cornelinux (commented):
Basically looks good to me.
Please check my comment about the removal of the param dict.

privacyidea/lib/token.py
timosturm: As far as I can tell, this does not make any difference. Because `upd_params = param` does not copy the dict, but only assigns a second name for reference. And python makes calls with 'pass by object-reference', so if `tokenobject.update()` changes `param` or `upd_params` would not make a difference because both reference the same dictionary.

plettich (approved):
looks good

PR #2647, opened 2021-03-02
make backup create ready for tar v1.30

### LINKED ISSUE #2646: pi-manage backup create is not compatible with tar v1.30
The argument order used for `tar` in

```
pi-manage backup create
```

is not compatible with tar v1.30 as used in Ubuntu 20.04. The `--exclude enckey` argument has no effect when used at the end.

The warning is

```
tar: The following options were used after any non-optional arguments in archive create or update mode.  These options are positional and affect only arguments that follow them.  Please, rearrange them properly.
tar: --exclude ‘/etc/privacyidea/enckey’ has no effect
tar: Exiting with failure stat
tar: Exiting with failure status due to previous errors
```

The backup is created, but enckey is always present in the created archive.

laclaro:
closes #2646.

This PR is against branch-3.5 since this bugfix should arrive as quick as possible.

Tested on Ubuntu 18.04 (tar 1.29) and 20.04 (tar 1.30)

### DIFF
--- a/pi-manage
+++ b/pi-manage
@@ -316,7 +316,8 @@ def create(directory="/var/lib/privacyidea/backup/",
 
     if not enckey:
         # Exclude enckey from backup
-        backup_call.append("--exclude={0!s}".format(enc_file))
+        # since tar v1.30 --exclude cannot be appended
+        backup_call.insert(1, "--exclude={0!s}".format(enc_file))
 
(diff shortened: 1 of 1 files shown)


pablo-knight (approved):
changes to pi-manage also working with ubuntu16
:)

PR #2654, opened 2021-03-03
Add some docs about import and Hash algo

### LINKED ISSUE #2257: Hash algo during import of seed files

cornelinux:
Describe which import uses which hash algorithm, or
how the hash algorithm during importing tokens seeds is
determined.

Closes #2257

### DIFF
--- a/doc/workflows_and_tools/import/index.rst
+++ b/doc/workflows_and_tools/import/index.rst
@@ -90,6 +90,9 @@ For TAN tokens it looks like this::
 
 The list of tans is a whitespace separated list.
 
+.. note:: The Hash algorithm (SHA1, SHA256, SHA512) is derived from the length of the **seed**.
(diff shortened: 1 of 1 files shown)


laclaro (approved):
Looks good to me, thank's

PR #2656, opened 2021-03-04
Fixed code quality issues

withshubh:
#### Description
Hi :wave: 
I ran the [DeepSource](https://deepsource.io) analysis on the forked copy of this repo and found some interesting [code quality issues](https://deepsource.io/gh/withshubh/privacyidea/issues/?category=recommended).

#### Summary of changes
- ~Remove length check in favour of truthiness of the object~
- ~Remove unnecessary use of comprehension~
- ~Remove unnecessary comprehension~
- Use identity check for comparison with True
- Added .deepsource.toml

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,19 @@
+version = 1
+
+test_patterns = ["tests/test_*.py"]
+
+exclude_patterns = [
+    "authmodules/**",
+    "contrib/**",
+    "deploy/**",
+    "migrations/**",
+    "po/**",
+    "tools/**"
+]
+
+[[analyzers]]
+name = "python"
+enabled = true
+
+  [analyzers.meta]
+  runtime_version = "3.x.x"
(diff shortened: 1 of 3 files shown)


cornelinux:
I confirm with the checks "is True".
But I would not want to merge other changes due to

1. readability (e.g. comprehension)
2. and probably python version compatibilities

So if you want to rework your changes to only incorporate "is True" checks, that would be fine.

withshubh:
Removed the other fixes! @cornelinux

cornelinux:
I am sorry, that you relied on your code checking tool. I think it produced a false positive, which we know so well from the plandemic.

The `filter` takes a `criterion`, which obivously does not work as a python expression. So in this case you have to use a "==" and not an "is".

That is the problem with the automatic code checking tools.

Thanks anyways for your efforts.

withshubh:
No problem! @cornelinux

PR #2660, opened 2021-03-07
Make the tokenwizard only visible in case of 2nd

### LINKED ISSUE #2632: Enrollment wizard menu entry is shown even if user already has a token assigned
If the webui policy `tokenwizard_2nd_token` is unset (users should not be able to use the token wizard if they already have a token enrolled), the `token enrollment wizard` menu entry is still shown in the UI.
The `token_wizard_2nd` variable is send with an empty list from the server but the UI checks for non-existence here:
https://github.com/privacyidea/privacyidea/blob/cd311f443352a049833557f2a87f17a939862092/privacyidea/static/components/token/views/token.html#L30

cornelinux:
The token wizard in the menu should only be visible
if the policy token_wizard_2nd is set.

Fixes #2632


plettich (approved):
looks good

PR #2666, opened 2021-03-11
Allow changing POOLING_LOOP_TIMEOUT of ldap3

### LINKED ISSUE #2662: reduce the time ldap3 waits between each round
The ldap3 pooling has a ``POOLING_LOOP_TIMEOUT``. (see 
https://ldap3.readthedocs.io/en/latest/server.html?highlight=serverpool#server-pool)

This is the time ldap3 waits between each round. But it looks like, as if this time is also waited, even if there is only one round.
The POOLING_LOOP_TIMEOUT defaults to 10.
We should either reduce this time or make it configurable.

An easy approach is to set it directly in the LDAPResolver.py:

~~~~python
import ldap3
from ldap3 import MODIFY_REPLACE, MODIFY_ADD, MODIFY_DELETE
from ldap3 import Tls
from ldap3.core.exceptions import LDAPOperationResult
from ldap3.core.results import RESULT_SIZE_LIMIT_EXCEEDED
import ssl

ldap3.set_config_parameter("POOLING_LOOP_TIMEOUT", 0)
~~~~

cornelinux:
The ldap3 config setting POOLING_LOOP_TIMEOUT is a system wide
setting, so it does not make sense to set this per LDAP resolver.
Thus we add PI_LDAP_POOLING_LOOP_TIMEOUT to pi.cfg, which is
the first LDAP resolver config in pi.cfg.

This commit does not change the default setting of currently
10 seconds of timeout between rounds. However, since the timeout
is also added if only one round is done, it might make sense
to set this to "0" in pi.cfg for performance reasons.

Closes #2662

### DIFF
--- a/doc/configuration/useridresolvers.rst
+++ b/doc/configuration/useridresolvers.rst
@@ -186,7 +186,7 @@ from the pool and will be re-inserted after the number of seconds specified in t
 If the pool is empty after a round, a timeout is added before the next round is started.
 The ldap3 module defaults system wide to 10 seconds before starting the next round.
 This timeout can be changed by setting ``PI_LDAP_POOLING_LOOP_TIMEOUT`` to an integer in seconds in ``pi.cfg``.
-If no reachable server has could been found after the number of rounds specified in the *retry rounds*,
+If no reachable server could be found after the number of rounds specified in the *retry rounds*,
 the request fails.
 
 By default, knowledge about unavailable pool servers is not persisted between requests.
--- a/doc/configuration/useridresolvers.rst
+++ b/doc/configuration/useridresolvers.rst
@@ -183,8 +183,10 @@ The *Server pool retry rounds* and *Server pool skip timeout* settings configure
 the LDAP server pool. When establishing a LDAP connection, the resolver uses a round-robin
 strategy to select a LDAP server from the pool. If the current server is not reachable, it is removed
 from the pool and will be re-inserted after the number of seconds specified in the *skip timeout*.
-If no server from the pool is reachable, the servers are queried again from the beginning. If
-a reachable server has not been found after the number of rounds specified in the *retry rounds*,
+If the pool is empty after a round, a timeout is added before the next round is started.
+The ldap3 module defaults system wide to 10 seconds before starting the next round.
+This timeout can be changed by setting ``PI_LDAP_POOLING_LOOP_TIMEOUT`` to an integer in seconds in ``pi.cfg``.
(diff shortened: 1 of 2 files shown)


doc/configuration/useridresolvers.rst
```diff
@@ -183,8 +183,10 @@ The *Server pool retry rounds* and *Server pool skip timeout* settings configure
 the LDAP server pool. When establishing a LDAP connection, the resolver uses a round-robin
 strategy to select a LDAP server from the pool. If the current server is not reachable, it is removed
 from the pool and will be re-inserted after the number of seconds specified in the *skip timeout*.
-If no server from the pool is reachable, the servers are queried again from the beginning. If
-a reachable server has not been found after the number of rounds specified in the *retry rounds*,
+If the pool is empty after a round, a timeout is added before the next round is started.
+The ldap3 module defaults system wide to 10 seconds before starting the next round.
+This timeout can be changed by setting ``PI_LDAP_POOLING_LOOP_TIMEOUT`` to an integer in seconds in ``pi.cfg``.
+If no reachable server has could been found after the number of rounds specified in the *retry rounds*,
```
plettich, line 189: If no reachable server could **be** found...

privacyidea/lib/resolvers/LDAPIdResolver.py
```diff
@@ -97,6 +97,12 @@
 SERVERPOOL_ROUNDS = 2
 # The number of seconds a non-responding server is removed from the server pool
 SERVERPOOL_SKIP = 30
+# The number of seconds that ldap3 waits if no server is left in the pool, before
+# starting the next round
+pooling_loop_timeout = get_app_config_value("PI_LDAP_POOLING_LOOP_TIMEOUT", 10)
+log.info("Setting system wide POOLING_LOOP_TIMEOUT to {0!s}.".format(pooling_loop_timeout))
```
plettich, line 103: Do we get this Info on each request or only when the server starts?

plettich (changes requested):
looks good, just some spelling

privacyidea/lib/resolvers/LDAPIdResolver.py
cornelinux: imho this happens during loading of this python module - thus only once at starttime.

doc/configuration/useridresolvers.rst
cornelinux: done in 26cb6fad

plettich (approved):
looks good

PR #2671, opened 2021-03-12
Add token condition for policies

### LINKED ISSUE #2590: Add tokentype to policy condition
In many cases we have several policy actions for different tokentypes. This is somehow annoying since it clutters the actions.
At other places we already discussed tokentypes in policies.

The problem is, that in certain cases policies do not refer to actions, that are about tokens. So this would not be applicable.

I think it is too complicated to link policy actions and allowed conditions. So we need to decide and define, how a tokentype condition should be handled, if it is not applicable. There are two possibilities to handle tokentypes at all:

* via extended attributes (while a lot of functionalities in the extended attributes do not make sense)
  On the other hand we could use the same logic like "take care that you define policies that make sense, otherwise an exceptoin gets raised". We could have an extended condition "token".  No database migration needed.
   We might allow all DB columns generically to be checked like "serial", "tokentype", "maxfail", "revoked", "locked" and even "rollout_state"...
* via a new specific condition. However, this very condition would be superfluous in 90% of the policies. We would need a database migration.

We already discussed this in #1947.
This will solve, what we wanted to do in #2142. 
We could also have scope dependent conditions in #2510!

# Condition Matching / Determining tokentype

So what should happen, if a tokentype can not be determined?
In the following cases:

* authentication, calling validate/check. Usually a token and a token type are available, but could be not...
   *Expected behaviour*: 
   a) I think the policy should not match. We *can* define a policy  without a token type condition, so this would be the default policy and the on with the tokentype definition would be the exlcusion, only if the specific tokentype does exist.
   b) OR An exception could be raised.
* defining a policy. Most probably not a tokentype contained, but could we define a policy for a tokentype?
      *Expected behaviour*:
   a) An exception should be raised.
* passthru (no token, no tokentype: would not make sense at all)
      *Expected behaviou*: 
   a) I think the policy should not match. We *can* define a policy without a token type definition, so this would be ok as a fallback. 
   b) An exception could be raised.

**I am thinking using a additional condition "token" provides the highest flexibility and does not break the current design.**

cornelinux:
The token condition can query any token DB column and compare it.
We also added the comparison of integer values to be bigger
or smaller.

Closes #2590

### DIFF
--- a/doc/policies/conditions.rst
+++ b/doc/policies/conditions.rst
@@ -30,7 +30,7 @@ consists of five parts:
 Sections
 ~~~~~~~~
 
-privacyIDEA implements three sections ``userinfo``, ``tokeninfo`` and ``HTTP Request Headers``.
+privacyIDEA implements three sections ``userinfo``, ``token``, ``tokeninfo`` and ``HTTP Request Headers``.
 
 ``userinfo``
 ^^^^^^^^^^^^
@@ -114,6 +114,21 @@ The tokeninfo condition works the same way as userinfo but matches the tokeninfo
 
 .. note:: Similar to the userinfo condition, a policy with an active tokeninfo condition will
    throw an exception whenever the token object cannot be determined (usually from the serial).
+
+``token``
+^^^^^^^^^
+
+The token condition works on the database columns of the token. This would be
+``description``, ``otplen``, ``count``, ``serial`` but most importantly
+also ``failcount`` and ``tokentype``.
+
+.. note:: A policy with an active tokeninfo condition will
+   throw an exception whenever the token object cannot be determined.
+   It will also throw an error, if the request ``Key`` does not exist
+   as a database column.
+
+.. note:: The matching is case sensitive. Note, that e.g. token types are
+   stored in lower case in the database.
 
 
 ``HTTP Request Header``
--- a/privacyidea/lib/policy.py
+++ b/privacyidea/lib/policy.py
@@ -812,14 +812,18 @@ class PolicyClass(object):
                 raise PolicyError(u"Unknown HTTP header key referenced in condition of policy "
                                   u"{!r}: {!r}".format(policy["name"], key))
         else:  # pragma: no cover
-            log.error(u"Policy {!r} has conditions on headers {!r}, but http header"
-                      u" is not available. This should not happen.".format(policy["name"], key))
-            raise PolicyError(u"Policy {!r} has conditions on headers {!r}, but http header"
-                        u" is not available".format(policy["name"], key))
+            log.error(u"Policy {!r} has conditions on HTTP headers, but HTTP header"
+                      u" is not available. This should not happen - possible "
+                      u"programming error {!s}.".format(policy["name"],
+                                                        ''.join(traceback.format_stack())))
+            raise PolicyError(u"Policy {!r} has conditions on headers {!r}, but HTTP header"
+                              u" is not available".format(policy["name"], key))
 
     @staticmethod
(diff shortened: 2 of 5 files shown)


privacyidea/lib/policy.py
```diff
@@ -809,6 +817,32 @@ def _policy_matches_request_header_condition(policy, key, comparator, value, req
...
+        """
+        if db_token and key in db_token.get():
+            try:
+                return compare_values(db_token.get(key), comparator, value)
+            except Exception as exx:
+                log.warning(u"Error during handling the condition on token {!r} "
+                            u"of policy {!r}: {!r}".format(key, policy['name'], exx))
+                raise PolicyError(
+                    u"Invalid comparison in the {!s} conditions of policy {!r}".format(type, policy['name']))
+        else:
```
laclaro, line 839: This else is reached if "not db_token OR key not in dbtoken", so also if we have _no_ dbtoken. This happens e.g. at /auth if any matching admin policy has a token condition. 
No login is possible and the error message is misleading, since the reason is that there is simply no token object available.

Please check in this else if there is a dbtoken and if not throw something similar to

"Policy {!r} has condition on token, but an according token object is not available"

as in `_policy_matches_info_condition`

doc/policies/conditions.rst
```diff
@@ -115,6 +115,21 @@ The tokeninfo condition works the same way as userinfo but matches the tokeninfo
...
+also ``failcount`` and ``tokentype``.
+
+.. note:: A policy with an active tokeninfo condition will
+   throw an exception whenever the token object cannot be determined.
+   It will also throw an error, if the request ``Key`` does not exist
+   as a database column.
+
+.. note:: The matching is case sensitive. Note, that e.g. token types are
+   stored in lower case in the database.
+
```
laclaro, line 132: Maybe add a note to not add a token condition to admin policies.

laclaro (changes requested):
This feature works as expected for user policies. If set on any matching admin policy, /auth is not possible. Can we circumvent this somehow? This also applies to tokeninfo conditions.

cornelinux:
> This feature works as expected for user policies. If set on any matching admin policy, /auth is not possible. Can we circumvent this somehow? This also applies to tokeninfo conditions.

You are right. Lately I stumbled upon this also with userinfo condition. (So it happens with userinfo, tokeninfo and token) I think this happens because the /auth endpoints checks for certain policies of an admin, to be able to do some ui settings.

I already thought about this and will come up with a generic solution in the code - where it actually has to be solved - since policy matching has different intentions at certain code sections.

privacyidea/lib/policy.py
cornelinux: I will change this to the flow of the other extended conditions.
However, in certain cases login will still not be possible (as it is also for other conditions.) This is simply a logical problem, which I will not address in the PR, which only deals with the token attributes. (I opened #2676 to do this)

doc/policies/conditions.rst
cornelinux: This is currently also true for other conditions (lile ``userinfo``).
I am planing to solve this in #2676 for 3.6. So I will not add a warning in the docs for 3.6, which will hopefully not be necessary when releasing 3.6 anymore.

privacyidea/lib/policy.py
cornelinux: done in 0f0588ce

doc/policies/conditions.rst
laclaro: okay

laclaro (approved):
Great feature!

PR #2673, opened 2021-03-15
Disable button if no scope is selected

### LINKED ISSUE #1888: Create policy button should be disabled when no scope is selected
Currently the button to create a policy is enabled as soon as the user has given the policy a name. However, if the user clicks the button before selecting a scope, the code will run into an exception. No error message is shown to the user, so from the users perspective, the button will have no function whatsoever. We should make sure the button only gets enabled, when a policy can actually be created.

cornelinux:
Closes #1888

### DIFF
--- a/privacyidea/static/components/config/views/config.policies.details.html
+++ b/privacyidea/static/components/config/views/config.policies.details.html
@@ -394,7 +394,7 @@
          ng-show="checkRight('policywrite')">
         <button ng-click="createPolicy()"
                 id="createPolicyButton"
-                ng-disabled="formPolicyAdd.$invalid"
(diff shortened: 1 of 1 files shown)


NNLemling (commented):
I've tested the whole thing on myself and it works fine. However, there is a second button below. The change should also be made for this in line 398.

cornelinux:
@NNLemling done in 5e0e1567

NNLemling:
sorry, i accidentally closed the pull request.

Everything works fine on my system now. So there shouldn't be anything against merging it.

cornelinux:
@NNLemling you need to approve it and merge it

NNLemling (approved):

PR #2674, opened 2021-03-16
2637/support tls1.3 in LDAP resolver

### LINKED ISSUE #2637: Support TLS v1.3 in LDAP Resolver
When verifying TLS in LDAP resolvers, we can choose to force a TLS protocol version, but not v1.3.

- Support TLS v1.3 enforcing
- We may be also able to force a TLS protocol version when NOT verifying the certificate

laclaro:
This PR introduces hardening changes for the LDAP resolver.

- TLS v1.3 is used as a default
- It is also possible to select a specific TLS version if 1.3 is not supported by the LDAP server
- Protocols are not negotiated with the LDAP server

If the resolverconfig key `TLS_FORCE_VERSION` is set, the resolverconfig `TLS_VERSION` is honored. Otherwise privacyIDEA will fall back to the default v1.3.

Previously, the TLS protocol version was always forced, once `TLS_VERSION` was set. In the WebUI this was always the case if certificate verification was used.
This means especially that we need a migration script which sets `TLS_FORCE_VERSION` for all LDAP resolvers where `TLS_VERSION` is also set. Otherwise people will lose their LDAP connectivity if they do not support TLS v1.3 as their expected default was
- TLS1.0 if `TLS_VERSION` is not set
- TLS1.2 at maximum if `TLS_VERSION` is set

Alternatively, we could default to "negotiation" which might be slightly more attackable. In that case, we won't need a migration and might cause an increase in security even.

- [x] Add tests for TLS parameters as we have none atm
- [x] Documentation
- [x] migration script to set TLS v1.2 for existing resolvers which previously had v1.0 as a default since TLS v1.3 is not supported by many LDAP servers. New resolvers will be created with v1.3 as default 

closes #2637.


privacyidea/lib/resolvers/LDAPIdResolver.py
```diff
@@ -983,14 +998,23 @@ def testconnection(cls, param):
...
-        if is_true(param.get("TLS_VERIFY")) \
-                and (ldap_uri.lower().startswith("ldaps") or
-                                    param.get("START_TLS")):
-            tls_version = int(param.get("TLS_VERSION") or ssl.PROTOCOL_TLSv1)
-            tls_ca_file = param.get("TLS_CA_FILE") or DEFAULT_CA_FILE
-            tls_context = Tls(validate=ssl.CERT_REQUIRED,
-                              version=tls_version,
-                              ca_certs_file=tls_ca_file)
+        if ldap_uri.lower().startswith("ldaps") or param.get("START_TLS"):
+            tls_force_version = is_true(param.get("TLS_FORCE_VERSION")) or False
```
cornelinux, line 1002: It is not clear to me, why we need the additional parameter TLS_FORCE_VERSION.
I was under the impression, that setting a distinct version actually forces to this version.
Lets talk

cornelinux (changes requested):
See my comment and also provide tests for the new code and documentation.

privacyidea/lib/resolvers/LDAPIdResolver.py
```diff
@@ -764,15 +767,26 @@ def loadConfig(self, config):
...
         self.tls_verify = is_true(config.get("TLS_VERIFY", False))
-        # Fallback to TLSv1. (int: 3, TLSv1.1: 4, v1.2: 5)
-        self.tls_version = int(config.get("TLS_VERSION") or ssl.PROTOCOL_TLSv1)
-
+        # Fallback to TLSv1. (int: 3, TLSv1.1: 4, v1.2: 5, TLS negotiation: 2)
+        self.tls_force_version = is_true(config.get("TLS_FORCE_VERSION", False))
+        self.tls_version = int(config.get("TLS_VERSION") or DEFAULT_TLS_PROTOCOL)
+        if self.tls_force_version and self.tls_version:
+            tls_version = self.tls_version
+            tls_options = TLS_OPTIONS_1_3 if tls_version == "2" else None
```
cornelinux, line 775: Please compare to ssl.PROTOCOL_TLS instead of "2".

cornelinux:
@laclaro please check the failing tests.

laclaro:
> @laclaro please check the failing tests.

Done for 3.5 in https://github.com/privacyidea/privacyidea/pull/2674/commits/5a0ecec5c96e105bd01933ddd14870b72e4819ab.

cornelinux:
An empty TLS version in the database shoud be treated as 1.0 as legacy support.

We need to discuss, if a migration script is necessary

doc/configuration/useridresolvers.rst
```diff
@@ -193,14 +193,18 @@ has not passed yet. If the *Per-process server pool* is enabled, knowledge about
 servers is persisted within each process. This setting may improve performance in situations in
 which a LDAP server from the pool is down for extended periods of time.
 
+TLS Version
+~~~~~~~~~~~
+
+When using TLS, you may specify the TLS version to use. Starting from version 3.6, privacyIDEA
+uses TLS v1.3 by default.
```
cornelinux, line 200: I think we should call it like

"...privacyIDEA offers TLS v1.3 as default value for new resovlers"

otherwise it could sound as if the configuration would have been changed.

migrations/versions/9155f0d3d028_.py
```diff
@@ -0,0 +1,64 @@
...
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '9155f0d3d028'
+down_revision = 'd5870fd2f2a4'
+
+from alembic import op
+from sqlalchemy import orm
+from privacyidea.models import Resolver, ResolverConfig
```
cornelinux, line 16: You need to copy the Class definition of the DB ``Resolver`` table and the ``ResolverConfig`` table here!

Otherwise the update script will fail if in later version the table ``Resolver`` or ``ResolverConfig`` would be enhanced.
I can explain this in detail.

migrations/versions/9155f0d3d028_.py
```diff
@@ -0,0 +1,64 @@
...
+                TLS_VERSION = TLS_VERSION_qry.Value
+            else:
+                TLS_VERSION = None
+            # For resolvers that had TLS_VERSION set but empty or not set at all use TLS v1.2
+            if (LDAPS or START_TLS):
+                if TLS_VERSION_qry is None:
+                    session.add(ResolverConfig(resolver_id=resolver_id, Key="TLS_VERSION", Value="5",
+                                               Type=u'int'))
+                elif TLS_VERSION == "":
+                    base_query.filter(ResolverConfig.Key == "TLS_VERSION").update({"Value": "5"})
```
cornelinux, line 54: ...also here.

privacyidea/lib/resolvers/LDAPIdResolver.py
```diff
@@ -764,15 +767,25 @@ def loadConfig(self, config):
         self.resolverId = self.uri
         self.authtype = config.get("AUTHTYPE", AUTHTYPE.SIMPLE)
         self.tls_verify = is_true(config.get("TLS_VERIFY", False))
-        # Fallback to TLSv1. (int: 3, TLSv1.1: 4, v1.2: 5)
-        self.tls_version = int(config.get("TLS_VERSION") or ssl.PROTOCOL_TLSv1)
-
+        # Fallback to TLSv1. (int: 3, TLSv1.1: 4, v1.2: 5, TLS negotiation: 2)
+        self.tls_version = int(config.get("TLS_VERSION") or DEFAULT_TLS_PROTOCOL)
```
cornelinux, line 771: I think the config should now (with the migration script) always contain a TLS_VERSION (at least 1.0 for the old configs).
I am a bit puzzled, if we would have to reset this here to ssl.PROTOCOL_TLSv1.

privacyidea/lib/resolvers/LDAPIdResolver.py
```diff
@@ -764,15 +767,25 @@ def loadConfig(self, config):
...
+        if self.tls_version:
+            tls_version = self.tls_version
+            tls_options = TLS_OPTIONS_1_3 if tls_version == int(ssl.PROTOCOL_TLS) else None
+        else:
```
cornelinux, line 775: I am not clear about this if-else branch.
In case of LDAPS or STARTTLS the line

    self.tls_version = int(config.get("TLS_VERSION") or DEFAULT_TLS_PROTOCOL)

will always contain a value.

In case of plain LDAP the self.tls_version will not be relevant anyways.
So I am not sure, if we need the if-else or if we need it in a different way.

privacyidea/lib/resolvers/LDAPIdResolver.py
```diff
@@ -764,15 +767,25 @@ def loadConfig(self, config):
...
+        if self.uri.lower().startswith("ldaps") or self.start_tls:
+            if self.tls_verify:
+                self.tls_context = Tls(validate=ssl.CERT_REQUIRED,
+                                       version=tls_version,
+                                       ssl_options=tls_options,
+                                       ca_certs_file=self.tls_ca_file)
+            else:
+                self.tls_context = Tls(validate=ssl.CERT_NONE,
+                                       version=tls_version,
+                                       ssl_options=tls_options)
```
cornelinux, line 788: Do I understand it correctly, that we now need a TLS context, even if the certificate is not verified, because TLS 1.3 does not work without a TLS context?

privacyidea/lib/resolvers/LDAPIdResolver.py
```diff
@@ -983,14 +996,22 @@ def testconnection(cls, param):
...
         serverpool_skip = int(param.get("SERVERPOOL_SKIP") or SERVERPOOL_SKIP)
-        if is_true(param.get("TLS_VERIFY")) \
-                and (ldap_uri.lower().startswith("ldaps") or
-                                    param.get("START_TLS")):
-            tls_version = int(param.get("TLS_VERSION") or ssl.PROTOCOL_TLSv1)
-            tls_ca_file = param.get("TLS_CA_FILE") or DEFAULT_CA_FILE
-            tls_context = Tls(validate=ssl.CERT_REQUIRED,
-                              version=tls_version,
-                              ca_certs_file=tls_ca_file)
+        if ldap_uri.lower().startswith("ldaps") or param.get("START_TLS"):
```
cornelinux, line 999: We have the same lengthy code to determine the TLS context in the LDAP resolver and in testconnection.

Do you think we can simplify this with a method like

     get_tls_context(...alll parameters to create the tls context...)

I am not sure myself, since the parameter list would probably get very lengthy.

cornelinux (changes requested):
Please check my comments and questions for me to understand and see what makes sense to be done.

There are a few things that need to be changed.

migrations/versions/9155f0d3d028_.py
```diff
@@ -0,0 +1,132 @@
...
+    Value = sa.Column(sa.Unicode(2000), default=u'')
+    Type = sa.Column(sa.Unicode(2000), default=u'')
+    Description = sa.Column(sa.Unicode(2000), default=u'')
+    __table_args__ = (sa.UniqueConstraint('resolver_id',
+                                          'Key',
+                                          name='rcix_2'),
+                      {'mysql_row_format': 'DYNAMIC'})
+
+
+class Realm(Base):
```
cornelinux, line 55: One last trick for migration scripts. In fact you do not need the **exact** copy of the ORM, only the parts, that are used.
I.e. in a migration script you could only map columns, that you actually need.

I this case you can remove the table definitions ``Realm`` and ``ResolverRealm``, since you are not accessing these tables in the logic of your code.

However, this is defined in the realationship in line 35 ``realm_list``. But you could also safely remove the ``realm_list`` definition from the ``Resolver``  class.

I would do this to shorten the code of the migration script.

Check it out. This comes in quite handy also for other external scripts. Also the linotp-migration script does it in similar ways, since not the comple ORM needs to be mapped.

cornelinux (commented):
The code is all fine now.

Check the failing tests.

I would recommend taking a look at my comment in the migration script, might give some insight for the future.

Also: I again got rather puzzled during my last review.

I think the important thing to note here is, that

**before version 3.6.** an empty or missing TLS_VERSION parameter is treated as TLS 1.

**starting with version 3.6** an empty or missing TLS_VERSION parameter is treated as TLS1 1.3.
The migration script fills up missing TLS_VERSION parameters from old versions, so that old configuration have an existing TLS_VERSION=1.0. 

Could you please already add such a note to the READ_BEFORE_UPDATE:md so that we are aware of this change. 
Thanks!

cornelinux:
@laclaro I took the liberty of fixing some minor aspects to avoid going round and round in circles over again.
Will merge it after successful test runs.

cornelinux (approved):

PR #2681, opened 2021-03-23
3.5 fix out of application context

cornelinux:
Fix two problems in the tests - not the code itself.


privacyidea/lib/resolvers/LDAPIdResolver.py
```diff
@@ -99,9 +99,13 @@
...
 # The number of seconds that ldap3 waits if no server is left in the pool, before
 # starting the next round
-pooling_loop_timeout = get_app_config_value("PI_LDAP_POOLING_LOOP_TIMEOUT", 10)
-log.info("Setting system wide POOLING_LOOP_TIMEOUT to {0!s}.".format(pooling_loop_timeout))
-ldap3.set_config_parameter("POOLING_LOOP_TIMEOUT", pooling_loop_timeout)
+try:
+    pooling_loop_timeout = get_app_config_value("PI_LDAP_POOLING_LOOP_TIMEOUT", 10)
+    log.info("Setting system wide POOLING_LOOP_TIMEOUT to {0!s}.".format(pooling_loop_timeout))
+    ldap3.set_config_parameter("POOLING_LOOP_TIMEOUT", pooling_loop_timeout)
+except RuntimeError:
```
plettich, line 106: I'd rather not fix issues from tests in the code.
Maybe we can move this call to the ``__init__`` function of the LDAPIdResolver, if there is no performance impact. We should always have an application context there.

plettich (changes requested):

privacyidea/lib/resolvers/LDAPIdResolver.py
cornelinux: I thought about putting it into `__init__`.
This also feels a bit wired, because init is called for each resolver, in each process. So the global LDAP3 setting is set several times.
I think the performance impact would not be...
...relevant.

plettich: I also thought about putting it into the ``app.py`` as all the global stuff is set there (and the configuration freshly read). But then we would have to import ``ldap3`` there as well which seemed a little too much.
We could use the technique used in the queue implementation but this also seems like overkill: https://github.com/privacyidea/privacyidea/blob/907a3c15fd2b3233062331912216b533012c952e/privacyidea/lib/queue.py#L34

cornelinux: Then I would prefer to put it into `__init__` leaving it in the LDAPResolver.py file.

Now, you?

cornelinux: see 844fc170

cornelinux:
I will create a new PR

PR #2686, opened 2021-03-26
2555/Enhance_diagnostic_report_tool

NNLemling:
- [x] add apache configuration
- [x] add system status (CPUs, RAM, HD)

I added the apache configurations to the apache_log and expanded the diag-tool with information about the cpu, hd and ram.

One file with all the information has now become three.

**1. privacyidea.config**
- with informations about the OS
- pi_cfg
- current_db_revision
- pi_versions
- pi_config
- FreeRADIUS_configurations
- system_status
- centos_auditlog and the apache logs including the apache config

**2. privacyidea.logfile**
- pi_logfile (It happens sometimes that customers have huge pi_log files. The log is now stored in a seperated file that gives us a better overview.

**3. privacyidea.auditlog**
- pi_auditlog (also stored in a seperated file)

The files are compressed and the customer sends us the tar.gz file at the end.

### DIFF
--- a/tools/privacyidea-diag
+++ b/tools/privacyidea-diag
@@ -155,7 +155,7 @@ system_status() {
 	log
 	log "SECTION: System Status"
 	log "======================"
-	log "CPU cores: $(grep -c ^processor /proc/cpuinfo | wc -l)"
+	log "CPU cores: $(grep -c ^processor /proc/cpuinfo)"
 	log "========================"
 	top -b -n 1 >> "$tempfile"
 	log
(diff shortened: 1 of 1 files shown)


tools/privacyidea-diag
```diff
@@ -156,51 +180,51 @@ apache_log() {
 		tail -50 /var/log/httpd/ssl_error_log  >> "$tempfile"
 		log
 		log "SECTION: httpd Access_log"
-		log "============================"
+		log "========================="
 		tail -50 /var/log/httpd/ssl_access_log  >> "$tempfile"
+		log
```
plettich, line 185: I think we should include at least the 100 last lines here, just to be on the safe side.

plettich (changes requested):
Some improvements, just contact me if there are questions.

plettich (changes requested):
just some minor changes

plettich (approved):
looks good

PR #2688, opened 2021-03-29
Fix some bugs from 3.5 merge

cornelinux:
fix out of app context in LDAP resolver
fix number of push tokens in test
fix remnant of OAuth token


plettich (approved):
Looks good

Issue #2613, opened 2021-02-04
Policy to change /token/init to "rollover"


laclaro:
Frequently, people would like to only allow one token per user (or per token type). This can be achieved by setting a max_token-type. However, often people would also like to automatically delete the old token whenever the user self-enrolls a new one.

At present, this can only be done using an event-handler script. 

My suggestion would be to implement a policy which brings this functionality to the privacyIDEA core.

- The policy could hold a list of token types to apply it to
- It should only affect user inits
- It should maybe be restricted to software tokens

laclaro:
Can be done with script https://github.com/privacyidea/scripts/pull/6.

laclaro:
Unfortunately, the scripts are not the best solution in some cases. Especially if you would like to restrict the number of tokens per user to 1. If, for instance users enroll their own token (singular) and for some reason the token is deactivated by some admin to lock the user. The user will not be able to enroll a new token.

This works with a pre-event handler, but this is unfortunately not very flexible. E.g. at the moment, pre event handler do not have access to the tokentype to be enrolled.

For the post-event handling script, the max_token restriction cannot be 1 since the user first enrolls a token and the script afterwards deletes all other tokens (e.g. by type).

## Alternatives
To realize the use case of the locked user with a post-event-handler-script, you can set the max_token to 2 and define an event handler, which enrolls a blocker-token on /token/disable by an admin and another which removes it on /token/enable.

cornelinux:
I think it would make more sense to enhance the event handler this way, that also other parameters from the request can be passed to the event handler.
THis way the event hanlder (script handler) would have the information about the to be enrolled tokentype.
This would also help for other scenarios.

cornelinux:
Thoughs from BAM:

* The seed regeneration must not happen with hardware tokens
* Should the regeneration occurr in a wizard?
* How many click should the user do? Regenerate in the token list (less clicks) or in the token details (more clicks)? Or at both locations

laclaro:
On regeneration of a token, we will need to take care of

- token type
- token realms

laclaro:
Buttons to Rollover will be displayed in the token list and the token details.

There might be be also a policy "wizard_token_rollover" which causes the wizard to be always displayed on user login to the WebUI and to rollover the first (and only?) token of the user.

laclaro:
2stepinit will not work, since capturing the input from a modal dialog is not easily possible, see:

[...]

PR #2692, opened 2021-04-09
2613/token regenerate token from list token and token details view

laclaro:
This PR adds the token rollover policy and adds the feature to the WebUI in token list and token details view.

working on #2613

- [x] Policy
- [x] WebUI
- [x] Documentation


privacyidea/static/components/token/views/dialog.regenerate_token.html
```diff
@@ -0,0 +1,57 @@
...
+
+            <div class="modal-content" ng-show="regenerateTokenStep === 1">
+                <div class="modal-header">
+                    <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                        <span aria-hidden="true">&times;</span></button>
+                    <h4 class="modal-title" id="myModalLabel"
+                        translate>Token secret regenerated</h4>
+                </div>
+                <div class="modal-body">
+                    <ng-include src="instanceUrl + '/static/components/token/views/token.enrolled.' + selectedToken.tokentype + '.html'">
```
cornelinux, line 47: We should only put the question, if the token should be rolled over in a modal dialog.
The rollover dialog itself should be a dedicated route.
This way, the rollover could also be addressed directly and we would follow the flat design.

To do so we need to add a route in static/components/token/states/states.js like this:

~~~~javascript
              .state('token.rollover', {
                    url: "/rollover/:tokenSerial",
                    templateUrl: tokenpath + "token.enrolled.hotp.html" + versioningSuffixProviderProvider.$get().$get(),
                    controller: "tokenDetailController",
                })
~~~~

However, we might need to add a wrapper html template so that we can handle different token types of the templates.

We could also put the javascript logic of rollover into its own controller.

cornelinux (changes requested):

privacyidea/static/components/token/controllers/tokenRolloverController.js
```diff
@@ -0,0 +1,104 @@
...
+        TokenFactory.getTokenForSerial($scope.enrolledToken.serial, function(data) {
+            $scope.enrolledToken.rollout_state = data.result.value.tokens[0].rollout_state;
+            // Poll the data after 2.5 seconds again
+            if ($scope.enrolledToken.rollout_state === "clientwait") {
+                $timeout($scope.pollTokenInfo, 2500);
+            }
+        })
+    };
+
+    $scope.sendClientPart = function () {
```
plettich, line 88: this doesn't work yet?

privacyidea/static/components/token/views/token.details.html
```diff
@@ -10,13 +10,24 @@ <h3 name="Token Details"><translate>Token details for {{ tokenSerial }}
...
                         ng-hide="!checkRight('delete') && hide_buttons">
                     <span class="glyphicon glyphicon-trash"></span>
                     <span translate>Delete</span>
                 </button>
+                </span>
+                <span>
+                <button class="btn btn-danger" ng-click="regenerateTokenAsk(token)"
+                        id="rolloverButton"
+                        ng-disabled="!regenerateTokenAllowed(token)"
+                        ng-hide="!regenerateTokenAllowed(token) && hide_buttons">
```
plettich, line 26: I would show the button only if the token regenerate is allowed. Otherwise we have the disabled button on every token type (even the ones which do not support rollover anyway) which might confuse people (I know about the 'hide_buttons' but still...).

privacyidea/static/components/token/views/token.list.html
```diff
@@ -130,7 +133,22 @@
             user_details_in_tokenlist">
                 <a ui-sref="config.resolvers.list">{{ token.resolver }}</a>
             </td>
+            <td ng-show="!checkPolicyEmpty(token_rollover)">
```
plettich, line 136: Same here, even if the button is disabled it might confuse people.

plettich (commented):
This all seems a little clunky, i guess we have to admit that we are no frontend/js/angular devs :-)
But it does what it should...

doc/policies/webui.rst
```diff
@@ -300,6 +300,22 @@ being disabled.
...
+type: str
+
+This is a whitespace separated list of tokentypes, for which a rollover button is
+displayed in the token details and in the token list. This button will generate a
+new token secret for the displayed token.
+
+This e.g. enables a user to transfer a softtoken to a new device while keeping the
+token number restricted to 1.
+
+This policy does not work for all token types. Use it with totp, hotp, push, paper or tan.
```
plettich, line 315: I can confirm that re-enrolling a push token works as well.

privacyidea/static/components/token/views/token.list.html
laclaro: done in 90b60853d1020aa950c45f486257760931a1b290

privacyidea/static/components/token/controllers/tokenRolloverController.js
laclaro: fixed in 90b60853d1020aa950c45f486257760931a1b290

privacyidea/static/components/token/controllers/tokenDetailController.js
```diff
@@ -386,5 +386,28 @@ myApp.controller("tokenDetailController", function ($scope,
...
+
+    $scope.regenerateTokenAllowed = function(token) {
+        if ( typeof(token) != 'undefined' ) {
+            if ($scope.checkEnroll() && (token.tokentype in $scope.token_rollover) &&
+                token.info.tokenkind === 'software' && token.rollout_state !== "clientwait") {
+                return true;
+            }
+        }
+        return false;
+    };
```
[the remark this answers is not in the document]
laclaro: would be nice to have, but I would leave it like this for now.

privacyidea/static/components/token/views/token.details.html
laclaro: done in https://github.com/privacyidea/privacyidea/commit/90b60853d1020aa950c45f486257760931a1b290

privacyidea/static/components/token/controllers/tokenRolloverController.js
plettich: If the 2step enrollment is forced, this works but if it is only allowed, it doesn't. It just enrolls a new token. But of course we would need to request the 2step enrollment in the modal before...

plettich (commented):
please see my comment below

doc/policies/webui.rst
```diff
@@ -300,6 +300,22 @@ being disabled.
...
+
+(Since privacyIDEA 3.6)
```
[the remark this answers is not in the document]
cornelinux: We decided that to be a webui policy. Since enrollment policies also work during enrollment on the API level.
This however, is a pure UI policy, since it displays a button, that leads to an API functionality, that works anyways.

laclaro:
closed in favor of https://github.com/privacyidea/privacyidea/pull/2741

PR #2694, opened 2021-04-09
2555/flexibility increases that there are no errors with debian / rhel

### LINKED ISSUE #2555: Enhance diagnostic report tool
The `privacyidea-diag` tool could use some enhancements:
- [x] check selinux status on CentOS (and, if enabled, include the system audit-log as well)
- [x] add the apache logs
- [x] add the database structure/current db revision
- [x] add FreeRADIUS configuration
- [x] add apache configuration
- [x] add system status (CPUs, RAM, HD)

NNLemling:
1. add error message for not supported OS
2. include debian/rehl to avoid errors

Closes #2555


pablo-knight (approved):
Check and tested on rhel, centos and ubuntu--> works

PR #2695, opened 2021-04-12
Add missing migration script

### LINKED ISSUE #1059: Allow auth_cache to cache only for a limited number of authentications

plettich:
If an authcache table already exists, the new 'auth_count' column must
be added to that table.

Closes #1059

### DIFF
--- a/migrations/versions/d415d490eb05_.py
+++ b/migrations/versions/d415d490eb05_.py
@@ -1,4 +1,4 @@
-"""Add auth_count column to authcache table
+"""v3.6: Add auth_count column to authcache table
 
 Revision ID: d415d490eb05
 Revises: 9155f0d3d028
--- a/migrations/versions/d415d490eb05_.py
+++ b/migrations/versions/d415d490eb05_.py
@@ -1,0 +1,33 @@
+"""Add auth_count column to authcache table
+
+Revision ID: d415d490eb05
+Revises: 9155f0d3d028
+Create Date: 2021-04-06 21:19:25.931603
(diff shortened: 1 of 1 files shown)


migrations/versions/d415d490eb05_.py
```diff
@@ -0,0 +1,33 @@
+"""Add auth_count column to authcache table
```
cornelinux, line 1: Can you please add in the first line at the beginning:

    v3.6: Add...

Since only the first line is displayed when running

    pi-manage db history

cornelinux (changes requested):
Please note my small comment about the comment.

migrations/versions/d415d490eb05_.py
plettich: Will do

plettich: Done in 53bf2249

cornelinux (approved):

Issue #2455, opened 2020-10-21
transfer configuration / pi-manage full config import/export


laclaro:
# Configuration transfer

pi-manage should be able to export and import the complete configuration to duplicate the configuration e.g. to duplicate the productive setup for development and testing purposes.

- [x] Implementation of the base functionality to export and import configurations https://github.com/privacyidea/privacyidea/pull/2696
- Config sections to be exported
   - [x] ca connectors https://github.com/privacyidea/privacyidea/pull/2822/commits/faa5c79a35f19ea7886cb95a06617333199ae124
   - [x] system config https://github.com/privacyidea/privacyidea/pull/2696
   - [x] events https://github.com/privacyidea/privacyidea/pull/2822
   - [x] machine resolvers https://github.com/privacyidea/privacyidea/pull/2822/commits/faa5c79a35f19ea7886cb95a06617333199ae124
   - [x] periodic tasks https://github.com/privacyidea/privacyidea/pull/2822/commits/faa5c79a35f19ea7886cb95a06617333199ae124
   - [x] policies https://github.com/privacyidea/privacyidea/pull/2822
   - [x] privacyidea servers https://github.com/privacyidea/privacyidea/pull/2822/commits/32567e80adb0d571236eb9a5663ed1c64b96fec3
   - [x] radius servers https://github.com/privacyidea/privacyidea/pull/2822/commits/bb5b61e491dfaa50a94bc7a8f033e9c382ce848f
   - [x] realms https://github.com/privacyidea/privacyidea/pull/2696
   - [x] resolvers https://github.com/privacyidea/privacyidea/pull/2696
   - [x] sms gateways https://github.com/privacyidea/privacyidea/pull/2822/commits/6862783b31abf43834ee65de006203f91d7960fe
   - [x] smtp servers https://github.com/privacyidea/privacyidea/pull/2822/commits/da4bc32cf90804e3db65fbf5e1a8a94b58595635 and https://github.com/privacyidea/privacyidea/pull/2822/commits/57d517f8fd5faaaaa1007e3fa75131a0f984613b

The actual data is not exported
- internal admins
- tokens
- token owners/assignments
- subscriptions
- audit
- stats

_Original issue text_
> In some cases we would like to transfer a privacyidea database from one development environment to another. Maybe these
> environments have different MySQL database connections (ip, username password, db-name).
>
[...]

Issue #2498, opened 2020-11-06
migrate pi-manage to click


laclaro:
similar to the privacyidea admin tool (https://github.com/privacyidea/privacyideaadm/issues/45), we would like to migrate pi-manage to click to have a more flexible and transparent menu.

PR #2696, opened 2021-04-15
First version of import/export tool

plettich:
- export/import functions can be registered
- pi-manage picks up these registered functions and creates subcommands
  with the appropriate parameters
- converting from/to YAML, json and python works

Working on #2455

### DIFF
--- a/pi-manage
+++ b/pi-manage
@@ -61,6 +61,7 @@ from getpass import getpass
 import gnupg
 import yaml
 import contextlib
+import argparse
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.asymmetric import rsa
 from cryptography.hazmat.primitives import serialization
@@ -109,25 +110,23 @@ DEFAULT_CONFTYPE_LIST = ("policy", "resolver", "event")
 
 app = create_app(config_name='production', silent=SILENT)
 manager = Manager(app)
-admin_manager = Manager(description='Create new administrators or modify existing '
+admin_manager = Manager(usage='Create new administrators or modify existing '
                                     'ones.')
-backup_manager = Manager(description='Create database backup and restore')
-realm_manager = Manager(description='Create new realms, delete existing realms '
+backup_manager = Manager(usage='Create database backup and restore')
+realm_manager = Manager(usage='Create new realms, delete existing realms '
                                     'or set the default realm')
-resolver_manager = Manager(description='Create new resolver')
-policy_manager = Manager(description='Manage policies')
-event_manager = Manager(description='Manage events')
-api_manager = Manager(description="Manage API keys")
-ca_manager = Manager(description="Manage Certificate Authorities")
-audit_manager = Manager(description="Manage Audit log")
-authcache_manager = Manager(description="Manage AuthCache")
-hsm_manager = Manager(description="Manage HSM")
-token_manager = Manager(description="Manage tokens")
-config_manager = Manager(description="Manage the privacyIDEA configuration")
-config_import_manager = Manager(description="import configuration",
-                                help="import configuration")
-config_export_manager = Manager(description="export configuration",
-                                help="export configuration")
+resolver_manager = Manager(usage='Create new resolver')
+policy_manager = Manager(usage='Manage policies')
+event_manager = Manager(usage='Manage events')
+api_manager = Manager(usage="Manage API keys")
+ca_manager = Manager(usage="Manage Certificate Authorities")
+audit_manager = Manager(usage="Manage Audit log")
+authcache_manager = Manager(usage="Manage AuthCache")
+hsm_manager = Manager(usage="Manage HSM")
+token_manager = Manager(usage="Manage tokens")
+config_manager = Manager(usage="Manage the privacyIDEA configuration")
+config_import_manager = Manager(usage="import configuration")
+config_export_manager = Manager(usage="export configuration")
 manager.add_command('db', MigrateCommand)
 manager.add_command('admin', admin_manager)
 manager.add_command('backup', backup_manager)
@@ -623,7 +622,7 @@ def profile(length=30, profile_dir=None):
     """
     Start the application in profiling mode.
     """
-    from werkzeug.contrib.profiler import ProfilerMiddleware
+    from werkzeug.middleware.profiler import ProfilerMiddleware
     if flask.has_request_context():
         print("WARNING: The app may behave unrealistically during profiling.")
     app.wsgi_app = ProfilerMiddleware(app.wsgi_app, restrictions=[length],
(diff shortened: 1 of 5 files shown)


privacyidea/lib/utils/export.py
```diff
@@ -0,0 +1,40 @@
...
+    def wrapped(func):
+        # TODO: this doesn't work since the logging configuration is not
+        #       initialized during imports
+        # log.debug('Registering export function {0!s} with name {1!s}'.format(func, name))
+        IMPORT_FUNCTIONS[name] = func
+        return func
+    return wrapped
+
+
+def register_export(func):
```
cornelinux, line 31: It is not clear to me, why you also have `register_export` and `register_import` which are not used anywhere - currently.

pi-manage
```diff
@@ -1286,8 +1289,130 @@ def import_tokens(file, tokenrealm=None):
...
+
+    # now create the command from the wrapper function
+    cmd = Command(exporter_wrapper(func))
+    # and add all the options
+    for o in opts:
+        cmd.add_option(o)
+    return cmd
+
+
+config_exporter_manager = Manager(help='registered exporter commands',
```
cornelinux, line 1350: I think the admin does not care about internal registration, so a help text like "list available exportable sections" might be more straightforward for the admin.

pi-manage
```diff
@@ -1286,8 +1289,130 @@ def import_tokens(file, tokenrealm=None):
...
+                                  description='registered exporters')
```
cornelinux, line 1351: see above, "available exporters" sounds better. Otherwise - I as an admin would wonder, how I could register new exporters.

pi-manage
```diff
@@ -1286,8 +1289,130 @@ def import_tokens(file, tokenrealm=None):
...
+
+    # now create the command from the wrapper function
+    cmd = Command(importer_wrapper(func))
+    # and add all the options
+    for o in opts:
+        cmd.add_option(o)
+    return cmd
 
+
+config_importer_manager = Manager(help='registered importer commands',
```
cornelinux, line 1408: see my above comment about the phrase "registered".

privacyidea/lib/utils/export.py
```diff
@@ -0,0 +1,40 @@
...
+
+import logging
+
+log = logging.getLogger(__name__)
+
+EXPORT_FUNCTIONS = {}
+IMPORT_FUNCTIONS = {}
+
+
+def register_export_name(name):
```
cornelinux, line 11: Currently it is not clear to me, why there is a `register_export_name` and a `register_export`.

I would prefer to drop the phrase "_name(name)". If we one day have exporters/importer all over the code, "name" is a bit unspecific.
Maybe the signature

~~~~python
def register_config_export(section):
~~~~

might be more intuitive, when stumbling over it somewhere in the code.

cornelinux (commented):
Please note my comments, basically about structure and naming.

privacyidea/lib/utils/export.py
plettich: This was from a previous test where i used the module name for the registration.
I'll make `register_export` and `register_import` the default with both functionality (passing a name or using the functions module name).

privacyidea/lib/utils/export.py
plettich: See above, I'll use `register_export` and `register_import`.

privacyidea/lib/utils/export.py
plettich: Done in 426389b

pi-manage
plettich: Rewrote manager in 426389b

pi-manage
plettich: The flask-script extension is not very intuitive when it comes to descriptions/help/usage. This should be better when we switch to click with #2498

privacyidea/lib/utils/export.py
plettich: Done in 426389b

pi-manage
plettich: Done in 426389b

privacyidea/lib/utils/export.py
```diff
@@ -0,0 +1,36 @@
...
+
+EXPORT_FUNCTIONS = {}
+IMPORT_FUNCTIONS = {}
+
+
+def register_export(name=None):
+    def wrapped(func):
+        # TODO: this doesn't work since the logging configuration is not
+        #       initialized during imports
+        # log.debug('Registering export function {0!s} with name {1!s}'.format(func, name))
```
cornelinux, line 15: Please check if you want to keep this many TODOs.

cornelinux (commented):
Only some comments.

pi-manage
```diff
@@ -1286,8 +1288,78 @@ def import_tokens(file, tokenrealm=None):
...
+                       help='Output format, default is \'python\'')
+@config_manager.option('-t', '--types', nargs='*', default=['all'],
+                       help='The types of configuration to export. By default create '
+                            'export using all available exporter types. Currently registered '
+                            'exporter types are: {0!s}'.format(', '.join(EXPORT_FUNCTIONS.keys())))
+def exporter(output, fmt, types):
+    """export configuration"""
+    print(output, file=sys.stderr)
+    print(fmt, file=sys.stderr)
+    print(types, file=sys.stderr)
```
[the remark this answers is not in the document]
plettich: Done in ef40697f

privacyidea/lib/utils/export.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
plettich: Done in ef40697f

pi-manage
```diff
@@ -1286,8 +1288,78 @@ def import_tokens(file, tokenrealm=None):
...
+        try:
+            d = inp_fmt_dict[tp](content)
+            break
+        except (SyntaxError, json.decoder.JSONDecodeError, yaml.error.YAMLError) as e:
+            continue
+    if not d:
+        print('Could not read input format! Accepting JSON, YAML or native Python.')
+        sys.exit(1)
+
+    for typ, content in d.items():
```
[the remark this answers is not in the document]
plettich: Done in ef40697f

privacyidea/lib/utils/export.py
plettich: Done in ef40697f

pi-manage
```diff
@@ -1288,75 +1288,82 @@ def import_tokens(file, tokenrealm=None):
...
+    if not data:
+        print('Could not read input format! '
+              'Accepting: {0!s}.'.format(', '.join(imp_fmt_dict.keys())),
+              file=sys.stderr)
         sys.exit(1)
 
-    for typ, content in d.items():
+    # we need to go through the importer functions based on priority
+    for typ, value in sorted(IMPORT_FUNCTIONS.items(), key=lambda x: x[1]['prio']):
         if typ in imp_types or 'all' in imp_types:
```
cornelinux, line 1363: I think this only needs to be:

~~~~python
if typ in imp_types:
~~~~
since in line `imp_types` was already set to the `.keys()` if 'all' as contained.
Thus, 'all' can not be contained in `imp_types` at all.

privacyidea/lib/utils/export.py
```diff
@@ -0,0 +1,36 @@
...
+
+import logging
+
+log = logging.getLogger(__name__)
+
+EXPORT_FUNCTIONS = {}
+IMPORT_FUNCTIONS = {}
+
+
+def register_export(name=None):
```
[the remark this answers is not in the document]
cornelinux: We are still missing a documentation like a docstring for the exporter and import, so that a developer knows how this is to be used.

"""
The register_export decorator is supposed to decorate a function that exports the configuration data from a given module like resolvers, realms, events...

The exporter function simply needs to return a dictionary contained the data.

@register_export
def export_events():
     .... some code ...
"""

And a docstring, so that a developer knows, how to use the import decorator.

"""
The decorator register_import is used to register an importer function, that can import data to a privacyidea section,
that was previously exported by the exporter.

Usually this is a dictionary. The exporter and importer of the section take care of handling the data correctly.

The priority can be used to define, which section are imported before other sections. Like resolvers need to be imported before realms, since realms depend on the resolvers.

example:

@register_import(prio=1)
def import_events():
 """

cornelinux (commented):
Looks good to me, I have only two minor comments, which you may consider:

1. The for loop, that imho does not necessarily require the 'all'-check and
2. some doc-strings for developers to understand how the exporter and importer are to be registered.

pi-manage
plettich: Done in f6f28b91

privacyidea/lib/utils/export.py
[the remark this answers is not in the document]
plettich: Done in f6f28b91

cornelinux (approved):

cornelinux:
#automerge after successful test :-)

PR #2700, opened 2021-04-19
add note about Origin-bound one-time codes

### LINKED ISSUE #2151: default SMS text format
There is a definition for SMS text by Apple for logging in to websites with 2FA:

https://github.com/wicg/sms-one-time-codes

This can of course be configured in privacyIDEA. But maybe we want to make this default, add a button or at least add a note in the documentation.

laclaro:
closes #2151

### DIFF
--- a/doc/policies/authentication.rst
+++ b/doc/policies/authentication.rst
@@ -187,6 +187,12 @@ Default: *<otp>*
    You should take care, that the *smstext* does not exceed this limit. SMS gateways could
    reject too long messages or the delivery could fail.
 
+.. note:: Some phones expect incoming OTPs as a so called
+   `origin-bound one-time code <https://github.com/wicg/sms-one-time-codes>`_.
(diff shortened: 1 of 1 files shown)


doc/policies/authentication.rst
```diff
@@ -187,6 +187,12 @@ Default: *<otp>*
    You should take care, that the *smstext* does not exceed this limit. SMS gateways could
    reject too long messages or the delivery could fail.
 
+.. note:: Some phones expect incoming OTPs as a so called
```
cornelinux, line 190: I would say "they can handle". Since the phone will also work with a simple OTP value. THe only thing is, that the provide provides an added value, if the text is sent in this format. The default format still works.

Also: Have you checked if a %0A will introduce a linebreak?

cornelinux (changes requested):
Only minor comment and question

cornelinux (approved):
Cool.
Right to the point - nothing more nothing less ;-)

PR #2703, opened 2021-04-27
Fix the DB migration branch

### LINKED ISSUE #2702: Reduce to one head in migration scripts
In the current development branch we ended up with two heads in the alembic history. Check with:

    pi-manage db history

We need to reorganize the migration scripts to only have one head in v3.6.

cornelinux:
We ended up with two heads. Instead of adding an
additional mergepoint this commit reorders the
DB migration scripts, since they are independent on each other.

Also remove a line break in an older migration script, which looks
a bit awkward in

   pi-manage db history

Fixes #2702


plettich (approved):
looks good

PR #2704, opened 2021-04-29
2642/fix add_user/resolver_in_response

NNLemling:
add_user/resolver_in_response will now evaluate the user/realm.

In addition, the previous tests have been adjusted

### DIFF
--- a/privacyidea/api/lib/postpolicy.py
+++ b/privacyidea/api/lib/postpolicy.py
@@ -329,7 +329,7 @@ def add_user_detail_to_response(request, response):
     content = response.json
 
     # Check for ADD USER IN RESPONSE
-    detail_pol = Match.action_only(g, scope=SCOPE.AUTHZ, action=ACTION.ADDUSERINRESPONSE)\
+    detail_pol = Match.user(g, scope=SCOPE.AUTHZ, action=ACTION.ADDUSERINRESPONSE, user_object=request.User)\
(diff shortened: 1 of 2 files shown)


privacyidea/api/lib/postpolicy.py
```diff
@@ -329,7 +329,7 @@ def add_user_detail_to_response(request, response):
     content = response.json
 
     # Check for ADD USER IN RESPONSE
-    detail_pol = Match.action_only(g, scope=SCOPE.AUTHZ, action=ACTION.ADDUSERINRESPONSE)\
+    detail_pol = Match.user(g, scope=SCOPE.AUTHZ, action=ACTION.ADDUSERINRESPONSE, user_object=request.User)\
         .policies(write_to_audit_log=False)
     if detail_pol and content.get("result", {}).get("value") and request.User:
```
plettich, line 334: I think we should check for `request.User` before calling `Match.user()`.
This whole function only makes sense if we have a user in the request anyway.

plettich (changes requested):

privacyidea/api/lib/postpolicy.py
plettich: After some extended thinking i've come to the conclusion that this is not necessary. I'll merge it now.

plettich (approved):

PR #2707, opened 2021-05-06
Enable fractions on MySQL DateTime type

### LINKED ISSUE #2706: MySQL DateTime must be told explicitely to save fractions of seconds
Since 9380eaee the audit log contains duration times of the request. Unfortunately MySQL by default doesn't save fractions in it's DateTime type. It has to be enabled explicitly.
With a compiled statement for SQLAlchemy the fractions for the DataTime object can be enabled only for MySQL.
The simple version would change all DataTime columns in all tables. This could be narrowed down to only use this for the audit table.
We need to check how other databases behave:
- [ ] Oracle
- [ ] Postgres

plettich:
Change database model and migration to enable fraction of seconds for
the MySQL DateTime type for the audit table.
Currently MySQL saves DateTime types with a precision of 1s which is
inaccurate for tracking request times in the audit log.

Fixes #2706


privacyidea/models.py
```diff
@@ -2654,7 +2662,7 @@ class Audit(MethodsMixin, db.Model):
     id = db.Column(db.Integer, Sequence("audit_seq"), primary_key=True)
     date = db.Column(db.DateTime)
```
cornelinux, line 2663: Don't we need to set a second precision here, too?
No, this is done by adding compiles decorator to the DateTime column type?

cornelinux (commented):
Just clarify one thing for me, thanks.

cornelinux (approved):

PR #2709, opened 2021-05-06
Remove usage of pkg_resources

### LINKED ISSUE #2708: Stamp database version when creating the db with pi-manage
Running `pi-manage createdb` does not stamp the database version thus we always need to call `pi-manage db stamp -d <path to migrations> head` afterwards.
We can call the `stamp()` function from [flask-migrate](https://flask-migrate.readthedocs.io/en/latest/) directly after creating the database.

plettich:
The pkg_resources package is kind of deprecated and [the use of
importlib_resources and importlib_metadata is encouraged](https://setuptools.readthedocs.io/en/latest/pkg_resources.html).
This PR also adds the version information to the PI-ASCII-art when
running `pi-manage` and determines the migration path from the installed
distribution files. That way we can stamp the database when running
`pi-manage createdb`.

Fixes #2708


requirements.txt
```diff
@@ -33,6 +33,8 @@ idna==2.10
 importlib-resources==3.3.1; python_version < '3.0'
 importlib-resources==3.2.1; python_version == '3.5'
 importlib-resources==5.0.0; python_version > '3.5'
+importlib_metadata==2.1.1; python_version <= '3.0'
+importlib_metadata==3.4.0; python_version > '3.5'
```
cornelinux, line 37: Why do we need to remove pkg_resource and introduce a *new dependency*?

cornelinux (changes requested):

requirements.txt
plettich: As stated in the link from the commit comment, this is the way to go...

pi-manage
```diff
@@ -559,14 +560,21 @@ def create_audit_keys(keysize=2048):
...
-    Initially create the tables in the database. The database must exist.
-    (SQLite database will be created)
+    Initially create the tables in the database. The database must exist
+    (an SQLite database will be created).
     """
     print(db)
     db.create_all()
+    if stamp:
+        # get the path to the migration directory from the distribution
+        p = [p.locate() for p in importlib_metadata.files('privacyidea') if
```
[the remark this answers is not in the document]
plettich: Done in 0c75cc8a

cornelinux (approved):

PR #2718, opened 2021-05-10
Fix enroll-button staying disabled

### LINKED ISSUE #2717: Enroll Button stays disabled after error during enrollment
# What happens

Up to version 3.5.2

If a user tries to enroll a token in the webui but the enrollment fails, the enroll-button stays disabled, if an error occurs.
The error can happen e.g. because the PIN policies do not match during enrollment.

# Why it happens

The reason for the disabled button is, the variable ``$scope.enrolling``, which is checked here
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/static/components/token/views/token.enroll.html#L122

The ``scope.enrolling`` is set to *true*, when the button is clicked here:
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/static/components/token/controllers/tokenControllers.js#L465

The reason for doing this is to avoid double clicking the enrollment button, which resulted in enrolling two tokens in the past!

# What we need to do

In case of an error the ``$scope.enrolling`` remains *true*. So the button is still disabled.
In case of a successful enrollment, we change to a different route, and thus the state of the button does not matter.
But in case of an errror we have to reset ``$scope.enrolling`` to *false*.

We can do this in ``token/factories/token.js enroll()`` like this
(here: https://github.com/privacyidea/privacyidea/blob/master/privacyidea/static/components/token/factories/token.js#L220):

~~~~javascript
 $http.post(tokenUrl + "/init", params,
                    {headers: {'PI-Authorization': AuthFactory.getAuthToken()}}
                ).then(function (response) { callback(response.data) }, function(error) { 
                          $rootScope.enrolling = false;
                          AuthFactory.authError(error.data); });
~~~~

cornelinux:
In case of an error during enrollment the enroll-button
stays disabled. This happens due to the $scope.enrolling
still being "true" in the webUI. We need to set this
to "false" again. To do so, we add a callback-function
for the error-case. This way we can reset the
enrolling variable in the correct scope.

Closes #2717

### DIFF
--- a/privacyidea/static/components/token/controllers/tokenControllers.js
+++ b/privacyidea/static/components/token/controllers/tokenControllers.js
@@ -471,7 +471,9 @@ myApp.controller("tokenEnrollController", ["$scope", "TokenFactory", "$timeout",
         $scope.form.validity_period_start = date_object_to_string($scope.form.validity_period_start);
         $scope.form.validity_period_end = date_object_to_string($scope.form.validity_period_end);
         TokenFactory.enroll($scope.newUser,
-            $scope.form, $scope.callback);
+            $scope.form, $scope.callback,
+            function (data) {$scope.enrolling=false;}
+            );
     };
 
     $scope.pollTokenInfo = function () {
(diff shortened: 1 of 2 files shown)


plettich (approved):
looks good.

PR #2722, opened 2021-05-12
2516/add more descriptive tooltip to token when assigning to machine

### LINKED ISSUE #2516: add more descriptive tooltip to token when assigning to machine
When managing a machine, I can add tokens via the "Attach a new token PIP" entry field. While any matching token id is displayed when entering data into the "serial" entry field, the tooltips of the listed tokens only show the token ID, as do the actual list entries.

Having many tokens, it is hard to validate to have hit the proper token just by the id.

To at least me, it would be helpful if some details of the token, like owning user and its "description" value would also be displayed. This could be implemented as according content of the tokens' tooltips or perhaps directly as a suffix to the according list entry.

creating a pop-up selection dialog could be considered the "full notch solution", with the ability to search via serial, user and description.

NNLemling:
Now also shows "user+realm" and "description" when entering data into the "serial" entry field if add token to a machine
and "description" if add a token to an existing user.

closes #2516

### DIFF
--- a/privacyidea/static/components/directives/controllers/directives.js
+++ b/privacyidea/static/components/directives/controllers/directives.js
@@ -211,7 +211,11 @@ myApp.directive('assignToken', function($http, $rootScope, tokenUrl,
                 serial: "*" + $viewValue + "*"}
             }).then(function ($response) {
                 scope.newTokenObject.loadedSerials = $response.data.result.value.tokens.map(function (item) {
-                    return item.serial + " (" + item.tokentype + ")";
+                serial_string = item.serial + " [" + item.tokentype + "] "
+                if (item.description !=''){
+                    serial_string += "[" + item.description + "] "
+                }
+                    return serial_string;
                 });
                 return scope.newTokenObject.loadedSerials;
             });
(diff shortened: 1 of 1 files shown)


privacyidea/static/components/directives/controllers/directives.js
```diff
@@ -211,7 +211,11 @@ myApp.directive('assignToken', function($http, $rootScope, tokenUrl,
                 serial: "*" + $viewValue + "*"}
             }).then(function ($response) {
                 scope.newTokenObject.loadedSerials = $response.data.result.value.tokens.map(function (item) {
-                    return item.serial + " (" + item.tokentype + ")";
+                serial_string = item.serial + " [" + item.tokentype + "] "
```
cornelinux, line 214: Would you change this back to rounded brackets, so we would keep the tokentype in rounded brackets and all additional information in square brackets.

privacyidea/static/components/directives/controllers/directives.js
```diff
@@ -245,7 +249,14 @@ myApp.directive('attachToken', function($http, tokenUrl,
                 params: {serial: "*" + $viewValue + "*"}
             }).then(function ($response) {
                 return $response.data.result.value.tokens.map(function (item) {
-                    return item.serial + " (" + item.tokentype + ")";
+                serial_string = item.serial + " [" + item.tokentype + "] "
```
cornelinux, line 252: see above.

cornelinux (commented):
Nice. Please consider my comments.

cornelinux (approved):

PR #2729, opened 2021-05-19
Only use no-tablespaces for mysqldump

### LINKED ISSUE #2728: restore breaks with pi-manage
When you want to restore a backup using pi-manage you will run into the error:

~~~~
mysql: [ERROR] unknown variable 'no-tablespaces=True'.
~~~~

This is due to the change introduced in #2454.

The `no-tablespaces` parameter is not valid for *all* mysql client tools. We can fix this by changing in `pi-manage` the creation of the mysql client config:

~~~~python
def _write_mysql_defaults(filename, username, password):
    """
    Write the defaults_file for mysql commands

    :param filename: THe name of the file
    :param username: The username to connect to the database
    :param password: The password to connect to the database
    :return:
    """
    with open(filename, "w") as f:
        f.write("""[client]
user={0!s}
password={1!s}
[mysqldump]
no-tablespaces=True""".format(username, password))
~~~~

Having an explicit section for mysqldump ensures, that the no-tablespaces is only used for `mysqldump` but not for `mysql`.

cornelinux:
The normal mysql client must not use the parameter
no-tablespaces. It will bail out with an error and
the restore will not work.

Fix #2728

### DIFF
--- a/pi-manage
+++ b/pi-manage
@@ -347,9 +347,10 @@ def _write_mysql_defaults(filename, username, password):
     """
     with open(filename, "w") as f:
         f.write("""[client]
-no-tablespaces=True
 user={0!s}
-password={1!s}""".format(username, password))
+password={1!s}
+[mysqldump]
+no-tablespaces=True""".format(username, password))
 
     os.chmod(filename, 0o600)
     # set correct owner, if possible


pablo-knight (approved):
works

PR #2733, opened 2021-05-20
fix testconnection in ldap resolver with TLS 1.3

### LINKED ISSUE #2735: Testing LDAP connections does not honor the TLS setting
Testing an LDAP connection from the UI does not take the set TLS version into account. Thus testing the connection works but the resolver itself failed with a TLS error (in case the LDAP server doesn't speak TLSv1.3).

laclaro:
fixes testconnection with TLS 1.3

Fixes #2735


plettich (approved):
looks good

Issue #2477, opened 2020-10-30
Nightly SQL database tests on MySQL and others


cornelinux:
We should run our tests also with MySQL and Postgres.

This could be done in a daily/nightly run in CircleCI.

Tests have to be adapted, since currently some tests have hardcoded sqlite databases like some audit tests.

See #2435

plettich:
Running the tests with MySQL requires some rewrite.
I'll re-open this issue to track the fixes.

plettich:
Nightly tests with MySQL/Postgres are running just fine.
Possible TODOs:
- Oracle (Local installation)
- MariaDB

PR #2734, opened 2021-05-20
Enable tests to run with real db

plettich:
Improve tests to run with a real db (first tests with MySQL)

Working on #2477

### DIFF
--- a/tests/test_api_monitoring.py
+++ b/tests/test_api_monitoring.py
@@ -1,9 +1,10 @@
-import json
+# -*- coding: utf-8 -*-
+
 from .base import MyApiTestCase
 from privacyidea.lib.monitoringstats import write_stats
 from privacyidea.lib.tokenclass import AUTH_DATE_FORMAT
+from privacyidea.models import db
 import datetime
-from flask import current_app
 
 
 class APIMonitoringTestCase(MyApiTestCase):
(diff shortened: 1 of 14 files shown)


tests/test_api_monitoring.py
```diff
@@ -13,13 +14,13 @@ def test_01_get_stats(self):
         # create some statistics
 
         write_stats("key1", 1)
-        write_stats("key2", "A")
+        write_stats("key2", 50)
```
cornelinux, line 17: Why would you remove the string as a stats value?
There are stats, I think, that can be strings.

tests/test_scripts.py
```diff
@@ -12,7 +12,7 @@
 SCRIPTS = [
     'creategoogleauthenticator-file',
     'getgooglecodes',
-    'privacyidea-convert-base32.py',
+#    'privacyidea-convert-base32.py',
```
cornelinux, line 15: Why is that? Should we or do we need to remove this? Or did we remove this?

cornelinux (changes requested):
Some clarification needed.

tests/test_scripts.py
plettich: This is what breaks the current tests. This script requests some input from stdin which confuses the test-runner. So i just commented it out for now.

tests/test_api_monitoring.py
plettich: Actually `write_stats()` already annotates the value as `int`:
https://github.com/privacyidea/privacyidea/blob/315c30927382bab04bc3ea76321be7e418a2e3e5/privacyidea/lib/monitoringstats.py#L68
Also the db model uses Integers:
https://github.com/privacyidea/privacyidea/blob/cb4296718c92db9d2ab06bb45d95cb19405555e3/privacyidea/models.py#L3018
I guess SQLite isn't too picky about the data types...

cornelinux: Oups. I was not aware of this. Thx.

cornelinux (approved):

plettich:
great, now lets see what the CircleCI mail tells us tomorrow :-)

PR #2736, opened 2021-05-21
Initial writeup for native YubiHSM support

cornelinux:
Works fine by running the internal test script.
Seems to have some issues with encoding when running
within privacyIDEA.


cornelinux:
We should change this in this way, that only the encryption key on the hard disk gets decrypted by the HSM.
See #3003

PR #2739, opened 2021-05-25
Added a small fix for the log messages

### LINKED ISSUE #2714: Misleading log-message when admin login fails
When an admin login fails due to a missing `password` parameter or a wrong password, the result of the `/auth/` request is a correct `"Authentication failure. Wrong credentials"`. But the privacyIDEA log contains the warning `A user 'admin' exists as local admin and as user in your default realm!` which is wrong.
If a `realm` is missing in the request, we set it to the default realm:
https://github.com/privacyidea/privacyidea/blob/895c967935cccb111ce594063cc50cb2a69f1e56/privacyidea/api/auth.py#L236
The warning gets printed if the admin authentication fails and the realm is the default (which is always the case for an internal admin):
https://github.com/privacyidea/privacyidea/blob/895c967935cccb111ce594063cc50cb2a69f1e56/privacyidea/api/auth.py#L312

timosturm:
Closes #2714
Applied the fix suggested by @cornelinux. As far as I can tell, that is the only change that can be made there.

### DIFF
--- a/privacyidea/api/auth.py
+++ b/privacyidea/api/auth.py
@@ -309,9 +309,11 @@ def get_auth_token():
                                                     superuser_realms=
                                                     superuser_realms)
         details = details or {}
-        if db_admin_exist(loginname) and realm == get_default_realm():
+        if db_admin_exist(loginname) and user_auth and realm == get_default_realm():
             # If there is a local admin with the same login name as the user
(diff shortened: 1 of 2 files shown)


plettich (approved):
Looks good. I've also added a test for the log message.

PR #2740, opened 2021-05-25
Fix the ValidateError in case of poll-only

### LINKED ISSUE #2720: Push without firebase
In order to use push without firebase, the SMS-gateway definition must allow empty fields for the firebase values, leaving only the following attributes as required fields:

* registration URL 
* time-to-live

Moreover, the rollout QR code should then be generated without the firebase values.

Maybe we do not need the SMS provider at all and we can simply do *Poll-Only* if no PushProvider is defined.

cornelinux:
The push token will raise a ValidateError if
the "res" is false, which would usually indicate,
that there is an error when submitting to firebase.

But since we do not submit to firebase there will be no
error, so we default the "res" to True.

For the polling endpoint the administrator sill needs to
set the push_poll_allow policies.

Closes #2720

### DIFF
--- a/privacyidea/lib/tokens/pushtoken.py
+++ b/privacyidea/lib/tokens/pushtoken.py
@@ -821,7 +821,8 @@ class PushTokenClass(TokenClass):
 
         attributes = None
         data = None
-        res = False
+        # Initially we assume there is no error from Firebase
+        res = True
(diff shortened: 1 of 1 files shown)


timosturm (approved):

cornelinux:
This PR has been merged but for some strange reason, this was not marked as merged at github.

PR #2741, opened 2021-05-25
Token rollover

cornelinux:
This is an alternative implementation of the token rollover.
It uses the normal insite token enrollment and also
allows to change data (key length, otp length) if the administrator
allows so.

Still missing the webui policy.

Working on #2613

### DIFF
--- a/doc/policies/webui.rst
+++ b/doc/policies/webui.rst
@@ -306,7 +306,7 @@ token_rollover
 type: str
 
 This is a whitespace separated list of tokentypes, for which a rollover button is
-displayed in the token details and in the token list. This button will generate a
+displayed in the token details. This button will generate a
 new token secret for the displayed token.
 
 This e.g. enables a user to transfer a softtoken to a new device while keeping the
--- a/doc/policies/webui.rst
+++ b/doc/policies/webui.rst
@@ -300,6 +300,20 @@ being disabled.
 
 (Since privacyIDEA 3.0)
 
+token_rollover
+~~~~~~~~~~~~~~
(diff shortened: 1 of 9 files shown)


cornelinux:
@laclaro I would like to discuss this.

doc/policies/webui.rst
```diff
@@ -300,6 +300,20 @@ being disabled.
 
 (Since privacyIDEA 3.0)
 
+token_rollover
+~~~~~~~~~~~~~~
+
+type: str
+
+This is a whitespace separated list of tokentypes, for which a rollover button is
+displayed in the token details and in the token list. This button will generate a
```
laclaro, line 309: this PR leaves the token list untouched.

privacyidea/lib/policy.py
```diff
@@ -2381,6 +2382,13 @@ def get_static_policy_definitions(scope=None):
                 'desc': _("The tokenwizard will be displayed in the token "
                           "menu, even if the user already has a token.")
             },
+            ACTION.TOKENROLLOVER: {
+                'type': 'str',
+                'desc': _('This is a whitespace separated list of tokentypes, '
+                          'for which a rollover button is displayed in the token '
+                          'details and in the token list.'),
```
laclaro, line 2389: see above. Please remove token list.

laclaro (changes requested):
Please remove the hints to a button in the token list as we do not implement it with this PR.

laclaro (approved):
I quickly corrected the descriptions.

PR #2745, opened 2021-05-26
Fix URL for polling

cornelinux:
We need to take the URL from the policies and not
from the firebase config anymore.


timosturm (approved):
Looks good to me

Issue #2670, opened 2021-03-12
 Error during fetching LDAP objects: TypeError("a bytes-like object is required, not 'str'")


FloMeyer:
### Top-level intent

What did you try to achieve?

### Steps to reproduce

1. using default objectGUID of Active Directory as UID type

### Expected outcome

What do you think, how the system should have worked?
normal

### Actual outcome

What did actually happen?

 Error during fetching LDAP objects: TypeError("a bytes-like object is required, not 'str'")

But the objectGUID is not string..
![image](https://user-images.githubusercontent.com/2670188/111000733-e1126b80-8382-11eb-8949-0aa226bc0a9b.png)
![image](https://user-images.githubusercontent.com/2670188/111000758-ebcd0080-8382-11eb-9d17-1c646ea658d3.png)

If i set UID Type to DN, everything works, but i dont want to use DN as UID Type. ObjectGUID is fine.

### Configuration

* **privacyIDEA version**: 3.5.1
* **Installation method**: from Ubuntu packages
* **Python version**:  3.8
* **Operating system**: Ubuntu 20
* **Webserver**: apache2
* **Token database**: MySQL

### Log file

`

[2021-03-12 21:12:47,629][123262][140672039892736][ERROR][privacyidea.app:1891] Exception on /token/ [GET]
Traceback (most recent call last):
  File "/opt/privacyidea/lib/python3.8/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/opt/privacyidea/lib/python3.8/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/opt/privacyidea/lib/python3.8/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/opt/privacyidea/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/opt/privacyidea/lib/python3.8/site-packages/flask/app.py", line 1948, in full_dispatch_request
    rv = self.preprocess_request()
  File "/opt/privacyidea/lib/python3.8/site-packages/flask/app.py", line 2242, in preprocess_request
    rv = func()
  File "/opt/privacyidea/lib/python3.8/site-packages/privacyidea/api/auth.py", line 402, in decorated_function
    return f(*args, **kwargs)
[...]

PR #2746, opened 2021-05-26
Fix handling of byte-array

plettich:
In some weird cases ldap3 seems to return the userid attribute as a
byte-array representing the ObjectGUID. In that case we try to convert
it to a real UUID.

Working on #2670

### DIFF
--- a/privacyidea/lib/resolvers/LDAPIdResolver.py
+++ b/privacyidea/lib/resolvers/LDAPIdResolver.py
@@ -426,7 +426,18 @@ class IdResolver (UserIdResolver):
                 # In order to ensure backwards compatibility for user mappings,
                 # we strip the curly braces from objectGUID values.
                 # If we are using ldap3 <= 2.4.1, there are no curly braces and we leave the value unchanged.
-                uid = uid.strip("{").strip("}")
+                try:
+                    uid = convert_column_to_unicode(uid).strip("{").strip("}")
+                except UnicodeDecodeError as e:
+                    # in some weird cases we sometimes get a byte-array here
+                    # which resembles a uuid. So we just convert it to one...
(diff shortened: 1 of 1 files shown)


plettich:
@laclaro please check if this fixes the issue for You, i still couldn't reproduce it in my setup.

privacyidea/lib/resolvers/LDAPIdResolver.py
```diff
@@ -426,7 +426,18 @@ def _get_uid(entry, uidtype):
...
                 # If we are using ldap3 <= 2.4.1, there are no curly braces and we leave the value unchanged.
-                uid = uid.strip("{").strip("}")
+                try:
+                    uid = convert_column_to_unicode(uid).strip("{").strip("}")
+                except UnicodeDecodeError as e:
+                    # in some weird cases we sometimes get a byte-array here
+                    # which resembles a uuid. So we just convert it to one...
+                    # We might run into endian-issues here depending on what
+                    # ldap3/AD returns.
+                    log.debug('Found a byte-array as uid ({0!s}), trying to '
```
cornelinux, line 436: Should we increase the log level, since this is s.th. that *should* not happen. So that if it happens it is not hidden by the debug-log-level.

cornelinux (commented):
Looks good to me.
I can remember that we had issues in the past since the uid was not able to perform the strip.
So at least the *convert* should cover those cases and we hopefully never run into the exception.

privacyidea/lib/resolvers/LDAPIdResolver.py
plettich: Sounds good to me, will change to `warning`

cornelinux (approved):

PR #2751, opened 2021-05-27
Be more specific about versions of dependencies.

### LINKED ISSUE #2750: Mention pinned dependencies in the docs more clearly.
The pypi installation should mention, that using the latest versions of the dependencies could lead to untested behaviour.

The user should install the requirements.txt before.

Thus we could also remove the note about 3.3.3, since this should actually not be contained in the version for 3.6.

cornelinux:
Closes #2750


laclaro (approved):
Thank you!

PR #2578, opened 2021-01-14
Highlight search terms in policy action search

laclaro:
closes #2577

PR #2756, opened 2021-06-01
remove highlighting due to possible html injections

laclaro:
rollback changes from #2578. Although highlighting is a nice feature, we should reconsider to make it more secure and prevent interpretation of the full descriptive texts (which breaks in some cases).


cornelinux (approved):
Looks good to me.
Why did you create a draft?  Shouldn't we merge it, now?

Issue #2738, opened 2021-05-22
Optimize push token handling in case of multiple tokens


jmozd:
**Is your feature request related to a problem? Please describe.**
see https://community.privacyidea.org/t/firebase-access-suddenly-broken-partially/2018/6 : user had *two* push tokens registered in PI, with only one assigned to hosts by the PI administrator. When the user removed the "extra" token from the according PI App, the user could no longer use the other (assigned) token for 2FA (Firebase reported 404 for the removed token ID, which then made PI server bail out on the request). The situation with the extra token didn't cause harm for months, until the user removed the extra token from the personal device / app.

**Describe the solution you'd like**
If Firebase reports 404 when trying to trigger a push token, and PI has more than one push token issued by the user, retry the request with the next push token or, if already triggered (we could see that the "correct" push token was indeed triggered by PI), simply ignore the 404 and return "ok".

**Describe alternatives you've considered**
Only current chance to resolve the issue was to remove the extra token from PI. Since simple user action (removing a particular extra push token from the app) will break operations non-obviously, PI might at least warn the user when registering an extra push token - and notify the admin of such a situation or offer a "consistency check" for the PI system, that the admin can trigger on occasions where dubious errors occur.

**Additional context**
The user logged in to a system, for which per PI rules the user had a push token assigned to that host. Actually that push token was triggered by PI server (and the user acknowledged the request), but PI server very early bailed out with 400 code (the the PAM module on the client system) when PI tried to trigger the *extra* push token (as it no longer was in any App, because the user deleted it).
I've been told by PI team years ago that multiple push tokens for a user might not be a good idea and we've never got it to work. But obviously, simple user action can break operations and without additional debugging code in PI server I would have never found out that an extra push token was part of the picture.

cornelinux:
Could be related to #1668

jmozd:
Yes, this seems to be the issue here, too. I only learned late into the problem that the user had a second push token registered to the user id, and had it added to PI App on a *separate* phone. And the problem may well have come up once the device with the extra token was full-reset, or the app de-installed from that device (I don't have full details).

The extra push token was *not* added to any host in PI, it just "existed" in the list of tokens the user had issued, alongside the "main" push token that had been added to hosts.

jmozd:
should we take this to issue #1668 and close this enhancement request as "duplicate"?

PR #2769, opened 2021-06-04
Add log warning on challenge creation

### LINKED ISSUE #1668: Trigger challenge fails if android app is deinstalled
# What did you try to do?

* Enrolled push token with android app
* Removed android app (push token is still in privacyIDEA)
* validate/triggerchallenge

Trigger challenge shows my tokens and their messages

HTTP 400 BAD REQUEST
result -> error -> code: 401
result -> error -> message: "ERR401: Failed to submit message to firebase service."

# Configuration

* privacyIDEA Version: 3.0.1 (master branch)
* Installation method: venv

## more details: 

* OS: Ubuntu 19.04
* Webserver: Apache2
* Tokendatabase: MySQL
* Android: latest beta from play store (2.0.1?)

# Log file

The attached log file is the response after I did /validate/triggerchallange 

This repeats 134 times with different lines (see log file)
```
  File "/usr/lib/python2.7/logging/handlers.py", line 77, in emit
    self.doRollover()
  File "/usr/lib/python2.7/logging/handlers.py", line 142, in doRollover
    os.rename(self.baseFilename, dfn)
OSError: [Errno 13] Permission denied
Logged from file before_after.py, line 88
```
This is the last line
```
<my pi ip> - - [11/Jun/2019 12:39:29] "POST /validate/triggerchallenge HTTP/1.1" 400 -
```

[privacyidea.log](https://github.com/privacyidea/privacyidea/files/3276051/privacyidea.log)

timosturm:
Add a warning to the log if submitting the challenge to firebase fails.
Closes #1668 and #2738


privacyidea/lib/tokens/pushtoken.py
```diff
@@ -860,9 +860,11 @@ def create_challenge(self, transactionid=None, options=None):
                 db_challenge.save()
                 self.challenge_janitor()
 
-            # If sending the Push message failed, we still raise an error.
+            # If sending the Push message failed, we still raise an error and a warning.
             if not res:
-                raise ValidateError("Failed to submit message to firebase service.")
+                log.warning("Failed to submit message to firebase service for token {0!s}."
+                            .format(self.token.serial))
+                raise ValueError("Failed to submit message to firebase service.")
```
cornelinux, line 867: Why did you change this to ValueError?

cornelinux (changes requested):
Please check, why the test is failing.

privacyidea/lib/tokens/pushtoken.py
timosturm: This really shouldn't be here because I already caught that one. I will fix it immediately.

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -860,8 +860,10 @@ def create_challenge(self, transactionid=None, options=None):
                 db_challenge.save()
                 self.challenge_janitor()
 
-            # If sending the Push message failed, we still raise an error.
+            # If sending the Push message failed, we still raise an error and a warning.
             if not res:
+                log.warning("Failed to submit message to firebase service for token {0!s}."
+                            .format(self.token.serial))
```
cornelinux, line 866: I am sorry for my late comment.
I was not quite sure about this.

But in theory a serial number could contain a special character or a German umlaut. If we print it in the log message this way, we will raise an exception.
So we should do

    log.warnung(u"Failed to .....".format())

To allow logging unicode.

Thanks a lot

cornelinux (changes requested):
Sorry. One last comment

privacyidea/lib/tokens/pushtoken.py
timosturm: This should be fixed now.

cornelinux (approved):

PR #2771, opened 2021-06-08
Add missing Sequence declarations to migration script models.

### LINKED ISSUE #2770: Migration from LinOTP fails on NOT NULL constraint on `token.id`
Attempting to use the privacyidea-migrate-linotp.py script with postgresql+psycopg2 for both the databases:

```
Adding 40 tokens...
Insert token records 0 to 39 ...
Failed to insert chunk: (psycopg2.errors.NotNullViolation) null value in column "id" violates not-null constraint
DETAIL:  Failing row contains (...)

...

(Background on this error at: http://sqlalche.me/e/13/gkpj)
  File "./privacyidea-migrate-linotp.py", line 579, in <module>
    main()
  File "./privacyidea-migrate-linotp.py", line 569, in main
    migrate(config_obj)
  File "./privacyidea-migrate-linotp.py", line 471, in migrate
    token_values[i]["id"] = token_serial_id_map[token_values[i]["serial"]]
KeyError: '...'
```

Looking in the new PrivacyIDEA database (created by `pi-manage createdb`), the table `token` has an `id` column which is set NOT NULL, but has no default value.  The migration script is trying to insert rows into this table, but is not providing a value for `id` fields.

If the `id` column is supposed to contain a synthetic primary key, it should probably be created with the type `serial` or `bigserial` in Postgres.

It turns out that the database does have sequences defined for these `id` columns, but they are not set as the column DEFAULT in the database.  To fix this, you can do the following for each table:

```
ALTER TABLE token ALTER COLUMN id SET DEFAULT nextval('token_seq');
ALTER SEQUENCE token_seq OWNED BY token.id;
```

I went through and set up the column defaults for `token`, `tokenrealm`, `tokenowner` and `tokeninfo`, and then the migration script was able to complete successfully.

direvus:
This brings the model definitionss in privacyidea-migrate-linotp.py into
line with the models in the main application, and also solves a bug with
database migration on Postgres.

Fixes #2770.

### DIFF
--- a/tools/privacyidea-migrate-linotp.py
+++ b/tools/privacyidea-migrate-linotp.py
@@ -241,7 +241,7 @@ def migrate(config_obj):
 
     resolver_config_table = Table("resolverconfig", metadata,
                                   Column("id", Integer,
-                                         Sequence('resolverconfig_seq'),
+                                         Sequence('resolverconf_seq'),
                                          primary_key=True),
                                   Column("resolver_id", Integer),
                                   Column("Key", Unicode(255), default=u""),
--- a/tools/privacyidea-migrate-linotp.py
+++ b/tools/privacyidea-migrate-linotp.py
@@ -52,7 +52,7 @@ migrated tokens.
 from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker
 from sqlalchemy.sql import select
-from sqlalchemy.schema import ForeignKey
+from sqlalchemy.schema import ForeignKey, Sequence
 from sqlalchemy import (Table, MetaData, Column, Integer, Unicode, Boolean,
                         UnicodeText)
 import sys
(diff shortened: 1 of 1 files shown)


tools/privacyidea-migrate-linotp.py
```diff
@@ -201,39 +202,47 @@ def migrate(config_obj):
...
-                           Column("id", Integer, primary_key=True),
+                           Column("id", Integer, Sequence('resolver_seq'),
+                                  primary_key=True),
                            Column("name", Unicode(255), default=u""),
                            Column("rtype", Unicode(255), default=u""))
 
     resolver_config_table = Table("resolverconfig", metadata,
-                                  Column("id", Integer, primary_key=True),
+                                  Column("id", Integer,
+                                         Sequence('resolverconfig_seq'),
```
plettich, line 244: Please change this to "resolverconf_seq", unfortunately this sequence name does not match the table name:
https://github.com/privacyidea/privacyidea/blob/ee2951c18607d82a77461abecd3807d77adc46a4/privacyidea/models.py#L1035

plettich (changes requested):
Looks good so far.

tools/privacyidea-migrate-linotp.py
direvus: Thanks, good pickup.  I have updated the PR.

plettich (approved):

PR #2776, opened 2021-06-10
allow placeholders in eventhandler usernotification email reply-to

NNLemling:
these options are now available

* tokenowner
* logged_in_user
* internal admin -> (choose names of internal admins)
* admin realm -> (choose admin realm)
* email -> (choose static email address)

### DIFF
--- a/migrations/versions/ff26585932ec_.py
+++ b/migrations/versions/ff26585932ec_.py
@@ -1,0 +1,39 @@
+"""v3.7: migrate old reply_to email
+
+Revision ID: ff26585932ec
+Revises: fa07bd604a75
+Create Date: 2021-07-15 14:17:17.624748
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = 'ff26585932ec'
+down_revision = 'fa07bd604a75'
+
+from alembic import op
+from sqlalchemy import orm
+from privacyidea.models import EventHandlerOption
+from privacyidea.lib.eventhandler.usernotification import NOTIFY_TYPE
+
+
+def upgrade():
+    bind = op.get_bind()
+    session = orm.Session(bind=bind)
+    try:
+        for row in session.query(EventHandlerOption).filter(EventHandlerOption.Key == 'reply_to'):
+            reply_email = row.Value
+            evh_id = row.eventhandler_id
+            row.Value = NOTIFY_TYPE.EMAIL
+            row.save()
+            EventHandlerOption(evh_id, 'reply_to email', reply_email).save()
+
+    except Exception as e:
+        session.rollback()
+        print(e)
+
+    finally:
+        session.commit()
+
+
+def downgrade():
+    pass
--- a/migrations/versions/ff26585932ec_.py
+++ b/migrations/versions/ff26585932ec_.py
@@ -14,6 +14,7 @@ from alembic import op
 from sqlalchemy import orm
 from privacyidea.models import EventHandlerOption
 from privacyidea.lib.eventhandler.usernotification import NOTIFY_TYPE
+from sqlalchemy import and_
 
 
 def upgrade():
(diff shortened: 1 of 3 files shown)


cornelinux:
@NNLemling can you please fix the pep8 comments.

privacyidea/lib/eventhandler/usernotification.py
```diff
@@ -224,6 +248,7 @@ def do(self, action, options=None):
         handler_def = options.get("handler_def")
         handler_options = handler_def.get("options", {})
         notify_type = handler_options.get("To", NOTIFY_TYPE.TOKENOWNER)
+        reply_to_type = handler_options.get("reply_to", NOTIFY_TYPE.TOKENOWNER)
```
cornelinux, line 251: Old installations do not have the "type" in ` reply_to`  but the destinct email address!

privacyidea/lib/eventhandler/usernotification.py
```diff
@@ -140,7 +140,31 @@ def actions(cls):
                 "reply_to": {
                     "type": "str",
                     "required": False,
-                    "description": _("The Reply-To header in the sent email.")},
+                    "description": _("The Reply-To header in the sent email."),
```
cornelinux, line 143: We have a problem here with old installations.

Installations before version 3.6 will have the dedicated email address in the `reply_to` database field.

~~~~
mysql> select * from eventhandleroption where eventhandleroption.Key = "reply_to";
+----+-----------------+----------+-----------------------+------+-------------+
| id | eventhandler_id | Key      | Value                 | Type | Description |
+----+-----------------+----------+-----------------------+------+-------------+
| 10 |               4 | reply_to | user@example.com |      |             |
+----+-----------------+----------+-----------------------+------+-------------+
1 row in set (0,00 sec)
~~~~

(See my comment below).

But we need to take care of this, that old configurations still work with version 3.6.
So basically there are two possibilities:

# A) Migrate the data

You need to migrate the evenhandleroption 

    reply_to -> reply_toEmail
    set reply_to = "Email"

# B) Name the fields this way, that it is backward compatible

* Introduce a new field "reply_to_type"
* Use "reply_to" for destinct email addresses
* And instead of having the field `"reply_to " + NOTIFY_TYPE.EMAIL` use normal `reply_to`, since this is the old field, that contains the email address.

cornelinux (changes requested):
You need to work on the update path from 3.5 to 3.6.

NNLemling:
I had discussed the two proposed solutions again with @plettich.

We have come to the conclusion that it is better to write a migration script for it.

We're going to do this together this week.

migrations/versions/ff26585932ec_.py
```diff
@@ -0,0 +1,39 @@
...
+from privacyidea.models import EventHandlerOption
+from privacyidea.lib.eventhandler.usernotification import NOTIFY_TYPE
+
+
+def upgrade():
+    bind = op.get_bind()
+    session = orm.Session(bind=bind)
+    try:
+        for row in session.query(EventHandlerOption).filter(EventHandlerOption.Key == 'reply_to'):
+            reply_email = row.Value
```
cornelinux, line 24: Can we please add a failsafe here to check, that the data is not migrated yet?
Otherwise we might really run into trouble if for any reason this script runs twice (since the DB is not stamped) and then a new configuration might be overwritten.

cornelinux (approved):
This looks good to me.

@plettich @NNLemling 
However, I have the impression, that we should (one day) somehow wrap the logic of identifiying the *type* and the actual *email address* in a separated function.
The problem is, that we have the same logic for *notification* and for *reply_to*, but it is two separated code sections. This means we could end up with a bug in one occasion. The user would expect both (notifcation and reply_to) to work the same, but it would not, since on code section would contain a bug, which the other code section does not.

Issue #2778, opened 2021-06-11
admin user may manage tokens of not-allowed-resolver-users


laclaro:
# Scenario
- There is one `all-users` realm with multiple user-resolvers, say users1 and users2.
- An admin policy which has conditions on realm `all-users` and resolver `users1`.

# Outcome
- admin users cannot assign tokens to users from resolvers other than `users1`
- admin users can enable/disable, delete or unassign tokens of users from resolver `users2`

# Expected behavior
The admin policy conditioned on the resolver should prevent all actions for tokens assigned to users coming from resolvers other than `users1`.

Those API actions are done via serial, so privacyIDEA does apparently not take the tokenowner into account properly.

laclaro:
PR https://github.com/privacyidea/privacyidea/pull/2779 adds a resolver check for admin action.

There could be a restriction for 
- user listing
- token listing
with a given user from a non-allowed resolver

cornelinux:
Depends on #2810

laclaro:
If a user of an admin acts on a token determined by its serial, we should generally determine the tokenowner

cornelinux:
You mean, if "an admin acts on a token"... ?

cornelinux:
In #2810 I add the resolving of the user, if only a token serial is given.
I.e. the request is enriched with the user object, who is the tokenowner of the token(serial).

This way all policies will match for such user - even if the admin only acts on the token.
Please check, I think this will also close this very issue.

PR #2779, opened 2021-06-11
add resolver check based on serial to the base action check

laclaro:
working on #2778

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -1161,6 +1161,16 @@ def check_base_action(request=None, action=None, anonymous=False):
             realm = get_realms_of_token(request.view_args.get("serial"),
                                         only_first_realm=True)
 
+        # if the resolver was not given in the request,
+        # get the resolver by the serial either from request or from view_args
+        if not resolver:
+            if params.get("serial"):
+                user_object = get_token_owner(params.get("serial"))
+                resolver = user_object.resolver if user_object.resolver else None
+            elif request.view_args and request.view_args.get("serial"):
+                user_object = get_token_owner(request.view_args.get("serial"))
+                resolver = user_object.resolver if user_object.resolver else None
+
     # In this case we do not pass the user_object, since the realm is also determined
     # by the pure serial number given.
     action_allowed = Match.generic(g, scope=role,


privacyidea/api/lib/prepolicy.py
```diff
@@ -1161,6 +1161,16 @@ def check_base_action(request=None, action=None, anonymous=False):
             realm = get_realms_of_token(request.view_args.get("serial"),
                                         only_first_realm=True)
 
+        # if the resolver was not given in the request,
+        # get the resolver by the serial either from request or from view_args
+        if not resolver:
```
cornelinux, line 1166: I think the resolver is already available from 1138.

Or - We could make the resolver present here by adding the resolver as a return value in `determine_logged_in_userparams` in line 1139.

privacyidea/api/lib/prepolicy.py
```diff
@@ -1161,6 +1161,16 @@ def check_base_action(request=None, action=None, anonymous=False):
...
+            if params.get("serial"):
+                user_object = get_token_owner(params.get("serial"))
+                resolver = user_object.resolver if user_object.resolver else None
```
cornelinux, line 1169: Fetching the user object from a token changes the logic in a way that we do not do.

Why would this be?

privacyidea/api/lib/prepolicy.py
```diff
@@ -1161,6 +1161,16 @@ def check_base_action(request=None, action=None, anonymous=False):
...
+            elif request.view_args and request.view_args.get("serial"):
```
cornelinux, line 1170: params contains all_data, so it would also contain the view_args. This branch of the if-statement would never be touched to my understanding.

cornelinux (changes requested):
I am not sure if this is the right approach.

This would also require tests to verify the actualy behaviour

cornelinux:
@laclaro I think what you are basically saying is, that if an admin acts on a tokenserial, we should determine the tokenowner, so that we can check for user conditions in policies. This is not only relevant for the resolver, but also for the username and the realm.

So maybe we should use a more generic approach and enhance `determine_logged_in_userparams` in this way, that it would also return the acted on user, if the params only contain a serial/token.

privacyidea/api/lib/prepolicy.py
cornelinux: OK, I understand the serial being part of the URL.

cornelinux:
@laclaro I think this might already be solved in the PR #2823

Issue #2784, opened 2021-06-17
handle token enrollments with client_wait, what to do with pending enrollments


cornelinux:
We need to identify tokens that can be in a not finally enrolled state.

This can be 

* push tokens (maybe we need to improve the docs of the TTL, or we can totally remove the TTL and set the default to 2minutes)
* 2step enrollment
* webauthn
* ...

We need to define, what we do with such pending tokens. (see pending certifiate requests?)

What will be do with a push token, that is not scanned and enrolled by the smartphone app in a certain time?

Can we simply do a documentation, where we describe how such tokens can be handeled with the tokenjanitor?
Should we involve the periodic tasks?

We also need to check, that when the token is enrolled finally, that th enrollment state is set correctly and does *not* remain in `client_wait`. (see #2763 )

Note: The TTL in the push token was originially ment to define, for how long the smartphone app should *try* to reach the privacyIDEA server during enrollment, to avoid endless loops. If on the server side the half enrolled push token is not removed, the QR code could be scanned again and enrolled later with another smartphone (if the first enrollment is not finished).
This can be also used as an interesting feature.

# Analysis

The token attribute (DB) `rollout_state` is used in the

* `pushtoken.py` for enrollment and waiting for the 2nd part of the smartphone.
* base class `tokenclass.py`  (for 2step enrollment of HOTP and TOTP)

(The TiQR token is actually enrolled in one step)

Possible settings of the rollout_state are currently:

* "clientwait" for 2step enrollment of HOTP and TOTP waiting for the user input and PUSH token waiting for the smartphone response
* "enrolled" only used in Push token, to mark that the token has been enrolled.

Where is the rollout_state already used?

* Event Handlers can already have a condition on the rollout_state.
* Policies can use the token attributes condition, since the rollout_state is a direct token database column.

# Todo

What should we enhance?

* [x] the tokenjanitor should be able to find tokens on the rollout state. Probably we should be able to combine rollout_state and other conditions in the token janitor.
* ~~[ ] ensure that no enrolled token remains in the state `client_wait`.~~
* [x] check with u2f and webauthn: These tokens use internally `self.init_step = [1|2]`. This is simply depending on the the token/init parameter passed. We should add a `rollout_state == clientwait` to the token in the database to indicate also in the database, that the token is not finally enrolled.

# Thoughts about enrollment states

Possible states can be defined as constants in an enrollment-state-classe:

[...]

PR #2789, opened 2021-06-20
Add filter for token attribute in token janitor

cornelinux:
We can now use the tokenattribute filter in the
token-janitor. It matches the columns of the Token
table directly.
The available columns are read from the database definition.

The filter creation method for tokeninfo is used.

This allows us to filter for the column "rollout_state".

Working on #2784

### DIFF
--- a/tools/privacyidea-token-janitor
+++ b/tools/privacyidea-token-janitor
@@ -51,6 +51,7 @@ from privacyidea.lib.importotp import export_pskc
 from privacyidea.lib.token import (get_tokens, remove_token, enable_token,
                                    unassign_token, import_token,
                                    get_tokens_paginated_generator)
+from privacyidea.models import Token
 from privacyidea.app import create_app
 from flask_script import Manager
 import re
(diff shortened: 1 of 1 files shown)


Sebastian-Lemling (commented):
It works well for me.

but in some cases, for example with "ID" or "maxfail", a string is expected but an integer is passed. This leads to an error message that should be intercepted if necessary.

Perhaps this error should also be corrected, because it is not possible to filter for attributes where an integer is passed.

cornelinux:
The problem is, I did this like with the tokeninfo. But the tokeninfo is alway a string. So the equal comparison with strings in tokeninf, that contain a number, works out fine with the string comparison. However, the integer fields in the tokenattributes do not.
I will add an extra comparison.

cornelinux:
@NNLemling I fixed it this way, that you can now run this:

Match exact integer:

    privacyidea-token-janitor find --tokenattribute id --tokenattribute-value 10

Match a regular expression string:

    privacyidea-token-janitor find --tokenattribute serial --tokenattribute-value 'TOTP.*'

NNLemling:
I just tested it and it works perfectly!

Sebastian-Lemling (approved):

PR #2791, opened 2021-06-21
Use clientwait with U2F and WebAuthn

cornelinux:
We also use the rollout_state "clientwait" for
webauthn and u2f tokens. This way the token janitor
can easily find tokens, that have not been successfully/finally
enrolled.

The logic has changed slightly, so that some testcases would
not run correctly anymore, seemingly the lib-testcases do
not reflect the real world use case in this scenario.

This is why we need to failsafe the name of the
token in the description.

Added a class ROLLOUTSTATE for future use and
to avoid typos in the state names.

Working on #2784

### DIFF
--- a/privacyidea/lib/tokenclass.py
+++ b/privacyidea/lib/tokenclass.py
@@ -134,7 +134,7 @@ class TOKENMODE(object):
 
 class ROLLOUTSTATE(object):
     CLIENTWAIT = 'clientwait'
-    PENDING = 'pending' # TODO: to be used, if the server needs to take action
+    PENDING = 'pending'  # TODO: to be used, if the server needs to take action
 
 
 class TokenClass(object):
--- a/privacyidea/lib/tokenclass.py
+++ b/privacyidea/lib/tokenclass.py
@@ -132,6 +132,11 @@ class TOKENMODE(object):
     OUTOFBAND = 'outofband'
 
 
+class ROLLOUTSTATE(object):
+    CLIENTWAIT = 'clientwait'
+    PENDING = 'pending' # TODO: to be used, if the server needs to take action
+
+
 class TokenClass(object):
 
     # Class properties
@@ -517,7 +522,7 @@ class TokenClass(object):
             otpKey = self.decode_otpkey(otpKey, otpkeyformat)
 
         if twostep_init:
-            if self.token.rollout_state == "clientwait":
+            if self.token.rollout_state == ROLLOUTSTATE.CLIENTWAIT:
                 # We do not do 2stepinit in the second step
                 raise ParameterError("2stepinit is only to be used in the "
                                      "first initialization step.")
@@ -543,7 +548,7 @@ class TokenClass(object):
             otpKey = getParam(param, "otpkey", required)
 
         if otpKey is not None:
-            if self.token.rollout_state == "clientwait":
+            if self.token.rollout_state == ROLLOUTSTATE.CLIENTWAIT:
                 # If we have otpkey and the token is in the enrollment-state
                 # generate the new key
                 server_component = to_unicode(self.token.get_otpkey().getKey())
@@ -558,7 +563,7 @@ class TokenClass(object):
 
         if twostep_init:
             # After the key is generated, we set "waiting for the client".
-            self.token.rollout_state = "clientwait"
+            self.token.rollout_state = ROLLOUTSTATE.CLIENTWAIT
 
         pin = getParam(param, "pin", optional)
         if pin is not None:
--- a/privacyidea/lib/tokenclass.py
+++ b/privacyidea/lib/tokenclass.py
@@ -134,7 +134,7 @@ class TOKENMODE(object):
 
 class ROLLOUTSTATE(object):
     CLIENTWAIT = 'clientwait'
-    PENDING = 'pending'  # TODO: to be used, if the server needs to take action
+    PENDING = 'pending'
 
 
 class TokenClass(object):
--- a/privacyidea/lib/tokens/u2ftoken.py
+++ b/privacyidea/lib/tokens/u2ftoken.py
@@ -29,7 +29,7 @@
 #
 from privacyidea.api.lib.utils import getParam, attestation_certificate_allowed
 from privacyidea.lib.config import get_from_config
-from privacyidea.lib.tokenclass import TokenClass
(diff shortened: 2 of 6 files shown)


privacyidea/lib/tokenclass.py
```diff
@@ -132,6 +132,11 @@ class TOKENMODE(object):
     OUTOFBAND = 'outofband'
 
 
+class ROLLOUTSTATE(object):
+    CLIENTWAIT = 'clientwait'
+    PENDING = 'pending'  # TODO: to be used, if the server needs to take action
```
plettich, line 137: Why is there a `TODO`? because it is currently not used? But this type will stay or will it change in the future?

privacyidea/lib/tokens/u2ftoken.py
```diff
@@ -323,13 +322,17 @@ def update(self, param, reset_failcount=True):
         :return: None
         """
         TokenClass.update(self, param)
-        description = "U2F initialization"
+        automatic_description = "U2F initialization"
```
plettich, line 325: i think we don't need a default here, `automatic_description` will always be overridden by `parse_registration_data()` (wither with `""` or with the CN).

privacyidea/lib/tokenclass.py
plettich: And what about a `finished` state (instead of `""`)? or is that too much clutter in the token class?

privacyidea/lib/tokens/webauthntoken.py
```diff
@@ -882,19 +884,23 @@ def update(self, param, reset_failcount=True):
...
             # Delete all challenges. We are still in enrollment, so there
             # *should* be only one, but it can't hurt to be thorough here.
             for challengeobject in challengeobject_list:
                 challengeobject.delete()
             self.challenge_janitor()
+            # Reset clientwait rollout_state
+            self.token.rollout_state = ""
         else:
-            self.set_description("WebAuthn initialization")
+            raise ParameterError("regdata and or clientdata provided but token not in clientwait rollout_state.")
```
plettich, line 903: like above please add the token serial

privacyidea/lib/tokens/webauthntoken.py
```diff
@@ -798,13 +797,16 @@ def update(self, param, reset_failcount=True):
         transaction_id = getParam(param, "transaction_id", optional)
         reg_data = getParam(param, "regdata", optional)
         client_data = getParam(param, "clientdata", optional)
-
-        if reg_data and client_data:
-            self.init_step = 2
-
+        automatic_description = DEFAULT_DESCRIPTION
```
plettich, line 800: same as above, `automatic_description` will be overridden in line 888

privacyidea/lib/tokens/webauthntoken.py
plettich: It could also be that one of `regdata` or `clientdata` is missing.

plettich (changes requested):

privacyidea/lib/tokenclass.py
cornelinux: I was aslo thinking about a state "enrolled". 
But then we could have tokens, that are "enrolled", since they are in the `rollout_state` "enrolled", but there are also tokens that are "enrolled", since these token types never used the `rollout_state`.
So I did not want to overcomplicate it is the first step.

But I will remove the "TODO".

privacyidea/lib/tokens/u2ftoken.py
```diff
@@ -348,18 +351,23 @@ def update(self, param, reset_failcount=True):
...
-        description = getParam(param, "description", default=description)
-        self.set_description(description)
+            # Reset rollout state
+            self.token.rollout_state = ""
+            # If no description has already been set, set the automatic description or the
+            # description given in the 2nd request
+            if not self.token.description:
+                self.set_description(getParam(param, "description", default=automatic_description))
+        else:
+            raise ParameterError("regdata provided but token not in clientwait rollout_state.")
```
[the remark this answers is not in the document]
cornelinux: will do

privacyidea/lib/tokens/u2ftoken.py
cornelinux: You mean from the point of the program flow?
I will check this.

privacyidea/lib/tokens/webauthntoken.py
cornelinux: will do

privacyidea/lib/tokens/webauthntoken.py
cornelinux: Will check and do.

privacyidea/lib/tokens/u2ftoken.py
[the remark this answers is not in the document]
cornelinux: Hm, checked it. 
We never do this. The ParameterError is raised and displayed to the user.  This occurs during the enrollment. So imho there is in fact no need for the serial.
This message will appear in the audit log, But there the serial is contained in the audit table.

privacyidea/lib/tokens/webauthntoken.py
cornelinux: see above. Since it is an exception, I do not add this.

The text tries to reflect this by useind "and or".

privacyidea/lib/tokens/webauthntoken.py
cornelinux: This will only be overwritten in line 888, 

    if web_authn_credential.attestation_cert

exist. If line 870 is false, it would be missing.
So I have to keep it here.

privacyidea/lib/tokenclass.py
cornelinux: Done in 0780585ec89240821dd72594f9aef4345fc4600c

privacyidea/lib/tokens/u2ftoken.py
cornelinux: Done in bb978a447947f51dc645c265c2f8f229ca60ee5a

privacyidea/lib/tokens/webauthntoken.py
plettich: You are right, missed that.

plettich (approved):

PR #2804, opened 2021-07-05
Decode urlencoded parameters

### LINKED ISSUE #2800: Content-Type x-www-form-urlencoded for POST not considered
The body of a POST request which has the Content-Type application/x-www-form-urlencoded is url encoded. It seems that it is not considered by the server because the body is not decoded and therefore the requests result in failure.

cornelinux:
In application/x-www-form-urlencoded params we need to urldecode
(unquote) the parameters.

Closes #2800

### DIFF
--- a/privacyidea/api/lib/utils.py
+++ b/privacyidea/api/lib/utils.py
@@ -37,6 +37,7 @@ import jwt
 import threading
 import six
 import re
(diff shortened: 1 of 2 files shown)


plettich (approved):
Looks good

PR #2807, opened 2021-07-06
fix doc and add webUI hint for custom user attributes

laclaro:
small change which adds a descriptive hint to the custom user attributes field.

In the documentation the typesetting of the example for the custom user attributes policy is fixed.


cornelinux (approved):
thanks :-)
I take the liberty to merge this into the branch-3.6, so we have this fix in 3.6.1.

PR #2808, opened 2021-07-07
Removed the versioned setting of CSS customization

geekmichael:
Removed the 'versioned' setting as it introduced an error of "No such file" if the CSS customization is enabled in pi.cfg


geekmichael:
When the pi.cfg has the customization directory configured and PI_CUSTOM_CSS is enabled, PrivacyIDEA will look for the file /usr/local/lib/python3.8/dist-packages/privacyidea/mydirectory/css/custom.css, which introduces an error:
flaskext.versioned.VersionedError: no such file: /usr/local/lib/python3.8/dist-packages/privacyidea/mydirectory/css/custom.css

plettich (approved):
I can confirm this behavior. Because the URL does not match the path of the file (i.e. due to an alias setting in the apache config) the Flask-Versioned plugin throws this error.
We could fix this by implementing a different Flask-Versioned driver which doesn't check the paths for existence.
Or we could just promote the `PI_CSS` setting:
https://privacyidea.readthedocs.io/en/latest/faq/customization.html?highlight=pi_css#providing-your-own-stylesheet-in-the-config-file

PR #2814, opened 2021-07-08
Add a simple package-management for the webUI

### LINKED ISSUE #1886: Add package management for the webui
We currently have no package management for modules used in the webui. Because of this, we tend to use older versions of modules and we have no central way of managing the dependencies of the client.

It would be nice to manage the client-side dependencies using npm. This would allow us the centrally manage out dependencies, however an it would add an extra step and extra complexity to the build process. We also have to make sure, that there can be no issues due to the availability of the npm-service.

To ensure the reliability of our productive installations, we should download all required packages as part of the build process and include them in our packages, this way we can use npm to manage our dependencies while maintaining full control of the application we actually ship.

Since we are also considering replacing AngularJS, it could be sensible to do this as part of a larger rewrite of the webui.

plettich:
In order to ease the management of contrib JS packages for the webUI,
a simple make target `update-contrib` is added to install the required
packages via npm. A helper script then copies the necessary files into
the privacyIDEA source tree.

closes #1886

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,14 +46,17 @@ pypi:
 	gpg --detach-sign -a --default-key ${SIGNING_KEY} dist/*.tar.gz
 	twine upload dist/*.tar.gz dist/*.tar.gz.asc
 
-
-depdoc:
-	#sfood privacyidea | sfood-graph | dot -Tpng -o graph.png	
-	dot -Tpng dependencies.dot -o dependencies.png
-
 doc-man:
 	(cd doc; make man)
 
 doc-html:
 	(cd doc; make html)
 
+NPM_VERSION := $(shell npm --version 2>/dev/null)
+
+update-contrib:
+ifdef NPM_VERSION
+	(cd privacyidea/static && npm install && ./update_contrib.sh)
+else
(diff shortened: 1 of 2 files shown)


cornelinux (approved):
I never installed angular and bootstrap via npm but downloaded the ready made packages, so first I was a bit confused why angular, jquery and others would be updated via npm.

Is there any drawback? What do you think?
Otherwise it looks good to me and if this does not bring anything else up in your mind, you can please merge this.

plettich:
> Is there any drawback? What do you think?

This is the current way how JS-Applications are distributed. The `package.json` and `package-lock.json` files in `privacyidea/static/` define the intended environment (aka requirements). With `npm install` all the required packages are installed. The script then copies the necessary files from the `node_modules` directory into the `contrib` directory in our source tree where they can be added and committed afterwards.

Updating the packages to a newer version is **not** included here. I use [`npm-check`](https://www.npmjs.com/package/npm-check) to get information about updates.

> Otherwise it looks good to me and if this does not bring anything else up in your mind, you can please merge this.

I'll add some notes about the process in the GH Wiki so we have some documentation.

PR #2816, opened 2021-07-09
Add a tokeninfo_ui_exclude policy-action

theodor-franke:
This PR adds the ability to create a policy containing a comma-separated-list that specifies fields that are excluded in the web-ui.

To test this you can add a policy under webui>misc>tokeninfo_ui_exclude with for example "tokenkind" and this info should not be displayed in the ui anymore.


privacyidea/lib/policy.py
```diff
@@ -314,6 +314,7 @@ class ACTION(object):
     TOKENREALMS = "tokenrealms"
     TOKENTYPE = "tokentype"
     TOKENINFO = "tokeninfo"
+    TOKENINFO_UI_EXCLUDE = "tokeninfo_ui_exclude"
```
cornelinux, line 317: The naming is not good, you are removing the tokeninfo on the API level, not only the UI.

privacyidea/lib/policy.py
```diff
@@ -2444,6 +2445,10 @@ def get_static_policy_definitions(scope=None):
                 'desc': _("This action adds a QR code in the enrollment page for "
                           "HOTP, TOTP and Push tokens, that lead to this given URL."),
                 'group': 'QR Codes'
+            },
+            ACTION.TOKENINFO_UI_EXCLUDE: {
```
cornelinux, line 2449: You are putting this in the scope "webui", although the information is removed from the API level. I would choose another scope, like admin or user. Since it is somehow an access right to see certain information.

cornelinux (changes requested):
Thank you for your PR. This is interesting.
I assume you want to hide some token information from the users.
Would you mind opening an issue for this and referring the issue in the PR.

I would like to discuss some aspects, if this is only a UI behaviour or if this is indeed an access right. Then a normal user might not be allowed to see certain tokeninfo fields, while an administrator might be allowed to do so.
If this is the case, we could think about adding this near to the right to set tokeninfo.

theodor-franke:
I created an Issue #2819

privacyidea/lib/policy.py
theodor-franke: i can rename it to `TOKENINFO_API_EXCLUDE` do you think this is a good name?

privacyidea/lib/policy.py
theodor-franke: Okay I will move it to the scope `user` my thought was that it has an effect over the UI so I included it under `webui`

privacyidea/lib/policy.py
```diff
@@ -314,6 +314,7 @@ class ACTION(object):
     TOKENREALMS = "tokenrealms"
     TOKENTYPE = "tokentype"
     TOKENINFO = "tokeninfo"
+    TOKENINFO_API_EXCLUDE = "tokeninfo_api_exclude"
```
laclaro, line 317: Please rename it to

HIDE_TOKENINFO = "hide_tokeninfo"

We have a similar policy "hide_audit_columns".

privacyidea/lib/policy.py
```diff
@@ -2016,7 +2017,11 @@ def get_static_policy_definitions(scope=None):
                           "attribute, set this to '*'. Use '*' with CAUTION! "
                           "For more details, check the documentation."),
                 'mainmenu': [],
-                'group': GROUP.USER}
+                'group': GROUP.USER},
+            ACTION.TOKENINFO_API_EXCLUDE: {
```
laclaro, line 2021: Please add this action to SCOPE.USER and SCOPE.ADMIN, and rename it, analog to `HIDE_AUDIT_COLUMNS` to `HIDE_TOKENINFO`, see:

https://github.com/privacyidea/privacyidea/blob/a6c8ce47fdf8746d25f2e4f3c2a9b7478043f820/privacyidea/lib/policy.py#L1749

privacyidea/lib/policy.py
```diff
@@ -2016,7 +2017,11 @@ def get_static_policy_definitions(scope=None):
...
+                'type': TYPE.STRING,
+                'desc': 'A comma-seperated list of token-info fields that get excluded in the web-ui'
```
laclaro, line 2023: Tokeninfos with multiple values are whitespace-separated. See https://github.com/privacyidea/privacyidea/pull/2500/files. If you set unique=False in the Math above, then you immediately have all the values.

Maybe write something like

"A whitespace-separated list of tokeninfo fields which are not displayed to the user."

and the same for the admin.

privacyidea/lib/token.py
```diff
@@ -445,7 +445,8 @@ def get_tokens(tokentype=None, realm=None, assigned=None, user=None,
 def get_tokens_paginate(tokentype=None, realm=None, assigned=None, user=None,
                 serial=None, active=None, resolver=None, rollout_state=None,
                 sortby=Token.serial, sortdir="asc", psize=15,
-                page=1, description=None, userid=None, allowed_realms=None, tokeninfo=None):
+                page=1, description=None, userid=None, allowed_realms=None, tokeninfo=None,
+                excluded_tokeninfo_fields=None):
```
laclaro, line 449: please rename to `hidden_tokeninfo_fields` in analogy to `hidden_columns` in audit.py.

privacyidea/api/lib/prepolicy.py
```diff
@@ -2104,3 +2104,24 @@ def required_piv_attestation(request, action=None):
...
+
+
+def check_excluded_tokeninfo_fields(request=None, action=None):
+    user_object = request.User
+    role = g.logged_in_user.get("role")
+    if role == ROLE.USER:
+        return True
+
+    pols = Match.user(g, scope=SCOPE.USER, action=ACTION.TOKENINFO_API_EXCLUDE, user_object=user_object)\
+        .action_values(unique=True)
```
laclaro, line 2116: If you set unique=False, you may specify the policy action values as white spaced list (no need to split by comma)

We should make this policy available for admins or users (see below). Check for both:

``` python
    pols = Match.admin_or_user(g, action=ACTION.TOKENINFO_API_EXCLUDE,
                               user_obj=request.User).action_values(unique=False)
```

privacyidea/api/lib/prepolicy.py
```diff
@@ -2104,3 +2104,24 @@ def required_piv_attestation(request, action=None):
...
+
+    # when there is a policy
+    if pols:
+        # get the excluded fields
+        action_string = list(pols)[0]
+        # create a list
+        action_list = action_string.split(';')
+        # set the fields as a list to the request-data
+        request.all_data['excluded_tokeninfo_fields'] = action_list if len(action_list) > 0 else None
```
laclaro, line 2125: If you have unique=False above, you can just pass the action values by 

``` python
request.all_data['hide_tokeninfo_fields'] = pols
```

no listing and splitting needed.

privacyidea/api/token.py
```diff
@@ -343,6 +344,7 @@ def get_challenges_api(serial=None):
 
 @token_blueprint.route('/', methods=['GET'])
 @prepolicy(check_admin_tokenlist, request)
+@prepolicy(check_excluded_tokeninfo_fields, request)
```
laclaro, line 347: Please rename to `check_hide_tokeninfo`

privacyidea/lib/token.py
```diff
@@ -531,6 +532,18 @@ def get_tokens_paginate(tokentype=None, realm=None, assigned=None, user=None,
...
+                # the iteration
+                tmp = {}
+                # Loop over all the key-value pairs and check if they are not excluded. if there are not in the
+                # exclusion-list we add them to the temp-dict.
+                for key in token_dict['info'].keys():
+                    if key not in excluded_tokeninfo_fields:
+                        tmp[key] = token_dict['info'][key]
+
+                # now we just replace the old info-dict with the newly created one.
+                token_dict['info'] = tmp
```
laclaro, line 546: The error occurs if you change the size of the dict keys during iteration. You can circumvent this by this code snipped and avoid your longer workaround:

```python
                for key in list(token_dict['info']):
                    if key in excluded_tokeninfo_fields:
                        token_dict['info'].pop(key)
```
You iterate not over the `.keys` themselves, but in the copied `list(token_dict['info'])`, which works.

laclaro (changes requested):
Thank you for this pull request. We like the feature. I commented for code modifications and an extension towards SCOPE.ADMIN.

In the end, we should also add some documentation to policy/admin.rst and policy/user.rst.

theodor-franke:
I think I changed everything requested. I will look into creating test and documentation the coming days.

privacyidea/api/lib/prepolicy.py
```diff
@@ -2106,22 +2106,28 @@ def required_piv_attestation(request, action=None):
...
+    #
+    # # when there is a policy
+    # if pols:
+    #     # get the excluded fields
+    #     action_string = list(pols)[0]
+    #     # create a list
+    #     action_list = action_string.split(';')
+    #     # set the fields as a list to the request-data
+    #     request.all_data['excluded_tokeninfo_fields'] = action_list if len(action_list) > 0 else None
+
```
laclaro, line 2126: Please rename `hidden_tokeninfo_fields` to `hidden_tokeninfo`.
Please delete the commented lines and add a short description of the decorator function in """ quotes.

Maybe something like:

``` python
"""
This decorator checks for the policy `hide_tokeninfo` and sets the 
`hidden_tokeninfo` parameter. The given tokeninfo keys will be
removed from the response.

Check ACTION.HIDE_TOKENINFO

The decorator wraps GET /token/

:param request: The request that is intercepted during the API call
:type request: Request Object
:param action: An optional Action
:type action: basestring
:returns: Always true. Modified the parameter request
"""
```

laclaro (changes requested):
Thank you for your changes. Add some documentation in doc/policies/admin.rst and doc/policies/user.rst.

Maybe something like:
```
hide_tokeninfo
~~~~~~~~~~~~~~

type: string

This species a blank separated list of tokeninfo keys, that should be removed 
from the response and thereforewill not be shown in the WebUI.

For example a value ``tokenkind auto_renew`` will hide these two keys.
```

For the test, you may nearly copy the test for hide_audit_columns
https://github.com/privacyidea/privacyidea/blob/99d10ff1b3cb27af5c9afe3d3e9dd9d1f076fec0/tests/test_api_lib_policy.py#L1509

plettich:
I'll merge to a local feature branch so i can add tests and docs

PR #2823, opened 2021-07-15
2810/consolidate parameters

cornelinux:
This simply consolidates the different parameters in the request.

Determining the `request.User` from the serial number in the request should be done in a different PR.

### DIFF
--- a/privacyidea/api/lib/utils.py
+++ b/privacyidea/api/lib/utils.py
@@ -238,10 +238,16 @@ def getLowerParams(param):
     return ret
 
 
-def get_all_params(param, body):
+def get_all_params(request):
     """
-    Combine parameters from GET and POST requests
+    Retrieve all parameters from a request, no matter if these are GET or POST requests
+    or parameters are contained as viewargs like the serial in DELETE /token/<serial>
+
+    :param request: The flask request object
     """
+    param = request.values
+    body = request.data
+    view_args = request.view_args
     return_param = {}
     for key in param.keys():
         return_param[key] = unquote(param[key])
@@ -253,6 +259,10 @@ def get_all_params(param, body):
             return_param[k] = v
     except Exception as exx:
         log.debug("Can not get param: {0!s}".format(exx))
+
+    # We maybe need to unquote the view_args
+    #return_param.update(request.view_args)
+    #return_param.update(request.json)
 
     return return_param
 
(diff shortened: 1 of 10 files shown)


privacyidea/api/lib/utils.py
```diff
@@ -238,21 +238,41 @@ def getLowerParams(param):
...
+        # Add the original JSON data
+        return_param.update(request.json)
+    elif body:
+        # In case of serialized JSON data in the body, add these to the values.
+        try:
+            json_data = json.loads(to_unicode(body))
+            for k, v in json_data.items():
+                return_param[k] = v
+        except Exception as exx:
+            log.debug("Can not get param: {0!s}".format(exx))
```
plettich, line 269: This debug message bugs me for quite some time now, it just doesn't say anything and pops up on every `GET` request.
Maybe we can check if `request.data` exists (i.e. not empty) before trying to parse it.
Or we could parse it only if the `request.method` and/or the mime-type specify that the body should be json-formatted.

privacyidea/api/lib/utils.py
```diff
@@ -238,21 +238,41 @@ def getLowerParams(param):
...
+
+    if request.view_args:
+        log.debug(u"Update params in request {0!s} {1!s} with view_args.".format(request.method,
+                                                                                 request.base_url))
+        # We add the unquoted view_args
+        return_param.update({key: unquote(value) for (key, value) in request.view_args.items()})
```
plettich, line 275: We should also bear in mind that some parameters/args/bodies can be set by the user and thus will be available in the param dictionary. Should we check for possible dangerous parameters?

plettich (changes requested):

cornelinux:
@plettich sry - I still had a local commit, which I did not push upstream.
I will work on your comments.

privacyidea/api/lib/utils.py
cornelinux: Will do

privacyidea/api/lib/utils.py
cornelinux: I do not see, how this is different from the way we handled everything before?

privacyidea/api/lib/utils.py
cornelinux: I checked through all the tests again, you should not see this log message in the future.

The function was heavily redesign. We first check for json.
And since GET-Requets do not contain a body and we check 262 `elif body`.
So actually we really should only hit this debug log, if there is a post request with a badly formatted body.

privacyidea/api/lib/utils.py
plettich: exactly...

plettich (approved):

Issue #2827, opened 2021-07-16
Update documentation, translations, etc for v3.7


plettich:
This is a meta issue to track changes in the documentation, translation, comments, etc for v3.7
Some ideas:
- [ ] cleanup Makefile in `doc/`, we do not build for all these different media anyway
- [ ] ...

PR #2828, opened 2021-07-16
2827/fix spellings

plettich:
This PR has rather a lot of changes but it is mostly spelling and formatting in the docs.
Most of the spelling errors found with `make spelling` but the output is rather overwhelming. For newly introduced changes our spellcheck-bot should do the trick.

Working on #2827


privacyidea/lib/tokens/webauthntoken.py
```diff
@@ -63,7 +63,7 @@
  * registered by administrators for users or
  * registered by the users themselves.
 
-Beware the WebAuthn tokens can only be used if the privacyIDEA server and
```
cornelinux, line 66: You should really be afraid of the WebAuthn token!

cornelinux (approved):

PR #2833, opened 2021-07-20
let token handler act on all tokens (at token_load)

### LINKED ISSUE #2698: Allow event handler to access token serial during import

When importing tokens we should make the serial numbers of the imported tokens available so that the event handler can act on all imported tokens.

laclaro:
The serials, which are processed by token_load were already present in the token event handler in the audit object.

This PR uses this data to act on all tokens which were loaded on import_token.

The many changed lines are just due to the changed indentation (adding a loop).

Closes #2698


cornelinux (changes requested):
Can you please add a testcase - maybe with a token eventhandler that disables loaded tokens?
You can create a dummy import file and eheck afterwards if all loaded tokens exist and are disabled.

Also I am not sure, if the list of the tokens is truncated at this point (or we may be before the DB write, so we could have a striing with all 1000 imported tokens)
Problem: The string in `g.audit_object.audit_data.get("serial")` could get very long and I am not sure, if it is truncated at a certain point.

laclaro:
I implemented a test for 100 imported tokens. I tested also with 1000 (which takes long!) and it also works.

cornelinux (approved):

cornelinux:
In `check_serial_valid` we actually check the serial number, which is not allowed to contain a comma.

So my change would allow a list of serials like

    serial1, serial2, serial3

but also

     serial1,serial2, serial3      ,    serial4

PR #2834, opened 2021-07-20
2577/highlight policy search

### LINKED ISSUE #2577: highlight search term in policy action search
While working making the changes for #2574, I found highlighting the search terms in the policy action search would be a big pro. This can be done using angular-internals.

> During the implementation of this feature, we could also highlight the search term using the angular module ui-utils as described here
> https://angular.de/artikel/mark-highlight-search-results/

_Originally posted by @laclaro in https://github.com/privacyidea/privacyidea/issues/2178#issuecomment-759389202_

laclaro:
closes #2577.

This PR adds a separate filter to escape the descriptions before highlighting

### DIFF
--- a/privacyidea/static/contrib/js/ngmodules/escape-text.js
+++ b/privacyidea/static/contrib/js/ngmodules/escape-text.js
@@ -1,0 +1,16 @@
+'use strict';
+
+/**
+ * Wraps the
+ * @param text {string} text which is escaped
+ */
+angular.module('escape_text',[]).filter('escape_text', function () {
+  return function (text) {
+    return text.toString()
+        .replace(/&/g, "&amp;")
+        .replace(/</g, "&lt;")
+        .replace(/>/g, "&gt;")
+        .replace(/"/g, "&quot;")
+        .replace(/'/g, "&#039;");
+  };
+});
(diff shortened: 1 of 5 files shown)


privacyidea/static/contrib/js/ngmodules/escape-text.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
```
cornelinux, line 1: Is this module self written?

In contrib we only have external modules. So if this is your own code, you need to put it somewhere else like add it to app.js or the contoller of the policies.

privacyidea/static/contrib/js/ngmodules/escape-text.js
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Wraps the
+ * @param text {string} text which is escaped
+ */
+angular.module('escape_text',[]).filter('escape_text', function () {
+  return function (text) {
+    return text.toString()
+        .replace(/&/g, "&amp;")
+        .replace(/</g, "&lt;")
```
cornelinux, line 11: Is this sufficient?
What if I pass urlencoded action descriptions like

~~~~
%3Cscript%3Ealert%28%22Got%20you%21%22%29%3C%2Fscript%3E
~~~~

cornelinux (commented):
Please check my comments.

privacyidea/static/contrib/js/ngmodules/escape-text.js
laclaro: it is indeed self-written.

plettich:
Closing in favor of #3027

PR #2837, opened 2021-07-20
enable smtps identified from the server URI

### LINKED ISSUE #2568: Allow SSL over SMTP and not only STARTTLS

laclaro:
closes #2568

I do not like this if elif else clause - any thoughts on that?

### DIFF
--- a/privacyidea/lib/smtpserver.py
+++ b/privacyidea/lib/smtpserver.py
@@ -18,7 +18,7 @@
 #
 #
 import six
-
+from six.moves.urllib.parse import urlparse
 from privacyidea.lib.framework import get_app_config_value
 from privacyidea.lib.queue import job, wrap_job, has_job_queue
 from privacyidea.models import SMTPServer as SMTPServerDB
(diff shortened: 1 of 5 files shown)


privacyidea/lib/smtpserver.py
```diff
@@ -103,9 +103,15 @@ def test_email(config, recipient, subject, body, sender=None,
         msg['To'] = ",".join(recipient)
         msg['Date'] = strftime("%a, %d %b %Y %H:%M:%S +0000", gmtime())
         msg['Reply-To'] = reply_to
-
-        mail = smtplib.SMTP(config['server'], port=int(config['port']),
-                            timeout=config.get('timeout', TIMEOUT))
+        if config['server'].startswith('smtps://'):
+            mail = smtplib.SMTP_SSL(config['server'][8:], port=int(config['port']),
```
cornelinux, line 107: Please add a test case (mock) for this!
Is it necessary to add some docs?

Have you checked if `SMTP_SSL` exists in Python 2.7.5?

cornelinux (changes requested):

privacyidea/lib/smtpserver.py
laclaro: Yes it is also in 2.7.

privacyidea/lib/smtpserver.py
```diff
@@ -103,15 +103,21 @@ def test_email(config, recipient, subject, body, sender=None,
         msg['To'] = ",".join(recipient)
         msg['Date'] = strftime("%a, %d %b %Y %H:%M:%S +0000", gmtime())
         msg['Reply-To'] = reply_to
-
-        mail = smtplib.SMTP(config['server'], port=int(config['port']),
-                            timeout=config.get('timeout', TIMEOUT))
+        if config['server'].lower().startswith('smtps://'):
```
plettich, line 106: We should use the builtin `urllib` for that. I'll add the changes.

plettich (changes requested):

privacyidea/lib/smtpserver.py
plettich: Done in 322d2707

cornelinux (approved):

cornelinux:
HI @plettich  I think you also need to approve to merge this :-)

plettich (approved):

PR #2838, opened 2021-07-20
Add a negating search for orphaned tokens

### LINKED ISSUE #2832: privacyidea-token-janitor --orphaned flag needs argument
It might be considered unusual that we call

`privacyidea-token-janitor find --orphaned 1`

instead of

`privacyidea-token-janitor find --orphaned`

The required change is small

```python
@manager.option('--orphaned', action='store_true',
                help='Whether the token is an orphaned token.')
```

However, in that case we break all existing cron jobs, so this behavior won't change too quickly.

cornelinux:
You can now search for not orphaned tokens
and you can search using "1" or "True".

Closes #2832


laclaro (approved):
Thank's. Good compromise with extra functionality.

PR #2850, opened 2021-08-04
sort policy output

### LINKED ISSUE #2802: sorted policy output from GET /policy
GET /policy delivers unsorted output. It might be nice to have always the same order (alphabetical?)

laclaro:
closes #2802


plettich:
Actually we do already sort the output namely by priority in `list_policies()`:
https://github.com/privacyidea/privacyidea/blob/e5df1eceb840d01d4587a11898a360e107466752/privacyidea/lib/policy.py#L671
The parameter is `True` by default.
Also i do think that the sorting should be done by the front-end (which has all the information like language, sort-order, etc.).

Issue #2783, opened 2021-06-16
Add a glossary


cornelinux:
We should add a glossary to the documentation.

There we can define words like *token*, *otp value*, *otp pin*, *time step*.
These might be rather clear to us developers but not to the admins.

We also need to define words, that currently seem to be used in different situations like *multi challenge*.

I think it is not necessary to define IT common words like *LDAP* or *RADIUS*. So we only need to define privacyIDEA specific words or words we defined ourselves.

Some more words and terms we need to not mix up:

* extended attribute mapping -> RADIUS
* extended conditions -> policies

laclaro:
- internal administrator
- opt pin
- otp value
- tokeninfo
- userinfo
- userstore
- audit
- token owner
- machine
- locked token
- revoked token
- orphaned token
- failcount
- userid
- split @ sign
- subscription
- rollout state
- otp key
- seed
- challenge
- multi challenge
- scope
- resolver
- realm
- event
- WebUI
- attribute mapping
-

PR #2859, opened 2021-08-06
initial glossary

laclaro:
working on #2783


cornelinux:
I changed the location of the glossary file, since we actually have no contents in the top level docs dir.
Also we might one day have several files as glossary or additional images.

cornelinux (approved):

doc/glossary/index.rst
```diff
@@ -0,0 +1,33 @@
...
+
+.. glossary::
+   :sorted:
+
+   OTP Value
+        A one-time password, which is generated by some mathematical algorithm, usually HMAC,
+        based on a seed. The term OTP value is used frequently by privacyIDEA to distinguish
+        the changing value from the :term:`OTP PIN`.
+        In terms of two factor authentication the OTP Value is actually the 2nd factor,
+        the possion factor, since it is usually only possible to calculage, if the user
```
cornelinux, line 14: calculate

doc/glossary/index.rst
```diff
@@ -0,0 +1,33 @@
...
+        is in the possession of the smartphone app or a hardware token. The OTP Value is calculated
+        using the secret cryptographic :term:`Seed`.
+
+   OTP PIN
+        The OTP PIN is the secret password with which the user authenticates against privacyIDEA.
+        The policy action :ref:`otppin_policy` sets the type of password. With this
+        password privacyIDEA will identify the tokens for which further actions are taken
+        (trigger a challenge or check a given :term:`OTP Value`).
+        In terms of two factor authentication the OTP PIN is the first factor, the knowledge.
```
cornelinux, line 23: We should add the information, that an OTP PIN is stored per token.
I.e. if a user has several tokens he can have different otp pins per token.
This is different if the otppinpolicy is set to userstore, when all tokens have the static factor from the user store

cornelinux (commented):
PLease chech my comments

PR #2862, opened 2021-08-10
raise TokenAdminError when assignment on /token/init

### LINKED ISSUE #2860: Raise error when the token already has a different owner on token_init
If you issue an init request 

`/token/init type=spass serial=MYSPASS pin=123 user=test realm=test`

but the token with serial MYSPASS is already assigned to a different user, the request succeeds, but the assignment will not happen, of course.

We may log a warning to make this behavior transparent.

laclaro:
closes #2860

### DIFF
--- a/privacyidea/lib/token.py
+++ b/privacyidea/lib/token.py
@@ -1052,6 +1052,16 @@ def init_token(param, user=None, tokenrealms=None,
                                                                   tokentype))
             log.error(msg)
             raise TokenAdminError("initToken failed: {0!s}".format(msg))
+        # prevent to init update a token changing the token owner
+        if user:
+            old_tokenowner = get_token_owner(serial)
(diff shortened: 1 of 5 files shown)


cornelinux:
I would put this at the location, where actually the owner of the token is handled:

~~~~python
    # Set the user of the token
    if user is not None and user.login != "":
        tokenobject.add_user(user)
~~~~

or even in `tokenclass.add_user()`. 

In the long run we are not sure, how we assign multiple tokens to the user.
So assign multiple users to a token should be basically possible. Currently it is somehow not.
So this should be handled in `add_user()`.

cornelinux:
@plettich I moved the check to a central location. All tests run successfully (locally) 
Please re-review.

plettich (approved):
Looks good

PR #2870, opened 2021-08-20
Fix upper realm names

### LINKED ISSUE #2869: upper case realm break login to webui
version 3.6 and above,

When logging in the webui with an upper case REALM name, we get this error.

~~~~
ERR905: Cannot pass user_object (<Administrator.netknights-lab@netknights-lab>) as well as user (Administrator), resolver (None), realm (NETKNIGHTS-LAB)in policy (None, 'user', 'enrollYUBIKEY')"}
~~~~

This is due to a user comparison in the policy module.

mitigation: Avoid this by using lower case realm.

cornelinux:
The user object check needs to be case-in-sensitive in the
realm.

Fixes #2869


plettich (approved):
looks good

PR #2874, opened 2021-08-25
Fix endianess off AD GUID

plettich:


privacyidea/lib/resolvers/LDAPIdResolver.py
```diff
@@ -437,14 +437,12 @@ def _get_uid(entry, uidtype):
...
-                    # which resembles a uuid. So we just convert it to one...
-                    # We might run into endian-issues here depending on what
-                    # ldap3/AD returns.
+                    # which resembles an uuid. So we just convert it to one...
                     log.warning('Found a byte-array as uid ({0!s}), trying to '
                                 'convert it to a UUID. ({1!s})'.format(binascii.hexlify(uid),
                                                                        e))
                     log.debug(traceback.format_exc())
-                    uid = str(uuid.UUID(bytes=uid))
+                    uid = str(uuid.UUID(bytes_le=uid))
```
cornelinux, line 445: Can we always safely assume that Windows / AD (which indicates `objectguid`) is little endian? I think so. Or should/need we make this configurable. However, in this function
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/resolvers/LDAPIdResolver.py#L169
(which is only used by Active Directory `objectguid`), again it is little endian, right?
So this is all fine?

cornelinux (commented):
Please simply assure me.

cornelinux (approved):

Issue #2852, opened 2021-08-04
Triggering incomplete push token


nilsbehlen:
1. Start pushtoken rollout
2. Do NOT scan the QR
3. Go to the token and activate it
4. Trigger it

This will result in the following stacktrace:
~~~~
Traceback (mostrecent call last):
  File "/opt/privacyidea/lib/python3.6/site-packages/flask/app.py", line
2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/opt/privacyidea/lib/python3.6/site-packages/flask/app.py", line
1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/opt/privacyidea/lib/python3.6/site-packages/flask/app.py", line
1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/opt/privacyidea/lib/python3.6/site-packages/flask/_compat.py",
line 39, in reraise
    raise value
  File "/opt/privacyidea/lib/python3.6/site-packages/flask/app.py", line
1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/opt/privacyidea/lib/python3.6/site-packages/flask/app.py", line
1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File
"/opt/privacyidea/lib/python3.6/site-packages/privacyidea/api/auth.py",
line 393, in decorated_function
    return f(*args, **kwargs)
  File
"/opt/privacyidea/lib/python3.6/site-packages/privacyidea/api/lib/postpolicy.py",
line 108, in policy_wrapper
    response = wrapped_function(*args, **kwds)
  File
"/opt/privacyidea/lib/python3.6/site-packages/privacyidea/api/lib/postpolicy.py",
line 108, in policy_wrapper
    response = wrapped_function(*args, **kwds)
  File
"/opt/privacyidea/lib/python3.6/site-packages/privacyidea/api/lib/decorators.py",
line 41, in function_wrapper
    response = wrapped_function(*args, **kwds)
  File
"/opt/privacyidea/lib/python3.6/site-packages/privacyidea/lib/decorators.py",
line 100, in check_user_or_serial_in_request_wrapper
    f_result = func(*args, **kwds)
  File
"/opt/privacyidea/lib/python3.6/site-packages/privacyidea/api/lib/prepolicy.py",
line 154, in policy_wrapper
    return wrapped_function(*args, **kwds)
  File
[...]

PR #2875, opened 2021-08-26
#2852/add event handler for custom user attributes

jona-samuel:
Implementation of the eventhandler for customuserattributes


privacyidea/lib/eventhandler/customuserattributeshandler.py
```diff
@@ -0,0 +1,87 @@
...
+        :param action:
+        :param options: Contains the flask parameters g,  attrkey and attrvalue
+        :type options: dict
+        :return:
+        """
+        ret = True
+        g = options.get("g")
+        attrkey = options.get("attrkey")
+        attrvalue = options.get("attrvalue")
+        logged_in_user = g.logged_in_user
```
cornelinux, line 79: I see we have been a bit unspecific here in the definition of the issue.
I think setting the attribute of the logged in user does very seldomly make sense.

I guess often we might want to set a custom attribute of a user, who authenticates at /validate/check or if an administrator acts on a user or on the token of a user.

So determining "logged_in_user" is not enough.

You should use the user from the request.
OR: we might even define in the event handler on which user we want to act: either on the "logged_in_user" or on the "tokenowner"; similar to the notification handler:
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/eventhandler/usernotification.py#L177

So the `actions` would be sth like:

~~~~python
"set_custom_user_attributes":{
                "user" : {
                    'type': 'str',
                    'required': True,
                    'value': ["logged in user", "tokenowner"] },
                "attrkey": {
                    'type': 'str',
                    'required': True,   
                    'description': _('The key of the attribute')},
                "attrvalue": {
                    'type': 'str',
                    'required': True,
                    'description': _('The value of the attribute')}
            },
  "delete_custom_user_attributes":{
      "user": {
                    'type': 'str',
                    'required': True,
                    'value': ["logged in user", "tokenowner"] },
        "attrkey": {
            'type': 'str',
            'required': True,   
            'description': _('The custom attribute of the user to delete.')},
~~~~

Please check the user notification how to determine the user object either from the logged in user or from the tokenowner.

privacyidea/lib/eventhandler/customuserattributeshandler.py
```diff
@@ -0,0 +1,87 @@
...
+        if action.lower() in ["set_custom_user_attributes",
+                              "delete_custom_user_attributes"]:
+            if action.lower() == "set_custom_user_attributes":
+                ret = logged_in_user.set_attribute(attrkey, attrvalue)
```
cornelinux, line 83: This must not only be the logged in user

privacyidea/lib/eventhandler/customuserattributeshandler.py
```diff
@@ -0,0 +1,87 @@
...
+            elif action.lower() == "delete_custom_user_attributes":
+                ret = User.delete_attribute(logged_in_user)
```
cornelinux, line 85: The usage is wrong!

You need to call it

     user.delete_attribute(attrkey)

cornelinux (changes requested):
The whole user handling is not sufficient.

Also the delete_attribute is wrong.

Then you will probably also have to adapt the tests.

tests/test_lib_events.py
```diff
@@ -2358,3 +2360,105 @@ def test_12_set_max_failcount(self):
...
+        a = user.attributes
+        self.assertEqual({'test': 'check'}, a)
+
+    def test_02_event_delete_attributes(self):
+
+        # Setup realm and user
+        self.setUp_user_realms()
+
+        g = FakeFlaskG()
+        audit_object = FakeAudit()
```
plettich, line 2400: again, the audit data is not necessary here.

tests/test_lib_events.py
```diff
@@ -2358,3 +2360,105 @@ def test_12_set_max_failcount(self):
...
+        b = user.attributes
+        self.assertEqual({}, b)
+
+    def test_03_event_set_attributes_tokenowner(self):
+
+        # Setup realm and user
+        self.setUp_user_realms()
+
+        g = FakeFlaskG()
+        audit_object = FakeAudit()
```
plettich, line 2432: and again...

tests/test_lib_events.py
```diff
@@ -2358,3 +2360,105 @@ def test_12_set_max_failcount(self):
...
+        audit_object.audit_data["serial"] = "SPASS01"
```
plettich, line 2433: i think we need to create a token and assign it to a user (which should not be the `logged_in_user`, we do not need a `logged_in_user` in this case). The user should then be passed as the `req.User`.

plettich (changes requested):
Some additional notes:
- please comment if You fixed a remark so the reviewer can mark them as resolved
- don't commit the `testuser.sqlite`, it shouldn't have been changed
- We need another real-world test with an API-Call and check the results
- We should also create _negative_ tests i.e. the user does not exists, etc.

privacyidea/lib/eventhandler/customuserattributeshandler.py
```diff
@@ -0,0 +1,110 @@
...
+    @property
+    def actions(cls):
+        """
+        This method returns a dictionary of allowed actions and possible
+        options in this handler module.
+
+        :return: dict with actions
+        """
+        actions = {
+            "set_custom_user_attributes": {
```
plettich, line 50: We should create constants for the actions like here:
https://github.com/privacyidea/privacyidea/blob/7bbf194aebfd2bbbee9b610ab22763ad5e6b32cb/privacyidea/lib/eventhandler/tokenhandler.py#L66

privacyidea/lib/eventhandler/customuserattributeshandler.py
```diff
@@ -0,0 +1,110 @@
...
+            logged_in_user = g.logged_in_user
+        except Exception:
+            logged_in_user = {}
+        tokenowner = self._get_tokenowner(request)
+        if user_type == USER_TYPE.TOKENOWNER and not tokenowner.is_empty():
+            user = tokenowner
+        elif user_type == USER_TYPE.LOGGED_IN_USER:
+            user = logged_in_user
+        else:
+            log.warning("Was not able to determine the recipient for the user "
```
plettich, line 98: We should also return here otherwise we get an error further down because the `user` does not exist.

plettich (changes requested):

privacyidea/lib/eventhandler/customuserattributeshandler.py
```diff
@@ -0,0 +1,87 @@
...
+
+from privacyidea.lib.eventhandler.base import BaseEventHandler
+from privacyidea.lib.user import User
+from privacyidea.lib import _
+import logging
+
+log = logging.getLogger(__name__)
+
+
+class VALIDITY(object):
```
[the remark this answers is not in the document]
jona-samuel: https://github.com/privacyidea/privacyidea/compare/%232852/Add_event_handler_for_custom_user_attributes
This is the commit

privacyidea/lib/eventhandler/customuserattributeshandler.py
```diff
@@ -0,0 +1,87 @@
...
+        This method returns a dictionary of allowed actions and possible
+        options in this handler module.
+
+        :return: dict with actions
+        """
+        actions = {
+            "set_custom_user_attributes":{
+                "attrkey": {
+                    'type': 'str',
+                    'discription': _('The key of the attribute')},
```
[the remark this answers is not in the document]
jona-samuel: https://github.com/privacyidea/privacyidea/compare/%232852/Add_event_handler_for_custom_user_attributes

privacyidea/lib/eventhandler/customuserattributeshandler.py
```diff
@@ -0,0 +1,87 @@
...
+                "attrvalue": {
+                    'type': 'str',
+                    'discription': _('The value of the attribute')}
+            },
+            "delete_custom_user_attributes":{
+                "user": {
+                    'type': 'user',
+                    'discription': _('An object from the class user')}
```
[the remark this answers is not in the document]
jona-samuel: https://github.com/privacyidea/privacyidea/compare/%232852/Add_event_handler_for_custom_user_attributes

privacyidea/lib/eventhandler/customuserattributeshandler.py
jona-samuel: https://github.com/privacyidea/privacyidea/compare/%232852/Add_event_handler_for_custom_user_attributes

privacyidea/lib/eventhandler/customuserattributeshandler.py
jona-samuel: https://github.com/privacyidea/privacyidea/compare/%232852/Add_event_handler_for_custom_user_attributes

privacyidea/lib/eventhandler/customuserattributeshandler.py
```diff
@@ -0,0 +1,87 @@
...
+        actions = {
+            "set_custom_user_attributes":{
+                "attrkey": {
+                    'type': 'str',
+                    'discription': _('The key of the attribute')},
+                "attrvalue": {
+                    'type': 'str',
+                    'discription': _('The value of the attribute')}
+            },
+            "delete_custom_user_attributes":{
```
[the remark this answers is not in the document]
jona-samuel: I fixed it in
https://github.com/privacyidea/privacyidea/compare/%232852/Add_event_handler_for_custom_user_attributes
but I forgot to add it to actions and fixed this in
https://github.com/privacyidea/privacyidea/compare/%232852/Add_event_handler_for_custom_user_attributes

plettich (changes requested):

privacyidea/lib/eventhandler/customuserattributeshandler.py
```diff
@@ -0,0 +1,125 @@
...
+            return ret
+
+        attrkey = options.get("attrkey")
+        attrvalue = options.get("attrvalue")
+        if action.lower() in ["set_custom_user_attributes",
+                              "delete_custom_user_attributes"]:
+            if action.lower() == "set_custom_user_attributes":
+                ret = user.set_attribute(attrkey, attrvalue)
+            elif action.lower() == "delete_custom_user_attributes":
+                ret = user.delete_attribute(attrkey)
```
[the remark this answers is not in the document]
jona-samuel: 40ed3b2

plettich:
Looks good so far, i'll add an API testcase then we can merge

plettich (approved):

cornelinux (approved):

cornelinux:
@plettich you requested me for review anew? Is there a special thing you want me to look into?

PR #2879, opened 2021-09-06
Add subscription checking for push tokens and plugins

### LINKED ISSUE #2801: enhance subscription definitions
We should add

* [x] "privacyidea-adfs"
* [x] "privacyidea-keycloak"
* [x] "simplesamlphp"
* [x] "privacyidea-simplesamlphp"

into subscription.py.

We can also enhance the subscription e.g. for push-tokens or the privacyIDEA authenticator app, which can be checked, when we provide additional enterprise features for the authenticator app.

There are two different types of subscriptions, the client-subscriptions (privacyidea-adfs, privacyidea-cp...) which are checked during authentication. And the kind of server or token subscriptions (privacyidea, privacyidea-app...) which are checked during login at the webui.
We could also rework the subscription definition in this way, that we also add the information, which "type" the subscription is.

In `lib/subscription.py:subscription_status()` we could return a list of tuples like:

    [ ("privacyidea": 3), ("privacyidea-app": 1) ]

indicating that the server subscription is fine, but the app subscription is exceeded.

# Note

"privacyIDEA-App" is the UserAgent that is used when the App communicates to the privacyIDEA server.
Currently we are issuing "privacyIDEA Authenticator".

The question is: *What do we want to do, when the  subscription for the App is exceeded?*

1. ~~Fail on Push authentication like with CP or~~
2.  Simply notify in the WebUI like with the privacyIDEA server subscription.

cornelinux:
Display a warning if there are push tokens used
without a subscription.
Display a warning if the subscription for push tokens
is exceeded/expired.

Closes #2801

### DIFF
--- a/privacyidea/api/lib/postpolicy.py
+++ b/privacyidea/api/lib/postpolicy.py
@@ -594,6 +594,8 @@ def get_webui_settings(request, response):
         content["result"]["value"]["show_seed"] = show_seed
         content["result"]["value"]["show_node"] = get_privacyidea_node() if show_node else ""
         content["result"]["value"]["subscription_status"] = subscription_status()
+        content["result"]["value"]["subscription_status_push"] = subscription_status("privacyidea authenticator",
+                                                                                     tokentype="push")
         content["result"]["value"]["qr_image_android"] = qr_image_android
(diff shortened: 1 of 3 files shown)


plettich (approved):
looks good

PR #2880, opened 2021-09-06
Extended policy conditions: HTTP enrivonment

### LINKED ISSUE #2510: Different behaviour for /validate/check and /auth endpoints -> policy condition

In loginmode=privacyIDEA users log in to the webui with their tokens.
This is basically a request against ``/auth`` with the same tokens like they are used in requests against ``/validate/check``.
However, sometimes it is required to have different login logic for these two endpoints, like:

* User is allowed to use EMAIL tokens to login to WebUI (/auth) but not to the VPN (/validate/check)
* User needs to use a PIN when login to WebUI (/auth) but not when login to VPN (/validate/check)

Usually this can be done using the client condition or HTTP headers in the policies. But sometimes this is deemed to complicated.

This sounds like a condition for the specific scopes ``authentication``   and ``authorization``.

Use client IPs ;-)

* Add a condition. But this condition would only hold for two scopes. So suddenly we would have scope dependent conditions. However, we also have scope dependent conditions, if we do tokentypes as conditions! #2590
* ~~Add an extended condition like the *http_header* and *userinfo*. This way it would not be to prominent. However, the user could modify the *http_header* when authenticating against the webui, so this is not really practical!~~

# Concept

In `g.request_headers.environ` we have a lot of environment invormation like the `PATH_INFO` being `validate/check`. This way we could distringuish between `/auth` and `/validate/check`.

image

We could add an additional extended condition (to userinfo, token, tokeninfo, http request header) being *HTTP environment*. 
This way we could have an extended condition `PATH_INFO=/validate/check` which would e.g. match the OTPPIN policy only in case of endpoint /validate/check and not /auth.

cornelinux:
Add HTTP environment as extended conditions.
This allows us to check for server software,
and URI path names.

Closes #2510

### DIFF
--- a/doc/policies/conditions.rst
+++ b/doc/policies/conditions.rst
@@ -30,7 +30,7 @@ consists of five parts:
 Sections
 ~~~~~~~~
 
-privacyIDEA implements three sections ``userinfo``, ``token``, ``tokeninfo``, ``HTTP Request Headers``
+privacyIDEA implements the sections ``userinfo``, ``token``, ``tokeninfo``, ``HTTP Request Headers``
 and ``HTTP Environment``.
 
 ``userinfo``
--- a/doc/policies/conditions.rst
+++ b/doc/policies/conditions.rst
@@ -30,7 +30,8 @@ consists of five parts:
 Sections
 ~~~~~~~~
 
-privacyIDEA implements three sections ``userinfo``, ``token``, ``tokeninfo`` and ``HTTP Request Headers``.
+privacyIDEA implements three sections ``userinfo``, ``token``, ``tokeninfo``, ``HTTP Request Headers``
+and ``HTTP Environment``.
 
 ``userinfo``
 ^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


doc/policies/conditions.rst
```diff
@@ -30,7 +30,8 @@ consists of five parts:
 Sections
 ~~~~~~~~
 
-privacyIDEA implements three sections ``userinfo``, ``token``, ``tokeninfo`` and ``HTTP Request Headers``.
+privacyIDEA implements three sections ``userinfo``, ``token``, ``tokeninfo``, ``HTTP Request Headers``
```
plettich, line 33: Please change this to "five sections".

plettich (changes requested):
just a minor change

doc/policies/conditions.rst
cornelinux: Will do "the sections" to be future proof.

cornelinux: Done in 753c230b

plettich (approved):
looks good

PR #2884, opened 2021-09-14
2753/check_tokeninfokey

### LINKED ISSUE #2753: tokenjanitor: check tokeninfo key exstistence
In some cases it might be useful to check just for the existence or not-existence of a certain tokeninfo value. By doing this, you filter out those tokens which have been processed (marked) earlier.

```shell
privacyidea-token-janitor find --has-not-tokeninfo-key import_time --action mark \
    --set-tokeninfo-key import_time --set-tokeninfo-value $(date --iso-8601=minutes)
```

NNLemling:
I created a new pullrequest for this issue because my old branch was created before the attribute changes what produced version conflicts.
Closes #2753

### DIFF
--- a/tools/privacyidea-token-janitor
+++ b/tools/privacyidea-token-janitor
@@ -229,8 +229,6 @@ def _compare_regex_or_equal(_key, given_regex):
 
 def build_tokenvalue_filter(key,
                             tokeninfo_value,
-                            has_not_tokeninfo_key,
-                            has_tokeninfo_key,
                             tokeninfo_value_greater_than,
                             tokeninfo_value_less_than,
                             tokeninfo_value_after,
@@ -243,6 +241,7 @@ def build_tokenvalue_filter(key,
     if the conjunction of all comparators returns True.
 
     :param key: user-supplied --tokeninfo-key= value
+    :param has_tokeninfo_key: user-supplied --has_tokeninfo_key= value
     :param has_not_tokeninfo_key: user-supplied --has_not_tokeninfo_key= value
     :param tokeninfo_value: user-supplied --tokeninfo-value= value
     :param tokeninfo_value_greater_than: user-supplied --tokeninfo-value-greater-than= value
@@ -254,10 +253,6 @@ def build_tokenvalue_filter(key,
     tvfilter = []
     if tokeninfo_value:
         tvfilter.append(_compare_regex_or_equal(key, tokeninfo_value))
-    if has_not_tokeninfo_key:
-        tvfilter.append(has_not_tokeninfo_key)
-    if has_tokeninfo_key:
-        tvfilter.append(has_tokeninfo_key)
     if tokeninfo_value_greater_than:
         tvfilter.append(_compare_greater_than(key, tokeninfo_value_greater_than))
     if tokeninfo_value_less_than:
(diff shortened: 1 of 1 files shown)


tools/privacyidea-token-janitor
```diff
@@ -227,6 +229,8 @@ def _compare_regex_or_equal(_key, given_regex):
 
 def build_tokenvalue_filter(key,
                             tokeninfo_value,
+                            has_not_tokeninfo_key,
+                            has_tokeninfo_key,
```
plettich, line 233: actually we do not need these parameters here since we check it after the database was queried. The filter is only for the database query.

tools/privacyidea-token-janitor
```diff
@@ -248,6 +254,10 @@ def build_tokenvalue_filter(key,
     tvfilter = []
     if tokeninfo_value:
         tvfilter.append(_compare_regex_or_equal(key, tokeninfo_value))
+    if has_not_tokeninfo_key:
+        tvfilter.append(has_not_tokeninfo_key)
+    if has_tokeninfo_key:
+        tvfilter.append(has_tokeninfo_key)
```
plettich, line 260: this makes no sense, we would need a compatible database query here.

plettich (changes requested):

tools/privacyidea-token-janitor
NNLemling: All changes fixed in commit 7d081601e5905194024ef14f932dd101012fe3a3.

NNLemling:
As it turned out, "tinyint" is passed as bool (0 = False, 1 True). As a result, this type was not intercepted in the "_compare_regex_or_equal" function and thus treated as a "string".

tools/privacyidea-token-janitor
```diff
@@ -237,7 +241,9 @@ def build_tokenvalue_filter(key,
     user-defined criterion matches.
     The filter matches a record if *all* comparator functions return True, i.e.
     if the conjunction of all comparators returns True.
+
     :param key: user-supplied --tokeninfo-key= value
+    :param has_not_tokeninfo_key: user-supplied --has_not_tokeninfo_key= value
```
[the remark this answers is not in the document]
plettich: Since the parameters are removed, we don't need the entries in the doc-string.

tools/privacyidea-token-janitor
```diff
@@ -405,52 +416,49 @@ def export_user_data(token_list, attributes=None):
...
 @manager.option('--tokentype', help='The tokentype to search.')
 @manager.option('--serial', help='A regular expression on the serial')
 @manager.option('--description', help='A regular expression on the description')
-@manager.option('--action', help='Which action should be performed on the '
-                                 'found tokens. {0!s}'.format("|".join(ALLOWED_ACTIONS)))
-@manager.option('--set-description', help='')
-@manager.option('--set-tokeninfo-key', help='')
-@manager.option('--set-tokeninfo-value', help='')
+@manager.option('--attributes', help='Extends the "listuser" function to display additional user attributes')
+@manager.option('--tokenrealms', help='A comma separated list of realms, to which the tokens '                                      'should be assigned.')
```
plettich, line 447: there is a newline missing here, or at least remove the whitespaces.
It is basically a good idea to limit the line length even if everybody has ultra-HD-widescreen-fancy displays nowadays.
I think we enforce something like 120 chars (even though i think it shouldn't be utilized all the time, keep to 80-100 chars for readability). I'll have to check why our PEP8 checker didn't say something here.

plettich (changes requested):

tools/privacyidea-token-janitor
NNLemling: fixed in c5d5479e336ee17e53e7e382af2a17c78eded56c.

cornelinux: nice

PR #2905, opened 2021-10-27
2755/tokenjanitor ignoring validate mac no check

### LINKED ISSUE #2755: Tokenjanitor ignoring "--validate_mac no_check"
Version: 3.5.1
OS: Ubuntu 18.04

i tried importing tokens into a new Setup because i don´t want to recycle the old VM but it always says it cant read because of validation.

>  /privacyidea-token-janitor loadtokens --validate_mac no_check --preshared_key_hex 0c577cd2f6a46a42d7651e13cdea8b98 --pskc tokens-01.xml
The following tokens were not read from the PSKC file because they could not be validated: dict_keys(

Sebastian-Lemling:
The vmac parameter was passed incorrectly, which is why the "validate_mac" was always set to "check_fail_hard". This led to problems if the VMAC was incorrect. 

fixes #2755

### DIFF
--- a/tools/privacyidea-token-janitor
+++ b/tools/privacyidea-token-janitor
@@ -617,7 +617,7 @@ def find(last_auth, assigned, active, tokeninfo_key, tokeninfo_value,
                 help='Import this PSKC file.')
 @manager.option('--preshared_key_hex', dest='preshared_key_hex',
                 help='The AES encryption key.')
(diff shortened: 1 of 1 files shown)


plettich (changes requested):
This also includes the changes from #2884.
We should cherry-pick these changes onto master.

PR #2906, opened 2021-10-28
2848/dropdb_doesnt_remove_alembic_version_table

Sebastian-Lemling:
Dropdb now also deletes the alembic_version table.

### DIFF
--- a/pi-manage
+++ b/pi-manage
@@ -68,7 +68,6 @@ from cryptography.hazmat.primitives import serialization
 import flask
 from six.moves import shlex_quote
 import importlib_metadata
-from sqlalchemy.ext.declarative import declarative_base
 
 from privacyidea.lib.sqlutils import delete_matching_rows
 from privacyidea.lib.security.default import DefaultSecurityModule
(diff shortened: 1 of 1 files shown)


pi-manage
```diff
@@ -584,16 +585,24 @@ def createdb(stamp=False):
...
-    This drops all the privacyIDEA database tables (except audit table).
+    This drops all the privacyIDEA database tables.
     Use with caution! All data will be lost!
 
     For safety reason you need to pass
         --dropit==yes
     Otherwise the command will not drop anything.
     """
     if dropit == "yes":
+        engine = db.session.get_bind()
```
plettich, line 596: We can rewrite this as
```python
    if dropit == "yes":
        print("Dropping all database tables!")
        db.drop_all()
        table_name = "alembic_version"
        db.reflect()
        table = db.Model.metadata.tables.get(table_name, None)
        if table:
            db.Model.metadata.drop_all(bind=db.engine,
                                       tables=[table],
                                       checkfirst=True)
    else:
        print("Not dropping anything!")
```
This way we reduce imports and use the given database object

plettich (changes requested):

pi-manage
NNLemling: fixed in commit 7dd23f.

plettich: The formatting seems reasonable so i'll merge all of it.

plettich (approved):

PR #2909, opened 2021-10-29
If only one realm is available make it the selected realm

### LINKED ISSUE #2908: Autoselect the realm, if only one is available
Given a realm admin, who configures a non-default realm. 

In the user view UI no realm is autoselected in the realm-dropdown.

## expected behaviour

If there is only one realm (non-default) realm, this realm should be autoselected in the user-realm dropdown.

I think this could be totally done in the UI.

cornelinux:
It could be that a realm admin is only allowed to see one realm,
which is not the default realm.
However, we make it the default selected realm in the WebUI,
so that the admin does not need to "select" this single realm
manually.

Closes #2908


plettich (approved):
Looks good. Can we use this as a component for #2912 ?

PR #2910, opened 2021-10-29
Fix removing of tokenowner when registration token is deleted

### LINKED ISSUE #2907: Remove tokenowner entry when a registration token is used.
If a user is using his registration token, the token is deleted but an entry in the `tokenowner` table remains with the `token_id=NULL`.

When automatically deleting the registration token we also need to remove the owner entry.

# Implementation

We need to clean up the token-deletion.

The deletion of all references (tokenowner, tokenrealm, challenges) is put in the DB token model delete method and the delete token calls are cleaned up.
This way we implicitly achieve, that tokenowners are removed, when the registration token is deleted.

cornelinux:
The token delete mechanism is cleaned up, so that all information
is deleted on the database model level.
If a token is deleted we take care to delete all relevant information
like tokenowner, tokenrealm...

Add test for checking if the tokenowner entry is removed after the
registration token is used and implicitly deleted.

Closes #2907

### DIFF
--- a/privacyidea/api/register.py
+++ b/privacyidea/api/register.py
@@ -168,7 +168,7 @@ def register_post():
     if not email_sent:
         log.warning("Failed to send registration email to {0!r}".format(email))
         # delete registration token
-        token.delete()
+        token.delete_token()
         # delete user
         user.delete()
         raise RegistrationError("Failed to send email!")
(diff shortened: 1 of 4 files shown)


plettich (approved):
Looks good

Issue #2919, opened 2021-11-19
Use secondary loginattributes everywhere


cornelinux:
An LDAPresolver can contain several login attributes. These can be used in the /validate/check request to authenticate the same user object.

However, currently the audit entry contains the login attribute that was used, not the first login attribute.
At other places these further (secondary) login attributes can not be used - like login at webUI, assigning tokens.

We should check if and how we can enhance the usage of further login attributes. Several question arise:

* Should there be on leading login attribute? 
* Should the userobject always be project on this leading login attribute, to ensure that everywhere the same user data is used?
* How will it work with policies, that contain usernames!? 
* ...

plettich:
We have the `used_login` attribute in the user object which contains the initial `login` parameter:
https://github.com/privacyidea/privacyidea/blob/deae3ef9dc99fc288a99303f1d690e788f0a20ba/privacyidea/lib/user.py#L92
The `login` attribute might change later during the `_get_user_from_userstore()` call:
https://github.com/privacyidea/privacyidea/blob/deae3ef9dc99fc288a99303f1d690e788f0a20ba/privacyidea/lib/user.py#L124

PR #2920, opened 2021-11-19
Check for matching login or used_login

plettich:
The `User` object provides the `login` and `used_login` attributes which
can differ when multiple login attributes are allowed.
This leads to a `PolicyError` in `match_policies()` since the `user`
parameter and the `user_object.login` parameter don't match.

Working on #2919

### DIFF
--- a/privacyidea/lib/policy.py
+++ b/privacyidea/lib/policy.py
@@ -720,7 +720,7 @@ class PolicyClass(object):
         """
         if user_object is not None:
             # if a user_object is passed, we check, if it differs from potentially passed user, resolver, realm:
-            if (user and user.lower() != user_object.login.lower()) \
+            if (user and user.lower() not in {user_object.login.lower(), user_object.used_login.lower()}) \
                     or (resolver and resolver.lower() != user_object.resolver.lower()) \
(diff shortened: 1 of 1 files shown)


privacyidea/lib/policy.py
```diff
@@ -720,7 +720,7 @@ def match_policies(self, name=None, scope=None, realm=None, active=None,
         """
         if user_object is not None:
             # if a user_object is passed, we check, if it differs from potentially passed user, resolver, realm:
-            if (user and user.lower() != user_object.login.lower()) \
+            if (user and user.lower() not in {user_object.login.lower(), user_object.used_login.lower()}) \
```
cornelinux, line 723: We have not used curly braces for sets anywhere else. This was a bit puzzling to me.
Is there a reason you chose curly braces instead of normal braces?

cornelinux (commented):
Please comment and enlighten me :-)

privacyidea/lib/policy.py
plettich: The idea for the `set` was, that we only have one element to compare if the `login` and `used_login` attributes are equal.

cornelinux (approved):

PR #2923, opened 2021-11-23
Be more specific about events and actions

cornelinux:
To avoid confusion if actions could trigger new actions
we are more specific about events, which trigger actions,
and actions, that can not trigger new actions.


plettich (approved):
Looks good

PR #2927, opened 2021-11-26
Add test for resolver policy condition

### LINKED ISSUE #2778: admin user may manage tokens of not-allowed-resolver-users

plettich:
Check that an administrator for a different resolver in the same realm
is not allowed to perform actions on tokens assigned to users from the
other resolver.

Closes #2778

### DIFF
--- a/tests/test_api_token.py
+++ b/tests/test_api_token.py
@@ -10,7 +10,8 @@ from privacyidea.lib.policy import (set_policy, delete_policy, SCOPE, ACTION,
                                     PolicyClass)
 from privacyidea.lib.token import (get_tokens, init_token, remove_token,
                                    get_tokens_from_serial_or_user, enable_token,
-                                   check_serial_pass, get_realms_of_token, assign_token)
+                                   check_serial_pass, get_realms_of_token,
+                                   assign_token, token_exist)
 from privacyidea.lib.resolver import save_resolver
 from privacyidea.lib.realm import set_realm
 from privacyidea.lib.user import User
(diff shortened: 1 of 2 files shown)


tests/test_api_token.py
```diff
@@ -126,6 +127,88 @@ def test_01_test_two_tokens(self):
...
+                                  [self.resolvername2])
+        self.assertEquals(len(failed), 0)
+        self.assertEquals(len(added), 1)
+
+        # create token delete policy for "testadmin" on resolver1
+        set_policy(name="pol-reso1",
+                   scope=SCOPE.ADMIN,
+                   action=','.join([ACTION.DELETE, ACTION.ASSIGN, ACTION.UNASSIGN,
+                                    ACTION.DISABLE, ACTION.ENABLE]),
+                   user=self.testadmin,
```
cornelinux, line 146: I think this is wrong. This policy matches only for the actions on the user who is called "self.testadmin".
You probably want to use the parameter "adminuser" of `set_policy`.
So I assume the following tests *could* also fail, since the user name does not match.

Please consider also adding a positive test.

cornelinux (changes requested):

tests/test_api_token.py
plettich: Correct, with the positive tests this makes sense now.
Fixed in 81cf2f74

cornelinux (approved):

Issue #2310, opened 2020-07-27
Boolean options saved as strings in the db might not work as expected


plettich:
Currently we save Eventhandler options in the database as strings (since they can be of several types).
When using boolean options this might lead to a bug since we never check the type.
- i.e. when an option is not set, there is no entry in the database -> `False`
- If the option is enabled, the value `"True"` will be written into the db. This will result in `bool("True")` -> `True`
- If the option is then disabled again, the value `"False"` will be written to the db: `bool("False")` -> `True`

This can be seen in the [`UserNotificationEventHandler`](https://github.com/privacyidea/privacyidea/blob/972b103011ee366982e52b7b7ccadf4d90541f06/privacyidea/lib/eventhandler/usernotification.py#L354).
The `EventHandlerOption` Table already contains a column `Type` which is currently unused.
We could check the action type whenever we use an option and cast accordingly or we can cast once when reading from the db (if we know the type at this point).

cornelinux:
We could use a function, that wraps all event handler actions, checking if this is a boolean action and running it through `is_true`.
This way we do not have to manually check, which action is boolean.

PR #2938, opened 2021-12-10
Fixing issue #2310

jona-samuel:
This is should fix issue #2310 
Now the usernotificationhandler gets a string from db and use the is_true function to get a bool from it


privacyidea/lib/eventhandler/usernotification.py
```diff
@@ -130,7 +131,7 @@ def actions(cls):
                     "description": _("Either send email as plain text or HTML."),
                     "value": ["plain", "html"]},
                 "attach_qrcode": {
-                    "type": "bool",
+                    "type": "str",
```
plettich, line 134: This should still be `bool` so that the UI displays a checkbox.

plettich (changes requested):
Please check all other event handlers as well for the usage of boolean options and add the `is_true()`.

cornelinux:
I created a new complete pull request #2994 also to avoid merge conflicts.

PR #2942, opened 2021-12-14
Move to current AngularJS Syntax

### LINKED ISSUE #1917: Move to new dependency injection syntax
We should replace Angular's old, deprecated dependency injection syntax with the new one.

plettich:
The syntax for dependency injections has changed in AngularJS so we'll
adapt to the new syntax.

Closes #1917


cornelinux (approved):
If you say so.
Looks more complicated, but consistent from a formatting point of view.

PR #2970, opened 2022-01-21
Fix NEW routes of most config objects

### LINKED ISSUE #2546: Split edit and create object in UI
In the UI the routes/states sometimes behave unexpected.

This is due to the fact, that "creating" an object and "editing" an object uses the same route and template.

This is true for the following cases.

Missing highlighted (active) route:
- Config -> SMTP Servers -> New SMTP-Server
         -> RADIUS Server
         -> privacyIDEA Server

Unclickable 'new' route when in details:
- Config -> SMTP Server
         -> RADIUS Server
         -> privacyIDEA Server
         -> SMTP Server
- Policies
- Event Handler
- Periodic Task

It would be great if:

* The *create_object* button would always be available
* A *named button* would be available during *editing* an object.

An example implementation is already here: https://github.com/privacyidea/privacyidea/compare/master...split_edit_and_create_policies

cornelinux:
The route/state of creating an object
was broken. Previouly it was the edit-state
without a given object parameter.
This some either broke or never worked correctly.

I added an additional state like
  config.radius.add
to explicitly reference, when a new configuration
would be created.
This fixed the problem with the link and the
problem with the highlighting.

Closes #2546

### DIFF
--- a/privacyidea/static/components/config/controllers/privacyideaServerController.js
+++ b/privacyidea/static/components/config/controllers/privacyideaServerController.js
@@ -51,7 +51,7 @@ myApp.controller("privacyideaServerController", ["$scope", "$stateParams", "info
 
     $scope.identifier = $stateParams.identifier;
     if ($scope.identifier) {
-        // We are editing an existing RADIUS Server
+        // We are editing an existing privacyIDEA Server
         $scope.getPrivacyideaServers($scope.identifier);
         } else {
         // This is a new privacyIDEA server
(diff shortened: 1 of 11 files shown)


plettich (changes requested):
There is still an issue when moving from a RADIUS-Server definition to adding a new RADIUS Server. The data from the old RADIUS-Server definition is pre-filled in the form.
This is due to a bug in the controller:
We would need to remove this block:
https://github.com/privacyidea/privacyidea/blob/b10610276e6667cadbf794fde18a0c3dfc15d806/privacyidea/static/components/config/controllers/radiusServerController.js#L33
and add the `$scope.identifier` as a parameter here:
https://github.com/privacyidea/privacyidea/blob/b10610276e6667cadbf794fde18a0c3dfc15d806/privacyidea/static/components/config/controllers/radiusServerController.js#L64

cornelinux:
Thanks a lot for noting!
actually the controller logic for the other server definitions was working - I adapted the RADIUS server logic the this of the other configs.
Done in 8f64de91

cornelinux:
Look @plettich, I shuffled the code a bit in 6079f973

PR #2974, opened 2022-01-22
Make all languages translateable in the Makefile

cornelinux:
This includes all languages, also the new ones in the translation of the WebUI and generation of translations.js.


Makefile
```diff
@@ -25,12 +25,13 @@ setversion:
...
+POS = $(wildcard po/*.po)
 translate:
 	grunt nggettext_extract
-	for language in de nl fr cs ; do \
-		(cd po; msgmerge $$language.po template.pot > tmp.po; mv tmp.po $$language.po) ; \
+	for language in $(POS); do \
+		(msgmerge $$language po/template.pot > tmp.po; mv tmp.po $$language) ; \
 	done
-	grunt nggettext_compile
+	grunt nggettext_compile --force
```
plettich, line 34: why is there a `--force` needed?

Makefile
```diff
@@ -25,12 +25,13 @@ setversion:
 	@echo "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!"
 	@echo "Please set a tag like:  git tag 3.17"
 
+POS = $(wildcard po/*.po)
 translate:
 	grunt nggettext_extract
-	for language in de nl fr cs ; do \
-		(cd po; msgmerge $$language.po template.pot > tmp.po; mv tmp.po $$language.po) ; \
+	for language in $(POS); do \
+		(msgmerge $$language po/template.pot > tmp.po; mv tmp.po $$language) ; \
```
plettich, line 32: We can use `msgmerge -U --backup=off $$language po/template.pot` to avoid the temporary file

plettich (changes requested):
I didn't take a closer look at the translations...

Makefile
cornelinux: This is a good question, maybe some of the myriads translations is incomplete, I am getting this warning:

~~~~
Running "nggettext_compile:all" (nggettext_compile) task
Warning: No Language header found! Used --force, continuing.
~~~~

Without the `--force` the Makefile would break.

cornelinux: OK, in my local copy I had some broken remnants like `de.po.po` from a previous Make run. I can remove the --force again.

Makefile
cornelinux: Done in 848585bec87dd5ea115a64c70fa38fc43154c3db

plettich (approved):

PR #2987, opened 2022-02-03
Check client definition

### LINKED ISSUE #2949: Check the client definition during creation of a policy

The function [check_ip_in_policy](https://github.com/privacyidea/privacyidea/blob/982329aa684ad096e7836e048d808aaca6ece4b9/privacyidea/lib/utils/__init__.py#L712) verifies if the given client IP is contained in the given IP settings list.
However, if the list contains an invalid IP definition, it raises a 500 Error as expected.

We should use this very same function (check_ip_in_policy) during the creation of a policy, so that we already raise an error during the saving of the policy.

We should check this on the library level [here](https://github.com/privacyidea/privacyidea/blob/2d240464961dfdc2f1a6560ed99fa051e543c6b7/privacyidea/lib/policy.py#L1338). We can ignore the outcome and simply use the fact, that an invalid definition raises an exception:

~~~~python
    if type(client) == list:
        client = ", ".join(client)
    # It does not matter, which IP we check. An invalid "client" definition will raise an exception
    check_ip_in_policy("127.0.0.1", client.split(","))
~~~~

None

**Additional context**

Do not forget the start with some tests in test_api_policy.

NNLemling:
- [x] Add function
- [x] Add test
- [x] Adjusted existing tests

Closes #2949

### DIFF
--- a/privacyidea/lib/policy.py
+++ b/privacyidea/lib/policy.py
@@ -188,6 +188,8 @@ from privacyidea.lib.utils.compare import compare_values, COMPARATOR_DESCRIPTION
 from privacyidea.lib.utils.export import (register_import, register_export)
 from privacyidea.lib.user import User
 from privacyidea.lib import _
+from netaddr import AddrFormatError
+from privacyidea.lib.error import privacyIDEAError
 import datetime
 import re
 import ast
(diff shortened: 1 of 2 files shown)


NNLemling:
All the python tests are failing.

It may be because no client is defined in most of the tests. In fact, it would be a lot of work to manually adjust all tests. Maybe someone could give me a tip on how to solve this more easily.

cornelinux:
OK, I think we should default the `client` to an empty list, since `check_ip_in_policy` expects a list.

~~~~python
client = client or []
~~~~

before calling check_ip_in_policy.

cornelinux:
I fixed it in bb6b2b12. When the testst are through, we can merge it.

cornelinux (approved):

NNLemling:
Cool! I actually had and tested a similar solution. More tests have been successfully completed but still not all. But your change is a little different. Hopefully, all tests will now pass. Thank you for the effort!

cornelinux:
@NNLemling we can take a look together at my last changes.

PR #2998, opened 2022-02-07
TUTI-6063:resolved SCIMresolver issue

sbandari-axiad:


privacyidea/lib/resolvers/SCIMIdResolver.py
```diff
@@ -69,19 +69,37 @@ def checkPass(self, uid, password):
...
+        params = {'filter':'userName eq "{0}"'.format(userid)}
+        res = self._search_users(self.resource_server,
+                                 self.access_token,
+                                 self.verify_tls,
+                                 params)
+        if len(res.get('resources', [])) > 1:
+            raise Exception('Could not find user: {0}'.format(userid))
+        if res.get('resources'):
+            for entry in res.get('resources'):
+                ret = self._fill_user_schema_1_0(entry, self.mapping)
```
cornelinux, line 102: This pull request is not working and not complete.
E.g. in this case the parameters of `_fill_user_schema_1_0` is wrong, since the method has no parameter for `self.mapping`!.

I assume, that you are working on a very different base branch!
So you probably either need to add more changes or rebase your branch.

Also, it would be great if you would open an issue and refer to this issue so that we know, what this pull request is ment to fix.

cornelinux (changes requested):
It woul be great, if you would also add tests.

sbandari-axiad:
As per @quoc-axiadids comment that this PR should not be in upstream privacy idea, so closing this PR and will create the PR in axiad/privacyidea repo.

PR #3001, opened 2022-02-09
Add more output to the HSM session handling

### LINKED ISSUE #3000: More log output for CKR_SESSION_HANDLE_INVALID
Sometimes we receive a `CKR_SESSION_HANDLE_INVALID` and we reinitialize the HSM connection.

The session handle to be invalid is a decision of the HSM, so we do not really control this. However, we could add some additional information, to understand why or that a session handle is becoming invalid (like due timeouts)

To do so, the AESHardwareSecurityModule can store to values:

During init of the HSM
(https://github.com/privacyidea/privacyidea/blob/740e6b92acfcc333a2f1ae05880b1a3cfb3a4e13/privacyidea/lib/security/aeshsm.py#L101)
we can store a timestamp `self.session_start_time`.
Each time the session is used (usually for decryption, sometimes for encryption) we could update the value `self.session_last_used-_time`.

During reinitialization we then could output these values to provide a better understanding to the administrator.

cornelinux:
We also add debug log output to know the time
that a hsm session was used.
And also how often (last use) the session handle was
used.

Closes #3000

### DIFF
--- a/privacyidea/lib/security/aeshsm.py
+++ b/privacyidea/lib/security/aeshsm.py
@@ -94,6 +94,8 @@ class AESHardwareSecurityModule(SecurityModule):  # pragma: no cover
         self.max_retries = config.get("max_retries", MAX_RETRIES)
         log.debug("Setting max retries: {0!s}".format(self.max_retries))
         self.session = None
+        self.session_start_time = datetime.datetime.now()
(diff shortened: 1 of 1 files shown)


NNLemling (approved):
looks good!

PR #3002, opened 2022-02-10
Allow to create human readable password (pw token)

### LINKED ISSUE #2864: Allow to define the contents of a generated pw-token
At the moment the pw token has very limited capabilities. You may already enroll a pw token via the api, but you are left with some seed://hexstring password.

I suggest to decouple the password generation (which honors some special contents policy) from the lost token process in
https://github.com/privacyidea/privacyidea/blob/99d10ff1b3cb27af5c9afe3d3e9dd9d1f076fec0/privacyidea/lib/token.py#L1845

I further suggest to return the password like for the registration token with some special field "password".

This change makes the pw token type more useful and may be one of several steps towards application-specific passwords.

cornelinux:
Just like the policies with registration codes
we can now define policies to create human readable
password tokens.

Closes #2864

### DIFF
--- a/privacyidea/lib/tokens/passwordtoken.py
+++ b/privacyidea/lib/tokens/passwordtoken.py
@@ -171,22 +171,6 @@ class PasswordTokenClass(TokenClass):
             param["otplen"] = len(param["otpkey"])
         TokenClass.update(self, param)
 
-    @log_with(log)
-    @check_token_locked
-    def set_otplen__(self, otplen=0):
-        """
-        sets the OTP length to the length of the password
-
-        :param otplen: This is ignored in this class
-        :type otplen: int
-        :result: None
-        """
-        secretHOtp = self.token.get_otpkey()
-        sp = PasswordTokenClass.SecretPassword(secretHOtp)
-        pw_len = len(sp.get_password())
-        TokenClass.set_otplen(self, pw_len)
-        return
-
     @log_with(log, log_entry=False)
     @check_token_locked
     def check_otp(self, anOtpVal, counter=None, window=None, options=None):
--- a/privacyidea/lib/tokens/passwordtoken.py
+++ b/privacyidea/lib/tokens/passwordtoken.py
@@ -149,10 +149,6 @@ class PasswordTokenClass(TokenClass):
         :type param: dict
         :return: None
         """
-        """
-        :param param:
-        :return:        
-        """
         if "genkey" in param:
             # Otherwise genkey and otpkey will raise an exception in
             # PasswordTokenClass
--- a/privacyidea/lib/tokens/passwordtoken.py
+++ b/privacyidea/lib/tokens/passwordtoken.py
@@ -21,16 +21,21 @@ This file contains the definition of the password token class
 import logging
 
 from privacyidea.lib.crypto import zerome, safe_compare
+from privacyidea.lib.utils import to_unicode
 from privacyidea.lib.tokenclass import TokenClass
 from privacyidea.lib.log import log_with
 from privacyidea.lib.decorators import check_token_locked
-from privacyidea.lib.utils import to_bytes
 from privacyidea.lib import _
 from privacyidea.lib.policy import SCOPE, ACTION, GROUP
+from privacyidea.api.lib.prepolicy import _generate_pin_from_policy
 
 
 optional = True
 required = False
+
+# We use an easier length of 12 for password tokens
+DEFAULT_LENGTH = 12
+DEFAULT_CONTENTS = 'cn'
 
 log = logging.getLogger(__name__)
 
@@ -41,6 +46,8 @@ class PasswordTokenClass(TokenClass):
     In addition, the OTP PIN can be used with this token.
     This Token can be used for a scenario like losttoken
     """
+
+    password_detail_key = "password"
 
     class SecretPassword(object):
 
@@ -74,6 +81,8 @@ class PasswordTokenClass(TokenClass):
 
     def __init__(self, aToken):
         TokenClass.__init__(self, aToken)
+        self.otp_len = DEFAULT_LENGTH
+        self.otp_contents = DEFAULT_CONTENTS
         self.hKeyRequired = True
         self.set_type(u"pw")
 
@@ -132,6 +141,7 @@ class PasswordTokenClass(TokenClass):
                 ret = res
         return ret
 
+    @log_with(log, log_entry=False)
     def update(self, param):
         """
         This method is called during the initialization process.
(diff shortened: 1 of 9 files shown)


cornelinux:
@plettich would you please review and then we could fix the merge conflict.
The problem is, that with every other pull request we will get a merge conflict, since most PRs have changes in lib/policy.py

tests/test_api_lib_policy.py
```diff
@@ -3022,14 +3022,49 @@ def test_36_init_registrationcode_length_contents(self):
...
+        # request, that matches the policy
+        req.all_data = {"user": "cornelius", "realm": "home", "type": "pw"}
+        init_token_length_contents(req)
+        # Check, if the tokenlabel was added
+        self.assertEqual(req.all_data.get("pw.length"), "6")
+        self.assertEqual(req.all_data.get("pw.contents"), "+n")
+        # delete policy
+        delete_policy("pw_length")
+        delete_policy("pw_contents")
+
```
plettich, line 3067: Please also add a test that if the token is neither `pw` nor `registration` we don't have the `lenght` and `content` parameter in `all_data`

tests/test_api_validate.py
```diff
@@ -3101,6 +3116,158 @@ def test_00_registrationtoken_with_validity_period(self):
...
+        with self.app.test_request_context('/token/init',
+                                           method='POST',
+                                           data={'user': 'cornelius',
+                                                 'type': 'registration'},
+                                           headers={'Authorization': self.at}):
+            res = self.app.full_dispatch_request()
+            self.assertTrue(res.status_code == 200, res)
+            data = res.json
+            detail = data.get("detail")
+            regcode = detail.get("registrationcode")
```
plettich, line 3135: we could also check here that the `regcode` has the default length and content

tests/test_api_validate.py
```diff
@@ -3101,6 +3116,158 @@ def test_00_registrationtoken_with_validity_period(self):
...
+                                           method='POST',
+                                           data={'user': 'cornelius',
+                                                 'type': 'registration',
+                                                 'otpkey': "hallo"},
+                                           headers={'Authorization': self.at}):
+            res = self.app.full_dispatch_request()
+            self.assertTrue(res.status_code == 200, res)
+            data = res.json
+            detail = data.get("detail")
+            regcode = detail.get("registrationcode")
```
plettich, line 3181: is the `otpkey` parameter ignored? please check.

tests/test_api_validate.py
```diff
@@ -3101,6 +3116,158 @@ def test_00_registrationtoken_with_validity_period(self):
...
+                                           method='POST',
+                                           data={"user": "cornelius",
+                                                 "pass": regcode}):
+            res = self.app.full_dispatch_request()
+            self.assertTrue(res.status_code == 200, res)
+            data = res.json
+            self.assertEqual("ACCEPT", data.get("result").get("authentication"))
+
+        # The registration code was generated. The passed otpkey was NOT used.
+        with self.app.test_request_context('/validate/check',
```
plettich, line 3194: Ok, we could have just checked the returned `regcode` :-)

tests/test_api_validate.py
```diff
@@ -3101,6 +3116,158 @@ def test_00_registrationtoken_with_validity_period(self):
...
+                                                 'type': 'pw',
+                                                 'genkey': 1,
+                                                 'pin': 'test'},
+                                           headers={'Authorization': self.at}):
+            res = self.app.full_dispatch_request()
+            self.assertEqual(200, res.status_code)
+            data = res.json
+            detail = data.get("detail")
+            serial = detail.get("serial")
+            password = detail.get("password")
```
plettich, line 3256: here we should also check that the password it matches the default length and content

plettich (changes requested):
Mostly test-related changes

privacyidea/lib/tokens/passwordtoken.py
```diff
@@ -132,23 +141,39 @@ def get_class_info(key=None, ret='all'):
...
+            param["otpkey"] = _generate_pin_from_policy(contents, size=int(size))
+        if "otpkey" in param:
+            param["otplen"] = len(param["otpkey"])
         TokenClass.update(self, param)
-        self.set_otplen()
 
     @log_with(log)
     @check_token_locked
-    def set_otplen(self, otplen=0):
+    def set_otplen__(self, otplen=0):
```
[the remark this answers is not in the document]
cornelinux: Oh, you are probably right. This is why I named it that funny, I was probably checking if I can get rid of it.

Will do.

tests/test_api_lib_policy.py
cornelinux: Will check and will do.

tests/test_api_validate.py
cornelinux: will do.

tests/test_api_validate.py
cornelinux: Will check.

tests/test_api_validate.py
cornelinux: will do.

tests/test_api_validate.py
cornelinux: Done

tests/test_api_validate.py
cornelinux: Done 
It looks like the default regcode contains special characters. But is so, this was a problem, that already existed earlier and we should recheck this and open another issue for that.

tests/test_api_lib_policy.py
cornelinux: Done

privacyidea/lib/tokens/passwordtoken.py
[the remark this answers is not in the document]
cornelinux: Done

tests/test_api_validate.py
cornelinux: There is nothing to do here, right?

tests/test_api_validate.py
plettich: But according to
https://github.com/privacyidea/privacyidea/blob/4dff583f3e099dd1479b2454115a83c6700e9f93/privacyidea/lib/tokens/registrationtoken.py#L41
the contents should have been "cn". Where did You encounter special characters?

plettich: Maybe this is the reason why the `APITokenTestCase.test_15_registration_code()` fails sometimes.

tests/test_api_validate.py
```diff
@@ -3101,6 +3118,170 @@ def test_00_registrationtoken_with_validity_period(self):
...
+                                                 'type': 'registration'},
+                                           headers={'Authorization': self.at}):
+            res = self.app.full_dispatch_request()
+            self.assertTrue(res.status_code == 200, res)
+            data = res.json
+            detail = data.get("detail")
+            regcode = detail.get("registrationcode")
+            self.assertEqual(DEFAULT_LENGTH_REG, len(regcode))
+            # Check if a number is contained
+            self.assertRegex(regcode, "[0-9]+")
```
plettich, line 3140: Please use `assertRegexpMatches()` here, even though it is deprecated, we can check with Python 2 as well.

tests/test_api_validate.py
```diff
@@ -3101,6 +3118,170 @@ def test_00_registrationtoken_with_validity_period(self):
...
+                                           headers={'Authorization': self.at}):
+            res = self.app.full_dispatch_request()
+            self.assertEqual(200, res.status_code)
+            data = res.json
+            detail = data.get("detail")
+            serial = detail.get("serial")
+            password = detail.get("password")
+            self.assertEqual(DEFAULT_LENGTH_PW, len(password))
+            # Check if a number is contained
+            self.assertRegex(password, "[0-9]+")
```
plettich, line 3268: See above

plettich (changes requested):

plettich (approved):
looks good

plettich:
Closed in favor of PR #3023

PR #3006, opened 2022-02-12
Handle the error in case an invalid modhex prefix is used.

### LINKED ISSUE #2896: Uncaught exception with Yubikey in "yubikey" mode
It looks like in yubikey mode we do not handle the exception correctly if the user input is wrong. 
Then the modhex decode will fail:

~~~~
  File "/opt/privacyidea/lib/python2.7/site-packages/privacyidea/lib/tokens/yubikeytoken.py", line 296, in check_otp
    log.debug("prefix: {0!r}".format(binascii.hexlify(modhex_decode(yubi_prefix))))
  File "/opt/privacyidea/lib/python2.7/site-packages/privacyidea/lib/utils/__init__.py", line 333, in modhex_decode
    return binascii.unhexlify(''.join([mod2HexDict[c] for c in m]))
TypeError: Odd-length string
~~~~

We need to check this, write a test and catch the exception.

cornelinux:
Closes #2896

### DIFF
--- a/privacyidea/lib/tokens/yubikeytoken.py
+++ b/privacyidea/lib/tokens/yubikeytoken.py
@@ -298,7 +298,11 @@ class YubikeyTokenClass(TokenClass):
 
         uid = msg_hex[0:12]
         log.debug("uid: {0!r}".format(uid))
-        log.debug("prefix: {0!r}".format(binascii.hexlify(modhex_decode(yubi_prefix))))
+        try:
+            log.debug("prefix: {0!r}".format(binascii.hexlify(modhex_decode(yubi_prefix))))
+        except binascii.Error:
+            log.info("Wrong OTP value. Not modhex compatible. Can not extract yubi_prefix.")
+            return -4
         # usage_counter can go from 1 – 0x7fff
         usage_counter = msg_hex[12:16]
         timestamp = msg_hex[16:22]
--- a/privacyidea/lib/tokens/yubikeytoken.py
+++ b/privacyidea/lib/tokens/yubikeytoken.py
@@ -298,11 +298,7 @@ class YubikeyTokenClass(TokenClass):
 
         uid = msg_hex[0:12]
         log.debug("uid: {0!r}".format(uid))
-        try:
-            log.debug("prefix: {0!r}".format(binascii.hexlify(modhex_decode(yubi_prefix))))
-        except binascii.Error:
-            log.info("Wrong OTP value. Not modhex compatible. Can not extract yubi_prefix.")
-            return -4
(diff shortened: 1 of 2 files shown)


privacyidea/lib/tokens/yubikeytoken.py
```diff
@@ -298,7 +298,11 @@ def check_otp(self, anOtpVal, counter=None, window=None, options=None):
 
         uid = msg_hex[0:12]
         log.debug("uid: {0!r}".format(uid))
-        log.debug("prefix: {0!r}".format(binascii.hexlify(modhex_decode(yubi_prefix))))
+        try:
+            log.debug("prefix: {0!r}".format(binascii.hexlify(modhex_decode(yubi_prefix))))
```
plettich, line 302: This is just a debug output, can't we just print the prefix as is instead of converting it to a hex-string?
The check if the prefix is valid should be done separately from the debug message.

tests/test_lib_utils.py
```diff
@@ -822,6 +822,8 @@ def test_28_yubikey_utils(self):
         self.assertEqual(modhex_decode('cbdefghijklnrtuv'),
                           binascii.unhexlify('0123456789abcdef'))
         self.assertEqual(modhex_decode('fjhbhrhrhv'), b'Hallo')
+        # fail with invalid modhex
+        self.assertRaises(binascii.Error, modhex_decode, 'nlltvcc')
```
plettich, line 826: We should do the same here:
`self.assertRaises((binascii.Error, TypeError), modhex_decode, 'nlltvcc')`

plettich (changes requested):

privacyidea/lib/tokens/yubikeytoken.py
cornelinux: Sounds sensible to me.

Will do.

privacyidea/lib/tokens/yubikeytoken.py
```diff
@@ -298,7 +298,11 @@ def check_otp(self, anOtpVal, counter=None, window=None, options=None):
...
+        except binascii.Error:
```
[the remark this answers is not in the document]
cornelinux: If we log the yubiprefix and not the modhex_decoded, we do not need to catch this exception.

privacyidea/lib/tokens/yubikeytoken.py
cornelinux: Done in 0bba2428

tests/test_lib_utils.py
cornelinux: Done in 0bba2428

plettich (approved):

PR #3007, opened 2022-02-14
improve documentation

### LINKED ISSUE #2986: Set `max_identifier_length` in SQLAlchemy
We need to set `max_identifier_length` in SQLAlchemy to avoid e.g. warning with Oracle:

We’re running PrivacyIDEA 3.6.3 with an httpd web server and Oracle 19c database. Seems every time we perform a database operation such as a user logging in the webUI and browsing around, we get the following error in the httpd error logs

SAWarning: Oracle version () is known to have a maximum identifier length of 128, rather than the historical default of 30. SQLAlchemy 1.4 will use 128 for this database; please set max_identifier_length=128 in create_engine() in order to test the application with this new length, or set to 30 in order to assure that 30 continues to be used. In particular, pay close attention to the behavior of database migrations as dynamically generated names may change. See the section 'Max Identifier Lengths' in the SQLAlchemy Oracle dialect documentation for background.

This is related to #2798

NNLemling:
set max_identifier_length = 128

Closes #2986

### DIFF
--- a/doc/installation/system/inifile.rst
+++ b/doc/installation/system/inifile.rst
@@ -26,7 +26,7 @@ The file should contain the following contents::
    # Your database
    SQLALCHEMY_DATABASE_URI = 'sqlite:////etc/privacyidea/data.sqlite'
    # Set maximum identifier length to 128
-   SQLALCHEMY_ENGINE_OPTIONS = {"max_identifier_length": 128}
+   # SQLALCHEMY_ENGINE_OPTIONS = {"max_identifier_length": 128}
    # This is used to encrypt the auth_token
    SECRET_KEY = 't0p s3cr3t'
    # This is used to encrypt the admin passwords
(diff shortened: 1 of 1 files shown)


doc/installation/system/inifile.rst
```diff
@@ -52,6 +54,14 @@ The file should contain the following contents::
 You may want to use the MySQL database or Maria DB. There are two possible
 drivers, to connect to this database. Please read :ref:`mysqldb`.
 
+``SQLALCHEMY_ENGINE_OPTIONS`` is a dictionary of keyword args to send
+to `create_engine() <https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy
+.create_engine>`_. `max_identifier_length <https://docs.sqlalchemy.org/en/14/core/engines
+.html#sqlalchemy.create_engine.params.max_identifier_length>`_ is the database’s
+configured maximum number of characters that may be used in a SQL identifier
+such as a table name, column name, or label name. ``max_identifier_length``
+should be set to *128* to avoid error messages.
```
cornelinux, line 63: Please replace

    ``max_identifier_length`` should be set to *128* to avoid error messages.

with

      For Oracle version 19 and above the `max_identifier_length` should be set to 128.

Since setting this value is only necessary for Oracle installations.

doc/installation/system/inifile.rst
```diff
@@ -25,6 +25,8 @@ The file should contain the following contents::
    SUPERUSER_REALM = ['super', 'administrators']
    # Your database
    SQLALCHEMY_DATABASE_URI = 'sqlite:////etc/privacyidea/data.sqlite'
+   # Set maximum identifier length to 128
+   SQLALCHEMY_ENGINE_OPTIONS = {"max_identifier_length": 128}
```
cornelinux, line 29: Please add this as comment

    # SQLALCHEMY_ENGINE_OPTIONS = {"max_identifier_length": 128}

cornelinux (changes requested):

doc/installation/system/inifile.rst
NNLemling: Fixed in 311611ae32b43837bfd4427d1cd6766ab188361c.

doc/installation/system/inifile.rst
NNLemling: Fixed in 311611ae32b43837bfd4427d1cd6766ab188361c.

cornelinux (approved):

PR #3014, opened 2022-02-20
Allow administrator to change the username of the user

### LINKED ISSUE #3009: Allow to change the username via the API
In an editable resolver the username can not be modified.

Well, the library function allows to do so. Also the UI has a lineedit field to modify the username.
However, it looks like the API uses the username to identify the user, who should be edited.

When the username is changed in the UI, then the API request contains this new username and this user can not be found.

It would be great to allow changing the username.
So we would have to change the API this way, that it either uses the userid or the **old** username.

cornelinux:
The admin who can edit a user can also change the username.

Closes #3009


NNLemling (approved):
Looks good!

PR #3015, opened 2022-02-22
#2895/Remove error code from the return_message of the sms and email token

### LINKED ISSUE #2895: SMS error message appended to return information
When an error occurs while sending an SMS (for SMS Token), the error message will be appended to the information returned to the client. This might get displayed to the user and thus disclose internal configuration:
https://github.com/privacyidea/privacyidea/blob/db7cb708e16e5226cdcc17e346e0eb65fd8f1b0e/privacyidea/lib/tokens/smstoken.py#L350
The same is true for the Email-Token.
We might want to reduce the information send/presented to the user and just log the error message in the internal log.

jona-samuel:
Remove error code from the return_message of the sms and email token

Closes #2895

### DIFF
--- a/privacyidea/lib/tokens/emailtoken.py
+++ b/privacyidea/lib/tokens/emailtoken.py
@@ -308,7 +308,8 @@ class EmailTokenClass(HotpTokenClass):
                         "EMail could not be sent: %r" % e)
                 log.warning(info)
                 log.debug(u"{0!s}".format(traceback.format_exc()))
-                return_message = info
+                return_message = info = ("The PIN was correct, but the "
+                        "EMail could not be sent")
                 if is_true(options.get("exception")):
                     raise Exception(info)
 
--- a/privacyidea/lib/tokens/emailtoken.py
+++ b/privacyidea/lib/tokens/emailtoken.py
@@ -252,7 +252,7 @@ class EmailTokenClass(HotpTokenClass):
         :param transactionid: the id of this challenge
         :param options: the request context parameters / data
             You can pass ``exception=1`` to raise an exception, if
-            the SMS could not be sent. Otherwise the message is contained in the response.
+            the SMS could not be sent.
         :return: tuple
             of (success, message, transactionid, attributes)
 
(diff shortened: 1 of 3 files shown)


tests/test_lib_tokens_sms.py
```diff
@@ -490,7 +490,6 @@ def test_21_failed_loading(self):
                                "HttpSMSProvider.HttpSMSProvider")
         c = token.create_challenge(transactionid)
         self.assertFalse(c[0], c)
-        self.assertTrue("Failed to load sms.providerConfig" in c[1], c[1])
```
plettich, line 493: We should check that the correct message was logged. This can be don with the `log_capture()`decorator from the [`testfixtures`](https://testfixtures.readthedocs.io/en/latest/logging.html#the-decorator) package.

plettich (changes requested):

privacyidea/lib/tokens/emailtoken.py
```diff
@@ -252,7 +252,7 @@ def create_challenge(self, transactionid=None, options=None):
         :param transactionid: the id of this challenge
         :param options: the request context parameters / data
             You can pass ``exception=1`` to raise an exception, if
-            the SMS could not be sent. Otherwise the message is contained in the response.
+            the SMS could not be sent.
```
plettich, line 255: this should be "Email" here

privacyidea/lib/tokens/emailtoken.py
```diff
@@ -304,9 +304,9 @@ def create_challenge(self, transactionid=None, options=None):
                     mimetype=mimetype)
 
             except Exception as e:
-                info = ("The PIN was correct, but the "
-                        "EMail could not be sent: %r" % e)
-                log.warning(info)
+                info = _("The PIN was correct, but the "
+                         "EMail could not be sent")
+                log.warning(info + "({0!s})".format(e))
```
plettich, line 309: We should have consistent strings. Here is no punctuation, in the SMSToken there is a full-stop. Maybe an exclamation mark would be fitting here.
Also there is no space between the info string and the added error message in the warning.

privacyidea/lib/tokens/smstoken.py
```diff
@@ -349,10 +349,10 @@ def create_challenge(self, transactionid=None, options=None):
...
                 transactionid = transactionid or db_challenge.transaction_id
             except Exception as e:
-                info = ("The PIN was correct, but the "
-                        "SMS could not be sent: %r" % e)
-                log.warning(info)
-                log.debug("{0!s}".format(traceback.format_exc()))
+                info = _("The PIN was correct, but the "
+                         "SMS could not be sent.")
+                log.warning(info + " ({0!s})".format(e))
+                log.debug(u"{0!s}".format(traceback.format_exc()))
```
plettich, line 355: the `u` is not needed here, there shouldn't be any unicode issues in the stack-trace.

plettich (changes requested):

plettich (approved):
Looks good

PR #3022, opened 2022-02-25
Ignore CA_FILE if TLS_VERIFY is disabled

### LINKED ISSUE #2892: Creating TLS context fails when no TLS_CA_FILE is given even if TLS_VERIFY is false
We are initializing the TLS context when creating the LDAPIdResolver here:
https://github.com/privacyidea/privacyidea/blob/28257da4cacdd0ed68ce4eb6e6b65e0c498ea884/privacyidea/lib/resolvers/LDAPIdResolver.py#L821
The `tls_ca_file` parameter is always set (either from the config or with `DEFAULT_CA_FILE`.
When creating the TLS context, the content of `tls_ca_file` is passed independent of the `tls_verify` setting:
https://github.com/privacyidea/privacyidea/blob/28257da4cacdd0ed68ce4eb6e6b65e0c498ea884/privacyidea/lib/resolvers/LDAPIdResolver.py#L529
Unfortunately testing the connection won't catch this error since the `_get_tls_context()` function is called with the `tls_ca_file` parameter correctly set to `None` in case it is missing:
https://github.com/privacyidea/privacyidea/blob/28257da4cacdd0ed68ce4eb6e6b65e0c498ea884/privacyidea/lib/resolvers/LDAPIdResolver.py#L1033

plettich:
In case TLS_VERIFY was disabled but a TLS_CA_FILE was given, creating
the TLS-Context threw an error.

Fixes #2892

### DIFF
--- a/privacyidea/lib/resolvers/LDAPIdResolver.py
+++ b/privacyidea/lib/resolvers/LDAPIdResolver.py
@@ -526,6 +526,7 @@ class IdResolver (UserIdResolver):
                 tls_ca_file = tls_ca_file or DEFAULT_CA_FILE
             else:
                 tls_verify = ssl.CERT_NONE
+                tls_ca_file = None
             tls_context = Tls(validate=tls_verify,
                               version=int(tls_version),
                               ssl_options=tls_options,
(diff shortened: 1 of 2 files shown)


privacyidea/lib/resolvers/LDAPIdResolver.py
```diff
@@ -817,11 +818,11 @@ def loadConfig(self, config):
         self.tls_verify = is_true(config.get("TLS_VERIFY", False))
         # Fallback to DEFAULT_TLS_PROTOCOL (TLSv1: 3, TLSv1.1: 4, v1.2: 5, TLS negotiation: 2)
         self.tls_version = int(config.get("TLS_VERSION") or DEFAULT_TLS_PROTOCOL)
-        self.tls_ca_file = config.get("TLS_CA_FILE") or DEFAULT_CA_FILE
```
cornelinux, line 820: Why remove the default file? (See the definition at the beginning of this file)

cornelinux (commented):
Please see my comment

privacyidea/lib/resolvers/LDAPIdResolver.py
plettich: The default file will be set in `_get_tls_context()` here:
https://github.com/privacyidea/privacyidea/blob/02f9e89507416e90369da6f5259729935624a834/privacyidea/lib/resolvers/LDAPIdResolver.py#L526
This is the same behavior as with `testconnection()` when no `TLS_CA_FILE` is set in the LDAP configuration.

PR #3027, opened 2022-03-03
2577/highlight policy search action

### LINKED ISSUE #2577: highlight search term in policy action search

plettich:
Closes #2577


privacyidea/static/components/config/views/config.policies.details.html
```diff
@@ -308,77 +331,76 @@ <h3 translate>Action</h3>
                                  ng-repeat="actionGroup in actionGroups">
                                 <table class="table table-hover">
                                     <tr ng-repeat="action in actions"
-                                        ng-show="(actionCheckBox[action.name] ||
-                                    !onlySelectedVisible) &&
-                                    (action.name.toLowerCase().match(action_filter.toLowerCase()) !== null ||
-                                    action.desc.toLowerCase().match(action_filter.toLowerCase()) !== null)"
+                                        ng-show="checkOpenGroup(action, action_filter)"
```
cornelinux, line 334: Where does this come from? 
Is it an internal function - waht does checkOpenGroup do?

privacyidea/static/package-lock.json
```diff
@@ -16,14 +16,12 @@
         "angular-sanitize": "^1.8.2",
         "angular-ui-bootstrap": "^2.5.6",
         "angular-ui-router": "^1.0.29",
-        "angular-ui-utils": "^0.1.1",
```
cornelinux, line 19: Why don't we need this anymore - was it a remnant?

privacyidea/static/templates/footer.html
```diff
@@ -28,7 +28,6 @@
 <script src="{{ instance }}/{{ "static/contrib/js/ngmodules/angular-idle.js" | versioned }}"></script>
 <script src="{{ instance }}/{{ "static/contrib/js/ngmodules/angular-gettext.js" | versioned }}"></script>
 <script src="{{ instance }}/{{  "static/components/translation/translations.js" | versioned }}"></script>
-<script src="{{ instance }}/{{ "static/contrib/js/ngmodules/highlight.js" | versioned }}"></script>
```
cornelinux, line 31: Interesting, so was the hightlight.js a remnant from a previous try or did we actually used it?
Just for me to understand: And this is now done by our own `filter`, right?

cornelinux (changes requested):
I am sorry - we have this cool highlighting in the policy actions where we search.
Hm, the event handler conditions have the same search mechansim. Should we do it there, too?

If not now, then in a new issue.

See #3062

privacyidea/static/components/config/views/config.policies.details.html
cornelinux: OK, thank you, found it.

privacyidea/static/package-lock.json
cornelinux: UNderstood. Only highlight.js was used.

cornelinux (approved):

PR #3037, opened 2022-03-10
Enable WebUI-login with Push-Token

### LINKED ISSUE #2893: privacyIDEA login with push-token doesn't work
Since 04285e30 the login to privacyIDEA with push token (and possibly other out-of-band tokens) does not work anymore.
First it fails here:
https://github.com/privacyidea/privacyidea/blob/8bca25e053bf763737ff578a3d8500a32265731d/privacyidea/static/components/login/controllers/loginControllers.js#L254
because the `attributes` variable does not exist anymore.
Second it won't poll the server for answered challenges since the `poll` variable isn't set:
https://github.com/privacyidea/privacyidea/blob/8bca25e053bf763737ff578a3d8500a32265731d/privacyidea/static/components/login/controllers/loginControllers.js#L272

plettich:
Due to changes in the return value of `create_challenge()` the
Push-Token (and also Webauthn/U2f-token) didn't work when logging in to
the WebUI (policy `login_mode`).
The loginController now checks the correct attributes in each challenge.
The Push-Token now returns the `hideResponseInput` attribute.

TODO:
- When using TOTP/HOTP with challenge-response, the input field is
  not shown (missing `hideResponseInput` attribute in response).
- The polling from the WebUI is not aborted when the login state is
  left.
- The Webauthn/U2F flow is not aborted when logged in with a different
  token.

Closes #2893

### DIFF
--- a/privacyidea/lib/tokens/pushtoken.py
+++ b/privacyidea/lib/tokens/pushtoken.py
@@ -821,7 +821,6 @@ class PushTokenClass(TokenClass):
                                                  ACTION.CHALLENGETEXT,
                                                  options) or DEFAULT_CHALLENGE_TEXT
 
-        reply_dict = {}
         data = None
         # Initially we assume there is no error from Firebase
         res = True
@@ -879,6 +878,7 @@ class PushTokenClass(TokenClass):
             if is_true(options.get("exception")):
                 raise ValidateError("The token has no tokeninfo. Can not send via Firebase service.")
 
+        reply_dict = {"attributes": {"hideResponseInput": self.client_mode != CLIENTMODE.INTERACTIVE}}
         return True, message, transactionid, reply_dict
 
     @check_token_locked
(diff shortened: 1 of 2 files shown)


cornelinux:
> * When using TOTP/HOTP with challenge-response, the input field is
>   not shown (missing `hideResponseInput` attribute in response).

@plettich is it already this way? Or is an effect of this PR? It is crucial that chalresp with TOTP and HOTP actually works in the UI, since this is widely used.

plettich:
> > * When using TOTP/HOTP with challenge-response, the input field is
> >   not shown (missing `hideResponseInput` attribute in response).
> 
> @plettich is it already this way? Or is an effect of this PR? It is crucial that chalresp with TOTP and HOTP actually works in the UI, since this is widely used.

If not combined with an out-of-band or webauthn/u2f token it works, otherwise the input field is not shown.

cornelinux (approved):

PR #3043, opened 2022-03-14
Use HSM for encrypting the key file

### LINKED ISSUE #3003: Use a HSM only for encrypting the encryption key.
Only store the Key Encryption Key in the HSM. The data encryption key is stored on the harddisk and decrypted by the KEK when the privacyIDEA process starts (the security module is initialized).

This would be a modified `aeshsm.py` module.

cornelinux:
The HSM is now only used to decrypt the encrypted key file.
The encryption keys are decrypted, loaded into memory
and kept there for the lifetime of the process.

Closes #3003

### DIFF
--- a/doc/installation/system/securitymodule.rst
+++ b/doc/installation/system/securitymodule.rst
@@ -105,3 +105,50 @@ this prefix, ``_`` and the key identifier (respectively ``token``,
 
 ``PI_HSM_MODULE_KEY_LABEL_VALUE`` is the label for ``value`` key
 (defaults to value based on ``PI_HSM_MODULE_KEY_LABEL`` setting).
+
+Encrypt Key Security Module
+---------------------------
+
+The Encrypt Key Security Module uses a hardware security module (HSM)
+to decrypt the encrypted encryption key. Within the HSM a private RSA key is
+used to decrypt an encrypted file like `/etc/privacyidea/enckey.enc`.
+
+With the first request to each process of the privacyIDEA server, the HSM is used
+to decrypt the encryption key. After that the encryption key is kept in memory during run time.
+
+To activate this module add the following to the configuration file
+(:ref:`cfgfile`)
+
+    PI_HSM_MODULE = "privacyidea.lib.security.encryptkey.EncryptKeyHardwareSecurityModule"
+
+Further attributes are
+``PI_HSM_MODULE_MODULE`` which takes the pkcs11 library. This is the full
+specified path to the shared object file in the file system.
+
+``PI_HSM_MODULE_SLOT`` is the slot on the HSM where the keys are
+located. This is an integer value.
+Alternatively you can specify ``PI_HSM_MODULE_SLOTNAME`` which would be the descriptive name
+of this slot.
+
+To use the correct key in this slot you can either specify the key by providing
+``PI_HSM_MODULE_KEYID`` with the integer id of the key or
+``PI_HSM_MODULE_KEYLABEL`  with the descriptive label of the key.
+
+.. note:: Some HSM fail to provide a correct keyid and it is necessary to use the key label.
+
+The last two mandatory attributes are ``PI_HSM_MODULE_PASSWORD`` which holds the password of the slot
+and ``PI_HSM_MODULE_ENCFILE`` which specifies the encrypted encryption key.
+
+You could e.g. use a Yubikey this way::
+
+    PI_HSM_MODULE = "privacyidea.lib.security.encryptkey.EncryptKeyHardwareSecurityModule"
+    PI_HSM_MODULE_MODULE = "/usr/lib/libykcs11.so"
+    PI_HSM_MODULE_SLOTNAME = "Yubico YubiKey"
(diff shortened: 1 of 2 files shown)


cornelinux:
This PR uses CKA_RSA_PKCS as padding mechanism which was quick and easy to achive for a PoC.
I think however, we should move to CKM_RSA_PKCS_PSS (If this is possible at all) or to CKM_RSA_PKCS_OAEP - which I initially skipped for PoC reasons since I need to define a template for OAEP. 
**Warning**: Some hardware devices or drivers might not support OAEP, so after all it could be we need to make it configurable if we want to allow to not use PKCS1 padding.

privacyidea/lib/security/encryptkey.py
```diff
@@ -0,0 +1,346 @@
...
+            config["keylabel"] = a
+        elif o in ("-i", "--infile"):
+            infile = a
+        elif o in ("-o", "--outfile"):
+            outfile = a
+        elif o in ("-e", "--encfile"):
+            config["encfile"] = a
+        elif o in ("-l", "--listpublic"):
+            listkeys = "public"
+        elif o in ("-l", "--listprivate"):
```
plettich, line 310: there are two `-l` parameter defined

privacyidea/lib/security/encryptkey.py
```diff
@@ -0,0 +1,346 @@
...
+            listkeys = "private"
```
plettich, line 311: there is no `help` option

privacyidea/lib/security/encryptkey.py
```diff
@@ -0,0 +1,346 @@
...
+
+    p = EncryptKeyHardwareSecurityModule(config, logout=False)
```
plettich, line 313: Currently this bails out with a python error if some configuration parameter is missing. Should we add some descriptions in case of error?

doc/installation/system/securitymodule.rst
```diff
@@ -104,4 +104,51 @@ this prefix, ``_`` and the key identifier (respectively ``token``,
...
+With the first request to each process of the privacyIDEA server, the HSM is used
+to decrypt the encryption key. After that the encryption key is kept in memory during run time.
+
+To activate this module add the following to the configuration file
+(:ref:`cfgfile`)
+
+    PI_HSM_MODULE = "privacyidea.lib.security.encryptkey.EncryptKeyHardwareSecurityModule"
+
+Further attributes are
+``PI_HSM_MODULE_MODULE`` which takes the pkcs11 library. This is the full
```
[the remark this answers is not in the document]
cornelinux: Willdo

privacyidea/lib/security/encryptkey.py
```diff
@@ -0,0 +1,346 @@
...
+
+
+__doc__ = """
+This is a PKCS11 Security module that decrypts a given encrypted key file
+in the HSM with an asymmetric private key.
+
+Please see the docstring of the SecurityModule below, on how to configure this
+module in pi.cfg.
+
+  python encryptkey.py --module /usr/lib/libykcs11.so --keyid 1 --slotname "Yubico YubiKey" \
```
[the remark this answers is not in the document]
cornelinux: Will do

privacyidea/lib/security/encryptkey.py
```diff
@@ -0,0 +1,346 @@
...
+            log.error("No keyid or keylabel defined.")
+            raise HSMException("No keyid or keylabel defined.")
+        self.keyid = self.config.get("keyid")
+        self.keylabel = self.config.get("keylabel")
+
+        if "password" in self.config:
+            self.password = self.config.get("password")
+        else:
+            log.error("No password specified.")
+            self.password = getpass()
```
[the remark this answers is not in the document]
cornelinux: No. It only works if the module is used as a script

privacyidea/lib/security/encryptkey.py
```diff
@@ -0,0 +1,346 @@
...
+                    for slot in slotlist:
+                        # Find the slot via the slotname
+                        slotinfo = self.pkcs11.getSlotInfo(slot)
+                        log.debug("Found slot '{}'".format(slotinfo.slotDescription))
+                        if slotinfo.slotDescription.startswith(self.slotname):
+                            self.slot = slot
+                            break
+
+            if self.slot not in slotlist:
+                raise HSMException("Slot {0:d} not present".format(self.slot))
```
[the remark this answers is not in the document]
cornelinux: Will do

privacyidea/lib/security/encryptkey.py
cornelinux: Will do

privacyidea/lib/security/encryptkey.py
cornelinux: You are right. I will add a comment to the docs, since noone would have the idea to run this module anyways.

privacyidea/lib/security/encryptkey.py
cornelinux: Hm, as this will be probably only used three times in two years I do not think so.

PR #3065, opened 2022-03-24
Make security module preloadable

cornelinux:
with apache2 and mod_wsgi we can now preload the encryption keys
before issuing the first request.

### DIFF
--- a/doc/installation/system/securitymodule.rst
+++ b/doc/installation/system/securitymodule.rst
@@ -160,3 +160,21 @@ If your key in the HSM is identified by a key label, then you can encrypt the ex
     python encryptkey.py --module /usr/lib/libykcs11.so --keylabel "my secret key" --slotname "Yubico YubiKey" \
                          --infile enckey --outfile enckey.enc
 
+Preloading of encryption keys
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This security module allows you to preload the encryption keys. I.e. privacyIDEA can use the HSM to decrypt
+the keys before the first request is sent to privacyIDEA. To do so, you need to modify the wsgi script
+(See :ref:`wsgiscript`) and add the parameter `init_hsm`::
+
+    application = create_app(config_name="production",
+                             config_file="/etc/privacyidea/pi.cfg", init_hsm=True)
+
+Moreover, you need to add the `WSGIImportScript` statement to your Apache2 configuration::
+
+    WSGIApplicationGroup %{GLOBAL}
+    WSGIImportScript /etc/privacyidea/privacyideaapp.wsgi process-group=privacyidea application-group=%{GLOBAL}
+
+.. note:: Please note, that this security module uses a lock file, to handle concurrent access to the HSM.
+   In certain cases of errors the log file could remain and not cleaned up.
+   Ensure, that the directory `/dev/shm/pilock/` does *not* exist at Apache2 startup.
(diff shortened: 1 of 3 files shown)


privacyidea/lib/security/encryptkey.py
```diff
@@ -73,10 +73,14 @@ def hsm_lock(timeout=DEFAULT_TIMEOUT):
...
         try:
             # Try to get the Lock, since we are acting.
+            log.debug("Requesting lock")
             os.mkdir(DIR)
+            log.debug("Lock acquired")
             yield
             # Cleanup
+            log.debug("Going to release lock")
             os.rmdir(DIR)
+            log.debug("Lock released")
```
plettich, line 83: We should add a `finally` block to delete the lock directory in the end.

plettich (changes requested):
Maybe the `finally` block will get rid of stale the lock directory

privacyidea/lib/security/encryptkey.py
cornelinux: Giving this a second thought it is not clear to me, if we really need a finally to do this or what we actually try to accomplish with a finally.

I suspect we would only want to ensure, that - if the `yield` raises some exceptions which we do not catch, we want to delete the lock file.
If we go through the loop for 15 seconds and simply can not aquire the lock, then we do not want to release the lock file, since another process very probably still has the lock.

So in my understanding it could look like this:

~~~~python
while timeout > 0:
    timeout =- 1
    try:
        lock()
        i_created_the_lock = True
        yield       # <- the problem is here, if we managed to lock, but bail out.
        release_lock()
        break
    except FileExistsError:
        # We can not get the lock and try again!
        sleep(1)
    except Exception:
         # Catch all other errors that occur in yield?
         if i_created_the_lock:
            release_lock() # Although I have not accomplished anything :-/
~~~~

OK, or we could do it like this:

~~~~python
while timeout > 0:
    timeout =- 1
    try:
        lock()
        i_created_the_lock = True
        yield       # <- the problem is here, if we managed to lock, but bail out.
        break
    except FileExistsError:
        # We can not get the lock and try again!
        sleep(1)
    finally:
         if i_created_the_lock:
            release_lock() # Although I have not accomplished anything :-/
~~~~

cornelinux: nicely done in 29a86dc2

doc/installation/system/securitymodule.rst
```diff
@@ -134,7 +134,10 @@ To use the correct key in this slot you can either specify the key by providing
 ``PI_HSM_MODULE_KEYID`` with the integer id of the key or
 ``PI_HSM_MODULE_KEYLABEL``  with the descriptive label of the key.
 
-The ``PI_HSM_MODULE_TIMEOUT`` can be used to define an integer value for a HSM lock timeout.
+The ``PI_HSM_MODULE_TIMEOUT`` can be used to define an integer value for a HSM lock timeout. The default is 15 seconds.
+
+Using the key ``PI_HSM_MODULE_LOCK_DIR`` you can define a different locking directory.
```
plettich, line 139: You might want to explain that the directory itself is the lock and should not exist before starting the service, otherwise people might create the directory beforehand.

privacyidea/lib/security/encryptkey.py
```diff
@@ -144,11 +152,15 @@ def __init__(self, config=None, logout=True):
         self.is_ready = False
 
         timeout = self.config.get("timeout") or DEFAULT_TIMEOUT
+        lock_dir = self.config.get("lock_dir") or DEFAULT_LOCK_DIR
```
plettich, line 155: the keys in the config are `PI_HSM_MODULE_LOCK_DIR` and `PI_HSM_MODULE_TIMEOUT`.

plettich (changes requested):

privacyidea/lib/security/encryptkey.py
cornelinux: Thats right. This is the way, how it works.

PR #3079, opened 2022-04-01
remove unnessecary input fields

### LINKED ISSUE #3081: make Firebase only depend on service account
With the new app version, the way Firebase works will be changed.

The old way was, that users had to create their own Firebase instance and specify values, so a connection could be established.

With the new app version, there will only be one Firebase instance, and users will only need to specify their service account

jacekwiczkowski:
With the new app version, the way Firebase works will be changed.

The old way was, that users had to create their own Firebase instance and specify values, so a connection could be established.

With the new app version, there will only be one Firebase instance, and users will only need to specify their service account

image

Fixes #3081


jacekwiczkowski:
#3081

jacekwiczkowski:
will fix tests later today

cornelinux:
Thanks a lot for making this easy. See my comment, we will probably refactor this PR with a new Provider.

PR #3091, opened 2022-04-08
Fix missing ttype validation

### LINKED ISSUE #3090: Bad error handling in /ttype/<type> endpoint
We observed a bad error handling when requesting the `/ttype/<type>` endpoint.

In specific, we faced the following error:
```
[ERROR][privacyidea.app:1892] Exception on /ttype/push"}. [GET]
...
AttributeError: 'NoneType' object has no attribute 'api_endpoint'
```
Actually, we could fix the problem but it would be nice to fix this upstream right away.

### Top-level intent

Access the `/ttype/<type>` endpoint.

### Steps to reproduce

1. Query `/ttype/test` endpoint
2. There will be a NoneType error in the logs.

### Expected outcome

Proper error handling

### Actual outcome

NoneType exception.

### Configuration

* **privacyIDEA version**: v3.6.3
* **Installation method**: (from Ubuntu packages, github, PyPI, ...)
* **Python version**: 3
* **Operating system**: linux
* **Webserver**: apache

### Log file

Happy2C0de:
Fixes #3090

### DIFF
--- a/privacyidea/lib/config.py
+++ b/privacyidea/lib/config.py
@@ -43,6 +43,7 @@ import threading
 from .log import log_with
 from ..models import (Config, db, Resolver, Realm, PRIVACYIDEA_TIMESTAMP,
                       save_config_timestamp, Policy, EventHandler)
+from privacyidea.lib.error import ParameterError
 from privacyidea.lib.framework import get_request_local_store, get_app_config_value, get_app_local_store
 from privacyidea.lib.utils import to_list
 from privacyidea.lib.utils.export import (register_import, register_export)
@@ -450,6 +451,9 @@ def get_token_class(tokentype):
         if tclass.get_class_type().lower() == tokentype.lower():
             tokenclass = tclass
             break
+
+    if tokenclass is None:
+        raise ParameterError(u"Invalid tokentype provided. ttype: {}".format(tokentype.lower()))
 
     return tokenclass
 
--- a/privacyidea/lib/config.py
+++ b/privacyidea/lib/config.py
@@ -43,7 +43,6 @@ import threading
 from .log import log_with
 from ..models import (Config, db, Resolver, Realm, PRIVACYIDEA_TIMESTAMP,
                       save_config_timestamp, Policy, EventHandler)
-from privacyidea.lib.error import ParameterError
 from privacyidea.lib.framework import get_request_local_store, get_app_config_value, get_app_local_store
 from privacyidea.lib.utils import to_list
 from privacyidea.lib.utils.export import (register_import, register_export)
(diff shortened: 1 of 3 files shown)


privacyidea/lib/config.py
```diff
@@ -43,6 +43,7 @@
 from .log import log_with
 from ..models import (Config, db, Resolver, Realm, PRIVACYIDEA_TIMESTAMP,
                       save_config_timestamp, Policy, EventHandler)
+from privacyidea.lib.error import ValidateError
```
cornelinux, line 46: I think we should use a `ParameterError`. The `ValidateError` is used, if the some error occurred during the authentication process.

In a normal authentication process you the application would not get the idea to ask for a non-existing tokentype, so this is rather an application passing wrong parameters.

privacyidea/lib/config.py
```diff
@@ -451,6 +452,9 @@ def get_token_class(tokentype):
             tokenclass = tclass
             break
 
+    if tokenclass is None:
+        raise ValidateError("invalid tokentype provided. ttype: {}".format(tokentype.lower()))
```
cornelinux, line 456: If I would call 

    /ttype/ööäääh

raising the ValidateError would raise an EncodingError. So you should do

~~~~python
raise ParameterError(u"Invalid tokentype provided. ttype: {}".format(tokentype.lower()))
~~~~

cornelinux (changes requested):
Please issue your PR against the branch-3.7, this way the fix will be published with the next release 3.7.1.

privacyidea/lib/config.py
Happy2C0de: Done, good input. I also thought about ParameterError but didn't find an explanation.
Changed.

privacyidea/lib/config.py
Happy2C0de: Adjusted

Happy2C0de:
@cornelinux 
I adjusted the error type and rebased the branch to `branch-3.7` in order to adjust the target branch.
LGTM

cornelinux (approved):

privacyidea/lib/config.py
```diff
@@ -451,6 +452,9 @@ def get_token_class(tokentype):
             tokenclass = tclass
             break
 
+    if tokenclass is None:
+        raise ParameterError(u"Invalid tokentype provided. ttype: {}".format(tokentype.lower()))
```
cornelinux, line 456: @Happy2C0de  It looks as if your change is too low level here, that causes several tests to fail!

Happy2C0de: @cornelinux 
I moved the error handling to `ttype.py`.
LGTM

PR #3092, opened 2022-04-09
Add notification key to allow better background notification handling

### LINKED ISSUE #3117: Change data message to notification message
Currently the push notification is sent as data message.
Check, if it is sensible to send the message as notification type.

Happy2C0de:
In order to allow a better background notification handling, one should add the `"notification"` key to the firebase message.

We had the problem that the message came to the device but it was not sent to the app in background mode. This is because all the information is sent in the `data` key. This doesn't trigger a proper notification on the device.

In order to have both, data and a proper notification in the background. We should add the given key.

See the specification:
https://firebase.google.com/docs/cloud-messaging/concept-options
e.g. using:
```json
{
  "message":{
    "notification":{
      "title":"Token",
      "body":"Token"
    },
    "data" : {...}
  }
}
```
> When in the background, apps receive the notification payload in the notification tray, and only handle the data payload when the user taps on the notification.

Closes #3117

### DIFF
--- a/privacyidea/lib/smsprovider/FirebaseProvider.py
+++ b/privacyidea/lib/smsprovider/FirebaseProvider.py
@@ -124,6 +124,10 @@ class FirebaseProvider(ISMSProvider):
             "message": {
                         "data": data,
                         "token": firebase_token,
+                        "notification": {
+                            "title": data.get("title"),
+                            "body": data.get("question")
+                        },
                         "android": {
                                     "priority": "HIGH",
                                     "ttl": "120s",


Happy2C0de:
@cornelinux @nilsbehlen 
If there is no specific reason why you didn't add the `"notification"` key, I would suggest do add it in order to have a proper notification handling.

nilsbehlen:
Hi,
when the app was originally created as a native app, the notification would have buttons to respond to the authentication request directly without opening the app. This can only be done using data messages. When you use notification message, you can not modify the notification because it is created by FCM, you can only process data in the background. Since v3 and the switch to flutter, we do not have buttons in the notification anymore, but i don't know why off the top of my head.

`We had the problem that the message came to the device but it was not sent to the app in background mode. This is because all the information is sent in the data key. This doesn't trigger a proper notification on the device.`
Was is on android or ios? Can you reproduce it?
I think we need to discuss further if we want to go back to notification messages.

cornelinux:
@nilsbehlen can we discuss the necessity internally?

nilsbehlen:
sure

cornelinux (approved):

PR #3096, opened 2022-04-11
Allow smartphone to decline a Push challenge

### LINKED ISSUE #2865: push_wait does not break, if the user presses decline on the smartphone.
The push_wait only exits the waiting loop, if the challenge was answered successfully, but not if the authentication request was declined on the smartphone:

https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/tokens/pushtoken.py#L916

(see https://community.privacyidea.org/t/connection-reject-does-not-work-as-expected-when-running-freeradius-as-a-service/2094/6?u=cornelinux)

We need to check, if the user presses decline on the smartphone, if this is only a smartphone internal functionality or if the challenge on the server gets modified/deleted. If so, we need to extend the if-statement in line 916, so that the loop also breaks if the user presses decline.

(I am not sure, if the decline button really communicates to the server, how the smartphone app is designed)

cornelinux:
The challenge then gets deleted in the database.

Closes #2865

### DIFF
--- a/privacyidea/lib/tokens/pushtoken.py
+++ b/privacyidea/lib/tokens/pushtoken.py
@@ -750,7 +750,7 @@ class PushTokenClass(TokenClass):
               Host: https://yourprivacyideaserver
 
               serial=<token serial>
-              nonce=<the actualce challenge>
+              nonce=<the actual challenge>
               decline=1
               signature=<signature over {nonce}|{serial}|decline
 
--- a/privacyidea/lib/tokens/pushtoken.py
+++ b/privacyidea/lib/tokens/pushtoken.py
@@ -567,6 +567,7 @@ class PushTokenClass(TokenClass):
             log.debug("Handling the authentication response from the smartphone.")
             challenge = getParam(request_data, "nonce")
             signature = getParam(request_data, "signature")
+            decline = is_true(getParam(request_data, "decline", default=False))
 
             # get the token_obj for the given serial:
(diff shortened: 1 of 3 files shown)


privacyidea/lib/tokens/pushtoken.py
```diff
@@ -732,7 +739,20 @@ def api_endpoint(cls, request, g):
...
+        - The smartphone can also decline the authentication request, by sending
+          a response to the server:
+
+            .. sourcecode:: http
+
+              POST /ttype/push HTTP/1.1
+              Host: https://yourprivacyideaserver
+
+              serial=<token serial>
+              nonce=<the actualce challenge>
```
nilsbehlen, line 753: typo

nilsbehlen (approved):

nilsbehlen:
How will older versions of the server behave if a smartphone sends the decline request?

privacyidea/lib/tokens/pushtoken.py
cornelinux: done in 7d8f20c4

nilsbehlen (approved):

cornelinux:
@elouin The reviewdog process is broken for some reasons for all PRs. I think sth. is wrong with the action definition, which runs in @plettich 's repo. So you may safely ignore the failing reviewdog.

elouin (approved):

PR #3104, opened 2022-04-25
mark errormessage strings for translation

### LINKED ISSUE #3066: untranslated API messages
### Top-level intent

What did you try to achieve?

Display localized error messages from privacyIDEA REST API in simpleSAMLphp. I guess there are two approaches, either do the translation on privacyIDEA server or ignore messages from server and use translated strings based on error codes.

### Steps to reproduce

1. use [simpleSAMLphp module privacyIDEA](https://github.com/privacyidea/simplesamlphp-module-privacyidea)
2. enter (for example) a wrong TOTP
3. an error message is shown, like `"Response did not match for 5 tokens"` ([source](https://github.com/privacyidea/privacyidea/blob/4ea557be5004feeed957fbee6cd78c1d417cc75f/privacyidea/lib/token.py#L2376))

### Expected outcome

The error message should be in the user's language (if available).

### Actual outcome

The error message is always in English, because the error messages are not being translated (in source code).

### Configuration

* **privacyIDEA version**: 3.7-dev4
* **Installation method**: github
* **Python version**: 3.9
* **Operating system**: debian
* **Webserver**: uwsgi

Sebastian-Lemling:
closes #3066

### DIFF
--- a/privacyidea/lib/token.py
+++ b/privacyidea/lib/token.py
@@ -150,7 +150,7 @@ def create_tokenclass_object(db_token):
         try:
             token_object = token_class(db_token)
         except Exception as e:  # pragma: no cover
-            raise TokenAdminError("create_tokenclass_object failed:  {0!r}".format(e),
+            raise TokenAdminError(_("create_tokenclass_object failed:  {0!r}").format(e),
                                   id=1609)
     else:
         log.error('type {0!r} not found in tokenclasses'.format(tokentype))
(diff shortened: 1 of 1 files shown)


privacyidea/lib/token.py
```diff
@@ -561,9 +561,9 @@ def get_one_token(*args, **kwargs):
     """
     result = get_tokens(*args, **kwargs)
     if not result:
-        raise ResourceNotFoundError(u"The requested token could not be found.")
+        raise ResourceNotFoundError(_(u"The requested token could not be found."))
```
jona-samuel, line 564: I think the 'u' is redundant, as the default string type is Unicode. In line 566 too

jona-samuel (changes requested):
There are translations missing in lines 1334 and 1873

privacyidea/lib/token.py
NNLemling: after consultation it was decided to leave the 'u' standing.

jona-samuel (approved):
Lucks good

Issue #3088, opened 2022-04-07
HTML sanitizer breaks download links for certificates


plettich:
The HTML sanitizer (introduced with #2577 ) marks the download links for certificates as unsafe. We need to add the `blob:` mechanism to the trusted URLs here:
https://github.com/privacyidea/privacyidea/blob/ad2749b3b8096abcf9e27beae11652f4f9c93dff/privacyidea/static/app.js#L158

PR #3108, opened 2022-04-27
Add blob method to sanitized URLs

plettich:
Fixes: #3088


plettich:
A `blob:` URL must be local, otherwise the browser doesn't allow it.

cornelinux (approved):

PR #3119, opened 2022-05-02
Add App-PIN setting for Push-Token

### LINKED ISSUE #3116: Force PIN policy for Push token
Currently there are only policies to force pin on the smartphone for HOTP and TOTP. There should be one for Push too. The app would then request the PIN/biometrics before the user is able to confirm a push authentication.

plettich:
Closes #3116

### DIFF
--- a/privacyidea/lib/tokens/pushtoken.py
+++ b/privacyidea/lib/tokens/pushtoken.py
@@ -335,6 +335,12 @@ class PushTokenClass(TokenClass):
                            'type': 'int',
                            'desc': _("The user may only have this maximum number of active Push tokens assigned."),
                            'group': GROUP.TOKEN
+                       },
+                       'push_' + ACTION.FORCE_APP_PIN: {
+                           'type': 'bool',
+                           'group': "PUSH",
+                           'desc': _('Enforce setting an app pin for the privacyIDEA '
+                                     'Authenticator App')
                        }
                    },
                    SCOPE.AUTH: {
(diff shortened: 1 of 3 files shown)


privacyidea/lib/tokens/pushtoken.py
```diff
@@ -335,6 +335,12 @@ def get_class_info(key=None, ret='all'):
                            'type': 'int',
                            'desc': _("The user may only have this maximum number of active Push tokens assigned."),
                            'group': GROUP.TOKEN
+                       },
+                       'push_' + ACTION.FORCE_APP_PIN: {
+                           'type': 'bool',
+                           'group': "PUSH",
+                           'desc': _('Enforce setting an app pin for the privacyIDEA '
```
nilsbehlen, line 342: I think this is a bit misleading since it is not a PIN for the whole app, but just for the token. I would word it like 'Enforce requesting the Smartphone PIN/Biometrics before Push requests can be accepted'

cornelinux: @nilsbehlen You are right. However, this has changed also for HOTP and TOTP...

@plettich ...I am not sure, if we want to change this *now*?

cornelinux (approved):

privacyidea/lib/tokens/pushtoken.py
plettich: @cornelinux @nilsbehlen I'll change it for the Push-Token, but for HOTP/TOTP we'll keep it that way for now and change it in v3.8.

plettich: Done in 5a7f9dd5

PR #3126, opened 2022-05-05
Fix JS error in pskc import

plettich:
The `translate` filter was not used correctly.

### DIFF
--- a/privacyidea/static/components/token/views/token.import.html
+++ b/privacyidea/static/components/token/views/token.import.html
@@ -87,7 +87,7 @@
                     check_fail_soft : 'Skip tokens that can not be verified',
                     check_fail_hard : 'Abort operation on unverifiable token'}"
                         value="{{ key }}">
-                    {{ desc | translate}}
+                    {{ desc }}
                 </option>
             </select>
         </div>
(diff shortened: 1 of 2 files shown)


privacyidea/static/components/token/views/token.import.html
```diff
@@ -82,11 +82,12 @@ <h3 translate>Import tokens from a file</h3>
...
-                <option ng-repeat="(key, desc) in {no_check: 'Do not verify the authenticity' | translate,
-                  check_fail_soft : 'Skip tokens that can not be verified' | translate,
-                  check_fail_hard : 'Abort operation on unverifiable token' | translate}"
+                <option ng-repeat="(key, desc) in {
+                    no_check: 'Do not verify the authenticity',
+                    check_fail_soft : 'Skip tokens that can not be verified',
+                    check_fail_hard : 'Abort operation on unverifiable token'}"
                         value="{{ key }}">
-                    {{ desc }}
+                    {{ desc | translate}}
```
cornelinux, line 90: I think this works out, if the strings 'Do not verify the authenticity' are already translated.

But if we change this string, grunt will not identify the string as translateable and we will not be able to translate changes in the future.

cornelinux (changes requested):
I am not sure, how we can solve this, we might have to define the strings with the `| translate` modifier as a variable, so that it can be translated and it does not break the for loop.

Or is grunt intelligent enough to identify which values `desc` can take?

cornelinux:
This does not make sense.
I have to remove it from 3.7.1.

cornelinux:
Reopening it, to at least fix the dropdown.

PR #3143, opened 2022-05-13
feat: WebAuthn alg policy options with RSA PKCS1

### LINKED ISSUE #3142: cannot enroll Windows Hello as WebAuthn token
### Top-level intent

Register Windows Hello (e.g. fingerprint) as a WebAuthn token.

### Steps to reproduce

1. Add token
2. type WebAuthn
3. look for Windows Hello prompt

### Expected outcome

Windows Hello should offer fingerprint or other screen lock options to be used as WebAuthn token. But it seems that Windows Hello only supports `RSASSA-PKCS1-v1_5 using SHA-256` (COSE -257) which is not included in any option of privacyIDEA's WebAuthn algorithm policy (options are 'ecdsa_preferred', 'ecdsa_only', 'rsassa-pss_preferred' and 'rsassa-pss_only').

The algorithm required by Windows Hello is implemented in [webauthn.py as RS256](https://github.com/privacyidea/privacyidea/blob/23f09b56e7957f3c6337514d495bbbcb898ba0ce/privacyidea/lib/tokens/webauthn.py#L188), there's just no way to use it because it is not listed in webauthntoken.py in the policy options.

### Actual outcome

Windows Hello only offers a security key (USB device such as YubiKey).

### Configuration

* **privacyIDEA version**: 3.7
* **Installation method**: pip
* **Python version**: 3.9
* **Operating system**: Debian
* **Webserver**: uwsgi

### Log file

No logs to show, happens on the frontend.

melanger:
allow Windows Hello to be used as a WebAuthn token

closes #3142

### DIFF
--- a/privacyidea/lib/tokens/webauthntoken.py
+++ b/privacyidea/lib/tokens/webauthntoken.py
@@ -432,13 +432,31 @@ PUBLIC_KEY_CREDENTIAL_ALGORITHM_PREFERENCE_OPTIONS = {
     'ecdsa_only': [
         COSE_ALGORITHM.ES256
     ],
+    'ecdsa_with_rsapkcs1': [
+        COSE_ALGORITHM.ES256,
+        COSE_ALGORITHM.RS256
+    ],
+    'ecdsa_preferred_with_rsapkcs1': [
+        COSE_ALGORITHM.ES256,
+        COSE_ALGORITHM.PS256,
+        COSE_ALGORITHM.RS256
+    ],
     'rsassa-pss_preferred': [
         COSE_ALGORITHM.PS256,
         COSE_ALGORITHM.ES256
     ],
     'rsassa-pss_only': [
         COSE_ALGORITHM.PS256
-    ]
+    ],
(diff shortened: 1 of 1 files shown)


melanger:
@cornelinux There is an issue with this solution, because the JavaScript library (webauthn-client) supports at most 2 algorithms, although there may be more:
https://github.com/privacyidea/privacyidea/blob/6a74ff77cebf429b47f5b2f67f3bedd06f020f1d/privacyidea/lib/tokens/webauthntoken.py#L994
Therefore even though I set the option to 3 algorithms, the JavaScript only gets two. I created https://github.com/privacyidea/webauthn-client/issues/13 for that.

melanger:
I have successfully tested the proposed new option `ecdsa_with_rsapkcs1` on Windows 11 with privacyIDEA v3.7 with this patch + patch from #3151 applied. I have tested various devices and all of them support only ES256 except Windows Hello which only supports RS256, so currently this option should fit all devices.

plettich:
Closing in favor of #3280

PR #3185, opened 2022-06-15
More settings in admin console (WebUI)

JGG-Apps:
Actually when we access to the Web console in admin, we didn't set any thing about the admin user environment himself, like:

- Timeout delay before auto logoff
- Force default language (don't want to use the web browser determine the language)
- Contact information (email) to inform if newer version is available
- Update alert
- Etc.


cornelinux:
This pull request does not contain what it claims.

PR #3193, opened 2022-06-17
Add user-object in DEL /user request

### LINKED ISSUE #3192: Make user available in user_delete
The event handler has no access on the user information in the event of `DELETE /user/`.
This is due to the fact, that the parameters (username, resolver) of this blueprint 
https://github.com/privacyidea/privacyidea/blob/0f085fd2ed59cd4567989c9bbe55ea01c29ced3a/privacyidea/api/user.py#L221
are not used to create the `request.User` object.

We could change the names of the parameters (but this would create a user object in the default realm!) or we could add the new logic to `before_after`.

We should add this to a 3.7 patch release.

cornelinux:
This way we can use all user information in the event handlers.

Fixes #3192

### DIFF
--- a/privacyidea/api/before_after.py
+++ b/privacyidea/api/before_after.py
@@ -102,11 +102,22 @@ def teardown_request(exc):
 
 @token_blueprint.before_request
 @audit_blueprint.before_request
-@user_blueprint.before_request
 @system_blueprint.before_request
 @user_required
 def before_user_request():
     before_request()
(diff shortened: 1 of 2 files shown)


plettich (approved):
Looks good.
In the long run we should use `user` and `resolver` as parameter for the endpoint and if this should break things this would be a good opportunity to introduce API versioning :-)

PR #3199, opened 2022-06-23
Add search filter for configured policies

### LINKED ISSUE #2788: Search configured policies
If you have configured a lot of policies you might wonder, if and how a certain setting is actually set.

"Do we have configured two step enrollment? If so, in which policy?"

A way to search the configured policies would be great.

Sebastian-Lemling:
- add pi-filter to the column "action" in the policy list view
- add general search window for quick search

closes #2788


cornelinux (approved):
Looks good to me. 
You could also drop the sorting later...

PR #3206, opened 2022-07-04
Fix number input type in policies

### LINKED ISSUE #3203: Webui policy logout_time with empty value cause 500 internal server error after relogin
### Steps to reproduce
Policy
1. under scope webui --> set policy logout_time without setting a value --> save
2. relogin --> cause 500 internal server error

### Expected outcome
Setting a default value if checkbox activated and value empty

### Actual outcome
empty value cause a 500 internal error after trying to relogin

plettich:
In the case of a number input type for a policy action, the input wasn't
correctly validated. Enforcing validation also fixes server errors when
the field was left empty.

Fixes #3203


pablo-knight (approved):
:) works

PR #3213, opened 2022-07-08
Always write passthru info to audit log

### LINKED ISSUE #3212: passthru policy in audit log
The passthru policy is only written to the audit log, if the password was correct. This can be misleading.
It should also be written to the audit log, if the passthru was actually checked.

See here: https://community.privacyidea.org/t/release-privacyidea-3-7-2/2560/7

We would simply have to move the `add_policy` calls to an upper location.
https://github.com/privacyidea/privacyidea/blob/e668d96b20c5d4480ddda30649dee5259471d7a4/privacyidea/lib/policydecorators.py#L296

cornelinux:
The name of the passthru policy will always be written to the audit log,
when passthru is checked - even if the credentials are wrong.

This information is important to admins to know, that also passthru
was checked but did not work.

Closes #3212

### DIFF
--- a/privacyidea/lib/policydecorators.py
+++ b/privacyidea/lib/policydecorators.py
@@ -290,10 +290,10 @@ def auth_user_passthru(wrapped_function, user_object, passw, options=None):
             policy_object.check_for_conflicts(pass_thru, "passthru")
             pass_thru_action = pass_thru[0].get("action").get("passthru")
             policy_name = pass_thru[0].get("name")
+            g.audit_object.add_policy([p.get("name") for p in pass_thru])
             if pass_thru_action in ["userstore", True]:
                 # Now we need to check the userstore password
                 if user_object.check_password(passw):
-                    g.audit_object.add_policy([p.get("name") for p in pass_thru])
                     return True, {"message": u"against userstore due to '{!s}'".format(
                                       policy_name)}
             else:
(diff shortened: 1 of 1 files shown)


privacyidea/lib/policydecorators.py
```diff
@@ -290,10 +290,10 @@ def auth_user_passthru(wrapped_function, user_object, passw, options=None):
             policy_object.check_for_conflicts(pass_thru, "passthru")
             pass_thru_action = pass_thru[0].get("action").get("passthru")
             policy_name = pass_thru[0].get("name")
+            g.audit_object.add_policy([p.get("name") for p in pass_thru])
```
plettich, line 293: Can't we just set `write_to_audit_log=True` in the `policies()` call above?
Ok, we would then have the `passhru` entry in the audit even if the user has a token...
Also shouldn't there be only one valid `passthru` action? Even if we have several matching policies with different priorities, only the first one will be used.

plettich (changes requested):

privacyidea/lib/policydecorators.py
cornelinux: > Ok, we would then have the `passhru` entry in the audit even if the user has a token...

This is why the policy is only added later to the audit.



> Also shouldn't there be only one valid `passthru` action? Even if we have several matching policies with different priorities, only the first one will be used.

We often (always?) do it this way. 
Why: In theory all these policies would lead to this decision. Even if currently only Policy Prio1. The bonus side effect is, that the admin knows, that Pol-Prio1 and Pol-Prio2 lead to this decision.
If we only would add Pol-Prio1 to the audit log, the admin might think to remove Pol-Prio1 with the expectation, that passthru does not work anymore. But, suprise, it will still work, due to Pol-Prio2, which was hidden before.

Bottomline: I would like to keep it this way.

plettich (approved):

PR #3214, opened 2022-07-09
3207/double webauthn

cornelinux:
Tests missing

### DIFF
--- a/privacyidea/lib/tokens/webauthn.py
+++ b/privacyidea/lib/tokens/webauthn.py
@@ -410,7 +410,8 @@ class WebAuthnMakeCredentialOptions(object):
                  icon_url=None,
                  authenticator_attachment=None,
                  authenticator_selection_list=None,
-                 location=None):
+                 location=None,
+                 credential_ids=None):
         """
         Create a new WebAuthnMakeCredentialOptions object.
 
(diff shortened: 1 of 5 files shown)


privacyidea/lib/tokens/webauthn.py
```diff
@@ -455,6 +458,19 @@ def __init__(self,
...
+        if credential_ids:
+            for cred_id in credential_ids:
+                self.exclude_credentials.append({
+                    "id": cred_id,
+                    "type": "public-key",
+                    "transports": [
+                        "usb",
+                        "ble",
+                        "nfc",
+                        "internal"
```
plettich, line 471: Maybe we should use the `DEFAULT_ALLOWED_TRANSPORTS` here:
https://github.com/privacyidea/privacyidea/blob/6a74ff77cebf429b47f5b2f67f3bedd06f020f1d/privacyidea/lib/tokens/webauthntoken.py#L417=

privacyidea/lib/tokens/webauthntoken.py
```diff
@@ -1004,6 +1021,9 @@ def get_init_detail(self, params=None, user=None):
             if (public_key_credential_creation_options.get("extensions") or {}).get("authnSel"):
                 response_detail["webAuthnRegisterRequest"]["authenticatorSelectionList"] \
                     = public_key_credential_creation_options["extensions"]["authnSel"]
+            if public_key_credential_creation_options.get("excludeCredentials"):
+                response_detail["webAuthnRegisterRequest"]["excludeCredentials"] = \
+                    public_key_credential_creation_options.get("excludeCredentials")
```
plettich, line 1026: Was there a reason why we remove these entries instead of just sending an empty list?

privacyidea/lib/tokens/webauthn.py
cornelinux: `DEFAULT_ALLOWED_TRONSPORTS` needs to be imported from `webauthntoken.py` which in turn imports `webauthn.py` (circular imports or reordering the constants)
I will currently use `TRANSPORTS` from within the `webauthn.py` file itself.

plettich (approved):

PR #3224, opened 2022-07-18
3115/allow_to_set_PKCS7_padding_in_picfg

### LINKED ISSUE #3115: Allow to choose padding in default security module
The padding of the defaul security module and the aeshsm.py is different.
Since the new encryptkey.py inherits the default security module, it will also have a different padding, than the aeshsm.py.

It would make sense to allow encryptkey.py to have the same padding like the aeshsm.py --- or make the padding configurable.

The `lib/security/default.py` would have to be changed this way to allow for the same padding like aeshsm.py:

~~~~python
 
from cryptography.hazmat.primitives.ciphers import algorithms
from cryptography.hazmat.primitives import padding 
 
 
def encrypt(self, data, iv, key_id=SecurityModule.TOKEN_KEY):
        if self.is_ready is False:
            raise HSMException('setup of security module incomplete')

        key = self._get_secret(key_id)

        # convert input to ascii, so we can securely append bin data for padding
        input_data = binascii.b2a_hex(data)
        padder = padding.PKCS7(algorithms.AES.block_size).padder()
        padded_data = padder.update(data) + padder.finalize()
        res = aes_cbc_encrypt(key, iv, padded_data)

        if self.crypted is False:
            zerome(key)
            del key
        return res

def decrypt(self, enc_data, iv, key_id=SecurityModule.TOKEN_KEY):
        if self.is_ready is False:
            raise HSMException('setup of security module incomplete')

        key = self._get_secret(key_id)
        output = aes_cbc_decrypt(key, iv, enc_data)
        # remove padding
        unpadder = padding.PKCS7(algorithms.AES.block_size).unpadder()
        data = unpadder.update(output) + unpadder.finalize()

        if self.crypted is False:
            zerome(key)
            del key

        return data
~~~~

To chose this padding over the old standard padding, we can add a config key:

    PI_HSM_PADDING = "PKCS7"

This might be related to #3005

NNLemling:
- [x] implement PKCS7 padding in [privacyidea/lib/security/default.py](https://github.com/privacyidea/privacyidea/blob/23a60012e30e1bef46ce375ee694ff6a0e20727f/privacyidea/lib/security/default.py)
- [x] Option for configuration in the pi.cfg
- [x] Implement tests
- [x] Add documentation 

closes #3115

### DIFF
--- a/doc/installation/system/inifile.rst
+++ b/doc/installation/system/inifile.rst
@@ -102,10 +102,15 @@ can be set, for example::
 Further information on possible parameters can be found in the
 `PassLib documentation <https://passlib.readthedocs.io/en/stable/lib/passlib.hash.html>`_.
 
-In ``PI_HSM_PADDING`` the padding can be set to "PKCS7". If ``PI_HSM_PADDING`` is not set, the default internal
-padding is used::
+In ``PI_HSM_PADDING`` the padding used by the `DefaultSecurityModule` or the `EncryptKeyHardwareSecurityModule` can be set to "PKCS7". 
+If ``PI_HSM_PADDING`` is not set, the default internal padding is used::
 
     PI_HSM_PADDING = "PKCS7"
+    
+.. note:: The `AESHardwareSecurityModule` always uses PKCS7 padding.
+    
+.. warning:: The two paddings are not compatible. Data encrypted with one padding can not decrypted with the other padding. 
+   So if you change the padding later during the cause of operation, it will not be possible to decrypt old data!
 
 Logging
 -------
--- a/doc/installation/system/inifile.rst
+++ b/doc/installation/system/inifile.rst
@@ -102,6 +102,11 @@ can be set, for example::
 Further information on possible parameters can be found in the
 `PassLib documentation <https://passlib.readthedocs.io/en/stable/lib/passlib.hash.html>`_.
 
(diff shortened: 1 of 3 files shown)


doc/installation/system/inifile.rst
```diff
@@ -102,6 +102,11 @@ can be set, for example::
 Further information on possible parameters can be found in the
 `PassLib documentation <https://passlib.readthedocs.io/en/stable/lib/passlib.hash.html>`_.
 
+In ``PI_HSM_PADDING`` the padding can be set to "PKCS7". If ``PI_HSM_PADDING`` is not set, the default internal
+padding is used::
```
cornelinux, line 106: privacyIDEA currently has three security Modules:

* DefaultSecurityModule
* AESHardwareSecurityModule
* EncryptKeyHardwareSecurityModule (new)

Please note, this is only valid for the `DefaultSecurityModule` and the `EncryptKeyHardwareSecurityModule`.

The `AESHardwareSecurityModule` will always use CKM_AES_CBC_PAD (which is PKCS7, in my understanding).
So this config value is only functional for the above two mentioned two Security Modules. 

Which is fine, but which I think we should mention here.

cornelinux (changes requested):
Please add the note about 2 of 3 modules.

cornelinux (approved):

PR #3244, opened 2022-07-27
Move to current SQLAlchemy v1.4

plettich:
This required removing sqlsoup and thus changing the SQL-Resolver to use
native SALAlchemy functionality.


plettich:
Closing in favor of #3535

PR #3245, opened 2022-07-28
Do not delete client, when policy is disabled

### LINKED ISSUE #3243: Unexpected results when disabling a policy from the webui
### Top-level intent

(from the webui)
A. Disabling a policy from the '/config/policies/list' page ('Active' column) wrongly deletes the contents of the 'Client' column.
B. Disabling a policy from the '/config/policies/details/<MyPolicy>' page is not taken into account after clicking on 'Disable' then 'Save Policy'.

### Steps to reproduce

A1. Create a new policy named MyPolicy
A2. Scope 'register', Priority '100'
A3. Enter an IP in the 'Client' field (for example 1.1.1.1)
A4. Save policy
A5. Disable MyPolicy from the list of policies

B1. Create a new policy named MyPolicy2
B2. Scope 'register', Priority '100'
B3. Enter an IP in the 'Client' field (for example 1.1.1.1)
B4. Save policy
B5. Edit MyPolicy2
B6. Click on 'Disable' then 'Save Policy'

### Expected outcome

A. Disabling a policy from the policy listing page should not delete the contents of the 'Client' column.
B. Disabling a policy from the edit policy page must be taken into account after clicking on 'Disable' then 'Save Policy'.

### Configuration

* **privacyIDEA version**: 3.7.2
* **Installation method**: from Ubuntu packages
* **Python version**: 3.8.10
* **Operating system**: Ubuntu 20.04
* **Webserver**: Apache/2.4.41
* **Token database**: MySQL

cornelinux:
A client logic check led to the problem, that the client
was deleted from a policy, when the policy was disabled or enabled.

The client information must not deleted from the policy.

Fix #3243

### DIFF
--- a/privacyidea/lib/policy.py
+++ b/privacyidea/lib/policy.py
@@ -1344,11 +1344,11 @@ def set_policy(name=None, scope=None, action=None, realm=None, resolver=None,
         resolver = ", ".join(resolver)
     if type(client) == list:
         client = ", ".join(client)
-    try:
-        client = client or ""
-        check_ip_in_policy("127.0.0.1", [c.strip() for c in client.split(",")])
-    except AddrFormatError:
-        raise privacyIDEAError(_("Invalid client definition!"), id=302)
+    if client is not None:
+        try:
+            check_ip_in_policy("127.0.0.1", [c.strip() for c in client.split(",")])
+        except AddrFormatError:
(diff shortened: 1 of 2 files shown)


NNLemling:
works as desired!

NNLemling (approved):

Issue #3056, opened 2022-03-18
Allow the HttpSMSProvider to send the data as a JSON string


plettich:
Some SMS Provider provide an API which only accepts the content type `application/json`.
Before sending the request, the data can be transformed into a JSON string here:
https://github.com/privacyidea/privacyidea/blob/f770837c683a32b17ae37e398c6319a62f6d89f2/privacyidea/lib/smsprovider/HttpSMSProvider.py#L145

cornelinux:
@jona-samuel Please check, if we also need to set the content-type in `requests`.

plettich:
@jona-samuel is this fixed with #3236?

plettich:
This is fixed with #3251

PR #3251, opened 2022-07-29
Don't send json data by default

plettich:
In case of a POST request to the sms-gateway we currently try to read
the given text as json data by default which can fail in some cases.
Also passing the ' json' parameter to 'requests' already sets the
'application/json' content-type header.

Working on #3056


plettich:
The failing Python 3.8 test case is the sporadically failing "TokenOutOfBandTestCase".
The Python 3.5 test cases fail because the output of the dictionary in the log is reverted. But since we want to fade out Python 3.5 in PI 3.8 anyway, i think we can ignore this here.

jona-samuel (approved):

PR #3271, opened 2022-08-19
Stop spamming log

bmwiedemann:
in case many keys are added


cornelinux:
@bmwiedemann Please check #3298
We can change this to debug.
But on the other hand please also add an info log about the ssh key being used.

Note, that we do use the info level not for debugging purposes but for normal information to understand what is happening in privacyIDEA. Thanks.

cornelinux (changes requested):
Please check my comment.

cornelinux:
I created a new PR #3324.
THanks a lot!

PR #3272, opened 2022-08-19
Use strong hash functions

bmwiedemann:
without this patch, the supplied salt forced the crypt function
to use the weak DES crypt method.


bmwiedemann:
https://docs.python.org/3/library/crypt.html#crypt.crypt writes
> If salt is not provided, the strongest method available in [methods](https://docs.python.org/3/library/crypt.html#crypt.methods) will be used.

plettich (approved):
Looks good

PR #3286, opened 2022-09-01
3285/add functionality for displaying multiple token serials in the audit log

### LINKED ISSUE #3285: Display of multiple token serials in the audit log
It would be nice if multiple token serials could also be displayed in the audit log.

NNLemling:
closes #3285

### DIFF
--- a/privacyidea/api/auth.py
+++ b/privacyidea/api/auth.py
@@ -309,6 +309,8 @@ def get_auth_token():
                                                     superuser_realms=
                                                     superuser_realms)
         details = details or {}
+        serials = ",".join([challenge_info["serial"] for challenge_info in details["multi_challenge"]]) \
+            if 'multi_challenge' in details else details.get('serial')
         if db_admin_exist(loginname) and user_auth and realm == get_default_realm():
             # If there is a local admin with the same login name as the user
             # in the default realm, we inform about this in the log file.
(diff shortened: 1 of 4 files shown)


privacyidea/api/auth.py
```diff
@@ -309,6 +309,8 @@ def get_auth_token():
                                                     superuser_realms=
                                                     superuser_realms)
         details = details or {}
+        serials = ",".join([challenge_info["serial"] for challenge_info in details["multi_challenge"]]) \
```
plettich, line 312: If You use `', '.join()` You'll have a space after the comma, looks better.

privacyidea/api/validate.py
```diff
@@ -416,10 +416,11 @@ def check():
                     # additional attributes
                     for k, v in ui.items():
                         result["attributes"][k] = v
-
+    serials = ",".join([challenge_info["serial"] for challenge_info in details["multi_challenge"]]) \
```
plettich, line 419: same as above

privacyidea/api/validate.py
```diff
@@ -416,10 +416,11 @@ def check():
...
+        if 'multi_challenge' in details else details.get('serial')
     g.audit_object.log({"info": log_used_user(user, details.get("message")),
                         "success": success,
-                        "serial": serial or details.get("serial"),
+                        "serial": serials,
```
plettich, line 423: We should also add a testcase for this and update the frontend code to create the links

plettich (changes requested):

privacyidea/api/auth.py
NNLemling: If a space is added after the "," the link to the token contains a space that no longer works.

privacyidea/api/validate.py
NNLemling: explanation above

privacyidea/api/validate.py
NNLemling: The frontend code actually handles the serials correctly. Links to the various tokens are generated and work.
serials2

plettich (approved):

PR #3290, opened 2022-09-12
UX improvement: reveal otp seeds

ziQ:
We had a lot of users confused about user experience on the enroll token UI. Few users recognized that "The OTP key" actually was a clickable link which could be used to reveal the OTP seed for import into Keepass or similar tools.

This pull request adds " (click to reveal)" to the UI to make this more intuitive for users.

20220912_12h27m02s_grim
20220912_12h27m39s_grim

(The shown token is a throw-away token, of course)

### DIFF
--- a/po/cs.po
+++ b/po/cs.po
@@ -3060,6 +3060,12 @@ msgstr ""
 msgid "The OTP key"
 msgstr "Sdílené tajemství (klíč)"
 
+#: privacyidea/static/components/token/views/token.enrolled.hotp.html:13
+#: privacyidea/static/components/token/views/token.enrolled.motp.html:10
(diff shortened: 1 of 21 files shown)


cornelinux:
This is a theme-typical accordeon. I would refrain from changing this in an individual way. I think there are two possibilities:

a) Enhance the accordeon in this way, that all have an arrorw down.
b) remove the accordeon around the otpseed anyways.

cornelinux:
We choose b)

**b) remove the accordeon around the otpseed anyways.**

@ziQ Could you please change the PR this way, that the accordean around the seed is removed?
(and also remove all the translations! ;-) Thanks!

cornelinux:
Closing for getting old

PR #3297, opened 2022-09-19
Do a softer subscription check

### LINKED ISSUE #3296: Subscription checking to hard
The subscription check for certain plugins will fail as soon as it is exceeded by only one token.

If a plugin with 5000 subscriptions is hit by a system with 20.000 users, the subscription check should be fine for 5000 cases out of 20.000.

cornelinux:
Subscription check should not fail for the subscribed users.

Closes #3296

### DIFF
--- a/tests/test_lib_subscriptions.py
+++ b/tests/test_lib_subscriptions.py
@@ -103,9 +103,14 @@ class SubscriptionApplicationTestCase(MyTestCase):
 
         init_token({"type": "spass"}, user=User("shadow", self.realm1))
         init_token({"type": "spass"}, user=User("nopw", self.realm1))
-        # Now we have three users with tokens, subscription will fail
-        self.assertRaises(SubscriptionError, check_subscription,
-                          "demo_application")
+        # Now we have three users with tokens, but only two are allowed. We fail with a probabiliy of 1/3
+        try:
+            for i in range(30):
+                check_subscription("demo_application")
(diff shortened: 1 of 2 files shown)


tests/test_lib_subscriptions.py
```diff
@@ -103,9 +103,14 @@ def test_03_check_subscription(self):
 
         init_token({"type": "spass"}, user=User("shadow", self.realm1))
         init_token({"type": "spass"}, user=User("nopw", self.realm1))
-        # Now we have three users with tokens, subscription will fail
-        self.assertRaises(SubscriptionError, check_subscription,
-                          "demo_application")
+        # Now we have three users with tokens, but only two are allowed. We fail with a probabiliy of 1/3
```
plettich, line 106: I think this is bound to produce randomly failing tests. We should mock out the randomness for the test and just check if we get the required error.

plettich (changes requested):

tests/test_lib_subscriptions.py
cornelinux: done in ae1b513d

plettich:
Now the tests fail with a different `random` error :-) I'll take a look.

plettich (approved):

plettich:
> Now the tests fail with a different `random` error :-) I'll take a look.

This will be fixed with #3303

PR #3304, opened 2022-09-26
fixing crypto test get_rand_digit_str

### LINKED ISSUE #3303: Tests for int -> str conversion fail since Python 3.10.7
Since Python 3.10.7 there is a limit for integer to string conversion:
https://docs.python.org/3/library/stdtypes.html#integer-string-conversion-length-limitation
The test for `get_rand_digit_str()` fail.

EthemGit:
Fixes #3303

### DIFF
--- a/tests/test_lib_crypto.py
+++ b/tests/test_lib_crypto.py
@@ -321,8 +321,8 @@ class RandomTestCase(MyTestCase):
         self.assertTrue(len(r) == 2, r)
         r = get_rand_digit_str(1001)
         self.assertTrue(len(r) == 1001, r)
-        r = get_rand_digit_str(19182)
-        self.assertTrue(len(r) == 19182)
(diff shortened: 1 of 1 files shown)


cornelinux:
Hm, where does this problem actually originate from?
I do not see the problem in `get_rand_digit_str`. Should we change the default or should we also add a failsafe here:

https://github.com/privacyidea/privacyidea/blob/2a66d4792289424e0643178bd902a2e6e5f96482/privacyidea/lib/crypto.py#L643

like:

~~~~Python

if length > 2500:
        raise ValueError("get_rand_digit_str only works for values < 2500 - or heck whatever it is!")
~~~~

Then we could also add a test for too high values.

cornelinux (changes requested):
Please check my comment

plettich (approved):

cornelinux (approved):

PR #3310, opened 2022-09-29
2996/rename createdb and dropdb

EthemGit:
- added new functions create_tables() and drop_tables() which call createdb() and dropdb()
- added deprecation warning to createdb() and dropdb() that these two functions will be removed in version 4.0
- changed createdb() to create_tables() and dropdb() to drop_tables() in docs

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -132,7 +132,7 @@ First You need to create a `config-file <https://privacyidea.readthedocs
 
 Then create the database and encryption key::
 
-    ./pi-manage createdb
+    ./pi-manage create_tables
     ./pi-manage create_enckey
 
 If You want to keep the development database upgradable, You should `stamp
--- a/README.rst
+++ b/README.rst
@@ -130,7 +130,7 @@ Running it
 First You need to create a `config-file <https://privacyidea.readthedocs
 .io/en/latest/installation/system/inifile.html>`_.
 
-Then create the database and encryption key::
+Then create the database tables and the encryption key::
 
     ./pi-manage create_tables
     ./pi-manage create_enckey
--- a/pi-manage
+++ b/pi-manage
@@ -48,12 +48,13 @@
 #
 # ./manage.py db init
 # ./manage.py db migrate
-# ./manage.py createdb
+# ./manage.py create_tables
 #
 from __future__ import print_function
 import os
 import sys
 import datetime
+import warnings
 from datetime import timedelta
 import re
 from subprocess import call, Popen, PIPE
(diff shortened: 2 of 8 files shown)


README.rst
```diff
@@ -132,7 +132,7 @@ First You need to create a `config-file <https://privacyidea.readthedocs
 
 Then create the database and encryption key::
```
cornelinux, line 133: Then we could also change this to 

"Then create the database tables and the encryption key"

pi-manage
```diff
@@ -579,6 +580,17 @@ def createdb(stamp=False):
         migration_dir = os.path.dirname(os.path.abspath(p[0]))
         fm_stamp(directory=migration_dir)
     db.session.commit()
+    warnings.warn("The function createdb() will be removed in Version 4.0", DeprecationWarning)
```
cornelinux, line 583: I would put it the other way round.

I would put the code into "create_tables" and in "createdb" simpley call `create_tables()`.

This way we do not need to move code in version 4.0.

pi-manage
```diff
@@ -579,6 +580,17 @@ def createdb(stamp=False):
...
+
+
+@manager.option('--stamp', '-s', help='Stamp database to current head revision.',
+                default=False, action='store_true')
+def create_tables(stamp=False):
+    """
+    Initially create the tables in the database. The database must exist
+    (an SQLite database will be created).
+    """
+    createdb(stamp=False)
```
cornelinux, line 593: When calling the productive function, you need to pass only "stamp".

Otherwise it will always be "False".

pi-manage
```diff
@@ -603,6 +615,20 @@ def dropdb(dropit=None):
...
+def drop_tables(dropit=None):
+    """
+    This drops all the privacyIDEA database tables.
+    Use with caution! All data will be lost!
+
+    For safety reason you need to pass
+        --dropit==yes
+    Otherwise the command will not drop anything.
+    """
+    dropdb(dropit=None)
```
cornelinux, line 631: You must only pass "dropit", not "=None".

cornelinux (changes requested):
Please move the code and fix the parameters.

README.rst
EthemGit: Done in: a5b50932686161a11509b56b4a98ac3665d0d493

pi-manage
EthemGit: Done: a5b50932686161a11509b56b4a98ac3665d0d493

pi-manage
EthemGit: Done in: a5b50932686161a11509b56b4a98ac3665d0d493

cornelinux (changes requested):
Please move the deprecation info from the option to the doctstring

cornelinux (approved):

PR #3312, opened 2022-09-29
WebUI policy for logout_redirect

jona-samuel:


privacyidea/webui/login.py
```diff
@@ -137,6 +137,15 @@ def single_page_application():
     except HSMException:
         hsm_ready = False
 
+    # A url to redirct to at logout
+    logout_redirect_url = Match.action_only(g, action=ACTION.LOGOUT_REDIRECT,
+                                                    scope=SCOPE.WEBUI) \
+        .action_values(unique=True, allow_white_space_in_action=True, write_to_audit_log=False)
+    if len(logout_redirect_url) and list(logout_redirect_url)[0]:
```
plettich, line 144: You can add a test like this:
https://github.com/privacyidea/privacyidea/blob/6d81530ab5e4f07a9cf3e037b98de2269f97c06f/tests/test_ui_login.py#L83

plettich (changes requested):

privacyidea/webui/login.py
jona-samuel: [3de128f](https://github.com/privacyidea/privacyidea/pull/3312/commits/3de128fa104cabbe23e5f5bf933b51783a419a70)

plettich (approved):

Issue #2375, opened 2020-09-07
/auth fails - LDAPServerPoolExhaustedError


laclaro:
An irresponsive LDAP resolver prevents an /auth login and exits with a error message for the irresponsive LDAP server

```python
LDAPServerPoolExhaustedError('no active server available in server pool after maximum number of tries')
```

laclaro:
This issue was not present in 3.3.3.

It occurs for the first time in 3.4dev6. The list of closed issues since 3.4dev5 ist this:
https://github.com/privacyidea/privacyidea/issues?q=is%3Aissue+closed%3A%3E2020-08-22+

The present error must be due to the pre-auth user object assignment.
https://github.com/privacyidea/privacyidea/blob/2fc1137f235f3153d35f48e7a4d5e9363fa60c0e/privacyidea/api/auth.py#L115

PR #3344, opened 2022-10-21
Make realm in auth-request case insensitive

### LINKED ISSUE #3329: Realm name is case sensitive when login to the WebUI
The login as an admin on the webGui is key sensitive.

Example:
 
xxxx@helpdesk => Login is functioning. User has the admin rights

xxxx@Helpdesk => Login is possible, but the user doesn't have the admin rights.

Privacyidea 3.7.3

* [ ] Fix login at webui (/auth)
* ~~[ ] We need check if the same problem exists in /validate/check or in other endpoints to identify the user~~
  This only affects the admin rights, the authentication in the realm works correctly. Realm are always lower case.

plettich:
While the user is found in the correct resolver/realm (by changing the realm to lower case), when querying for the admin rights the original realm was used and so no policy matched the realm and the admin didn't have any rights.
Fixing this by changing the realm to lower case.

Closes #3329

### DIFF
--- a/privacyidea/api/auth.py
+++ b/privacyidea/api/auth.py
@@ -291,6 +291,7 @@ def get_auth_token():
         user_obj = User()
         g.audit_object.log({"success": True,
                             "user": "",
+                            "realm": "",
                             "administrator": username,
                             "info": "internal admin"})
 
--- a/privacyidea/api/auth.py
+++ b/privacyidea/api/auth.py
@@ -114,7 +114,7 @@ def before_request():
         # On endpoints like /auth/rights, this is not available
         loginname, realm = split_user(username)
         # overwrite the split realm if we have a realm parameter. Default back to default_realm
-        realm = getParam(request.all_data, "realm", default=realm) or realm or get_default_realm()
+        realm = getParam(request.all_data, "realm", default=realm) or get_default_realm()
         # Prefill the request.User. This is used by some pre-event handlers
         try:
             request.User = User(loginname, realm)
@@ -226,14 +226,15 @@ def get_auth_token():
         raise AuthError(_("Authentication failure. Missing Username"),
                         id=ERROR.AUTHENTICATE_MISSING_USERNAME)
 
-    loginname, realm = split_user(username)
-
-    # overwrite the split realm if we have a realm parameter
-    if realm_param:
-        realm = realm_param
-
-    # and finally check if there is a realm
-    realm = realm or get_default_realm()
+    user_obj = request.User
+    if not user_obj:
+        # The user could not be resolved, but it could still be a local administrator
+        loginname, realm = split_user(username)
+        realm = (realm_param or realm or get_default_realm()).lower()
+        user_obj = User()
+    else:
+        realm = user_obj.realm
+        loginname = user_obj.login
 
     # Failsafe to have the user attempt in the log, whatever happens
     # This can be overwritten later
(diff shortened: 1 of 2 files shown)


privacyidea/api/auth.py
```diff
@@ -226,23 +226,24 @@ def get_auth_token():
         raise AuthError(_("Authentication failure. Missing Username"),
                         id=ERROR.AUTHENTICATE_MISSING_USERNAME)
 
-    loginname, realm = split_user(username)
```
cornelinux, line 229: Why overcomplicating things and not simply to a
~~~~Python
    loginname, realm = split_user(username)
    realm = realm.lower()
~~~~

or adding a `lower()` to split_username.
Verifying all the changes and ensuring no side effect, is near to impossible.

What is the reason for moving the creation of a user object around further below in the code?

privacyidea/api/auth.py
```diff
@@ -114,7 +114,7 @@ def before_request():
         # On endpoints like /auth/rights, this is not available
         loginname, realm = split_user(username)
         # overwrite the split realm if we have a realm parameter. Default back to default_realm
-        realm = getParam(request.all_data, "realm", default=realm) or realm or get_default_realm()
+        realm = getParam(request.all_data, "realm", default=realm) or get_default_realm()
```
cornelinux, line 117: wouldn't a a 

    realm = realm.lower()


be sufficient?

cornelinux (changes requested):
I think this is far to complicated and imposes the risk of unverified side effects!

privacyidea/api/auth.py
plettich: here we have no problems with the case of the realm. The realm will be converted to lower-case automatically when we create the `User` object.

privacyidea/api/auth.py
plettich: Because the user object is initially created in the `before_request()` and added to the request object here:
https://github.com/privacyidea/privacyidea/blob/fa643349df8fabdf329caf4091b70eaafaade74a/privacyidea/api/auth.py#L120
This way we can remove a second ldap request when the user object is created again later.
But we can move this change to a separate issue and just tackle the case-sensitivity here.

cornelinux: > This way we can remove a second ldap request when the user object is created again later.

Ok, I understand. Will look into it again.

privacyidea/api/auth.py
```diff
@@ -289,6 +288,7 @@ def get_auth_token():
         log.info("Local admin '{0!s}' successfully logged in.".format(username))
         # This admin is not in the default realm!
         realm = ""
+        user_obj = User()
```
cornelinux, line 291: To my understanding you are cleaning up the user_obj.
This could be if there is a user in the default realm with the same login name as a local admin.

However, I *think* I remeber we had problems in the past in this case, why we did a "late" user resolving (which would contradict the before_request).

I remember we had problems in this case, the LDAP is not accessable and due to an early user_obj creating the login for the local admin failed or timed out.
So if you move the user_obj creation to the beginning of this function, the code might look cleaner, but we need to verify that we do not reintroduce this problem again! This is why I am so hesitant about this.

cornelinux (changes requested):
OK, I checked and understood that the code and the code changes will work.
However,  I am not sure, if the changes are robust or if they will cause problems with non-reachable, breaking resolvers in the default realm.

With the current code, the local admin is able to login no matter which problems with the resolvers exist! We need to keep this functionality!

privacyidea/api/auth.py
plettich: The creation of the `User` object in the `before_request` was added for pre-event handler.
This failed for an admin log-in if the resolver was broken. Since #2375 the log-in doesn't fail but if a resolver in the default realm is broken, we still need to wait for the timeout.
We could check in the `before_request` if the user is an internal admin (e.g. a `username` without a realm) and then bypass the creation of the `User` object.
I still need to check what happens if there is a user with the same name in the default realm.

plettich: I've checked that the internal admin check still works despite a user with a same name in the default realm. Also the admin can still log-in if the resolver doesn't work, it just takes time to wait for the timeout from the resolver.

plettich: Also the local admin vs. user check works as expected.

cornelinux (approved):

PR #3355, opened 2022-10-27
Add stripping of end tags

cornelinux:
Lately migrating and reencrypting from
LinOTP 2.7.2 required these end tags.


plettich (approved):
This is the old legacy AES padding which we removed some time ago.

PR #3361, opened 2022-11-02
preferred client mode

### LINKED ISSUE #3044: Add preferred client_mode
The Challenge in challenge response can contain multiple possible authentications with different token types which result in different [client modes](https://privacyidea.readthedocs.io/en/latest/tokens/authentication_modes.html). [Possible values are interactive, webauthn, u2f, poll](https://github.com/privacyidea/privacyidea/blob/2b075edc8ff0f4f481de4c36411dc648f114ce01/privacyidea/lib/tokenclass.py#L137).

The response should contain an information, which mode should be preferred by a client/plugin.
I.e. a user could trigger challenge response for an `interactive` token and a `poll` token.

The response should contain this information like this:

~~~~json
{
    "detail": {
        "preferred_client_mode": "webauthn",
        "multi_challenge": [
            {
                "client_mode": "webauthn",
                "image": "",
                "message": "Please confirm with your WebAuthn token (Generic WebAuthn Token)",
                "serial": "WAN00023132",
                "transaction_id": "09250011178417540855",
                "type": "webauthn"
            }
        ],
~~~~

The multi_challenge section contains a list of challenge dictionaries.https://github.com/privacyidea/privacyidea/blob/8eeaa9d44999a412f4bc598b2fe54c6d552e6d46/privacyidea/lib/tokenclass.py#L137
So we should put the `preferred_client_mode` in the detail top-level. 
It can contain one value. It can contain the same values like the `client_mode`.

# Backend Implementation

Think about it - it actually does not change the authentication from a server perspective but only how the client/plugin behaves. Should we add a new policy scope for this?

~~* [ ] check if there are other policy actions, that only tell the client what to do.~~

* scope: authentication
* action: preferred_client_mode: type: str
 (https://github.com/privacyidea/privacyidea/blob/8eeaa9d44999a412f4bc598b2fe54c6d552e6d46/privacyidea/lib/tokenclass.py#L137) 

The string can contain a whitespace separated list like "interactive webauthn". The admin needs to write these values manually.
If no policy for the client_mode is set, we use the default `interactive`.
If no `interactive` is contained in multi_challenge, then we do default to `webauthn`.
If not `webauthn` is contained, then we default to `poll`.

@nilsbehlen : Do you always want the preferred client mode.

* [x] Implement the new return value `preferred_client_mode` in the response
* [x] Implement the order of default
* [x] add policy
* [x] test policy
* [x] documentation

jona-samuel:
This is the first version of the preferred client mode.
-To Do: more tests, error handling and add docu

Working on #3044

### DIFF
--- a/privacyidea/api/lib/postpolicy.py
+++ b/privacyidea/api/lib/postpolicy.py
@@ -48,6 +48,7 @@ from privacyidea.lib.error import PolicyError, ValidateError
 from flask import g, current_app, make_response
 from privacyidea.lib.policy import SCOPE, ACTION, AUTOASSIGNVALUE, AUTHORIZED
 from privacyidea.lib.policy import DEFAULT_ANDROID_APP_URL, DEFAULT_IOS_APP_URL
+from privacyidea.lib.policy import DEFAULT_PREFERRED_CLIENT_MODE
 from privacyidea.lib.policy import Match
 from privacyidea.lib.token import get_tokens, assign_token, get_realms_of_token, get_one_token
 from privacyidea.lib.machine import get_hostname, get_auth_items
@@ -331,15 +332,16 @@ def preferred_client_mode(request, response):
     :return:
     """
     content = response.json
+    user_object = request.User
 
     # get the preferred client mode from a policy definition
-    detail_pol = Match.action_only(g, scope=SCOPE.AUTH, action=ACTION.PREFERREDCLIENTMODE)\
-        .action_values(unique=True)
+    detail_pol = Match.user(g, scope=SCOPE.AUTH, action=ACTION.PREFERREDCLIENTMODE, user_object=user_object)\
+        .action_values(allow_white_space_in_action=True, unique=True)
 
     if detail_pol:
-        preferred_client_mode_list = detail_pol.get("test")[0].split(", ")
+        preferred_client_mode_list = list(detail_pol.values())[0][0].split(" ")
     else:
-        preferred_client_mode_list = ['poll', 'webauthn', 'interactive', 'u2f']
+        preferred_client_mode_list = DEFAULT_PREFERRED_CLIENT_MODE
     if content.get("detail"):
         detail = content.get("detail")
         if detail.get("multi_challenge"):
@@ -347,20 +349,17 @@ def preferred_client_mode(request, response):
             l = []
             for x in multi_challenge:
                 l.append(x.get("client_mode"))
-            l=list(dict.fromkeys(l))
 
-            if preferred_client_mode_list[0] in l:
-                preferred = preferred_client_mode_list[0]
+            try:
+                preferred = [x for x in preferred_client_mode_list if x in l][0]
                 content.setdefault("detail", {})["preferred_client_mode"] = preferred
-            elif preferred_client_mode_list[1] in l:
-                preferred = preferred_client_mode_list[1]
-                content.setdefault("detail", {})["preferred_client_mode"] = preferred
-            elif preferred_client_mode_list[2] in l:
-                preferred = preferred_client_mode_list[2]
-                content.setdefault("detail", {})["preferred_client_mode"] = preferred
-            elif preferred_client_mode_list[3] in l:
-                preferred = preferred_client_mode_list[3]
-                content.setdefault("detail", {})["preferred_client_mode"] = preferred
+            except IndexError as err:
+                content.setdefault("detail", {})["preferred_client_mode"] = 'interactive'
+                log.error('There was no except client mode in the multi-challenge. The preferred mode is'
+                          ' set to interactive. Please check your policy. Error:'.format(err))
+            except Exception as err:
+                content.setdefault("detail", {})["preferred_client_mode"] = 'interactive'
+                log.error('something with the preferred client mode got wrong:{0}'.format(err))
 
     response.set_data(json.dumps(content))
     return response
--- a/privacyidea/api/lib/postpolicy.py
+++ b/privacyidea/api/lib/postpolicy.py
@@ -318,6 +318,43 @@ def no_detail_on_success(request, response):
     return response
 
 
+def preferred_client_mode(request, response):
+    """
+    description
+    """
+    content = response.json
+
+    # get the preferred client mode from a policy definition
+    detail_pol = Match.action_only(g, scope=SCOPE.AUTH, action=ACTION.PREFERREDCLIENTMODE)\
+        .action_values(unique=True)
+
+    if detail_pol:
+        preferred_client_mode_list = detail_pol.get("test")[0].split(", ")
+    else:
+        preferred_client_mode_list = ['poll', 'webauthn', 'interactive', 'u2f']
+    if content.get("detail"):
+        detail = content.get("detail")
+        if detail.get("multi_challenge"):
+            multi_challenge = detail.get("multi_challenge")
+            l = []
+            for x in multi_challenge:
+                l.append(x.get("client_mode"))
+            l=list(dict.fromkeys(l))
+
+            if preferred_client_mode_list[0] in l:
+                preferred = preferred_client_mode_list[0]
+            elif preferred_client_mode_list[1] in l:
+                preferred = preferred_client_mode_list[1]
+            elif preferred_client_mode_list[2] in l:
+                preferred = preferred_client_mode_list[2]
+            elif preferred_client_mode_list[3] in l:
+                preferred = preferred_client_mode_list[3]
+            content.setdefault("detail", {})["preferred_client_mode"] = preferred
+
+    response.set_data(json.dumps(content))
+    return response
+
+
 def add_user_detail_to_response(request, response):
     """
     This policy decorated is used in the AUTHZ scope.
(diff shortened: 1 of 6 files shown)


privacyidea/lib/policy.py
```diff
@@ -2278,6 +2279,10 @@ def get_static_policy_definitions(scope=None):
                           'allow authentication with the same credentials for a '
                           'certain amount of time. '
                           'Specify timeout like 4h or 4h/5m.')
+            },
+            ACTION.PREFERREDCLIENTMODE: {
+                'type': 'str',
+                'desc': _('preferred client mode')
```
cornelinux, line 2285: It would be nice to add possioble values here:

~~~Python
   'value' : [...] 
~~~~

and take them from the client mode definition.

Ah, oh wait. We said we could add a list of preferred values.

So please enhace the `'desc'` and explain, which modes are available. Compare the action `AUTH_CACHE` above. There should be more hints!

privacyidea/api/validate.py
```diff
@@ -207,6 +207,7 @@ def offlinerefill():
 @postpolicy(check_tokentype, request=request)
 @postpolicy(check_serial, request=request)
 @postpolicy(autoassign, request=request)
+@postpolicy(preferred_client_mode, request=request)
```
cornelinux, line 210: I am not sure if we also should add this to triggerchallenge!?
Maybe we need to discuss this with @nilsbehlen

privacyidea/api/lib/postpolicy.py
```diff
@@ -318,6 +318,54 @@ def no_detail_on_success(request, response):
...
+    :return:
+    """
+    content = response.json
+
+    # get the preferred client mode from a policy definition
+    detail_pol = Match.action_only(g, scope=SCOPE.AUTH, action=ACTION.PREFERREDCLIENTMODE)\
+        .action_values(unique=True)
+
+    if detail_pol:
+        preferred_client_mode_list = detail_pol.get("test")[0].split(", ")
```
cornelinux, line 340: Lists in policies are whitespace separated.
So an admin would configure the policy like:

      "webauthn poll"

Hm, what is this with `get("test")`. THis does not make any sense.
Please see the policy `challenge_response` (https://github.com/privacyidea/privacyidea/blob/9bee74fb74a492ee97ca4ee9ec147312f8800108/privacyidea/lib/policydecorators.py#L122) to understand values in one policy.

Hm, it is still not clear, how we should handle multiple policies in this case:

pol1:  webauthn poll
pol2:   pol

privacyidea/api/lib/postpolicy.py
```diff
@@ -318,6 +318,54 @@ def no_detail_on_success(request, response):
...
+    set it as preferred client mode
+
+    :param request:
+    :param response:
+    :return:
+    """
+    content = response.json
+
+    # get the preferred client mode from a policy definition
+    detail_pol = Match.action_only(g, scope=SCOPE.AUTH, action=ACTION.PREFERREDCLIENTMODE)\
```
cornelinux, line 336: Please use `Match.user`m this way we also use user conditions in the policy.

privacyidea/api/lib/postpolicy.py
```diff
@@ -318,6 +318,54 @@ def no_detail_on_success(request, response):
...
+        .action_values(unique=True)
+
+    if detail_pol:
+        preferred_client_mode_list = detail_pol.get("test")[0].split(", ")
+    else:
+        preferred_client_mode_list = ['poll', 'webauthn', 'interactive', 'u2f']
```
cornelinux, line 342: I would preferably define this somewhere (near lib/policy.py?) as a constant.

privacyidea/api/lib/postpolicy.py
```diff
@@ -318,6 +318,54 @@ def no_detail_on_success(request, response):
...
+    if content.get("detail"):
+        detail = content.get("detail")
+        if detail.get("multi_challenge"):
+            multi_challenge = detail.get("multi_challenge")
+            l = []
+            for x in multi_challenge:
+                l.append(x.get("client_mode"))
+            l=list(dict.fromkeys(l))
```
cornelinux, line 350: What is this for? I think if we wanted to reduce the list, we did it usually via a set:

    list(set(l))

On the other hand, since you are only testing if it is **in** `l`, then reducing the list is not necessary.

privacyidea/api/lib/postpolicy.py
```diff
@@ -318,6 +318,54 @@ def no_detail_on_success(request, response):
...
+                content.setdefault("detail", {})["preferred_client_mode"] = preferred
+            elif preferred_client_mode_list[1] in l:
+                preferred = preferred_client_mode_list[1]
+                content.setdefault("detail", {})["preferred_client_mode"] = preferred
+            elif preferred_client_mode_list[2] in l:
+                preferred = preferred_client_mode_list[2]
+                content.setdefault("detail", {})["preferred_client_mode"] = preferred
+            elif preferred_client_mode_list[3] in l:
+                preferred = preferred_client_mode_list[3]
+                content.setdefault("detail", {})["preferred_client_mode"] = preferred
```
cornelinux, line 363: You could move the four `content.setdefault`s out of the if branch to only one call at the end.

cornelinux: You could simplify getting the preferred value like this:

~~~~Python
preferred = [x for x in preferrend_client_mode_list if x in l][0]
~~~~
This would replace the lines 352 to 362.

cornelinux (changes requested):
PLease check my comments

privacyidea/api/lib/postpolicy.py
cornelinux:
~~~~python
Match.user().action_values(allow_white_space_in_action=True)
~~~~

`allow_white_space_in_action` would make a string like "poll interactive" *one* action value.
Otherwise it would be split into two action values "poll" and "interactive".

privacyidea/lib/policy.py
jona-samuel: [34ee918](https://github.com/privacyidea/privacyidea/pull/3361/commits/34ee918173549cdef4dfeb1e000951457fff1cab)

privacyidea/api/lib/postpolicy.py
jona-samuel: [34ee918](https://github.com/privacyidea/privacyidea/pull/3361/commits/34ee918173549cdef4dfeb1e000951457fff1cab)

privacyidea/api/lib/postpolicy.py
jona-samuel: [34ee918](https://github.com/privacyidea/privacyidea/pull/3361/commits/34ee918173549cdef4dfeb1e000951457fff1cab)

privacyidea/api/lib/postpolicy.py
jona-samuel: [34ee918](https://github.com/privacyidea/privacyidea/pull/3361/commits/34ee918173549cdef4dfeb1e000951457fff1cab)

privacyidea/api/lib/postpolicy.py
jona-samuel: [34ee918](https://github.com/privacyidea/privacyidea/pull/3361/commits/34ee918173549cdef4dfeb1e000951457fff1cab)

privacyidea/api/lib/postpolicy.py
jona-samuel: [34ee918](https://github.com/privacyidea/privacyidea/pull/3361/commits/34ee918173549cdef4dfeb1e000951457fff1cab)

privacyidea/lib/policy.py
```diff
@@ -2278,6 +2280,12 @@ def get_static_policy_definitions(scope=None):
                           'allow authentication with the same credentials for a '
                           'certain amount of time. '
                           'Specify timeout like 4h or 4h/5m.')
+            },
+            ACTION.PREFERREDCLIENTMODE: {
+                'type': 'str',
+                'desc': _('You can set the client modes in the order that you prefer. '
+                          'For example: "interactive webauthn poll u2f". Except'
```
cornelinux, line 2287: "Accepted values"

privacyidea/api/lib/postpolicy.py
```diff
@@ -318,6 +319,52 @@ def no_detail_on_success(request, response):
...
+    """
+    content = response.json
+    user_object = request.User
+
+    # get the preferred client mode from a policy definition
+    detail_pol = Match.user(g, scope=SCOPE.AUTH, action=ACTION.PREFERREDCLIENTMODE, user_object=user_object)\
+        .action_values(allow_white_space_in_action=True, unique=True)
+
+    if detail_pol:
+        preferred_client_mode_list = list(detail_pol)[0].split(", ")
```
cornelinux, line 342: This will fail, if the admin enters a policy like `interactive,poll`.
I.e. no whitespace after the comma.

So you should do

~~~~python
preferred_client_mode_list = [cm.strip() for cm inlist(detail_pol)[0].split(",")]
~~~~

cornelinux (changes requested):

cornelinux:
I will fix this and merge.

PR #3366, opened 2022-11-06
Add enrollment policies for adding email and OU

### LINKED ISSUE #3327: email address in CSR causes request to fail
Adding an email to the subject results in a denied CSR with the MSCA

https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/tokens/certificatetoken.py#L472

So we need to define, if the CSR should contain the email address. This might be related to #2159 

In #2159 we propose a policy scope `ca`.

We could add a policy:

* scope: ca
* action: request_subject_components = email

However, this could be dependent on the template type.

Also look into localCA to check the additional value here

As a condition for such a policy, we would also need:

* Name of the CA issued to
* template name issued to.

So this would be #3364.

However, I think in a first step it is fine to create a simple policy - either in scope CA or in scope enrollment.

* [x] documentation
* [x] implement decorator `policy ca_request_subject_component = email` in scope enrollment 
* [x] test
* [x] Implement adding of components to request
* [x] test

cornelinux:
The subject of a certificate request may contain
emailAddress and organizationalUnit

Closes #3327

### DIFF
--- a/privacyidea/lib/tokens/certificatetoken.py
+++ b/privacyidea/lib/tokens/certificatetoken.py
@@ -69,6 +69,7 @@ class ACTION(BASE_ACTION):
     REQUIRE_ATTESTATION = "certificate_require_attestation"
     CA_CONNECTOR = "ca_connector"
     CERTIFICATE_TEMPLATE = "certificate_template"
+    CERTIFICATE_REQUEST_SUBJECT_COMPONENT = "certificate_request_subject_component"
 
 
 class REQUIRE_ACTIONS(object):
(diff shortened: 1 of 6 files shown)


privacyidea/lib/tokens/certificatetoken.py
```diff
@@ -316,6 +317,12 @@ def get_class_info(key=None, ret='all'):
                            'type': 'str',
                            'desc': _("The template that should be used to issue a certificate."),
                            'group': GROUP.TOKEN
+                       },
+                       ACTION.CERTIFICATE_REQUEST_SUBJECT_COMPONENT: {
+                           'type': 'str',
+                           'desc': _("This takes a space separated list of elemtens to be added to the subject. "
```
cornelinux, line 323: Fix typo

PR #3371, opened 2022-11-09
933/limit the number of challenge/response sent to avoid being spammed

### LINKED ISSUE #933: Limit the number of Challenge/Response sent, to avoid being spammed
The SMS could be limited to a certain number of SMS sent per time interval.

## Background info

SMS needs to be triggered by the user's password or by an administrator. In certain cases SMS could be triggered without any proof, i.e. anyone could trigger SMS simply by knowing the username.
The user would be spammed with SMS.

So we could limit the sending of SMS like ``5sms/5minutes``.

### Downside 

An attacker could prevent the user from being able to login. The attacker would "use" all 5 SMS per 5 minutes, and the user would not be able to login for 5 minutes. And again for 5 minutes. And again...

NNLemling:
closes #933

### DIFF
--- a/doc/policies/authentication.rst
+++ b/doc/policies/authentication.rst
@@ -391,20 +391,23 @@ type: bool
 If a user authenticates successfully all failcounter of all of his tokens
 will be reset. This can be important, if using empty PINs or *otppin=None*.
 
-always_increase_failcounter
+increase_failcounter_on_challenge
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 type: bool
 
-If a user logs in with their correct userstore password or a suitable tokenpin,
-the failcounter will still be increased for the time being.
+The normal behaviour is: to not increase the failcounter in case of challenge response.
+
+If this policy is activated the failcounter is increased for each token for which a challenge
+is triggered, even if the correct password or userstore pin is entered.
 
 The reason for this is that an attack can no longer trigger an infinite number
 of SMS or emails, for example. Because once the maximum failcounter has been reached,
-no further tokens can be triggered.
+no further challenges for these tokens can be triggered.
 
-.. note:: It should be noted that the failcounter of all tokens from the user are increased.
-   It makes sense to use this policy together with :ref:`reset_all_user_tokens`.
+.. note:: It should be noted that for all tokens for which a challenge has been generated,
+   the failcounter will be incremented. In the case of validate/triggerchallenge, the failcounters are increased for all tokens.
+   In some cases it makes sense to use this policy together with :ref:`reset_all_user_tokens`.
 
 .. _policy_auth_cache:
 
--- a/doc/policies/authentication.rst
+++ b/doc/policies/authentication.rst
@@ -399,7 +399,7 @@ type: bool
 The normal behaviour is: to not increase the failcounter in case of challenge response.
 
 If this policy is activated the failcounter is increased for each token for which a challenge
-is triggered, even if the correct password or userstore pin is entered.
+is triggered.
 
 The reason for this is that an attack can no longer trigger an infinite number
 of SMS or emails, for example. Because once the maximum failcounter has been reached,
--- a/doc/policies/authentication.rst
+++ b/doc/policies/authentication.rst
@@ -381,6 +381,7 @@ For more information on configuring U2F see :ref:`u2f_token`.
 
 .. [#pythonre] https://docs.python.org/2/library/re.html
 
+.. _reset_all_user_tokens:
 
 reset_all_user_tokens
 ~~~~~~~~~~~~~~~~~~~~~
@@ -390,6 +391,20 @@ type: bool
 If a user authenticates successfully all failcounter of all of his tokens
 will be reset. This can be important, if using empty PINs or *otppin=None*.
 
+always_increase_failcounter
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+type: bool
+
+If a user logs in with their correct userstore password or a suitable tokenpin,
+the failcounter will still be increased for the time being.
+
+The reason for this is that an attack can no longer trigger an infinite number
+of SMS or emails, for example. Because once the maximum failcounter has been reached,
+no further tokens can be triggered.
+
+.. note:: It should be noted that the failcounter of all tokens from the user are increased.
+   It makes sense to use this policy together with :ref:`reset_all_user_tokens`.
 
 .. _policy_auth_cache:
 
--- a/privacyidea/lib/policy.py
+++ b/privacyidea/lib/policy.py
@@ -357,6 +357,7 @@ class ACTION(object):
     CLIENTTYPE = "clienttype"
     REGISTERBODY = "registration_body"
     RESETALLTOKENS = "reset_all_user_tokens"
+    ALWAYSINCREASEFAILCOUNTER = "always_increase_failcounter"
     ENROLLPIN = "enrollpin"
     MANAGESUBSCRIPTION = "managesubscription"
     SEARCH_ON_ENTER = "search_on_enter"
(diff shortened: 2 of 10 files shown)


NNLemling:
TODO:
----------

- [ ] Appropriate name for policy wanted!
- [ ] Tests are not working yet
- [ ] Policy dont work with validate/triggerchallenge
- [ ] Supplement / adapt documentation if necessary

doc/policies/authentication.rst
```diff
@@ -390,6 +391,20 @@ type: bool
 If a user authenticates successfully all failcounter of all of his tokens
 will be reset. This can be important, if using empty PINs or *otppin=None*.
 
+always_increase_failcounter
```
cornelinux, line 394: I would make clear, that this is in case of challenge response!

So the new policy should be named s.th. like `increase_failcounter_on_challenge_response`. (Or do I get s.th. wrong?)

doc/policies/authentication.rst
```diff
@@ -390,6 +391,20 @@ type: bool
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+type: bool
+
```
cornelinux, line 398: Make clear, that it is about challenge response:

The normal behaviour is: to not increase the failcounter in case of challenge response.
If this policy is activated, the failcounter will be increased if the correct pin or password was given and a challenge was triggered.

privacyidea/lib/policy.py
```diff
@@ -2296,6 +2297,11 @@ def get_static_policy_definitions(scope=None):
                 'desc': _('If a user authenticates successfully reset the '
                           'failcounter of all of his tokens.')
             },
+            ACTION.ALWAYSINCREASEFAILCOUNTER: {
+                'type': 'bool',
+                'desc': _('If a user enters a correct pin or userstore password,'
+                          '  increases the failcounter of all of his tokens.')
```
cornelinux, line 2303: No, not all of his tokens.

Only these tokens, for which a challenge was triggered!

A user could have non-chal-resp tokens, and for those tokens there is no need to increase the failcounter!

privacyidea/lib/policydecorators.py
```diff
@@ -660,3 +660,32 @@ def reset_all_user_tokens(wrapped_function, *args, **kwds):
...
+    wrong_pin_or_otp = ['wrong otp pin', 'Response did not match the challenge.']
+    r = wrapped_function(*args, **kwds)
+    # A suitable tokenpin or/and the correct userstore password is entered
+    if not r[0] and g and allow_increase_failcount:
+        token_owner = tokenobject_list[0].user
+        increase_all = Match.user(g, scope=SCOPE.AUTH, action=ACTION.ALWAYSINCREASEFAILCOUNTER,
+                               user_object=token_owner if token_owner else None).policies()
+        if increase_all and r[1]['message'] not in wrong_pin_or_otp:
+            log.debug("Increases the failcounter of all tokens from {0!s}".format(token_owner))
+            for tok_obj_increase in tokenobject_list:
```
cornelinux, line 686: You must not increase the failcounter of all tokens.
You need to check if you have information in `r`, that tells you which tokens actually created a chal_resp.
And then you should only increase these tokens.

OR! You can do this within the check_token_list, which may be the better way to do it.
This would happen in this elif-branch:
https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/token.py#L2381

Maybe you should first define and create a working test and then it is easier to verify the correct behaviour.

privacyidea/lib/token.py
```diff
@@ -1972,7 +1973,8 @@ def check_realm_pass(realm, passw, options=None,
             return False, reply_dict
 
         return check_token_list(tokenobject_list, passw, options=options,
-                                allow_reset_all_tokens=False)
+                                allow_reset_all_tokens=False,
+                                allow_increase_failcount=False)
```
cornelinux, line 1977: Actually you do not need to add "False", if it is False by default anyways

cornelinux (commented):
Let's look into this togeather.
I think it might be easier to increase the failcounter in the check_token_list function.

doc/policies/authentication.rst
NNLemling: fixed in 4fe9f7f46e86b88dcd365672f86288d621345875.

doc/policies/authentication.rst
NNLemling: fixed in 4fe9f7f46e86b88dcd365672f86288d621345875.

privacyidea/lib/policy.py
NNLemling: fixed in 4fe9f7f46e86b88dcd365672f86288d621345875.

privacyidea/lib/policydecorators.py
NNLemling: fixed in 4fe9f7f46e86b88dcd365672f86288d621345875.

privacyidea/lib/token.py
NNLemling: fixed in 4fe9f7f46e86b88dcd365672f86288d621345875.

doc/policies/authentication.rst
```diff
@@ -391,20 +391,23 @@ type: bool
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 type: bool
 
-If a user logs in with their correct userstore password or a suitable tokenpin,
-the failcounter will still be increased for the time being.
+The normal behaviour is: to not increase the failcounter in case of challenge response.
+
+If this policy is activated the failcounter is increased for each token for which a challenge
+is triggered, even if the correct password or userstore pin is entered.
```
cornelinux, line 402: The "even" is superflous. Since the correct password or pin is the requirement for triggering the challenge.

privacyidea/lib/policy.py
cornelinux: I think this still needs to be fixed?

"Increase the failcounter for all the tokens, for which a challenge has been triggered."

cornelinux (changes requested):
Please check my minor comments

cornelinux (changes requested):
Also check your failing tests.

You have a faulty import of  "always_increase_failcounter".

doc/policies/authentication.rst
NNLemling: fixed in 770f80617c7499699940c14e4b075104444da030.

NNLemling:
> Also check your failing tests.

this should now be fixed in 66d31d5080f75060463c266667261effa9e106b2.

privacyidea/lib/policy.py
NNLemling: fixed in 770f80617c7499699940c14e4b075104444da030.

tests/test_lib_policydecorator.py
```diff
@@ -58,7 +58,12 @@ def fake_check_otp(dummy, pin, user=None, options=None):
         return pin == "FAKE"
 
     @staticmethod
-    def fake_check_token_list(tokenobject_list, passw, user=None, options=None, allow_reset_all_tokens=True,
+    def fake_check_token_list_reset(tokenobject_list, passw, user=None, options=None, allow_reset_all_tokens=True,
+                              result=False):
+        return result, "some text"    \
+
+    @staticmethod
+    def fake_check_token_list_increase(tokenobject_list, passw, user=None, options=None, allow_increase_failcount=True,
```
cornelinux, line 66: You are not using this method.
But worst of all you are not testing the increase at all.

Please ensure, that you test the functionality in the `check_token_list` (line 2390)

cornelinux (changes requested):

tests/test_lib_policydecorator.py
NNLemling: # You are not using this method.
fixed in a79404d8c496e829974bb3149e58145465639236.

cornelinux (approved):

PR #3380, opened 2022-11-15
3356/Adding missing sequence to migration scripts

jona-samuel:
I hope this works

### DIFF
--- a/migrations/versions/888b56ed5dcb_.py
+++ b/migrations/versions/888b56ed5dcb_.py
@@ -10,12 +10,28 @@ Create Date: 2021-02-10 12:17:40.880224
 revision = '888b56ed5dcb'
 down_revision = 'd415d490eb05'
 
-from alembic import op
+from alembic import op, context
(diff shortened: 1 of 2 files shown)


migrations/versions/888b56ed5dcb_.py
```diff
@@ -10,12 +10,28 @@
...
+
+
+def create_seq(seq):
+    if dialect_supports_sequences():
+        op.execute(CreateSequence(seq))
 
 
 def upgrade():
     try:
+        seq = Sequence('customuserattribute')
```
cornelinux, line 30: The sequence should always be named "customeruserattriute_seq", cincludeing _seq.

cornelinux (changes requested):
We need a new migration script, that only contain the creation of the sequence.

The problem might be to create an empty migration script.

Usually the migration script is created via:

pi-manage db migrate

Maybe it works with

pi-manage db revision

PR #3384, opened 2022-11-24
Adding missing sequence to migration scripts

### LINKED ISSUE #3356: Missing sequence for `customuserattribute` table in migration
When adding the `customuserattribute` table during migration, the sequence for the `id` is missing (needed by e.g. PostgreSQL):
https://github.com/privacyidea/privacyidea/blob/5cc92448833bf5de3bb0ab10cd669496cfcef7d0/migrations/versions/888b56ed5dcb_.py#L20
We should check if other tables added through migration scripts are missing the sequence definition.

jona-samuel:
The script is missing the downgrade options

### DIFF
--- a/migrations/versions/a28f2733897b_.py
+++ b/migrations/versions/a28f2733897b_.py
@@ -1,0 +1,57 @@
+"""Sequence for cusstomuserattribute, tokengroup and tokentokengroup
+
+Revision ID: a28f2733897b
+Revises: 89e57ed16379
+Create Date: 2022-11-24 11:05:42.572284
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = 'a28f2733897b'
+down_revision = '89e57ed16379'
+
+from alembic import op, context
+import sqlalchemy as sa
+from sqlalchemy.schema import Sequence, CreateSequence
+
+def dialect_supports_sequences():
+    migration_context = context.get_context()
+    return migration_context.dialect.supports_sequences
+
+
+def create_seq(seq):
+    if dialect_supports_sequences():
+        op.execute(CreateSequence(seq))
+
+
+def upgrade():
+    try:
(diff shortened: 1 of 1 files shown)


migrations/versions/a28f2733897b_.py
```diff
@@ -0,0 +1,57 @@
+"""Sequence for cusstomuserattribute, tokengroup and tokentokengroup
```
cornelinux, line 1: Please add a version prefix here. I think it is "v3.8: ...."

migrations/versions/a28f2733897b_.py
```diff
@@ -0,0 +1,57 @@
...
+    if dialect_supports_sequences():
+        op.execute(CreateSequence(seq))
+
+
+def upgrade():
+    try:
+        seq = Sequence('customuserattribute_seq')
+        try:
+            create_seq(seq)
+        except Exception as _e:
```
cornelinux, line 32: I think in this case we can skipt the inner exception and simpy to

~~~~python
try:
    seq = Sequence(...)
    create_seq(seq)
except Exeption as exx:
    print(exx)
~~~~

in all cases.

cornelinux (changes requested):

migrations/versions/a28f2733897b_.py
jona-samuel: 79cdc78

migrations/versions/a28f2733897b_.py
jona-samuel: 1b2bcbe

cornelinux (approved):

Ma27:
This is actually causing issues when trying to upgrade to 3.8, see also https://github.com/privacyidea/privacyidea/issues/3447

PR #3396, opened 2022-12-02
Speed Up MS CA Connector

### LINKED ISSUE #3365: Speed up CA Connector
Currently the CA connector is created on each request.
In the MSCA connector this means, the templates are fetched from the CA.

This could be done similar to the Policy object in the request? We could put this in the SharedConfigClass.

cornelinux:
The CA config is cashed in the SharedConfigObject. Thus we do not need to reread the templates on each request which speeds up the CA connector read dramatically.

Closes #3365

### DIFF
--- a/privacyidea/lib/caconnector.py
+++ b/privacyidea/lib/caconnector.py
@@ -42,6 +42,7 @@ from sqlalchemy import func
 from .crypto import encryptPassword, decryptPassword
 from privacyidea.lib.utils import (sanity_name_check, get_data_from_params, fetch_one_resource)
 from privacyidea.lib.utils.export import (register_import, register_export)
+from privacyidea.lib.config import get_config_object
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


privacyidea/lib/caconnector.py
```diff
@@ -329,3 +316,11 @@ def import_caconnector(data, name=None):
...
         log.info('Import of caconnector "{0!s}" finished,'
                  ' id: {1!s}'.format(res_data['caconnector'], rid))
+
+
+def get_all_caconnectors():
+    """
+    Shorthand to retrieve all caconnectors of the request-local config object
+    """
+    conf = get_config_object()
+    return conf.caconnectors
```
NNLemling, line 326: Add newline at the end.

Sebastian-Lemling (commented):
Please check my comment.

privacyidea/lib/caconnector.py
cornelinux: done in e177b2ba

Sebastian-Lemling (approved):

Issue #3408, opened 2022-12-12
Can't save SMTP without password 


maxbeth:
Version: master branch

If i save a new SMTP Server without user and password, i get an error
 ERR707: Failed to encrypt password!

cornelinux:
I can confirm this in version 3.8.

PR #3414, opened 2022-12-13
Solve None Password #3408

maxbeth:
encryptPassword is called with None at the first save SmtpServer, with an empty password.

### DIFF
--- a/privacyidea/lib/crypto.py
+++ b/privacyidea/lib/crypto.py
@@ -313,6 +313,8 @@ def encryptPassword(password):
     """
     hsm = get_hsm()
     try:
+        if password is None:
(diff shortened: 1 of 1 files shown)


privacyidea/lib/crypto.py
```diff
@@ -313,6 +313,8 @@ def encryptPassword(password):
     """
     hsm = get_hsm()
     try:
+        if password is None:
+            password = ""
```
cornelinux, line 317: You can do this more elegantly by one line:

    password = password or ""

cornelinux (changes requested):

cornelinux (approved):

PR #3417, opened 2022-12-15
Fix broken login from UI

### LINKED ISSUE #3416: Login with user@realm in webui not possible
This is a problem on the master branch we are currently working on.

plettich:
When a user logs in via the WebUI, there is always a `realm` parameter added to the `/auth` request, even if no realm is given (empty string). With the changes from fa643349 this lead to the user not being found if the login was `user@realm` and `realm` wasn't the default realm.

Closes #3416

### DIFF
--- a/tests/test_api_auth.py
+++ b/tests/test_api_auth.py
@@ -52,6 +52,9 @@ class AuthApiTestCase(MyApiTestCase):
             self.assertEqual(4031, result['error']['code'], result)
             self.assertEqual('Authentication failure. Wrong credentials',
                              result['error']['message'], result)
+        aentry = self.find_most_recent_audit_entry(action='POST /auth')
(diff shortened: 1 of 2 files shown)


tests/test_api_auth.py
```diff
@@ -120,6 +123,8 @@ def test_01_auth_with_split(self):
             # realm1 should be the default realm
             self.assertEqual('realm1', result['value']['realm'], result)
 
+        # TODO: Add test with empty realm parameter and user@realm not in default realm
```
cornelinux, line 126: Are you still planning on doing this? Like on Friday?

cornelinux (commented):
Looks good. Would you also add the TODO test?

cornelinux (approved):

Issue #3419, opened 2022-12-17
new grouping for system / config menu


maxbeth:
I think some points from the system should in the config, elsewhere it is necessary to switch to the system for select smtp/radius/... view

cornelinux:
What is your suggestion in plain English?

Please before issuing a pull request, describe your suggestion and what you want to do. 
So that we have a concept against which we can review the pull reuqest.

maxbeth:
I can't switch with a single click between SMTP-Server/RADIUS-Server/privacyIDEA Server/SMS-Gateway/Tokengroup.
I always have to select the System menu to show the menu.

cornelinux:
The logical difference between the configuration under `system` and the other configurations like `policies`, `events`, `realms` is, that everything under system is nothing that works for its own. It does not change the way privacyIDEA works.
There are only objects, that are configured to be used in other locations.
So it totally makes sense to group them underneath `system`.
I would even put `certificate authoritzies` under `system`.

However, I understand the usage concern, that switch from e.g. `sms gateway` to `radius server` is not easily possible. There need to be other ways than this clotting.

cornelinux:
I think s.th. like this might make sense:

Top menu:  | Dashboard  |   Token  | User |  **Config**  | Audit
Sub menu:   |  **System**  |  Policies  |   Events  |  Tasks ...
Sub System menu:   |  Systemconfig  |  SMTP  |   RADIUS   | privacyidea-server  |  ...

PR #3420, opened 2022-12-17
3419/system2config

maxbeth:
#3419 new menustructur to reduce the clicks


cornelinux:
I am sorry. but this is totally nasty :-/ 
it breaks the overall design and kills the overview.

Issue #2588, opened 2021-01-15
RADIUS server definition with round robin


cornelinux:
For using RADIUS tokens or RADIUS passthru we could configure multiple RADIUS server to do round robin.
The RADIUS server could just be a comma separated list in the RADIUS server definition.

(This superceeds #819)

PR #3422, opened 2022-12-19
2588/radius robin

maxbeth:
For better redundancy, the radius can now round-robin.
Issue #2588

### DIFF
--- a/privacyidea/api/radiusserver.py
+++ b/privacyidea/api/radiusserver.py
@@ -118,7 +118,7 @@ def test():
     """
     param = request.all_data
     identifier = getParam(param, "identifier", required)
-    servers = getParam(param, "server", required)
+    server = getParam(param, "server", required)
     port = int(getParam(param, "port", default=1812))
     secret = getParam(param, "secret", required)
     retries = int(getParam(param, "retries", default=3))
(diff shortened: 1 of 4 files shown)


privacyidea/api/radiusserver.py
```diff
@@ -127,9 +127,16 @@ def test():
     password = getParam(param, "password", required)
     dictionary = getParam(param, "dictionary",
                           default="/etc/privacyidea/dictionary")
-
-    r = test_radius(identifier, server, secret, user, password, port=port,
+    r = 0
+    for server in servers.split(','):
```
cornelinux, line 131: Please do

    for server in [x.strip() for x in servers.split(','):

This way we will also be able to handle lists like

1.2.3.4:1812, 2.3.4.5:1645

which contain whitespaces.

privacyidea/api/radiusserver.py
```diff
@@ -127,9 +127,16 @@ def test():
...
+        if len(serveraddr) >= 2:
+            serverport = int(serveraddr[1])
+        else:
+            serverport = port
+        r = test_radius(identifier, serveraddr[0], secret, user, password, port=serverport,
                     dictionary=dictionary, retries=retries, timeout=timeout)
-    g.audit_object.log({'success': r > 0,
+        g.audit_object.log({'success': r > 0,
                         'info':  r})
     return send_result(r > 0)
```
cornelinux, line 142: Hm, I am not sure, what the result should be here.

So if the first server fails, but the 2nd succeeds, we get a success.
If the first is successful and the 2nd fails, we get a fail.

cornelinux (changes requested):
I have not checked the radiustoken.py, yet.
But please see my comments in radiusserver.py.
Also add some docs about redundancy to 
https://github.com/privacyidea/privacyidea/blob/master/doc/configuration/radiusserver_config.rst

privacyidea/api/radiusserver.py
maxbeth: In the privacyidea//lib/tokens/radiustoken.py is a strip also

privacyidea/api/radiusserver.py
maxbeth: I return now the first response

cornelinux:
THanks a lot for the input.
We are thinking about reworking the whole radius implementation.
So I am closing this for now.

Issue #3423, opened 2022-12-19
delete a assigned tokengroup and the tokenlist can't be shown


maxbeth:
version: master
Create a tokengroup.
Create token and assign the tokengroup
delete the tokengroup 
the tokenlist is empty and a none valid json-response error raise

plettich:
I can not reproduce this, actually, i get an internal server error due to the fact that the database entry for the tokengroup has a foreign key constraint to the `tokentokengroup` table. So in fact it should only be able to be deleted, when no tokens are assigned to the tokengroup anymore. We should make sure to check the constraint like you can not delete a resolver when it is still part of a realm.

PR #3424, opened 2022-12-19
3423/missing tokengroup

maxbeth:
Fix the #3423 the missing token group is shown with "Missing <TokenTokenGroup #id>"


maxbeth:
OK, it is a better way to avoid to delete a used tokengroup

Issue #3425, opened 2022-12-19
Whole header from bootstrap accordion clickable


maxbeth:
Only the small headertext is clickable.
For a better UX the whole Header should be clickable.

PR #3426, opened 2022-12-19
3425/accordion header clickable

maxbeth:
Idea for #3425


plettich:
Hi @maxbeth, thanks again for the PR. I took Your changes and adjusted them for our needs.
Closing this in favor of #3522

PR #3437, opened 2022-12-31
3146/upgrade privacyIDEA diag tool

### LINKED ISSUE #3146: Improve privacyidea-diag script for use with nginx
Currently the `privacyidea-diag` script only gathers configuration and log file for the apache webserver. We should enable collecting data for nginx as well (and avoid sputtering out errors due to missing apache files).
We should also check if the script is running as `root` to avoid complications and missing files.

Some more ideas for the diag-script:
- [x] add nginx logs if available
- [x] check if script is run as root
- [x] add smtp/privacyIDEA/RADIUS server configuration to the diagnostics
- [x] check for external logging config (and avoid error when the PI logfile doesn't exist)
- [x] check if apache logs exist before appending them
- [x] create different files for system config and privacyIDEA config

NNLemling:
closes #3146

### DIFF
--- a/tools/privacyidea-diag
+++ b/tools/privacyidea-diag
@@ -31,6 +31,11 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
+if [[ $(id -u) != 0 ]]; then
+  echo "Script must be run as root user"
+  exit 1
+fi
+
 PICFG="/etc/privacyidea/pi.cfg"
 
 if [ "$#" -gt 0 ]; then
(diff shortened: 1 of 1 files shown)


tools/privacyidea-diag
```diff
@@ -56,6 +56,16 @@ case $OS in
...
+check_root() {
+	log "SECTION: Check Root"
+	log "==================="
+  if [ "$(id -u)" == 0 ]; then
+    log "This script is running as root."
+  else
+    log "This script is not running as root."
+  fi
+}
+
```
plettich, line 68: i was thinking more of something like
```
if [[ $(id -u) != 0 ]]; then
  echo "Script must be run as root user"
  exit 1
fi
```
right at the beginning of the script.

plettich (changes requested):
please see my comment

tools/privacyidea-diag
NNLemling: fixed in 808035780551993c88b00257271d96f89bbac166

plettich (approved):

PR #3443, opened 2023-01-05
Send correct QR-code image to client

### LINKED ISSUE #3427: Wrong QR-Code sent during enroll_via_validate
If `enroll_via_multichallenge` is active, the QR-code sent after the first request only contains the token seed, not the complete `otpauth://` uri in case of enrolling HOTP/TOTP token.
When calling [`get_init_detail()`](https://github.com/privacyidea/privacyidea/blob/b1e6c1d073190d68e5171bd5870cec99edda1ab8/privacyidea/lib/tokens/hotptoken.py#L835) and passing the user object, the complete QR-code is available in the dictionary as `googleurl` and `oathurl` (See [`POST /token/init`](https://privacyidea.readthedocs.io/en/latest/modules/api/token.html?highlight=google#post--token-init)).

plettich:
The QR-code sent to the client after enrollment during validate/check was incorrect, it just contained the token secret without the necessary HOTP/TOTP parameter.

Fixes #3427

### DIFF
--- a/privacyidea/lib/tokens/hotptoken.py
+++ b/privacyidea/lib/tokens/hotptoken.py
@@ -836,7 +836,6 @@ class HotpTokenClass(TokenClass):
         detail["transaction_ids"] = [c[2]]
         chal = {"transaction_id": c[2],
                 "image": init_details.get("googleurl").get("img"),
-                "token_uri": init_details.get("googleurl").get("value"),
                 "client_mode": CLIENTMODE.INTERACTIVE,
                 "serial": token_obj.token.serial,
                 "type": token_obj.type,
--- a/privacyidea/lib/tokens/hotptoken.py
+++ b/privacyidea/lib/tokens/hotptoken.py
@@ -832,10 +832,11 @@ class HotpTokenClass(TokenClass):
         # Create a challenge!
         c = token_obj.create_challenge()
         # get details of token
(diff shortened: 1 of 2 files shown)


privacyidea/lib/tokens/hotptoken.py
```diff
@@ -832,10 +832,11 @@ def enroll_via_validate(cls, g, content, user_obj):
         # Create a challenge!
         c = token_obj.create_challenge()
         # get details of token
-        init_details = token_obj.get_init_detail()
+        init_details = token_obj.get_init_detail(user=user_obj)
         detail["transaction_ids"] = [c[2]]
         chal = {"transaction_id": c[2],
-                "image": init_details.get("otpkey").get("img"),
+                "image": init_details.get("googleurl").get("img"),
+                "token_uri": init_details.get("googleurl").get("value"),
```
cornelinux, line 839: I would remove this. This is currently not documented and defined.
I would like to avoid using adding an undocumented feature here.

Also there are no defined cases, where this text represantation can actually be used.

cornelinux (changes requested):

privacyidea/lib/tokens/hotptoken.py
plettich: Done in 0fe459c1

cornelinux (approved):

PR #3473, opened 2023-01-20
3454/UserID from SQLIdResolver is returned as the type defined in the database

jona-samuel:
Convert userid to Unicode in SQLIdResolver.py


cornelinux (approved):
Well, I checked this with my local SQL configuration it looks to work out fine.

PR #3480, opened 2023-01-20
Fix the wrong naming of ca_connector

### LINKED ISSUE #3479: certifcate_ca_connector Policy does not work
It looks like the policy "certificate_ca_connector" policy does not overwrite the "ca" parameter sent in the `/token/init` request.

The problem is, that the policy actions "ca_connector" are fetched here:

https://github.com/privacyidea/privacyidea/blob/66d31d5080f75060463c266667261effa9e106b2/privacyidea/api/lib/prepolicy.py#L622-L625

cornelinux:
The policy action that is saved is certificate_ca_connector. So when fetching the policy in the prepolicy, we will not find the configuration and query the wrong ca.

Also add a failsafe warning to the log file, if a wrong CA connector name is specified.

Fixes #3479

### DIFF
--- a/privacyidea/lib/caconnector.py
+++ b/privacyidea/lib/caconnector.py
@@ -292,6 +292,9 @@ def get_caconnector_object(connector_name):
                     connector_config[conf.Key] = value
                 c_obj.set_config(connector_config)
 
+    if not c_obj:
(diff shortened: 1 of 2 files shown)


plettich (approved):
Not sure how the wrong name got there in the first place but if it works...

PR #3487, opened 2023-01-25
Update compiled translation files for v3.8

plettich:
Also fix some minor translation glitches in the view files.


cornelinux (approved):
We would have to verify this within weblate.

PR #3488, opened 2023-01-25
3478/validate check enroll hotp totp parameters

### LINKED ISSUE #3478: tokenlabel policies do not work in multichallenge-enroll
When enrolling e.g. HOTP token via multichallenge-enroll, the tokenlabel policy is not used.

Maybe also other policies like *pin-protect token on smartphone* do not work.

Take the config from the decorator `init_tokenlabel` (https://github.com/privacyidea/privacyidea/blob/66d31d5080f75060463c266667261effa9e106b2/privacyidea/api/lib/prepolicy.py#L561) and add it to the QR code during validate-check-enrollment.

This is done in here: https://github.com/privacyidea/privacyidea/blob/0fe459c1054e03ea578bd809d71e7599cfc7414b/privacyidea/lib/tokens/hotptoken.py#L835

We have to pass the `params`:

- tokenlabel
- tokenissuer
- appimageurl
- force_app_pin

Maybe we could extract this code 
https://github.com/privacyidea/privacyidea/blob/66d31d5080f75060463c266667261effa9e106b2/privacyidea/api/lib/prepolicy.py#L582-L603

and move it to a helper function that we call from `enroll_via_validate` to be able to pass these to `get_init_details` in line 835.

We  habe to check that this works with 
* [x] HOTP, like tokenlabel, tokenissuer
* [x] TOTP like tokenlabel, tokenissuer
* [ ] Push  like forcepin and enrollment thinks like verify_ssl. Need to make this decorator https://github.com/privacyidea/privacyidea/blob/eddfc338f5e09e85b9148dd72eb52c2c91cf4d17/privacyidea/api/lib/prepolicy.py#L1475 available in the validate-check-enroll method.
* [x] Email ???
* [x] SMS? ???

# Redesign

I think we have to da a major redesign. Other parameters like the otp_size, keyisze are also not used.
These are all passed via the API parameters.
We also need to ensure, that these parameters from policies are passed to the class method `enroll_via_validate`.

Or can we simply use the init-decorators on validate/check? and add the parameters to the reqeuest? We would "polute" the parameters of validate/check with parameters that are not used in 99% of the time and potentially slow down authentication.

Version 1: https://github.com/privacyidea/privacyidea/tree/3478/v1-parameters-in-enroll-via-validate with helper function

Version 2:  https://github.com/privacyidea/privacyidea/tree/3478/v2-use-parameters-for-validate-check-enrollment with reusing decorators

cornelinux:
As I would like to also add the PUSH parameters, this is still work in progress.

This closes #3478


cornelinux:
@plettich  Please review. If this slightly ugly design is acceptable for now, then I would to the push token in a similar way. Which might be quite easy since it has one single decorator.

cornelinux:
@plettich I would like to restink the location of the code before merging. It feels like the file lib/policy.py is a bad location for the two new helper functions get_init_tokenlabel_params and get_pushtoken_add_config. lib/policy.py contains basic policy functions.

These two are quite specifc functions for rollout.
I am not sure if I would put the into api/lib/prepolicy.py (where the code originially was from) or into the tokenclass files or in a prepolicy_helper.py module...

Where would someone search for such code?

tests/test_api_push_validate.py
```diff
@@ -396,17 +396,6 @@ def test_10_enroll_push(self):
         # Set Policy scope:auth, action:enroll_via_multichallenge=push
         set_policy("pol_multienroll", scope=SCOPE.AUTH,
                    action="{0!s}=push".format(ACTION.ENROLL_VIA_MULTICHALLENGE))
-        # We still do a passthru, since we do not have a firebase config, yet.
-        with self.app.test_request_context('/validate/check',
```
cornelinux, line 400: I remove this test, since it only occurred in an not complete configuration.
We now also raise an error, if the firebase config is missing, which also makes more sense.

privacyidea/api/lib/policyhelper.py
```diff
@@ -0,0 +1,117 @@
...
+import logging
+from privacyidea.lib.log import log_with
+from privacyidea.lib.policy import Match, SCOPE, ACTION
+from privacyidea.lib.error import PolicyError
+
+log = logging.getLogger(__name__)
+
+
+@log_with(log)
+def get_init_tokenlabel_parameters(g, params=None, token_type="hotp", user_object=None):
```
plettich, line 37: since this is not only for the tokenlabel, we should rename the function to something like `get_init_authenticator_paramsters()`

privacyidea/lib/tokens/hotptoken.py
```diff
@@ -832,7 +833,10 @@ def enroll_via_validate(cls, g, content, user_obj):
         # Create a challenge!
         c = token_obj.create_challenge()
         # get details of token
-        init_details = token_obj.get_init_detail(user=user_obj)
+        enroll_params = get_init_tokenlabel_parameters(g, user_object=user_obj,
+                                                       token_type=cls.get_class_type())
+        init_details = token_obj.get_init_detail(params=enroll_params,
+                                                 user=user_obj)
```
plettich, line 839: since we don't go through [`create_challenges_from_tokens()`](https://github.com/privacyidea/privacyidea/blob/630066ef53e4bf8916a3c341dd91fe1768f5713c/privacyidea/lib/token.py#L2116) we are missing the `messages` attribute here.

plettich (commented):
just some minor remarks, we can change them later as well.

privacyidea/api/lib/policyhelper.py
cornelinux: In a perfect world you are right.

I chose this name, since the decorator is actually called "init_tokenlabel". So I wanted to match the actual decorator - even if the naming of the decorator is also not perfect.

privacyidea/lib/tokens/hotptoken.py
cornelinux: Done in fb0bb613

cornelinux: in 11dc6504 it is even a list!

plettich (approved):

Issue #3337, opened 2022-10-17
Optimize URL-decoding


plettich:
If a remote token is used without checking the pin locally, the password is sent as is to the remote server. A percent sign in the password (and/or user?) will break authentication since it is not properly encoded here:
https://github.com/privacyidea/privacyidea/blob/c08154951bb02ddc1794157bd19c7d699d20d5e9/privacyidea/lib/privacyideaserver.py#L68
`realm`, `resolver`, `serial` and `transaction_id` shouldn't be a problem since they shouldn't contain a percent sign.
Normally, `requests` and `Flask` should take care of the data encoding/decoding when the correct content type (`application/x-www-form-urlencoded`) is set. But since we do an extra decoding step here:
https://github.com/privacyidea/privacyidea/blob/93b7aa9896f3714cab47d9f0d4ce35a5a75fa67b/privacyidea/api/lib/utils.py#L255
we need to encode the data before sending.

cornelinux:
I am not sure if all applications, plugins (3rd party plugins) always handle everything correctly.

So we might add a mapping or a policy to define, for which requests we should do urldecoding and for which one not.

plettich:
> I am not sure if all applications, plugins (3rd party plugins) always handle everything correctly.

The perl LWP module (FreeRADIUS) does it correctly.

> 
> So we might add a mapping or a policy to define, for which requests we should do urldecoding and for which one not.

If we receive a (POST-) request with the content type `application/x-www-form-urlencoded` we must assume, that the sender did the correct encoding and that Flask did take care of the decoding before passing the data to us.
Of course we need to check all plugins/SDKs for the correct behavior.

cornelinux:
> If we receive a (POST-) request with the content type application/x-www-form-urlencoded we must assume,

The requirement to an assumption is the beginning of the downtime ;-)

I am not talking about LWP module or whatsoever.

I am talking about people who develop arbitrary plugins in an uncommon way and maybe send `form-urlendoded` but do not encode.

Having  such a mapping would enable us, if a "customer" uses a broken 3rd party plugin to solve this problem in the server, without touching the untouchable 3rd party plugin.

cornelinux:
Need to rediscuss

cornelinux:
The decoding happens in api/lib/utils.py here:
https://github.com/privacyidea/privacyidea/blob/93b7aa9896f3714cab47d9f0d4ce35a5a75fa67b/privacyidea/api/lib/utils.py#L241

[...]

PR #3489, opened 2023-01-25
Check if data needs to be unquoted

plettich:
Based on the user-agent string, check if the given data needs to be unquoted again.

Working on #3337

### DIFF
--- a/privacyidea/api/lib/utils.py
+++ b/privacyidea/api/lib/utils.py
@@ -37,6 +37,7 @@ import jwt
 import threading
 import six
 import re
+from copy import copy
 from six.moves.urllib.parse import unquote
 from flask import (jsonify,
                    current_app)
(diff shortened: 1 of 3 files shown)


privacyidea/api/lib/utils.py
```diff
@@ -238,6 +239,32 @@ def getLowerParams(param):
...
+    :return: New dictionary with the possibly unquoted values
+    :rtype: dict
+    """
+    # if no user agent is available, we assume that we must unquote the data
+    if not request.user_agent.string:
+        return {key: unquote(value) for (key, value) in data.items()}
+
+    ua_match = re.match(r'^(?P<agent>[a-zA-Z0-9_-]+)(/(?P<version>\d+[\d.]*)(\s.*)?)?',
+                        request.user_agent.string)
+    if ua_match and not ua_match.group('agent') in ['privacyIDEA-LDAP-Proxy']:
```
cornelinux, line 262: Would you please put this in a constant at the beginning of this file near "TRUSTED_JWT_ALGOS" like

    NOT_QUOTED_USER_AGENTS = ["privacyIDEA-LDAP-Proxy"]

cornelinux (commented):
Just one minor nitpicking.

You can then silently merge it.

privacyidea/api/lib/utils.py
plettich: Done in e173060

cornelinux (approved):

PR #3492, opened 2023-01-26
Fixing tags in email token

jona-samuel:


cornelinux:
Please remove the non-existing tags from the docs.
https://github.com/privacyidea/privacyidea/blob/4ce99349a24e92959961cd2deb30d1a26b86e442/doc/policies/authentication.rst#L165

privacyidea/lib/tokens/emailtoken.py
```diff
@@ -417,7 +418,7 @@ def _get_auto_email(options):
         return autosms
 
     @log_with(log)
-    def _compose_email(self, message="<otp>", subject="Your OTP", mimetype="plain"):
+    def _compose_email(self, options, message="<otp>", subject="Your OTP", mimetype="plain"):
```
cornelinux, line 421: I recommend to add new parameters in a function always at the end as keyword parameters.
This way you can avoid to run into problems if there are function calls, that you do not update.

~~~~python
def _compose_email(self, message="<otp>", subject="Your OTP", mimetype="plain", options=None):
~~~~

Also, do not use empty dicts as default values.
This is why you need to initialize the options, in case it is not set. (see below)

privacyidea/lib/tokens/emailtoken.py
```diff
@@ -435,6 +436,7 @@ def _compose_email(self, message="<otp>", subject="Your OTP", mimetype="plain"):
         recipient = self._email_address
         otp = self.get_otp()[2]
         serial = self.get_serial()
```
cornelinux, line 438: Add the initialization of the options

~~~~python
options = options or {}
~~~~

cornelinux (changes requested):

cornelinux (approved):

Issue #3490, opened 2023-01-25
Apple WebAuthn/Passkey Enrollment fails


fritterhoff:
### Top-level intent

I am trying to enroll a WebAuthN Token using Apple Passkeys & Mac OSX Safari. This fails due to an error. The frontend only claims: "ERR404: Could not enroll webauthn token!". The backend log provide the additional information "Enrollment of webauthn token failed: Registration rejected. Error: Unable to verify client extensions.!"

Until now I tried several config flags starting with different webauthn_authenticator_attestation_levels or webauthn_authenticator_attestation_forms. Sadly I did not find any working solution.  

### Steps to reproduce

1. Try to enroll WebAuthN on Mac OS X using Safari and Apple Passkey
<img width="607" alt="grafik" src="https://user-images.githubusercontent.com/32478819/214620381-9bcc4420-0d9a-41ee-b799-c90de26034a4.png">

### Expected outcome

The Key should get registered.

### Actual outcome

It fails.

### Configuration

* **privacyIDEA version**: 3.8
* **Installation method**: PyPi
* **Python version**: Python 3.10.6
* **Operating system**: Mac OS X
* **Webserver**: Apache 2
* **Token database**: (MySQL, PostgreSQL, ...) sqlite (I assume this should have no effect ;-) )

### Log file
```
[2023-01-25 17:35:41,470][299356][140466881979968][INFO][privacyidea.lib.user:262] user 'professorin' found in resolver 'default'
[2023-01-25 17:35:41,470][299356][140466881979968][INFO][privacyidea.lib.user:264] userid resolved to '186545d6-cd5e-1035-8ff4-f189f3efbfdd'
[2023-01-25 17:35:41,586][299356][140466881979968][INFO][privacyidea.lib.crypto:116] Requesting secret key - verify the usage scope and zero + free
[2023-01-25 17:35:41,587][299356][140466881979968][INFO][privacyidea.lib.crypto:116] Requesting secret key - verify the usage scope and zero + free
[2023-01-25 17:35:41,587][299356][140466881979968][INFO][privacyidea.lib.crypto:116] Requesting secret key - verify the usage scope and zero + free
[2023-01-25 17:35:41,587][299356][140466881979968][INFO][privacyidea.lib.crypto:116] Requesting secret key - verify the usage scope and zero + free
[2023-01-25 17:35:41,588][299356][140466881979968][INFO][privacyidea.lib.crypto:116] Requesting secret key - verify the usage scope and zero + free
[...]

PR #3497, opened 2023-01-27
ignore rk client extension

fritterhoff:
This PR provides a simple fix for #3490


plettich:
Thanks for the PR, i'll have to look into it to make sure we don't break something else.

PR #3511, opened 2023-02-05
Add Set and !Set comparators to Token and user info

cbergmann:
This PR allows to test for the existence of a user- and tokeninfo key.
My requirement was that the first token a User enrolls must be a TAN token (for recovery). Thereafter the user is only allowed to enroll and delete webAuthn and TOTP tokens.
I implemented it the following way:
 One Policy only matches if the user has userinfo "enrolled" not set and only allows enrollment of  TAN tokens.
 Another Policy one matches if the user has userinfo "enrolled" set and its value is "1" and allows enrollment of the other two tokens.
 Via events the userinfo "enrolled" is set to 1 whenever a tan token is enrolled and deleted when the last token is deleted.


cornelinux:
Please create an issue, so that a pull request can refer to the issue.
See https://github.com/privacyidea/privacyidea/blob/master/CONTRIBUTING.md#develop

Also add documentation here https://github.com/privacyidea/privacyidea/blob/master/doc/policies/conditions.rst

Issue #3512, opened 2023-02-05
New comperator to test if given user- or tokeninfo is set


cbergmann:
My requirement was that the first token a User enrolls must be a TAN token (for recovery). Thereafter the user is only allowed to enroll and delete webAuthn and TOTP tokens.

My PR allows to test for the existence of a user- and tokeninfo key.

With this I could fulfill my requirement in the following way:
One Policy only matches if the user has userinfo "enrolled" not set and only allows enrollment of TAN tokens.
Another Policy one matches if the user has userinfo "enrolled" set and its value is "1" and allows enrollment of the other two tokens.
Via events the userinfo "enrolled" is set to 1 whenever a tan token is enrolled and deleted when the last token is deleted.

On other option would have been to implement a policy rule to test the number of enrolled tokens like the option  user_token_number in event conditions.

cornelinux:
I think your approach of the PR is not generic and only has your own problem in mind.

The implementation is also counter intuitive, since the value is not evaluated. Each admin will wonder what to enter in the value field for the extended condition.

Currently the comparison fails, if the key does not exist, since the condition can not be evaluated.

My generic approach would be to add a flag to the extended conditions that will tell how to evaluate the condition, if the condition can not be evaluated (e.g. the key does not exist) ;-)

Something like a checkbox 

   [ ] evaluate to true if key does not exist.

or a triple state

   true, rais exception (current behaviour), false

This would not only solve your problem but all others.

cbergmann:
hi, I just pushed the first rough implementation idea of a solution to this problem to my fork. Before I invest more time into tests and documentation I would be interested in your opinion on my idea. It would be nice if you could review the [commit](https://github.com/privacyidea/privacyidea/commit/32f212270e21a703471ef82b7d55aa14c8313036).

Kind regards
Clemens

jelinaunger:
Done in #4346

PR #3513, opened 2023-02-05
3512/New comperator to test if given user- or tokeninfo is set

cbergmann:
New comperator to test if given user- or tokeninfo is set
Issue #3512


cornelinux:
This PR does not work alone but only in combination with the closed one.
Can you please check my comment in the issue. Thanks.

PR #3527, opened 2023-02-19
3419/restructure config menu

### LINKED ISSUE #3419: new grouping for system / config menu

cornelinux:
Closes #3419


plettich (approved):
Looks good.
The translation is broken probably because the file of the origin strings changed. It will be fixed with the next translation update.

Issue #3507, opened 2023-02-02
Improve doc for validate-check-enrollment


cornelinux:
We should add information to the documentation of validate-check-enrollment some notes, that in this rollout type certain policies (like force description) do not work.

* [x] force_description
* [x] Set pin
* [x] verify-enrollment

NNLemling:
In the event that we come up with other policies that are affected here, I'll leave the issue open.

PR #3530, opened 2023-02-22
adjust enroll_via_multichallenge documentation

NNLemling:
working on #3507.


jona-samuel (approved):

cornelinux:
@NNLemling 

The policy "verify_enrollment" (https://privacyidea.readthedocs.io/en/latest/policies/enrollment.html#verify-enrollment) is also not used/checked, since it does not make any sense, since the user verified the enrollment explicitly by authenticating with an OTP value.

Please add this and then this issue can be closed for now.

PR #3531, opened 2023-02-22
3335\Adding set token application to tokenhandler

jona-samuel:


privacyidea/lib/eventhandler/tokenhandler.py
```diff
@@ -86,6 +87,12 @@ class ACTION_TYPE(object):
     SET_RANDOM_PIN = "set random pin"
     ADD_TOKENGROUP = "add tokengroup"
     REMOVE_TOKENGROUP = "remove tokengroup"
+    ATTACH_APPLICATION = "attach application"
+
+class TOKEN_APPLICATIONS(object):
+    SSH = "ssh"
+    OFFLINE = "offline"
+    LUKS = "luks"
```
cornelinux, line 95: There is the function 
get_application_types() in applications/base.py which would return the defined list of the available applications.
However, since further below we need these values in "visibleValue", I am not sure, if this is worth it.

So maybe it is the most sensible way to define it here, again. ...and keep it this way.

privacyidea/lib/eventhandler/tokenhandler.py
```diff
@@ -312,7 +319,66 @@ def actions(cls):
...
                                 "value": [tg.name for tg in get_tokengroups()]
                             }
-                        }
+                        },
+                   ACTION_TYPE.ATTACH_APPLICATION:
+                       {"machine ID":
+                            {
+                                "type": "string",
+                                "required": True,
+                                "description": _("The ID of the machine you want to attach the token to")
```
cornelinux, line 328: offline does not attach to a specific token anymore.
With offline, this can be left empty and it then will be connected to "any machine".

So it should not be required=True.

privacyidea/lib/eventhandler/tokenhandler.py
```diff
@@ -312,7 +319,66 @@ def actions(cls):
...
+                            {
+                                "type": "string",
+                                "required": True,
+                                "description": _("Set a token application like 'offline' or 'SSH'. Note: Not all tokens"
+                                                 " work well with all applications!"),
+                                "value": [TOKEN_APPLICATIONS.SSH, TOKEN_APPLICATIONS.OFFLINE, TOKEN_APPLICATIONS.LUKS]
+                            },
+                        "count":
+                            {
+                                "type": "string",
```
cornelinux, line 341: Please add a description like 

~~~Python
"description": _("The number of offline OTP values available")
~~~

privacyidea/lib/eventhandler/tokenhandler.py
```diff
@@ -312,7 +319,66 @@ def actions(cls):
...
+                                "visibleIf": "application",
+                                "visibleValue": TOKEN_APPLICATIONS.OFFLINE,
+                                "required": False
+                            },
+                        "rounds":
+                            {
+                                "type": "string",
+                                "visibleIf": "application",
```
cornelinux, line 349: Please add a description like 

~~~Python
"description": _("The number of rounds for password hashing")
~~~

privacyidea/lib/eventhandler/tokenhandler.py
```diff
@@ -312,7 +319,66 @@ def actions(cls):
...
+                                "visibleValue": TOKEN_APPLICATIONS.OFFLINE,
+                                "required": False
+                            },
+                        "user":
```
cornelinux, line 353: With offline we do not use the user anymore. You can remove this.

tests/test_lib_events.py
```diff
@@ -2426,6 +2426,54 @@ def test_13_tokengroup(self):
...
+        req.all_data = {"serial": "offHOTP", "type": "hotp"}
+        resp = Response()
+        resp.data = """{"result": {"value": true}}"""
+
+        # The count window of the token will be set to 123
+        options = {"g": g,
+                   "request": req,
+                   "response": resp,
+                   "handler_def":{
+                       "options": {"machine ID": 0,
```
cornelinux, line 2460: Please remove the "machine ID"  here.

tests/test_lib_events.py
```diff
@@ -2426,6 +2426,54 @@ def test_13_tokengroup(self):
...
+
+        t_handler = TokenEventHandler()
+        res = t_handler.do("attach application", options=options)
+        self.assertTrue(res)
+
+        # check if the options were set.
+        token_obj = get_tokens(serial="offHOTP")[0]
+        self.assertEqual(token_obj.token.machine_list[0].application, "offline")
+        self.assertEqual(token_obj.token.machine_list[0].option_list[0].mt_key,
+                         "count")
```
cornelinux, line 2474: Please check, that the following has been set in the option list:

hostname -> "any host"
machine_id -> "any machine"

cornelinux (changes requested):
a few comments about the options, and please add an option test.

privacyidea/lib/eventhandler/tokenhandler.py
cornelinux: @jona-samuel please fill in the answers in regards to my comments.

cornelinux (approved):

Issue #2798, opened 2021-06-26
Upgrade SQLAlchemy to v2.4


plettich:
Currently SQLAlchemy v2.4 doesn't work out of the box, we need to check all the possible implications when switching.
- [ ] find all ocurences where SQLAlchemy v2.4 does not work
  - Remove sqlsoup and replace with SQLAlchem directly
    - [ ] SQLResolver
- [ ] Fix

This issue was brought up in #2790

cornelinux:
@plettich  In regards to replaceing SQLSoup we might also look into #3130 and consider adding a table schema in the SQL resolver config. This might also speed up resolution.

PR #3535, opened 2023-02-23
Remove SQLSoup

plettich:
In order to switch to SQLAlchemy > 1.3 we need to remove SQLSoup to access the user database in the SQLResolver.
SQLAlchemy can do this job alone.

Working on #2798


privacyidea/lib/token.py
```diff
@@ -1071,34 +1071,38 @@ def init_token(param, user=None, tokenrealms=None,
...
 
     except Exception as e:  # pragma: no cover
-        log.error('token create failed!')
+        log.error('token create failed: {0!s}'.format(e))
         log.debug("{0!s}".format(traceback.format_exc()))
-        raise TokenAdminError(_("token create failed {0!r}").format(e), id=1112)
+        # delete the newly created token from the db
+        if token_count == 0:
+            db_token.delete()
+        raise
```
cornelinux, line 1104: I would reccomend raising the adminError again.

privacyidea/lib/resolvers/SQLIdResolver.py
```diff
@@ -449,17 +383,17 @@ def getUserList(self, searchDict=None):
...
+            conditions.append(self.TABLE.columns[column].like(value))
 
         conditions = self._append_where_filter(conditions, self.TABLE,
                                                self.where)
         filter_condition = and_(*conditions)
 
-        result = self.session.query(self.TABLE).\
-            filter(filter_condition).\
-            limit(self.limit)
+        result = self.session.execute(select(self.TABLE).
```
cornelinux, line 392: Is using `execute` OK or could it be prone to security issues?

privacyidea/lib/resolvers/SQLIdResolver.py
```diff
@@ -728,15 +653,13 @@ def delete_user(self, uid):
...
-            conditions = []
-            conditions.append(self._get_userid_filter(uid))
+            conditions = [self._get_userid_filter(uid)]
             conditions = self._append_where_filter(conditions, self.TABLE,
                                                    self.where)
             filter_condition = and_(*conditions)
-            user_obj = self.session.query(self.TABLE).filter(
-                filter_condition).first()
-            self.session.delete(user_obj)
+            self.session.execute(self.TABLE.delete().filter(filter_condition))
```
cornelinux, line 660: Is it `delete().filter()` or should it be `filter(...).delete()`.

or `...execute(delete(self.TABLE).filter(...)` ?

cornelinux (changes requested):
Please check some little comments.

privacyidea/lib/resolvers/SQLIdResolver.py
plettich: This is the [SQLAlchemy 2.0 style](https://docs.sqlalchemy.org/en/14/glossary.html#term-2.0-style), so basically for the future :-)

privacyidea/lib/resolvers/SQLIdResolver.py
plettich: You are right, according to the new style this should be something like
`self.session.execute(delete(self.TABLE).where(filter_condition))`
Will fix this

privacyidea/lib/token.py
plettich: The problem is, that before we threw that error only if the commit to the db failed (`db_token.save()`). Now we call several functions in the `try:` block which throw several different errors (`ParameterError`, `TokenAdminError`, ...). Since we don't want to loose there errors we catch them, clean up the database and re-raise them for further processing.

privacyidea/lib/resolvers/SQLIdResolver.py
plettich: Done in 54797b58

cornelinux (approved):

PR #3546, opened 2023-03-02
Merge remaining commits from branch 3.8

plettich:


cornelinux:
@plettich Could it be that the two tests are failing due to the changes in SQLAlechemy? Should we merge it and then fix it in the master branch?

plettich:
> @plettich Could it be that the two tests are failing due to the changes in SQLAlechemy? Should we merge it and then fix it in the master branch?
Yes, we have some weird (and unpredictable) errors due to the SQLAlchemy update. 
Hopefully this goes away when #3547 is done.

cornelinux (approved):

PR #3550, opened 2023-03-03
3493/adjust_policydecorators

### LINKED ISSUE #3493: multiselect policies in webUI
Since we now have a multiselect in policies, we could adapt some policies like

* challenge_response
* ...

That already take a list of e.g. tokentypes.
Since the multiselect and the existing policy both save this as a whitespace separated list, this should simply be an adaption of the policy definition here:
https://github.com/privacyidea/privacyidea/blob/d166281ce3fc226242952be9384814d78fa0d3c2/privacyidea/lib/policy.py#L2237

To generate the list of tokenclasses for the multiselect we could use a check of the class attribute "mode" to see if this can be a challene response token.

NNLemling:
closes #3493

### DIFF
--- a/privacyidea/lib/policydecorators.py
+++ b/privacyidea/lib/policydecorators.py
@@ -121,13 +121,13 @@ def challenge_response_allowed(func):
                                                  action=ACTION.CHALLENGERESPONSE, user_object=user_object)\
                 .action_values(unique=False, write_to_audit_log=False)
             log.debug("Found these allowed tokentypes: {0!s}".format(list(allowed_tokentypes_dict)))
-            # allowed_tokentypes_dict.keys() is a list of actions from several policies.
+            allowed_tokentypes_dict = {k.lower(): v for k, v in allowed_tokentypes_dict.items()}
+            token = token.get_tokentype().lower()
             chal_resp_found = False
-            for toks in allowed_tokentypes_dict:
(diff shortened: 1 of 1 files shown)


privacyidea/lib/policydecorators.py
```diff
@@ -121,13 +121,13 @@ def challenge_response_wrapper(*args, **kwds):
...
+            allowed_tokentypes_dict = {k.lower(): v for k, v in allowed_tokentypes_dict.items()}
+            token = token.get_tokentype().lower()
             chal_resp_found = False
-            for toks in allowed_tokentypes_dict:
-                if token.get_tokentype() in allowed_tokentypes_dict:
-                    # This token is allowed to to chal resp
-                    chal_resp_found = True
-                    g.audit_object.add_policy(allowed_tokentypes_dict.get(toks))
+            if token in allowed_tokentypes_dict:
+                # This token is allowed to to chal resp
```
plettich, line 128: ... to **do** ...
And i would write out `challenge-response`

plettich (changes requested):

privacyidea/lib/policydecorators.py
NNLemling: Fixed in 9776fd5193d12754747fb250ae94f7a99bb7d09a.

plettich (approved):

Issue #3261, opened 2022-08-09
Server response fit to multiple challenges


lukasmatusiewicz:
By the server response with a prompt for a new PIN, token type is still `webauthn`. That forces us to adjust the code to avoid triggering webauthn tokens if there is no sign request.

Maybe this can be a good idea:
if the server has an unordinary challenge for a specific token, type changes to: e.g. `newPIN`.

That seems more useful because only one thing that is telling us what the server expects is the message, which can differ by the language.

lukasmatusiewicz:
![grafik](https://user-images.githubusercontent.com/77617779/183588100-db2712eb-80e7-44e4-b5f3-7d9b7977a7d6.png)

cornelinux:
OK, in such a case we could remove the tokentype, or move the "type" so s.th. like "old_type", and -- as you said -- fill the "type" with s.th. other (None).

jona-samuel:
Another way could be, that we allow including the challenge_type to the challenge.

plettich:
After looking at it closer we just need to set the client mode to "interactive" in the second step.
This must be done here:
https://github.com/privacyidea/privacyidea/blob/28f138522618915950769a0fe196720f3b417393/privacyidea/lib/challengeresponsedecorators.py#L62
since all other challenges are dismissed in case of a challenge-response decorator.
The Plug-In must not use the token type to "guess" the current flow.

PR #3553, opened 2023-03-06
3261/adding challenge_type to multi_challenge

jona-samuel:
This adds a new entry to the reply_dict of the multi_challenge with information about the challenge type.
More info in #3261

### DIFF
--- a/privacyidea/lib/challengeresponsedecorators.py
+++ b/privacyidea/lib/challengeresponsedecorators.py
@@ -59,9 +59,14 @@ def _create_challenge(token_obj, challenge_type, message, challenge_data=None):
     db_challenge.save()
     token_obj.challenge_janitor()
     reply_dict = {}
+    if ((challenge_type == CHALLENGE_TYPE.PIN_RESET) or (challenge_type == CHALLENGE_TYPE.RESYNC)):
+        challenge = challenge_type
+    else:
+        challenge = None
     reply_dict["multi_challenge"] = [{"transaction_id": db_challenge.transaction_id,
                                       "message": message,
                                       "attributes": None,
+                                      "challenge_type": challenge,
                                       "serial": token_obj.token.serial,
                                       "type": token_obj.token.tokentype}]
     reply_dict["message"] = message
--- a/privacyidea/lib/challengeresponsedecorators.py
+++ b/privacyidea/lib/challengeresponsedecorators.py
@@ -62,7 +62,7 @@ def _create_challenge(token_obj, challenge_type, message, challenge_data=None):
     if ((challenge_type == CHALLENGE_TYPE.PIN_RESET) or (challenge_type == CHALLENGE_TYPE.RESYNC)):
         challenge = challenge_type
     else:
-        challenge = None
+        challenge = "standard"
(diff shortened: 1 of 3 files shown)


lukasmatusiewicz (approved):

privacyidea/lib/challengeresponsedecorators.py
```diff
@@ -59,9 +59,14 @@ def _create_challenge(token_obj, challenge_type, message, challenge_data=None):
     db_challenge.save()
     token_obj.challenge_janitor()
     reply_dict = {}
+    if ((challenge_type == CHALLENGE_TYPE.PIN_RESET) or (challenge_type == CHALLENGE_TYPE.RESYNC)):
+        challenge = challenge_type
+    else:
+        challenge = "standard"
```
plettich, line 65: This should be added to the [`CHALLENGE_TYPE` class](https://github.com/privacyidea/privacyidea/blob/28f138522618915950769a0fe196720f3b417393/privacyidea/lib/challengeresponsedecorators.py#L46), maybe as `generic_challenge` or `standard_challenge` (@cornelinux should know, why the values are called `generic_*`).

tests/test_lib_token.py
```diff
@@ -1934,6 +1934,7 @@ def test_01_successfully_change_pin(self):
                                                   "g": g})
         self.assertFalse(r)
         self.assertEqual("Please enter a new PIN", reply_dict.get("message"))
+        self.assertEqual("generic_pin_reset", reply_dict.get('multi_challenge')[0].get('challenge_type'))
```
plettich, line 1937: We should also test somewhere, that the `standard` challenge type is returned if there is no PIN-reset or resync-challenge.

plettich (changes requested):

privacyidea/lib/challengeresponsedecorators.py
jona-samuel: @lukasmatusiewicz is 'generic_challenge' instead of 'standard' fine for you?

jona-samuel: 7d835b180677b94d0de6d320c9d8841caff697a1

tests/test_lib_token.py
jona-samuel: 7d835b180677b94d0de6d320c9d8841caff697a1

privacyidea/lib/token.py
```diff
@@ -2092,6 +2092,7 @@ def create_challenges_from_tokens(token_list, reply_dict, options=None):
             message_list.append(message)
             if r_chal:
                 challenge_info = challenge_info or {}
+                challenge_info["challenge_type"] = "generic_challenge"
```
plettich, line 2095: please add the `generic_challenge` to the `CHALLENGE_TYPE` class here:
https://github.com/privacyidea/privacyidea/blob/28f138522618915950769a0fe196720f3b417393/privacyidea/lib/challengeresponsedecorators.py#L46

plettich (changes requested):

jona-samuel:
this is replaced with #3569

Issue #2128, opened 2020-03-23
Remove Support for Python 2.7


plettich:
Keeping the support for Python 2.7 makes the code more and more unmaintainable.
We need to jump through hoops to keep the requirements working, etc.
We should check if and how the main distributions support Python 3 and if we can fade out support for Python 2.

Steps to take:

* [x] remove note in setup.py and requirements.txt
* [x] cleanup code to remove the special code for 2.7. (see `six.PY2`) (@NNLemling ) 
* [x] remove all `future` imports (@plettich )
* [x] rename Regex assertions: https://docs.python.org/3.8/library/unittest.html#deprecated-aliases  (@plettich )
* [x] unicode/("u")  (@jona-samuel )
* [x] six.stringtype (@NNLemling )
* [x] remove all `six` imports (@plettich )
* [x] check if we can remove the different `unicode` functions:
  - `hexlify_and_unicode`
  - `b64encode_and_unicode`
  - `to_unicode`
  - `__unicode__`
* [x] This can also be removed:
  https://github.com/privacyidea/privacyidea/blob/3c9a0e25a304d586d71f0192b0abae2f3ae219df/privacyidea/lib/tokens/totptoken.py#L301
* [x] check for `ImportError` occurrences if they can be removed

cornelinux:
We should even drop official support for python <=3.5.
We should check, if we can already drop this in 3.8.

"We plan to remove the support". Add this to the release note for 3.7.1.

cornelinux:
Will  be removed in 3.9 as communicated to users.

plettich:
I think this is done.

PR #3554, opened 2023-03-06
Remove remaining six imports

plettich:
Working on #2128

### DIFF
--- a/privacyidea/lib/auditmodules/sqlaudit.py
+++ b/privacyidea/lib/auditmodules/sqlaudit.py
@@ -53,7 +53,6 @@ from sqlalchemy.sql import expression
 from sqlalchemy.ext.compiler import compiles
 import datetime
 import traceback
-from six import string_types
 from privacyidea.models import audit_column_length as column_length
 from privacyidea.models import Audit as LogEntry
 from sqlalchemy import create_engine
@@ -196,7 +195,7 @@ class Audit(AuditBase):
         for column, l in self.custom_column_length.items():
             if column in self.audit_data:
                 data = self.audit_data[column]
-                if isinstance(data, string_types):
+                if isinstance(data, str):
                     if column == "policies":
                         # The policies column is shortened per comma entry
                         data = truncate_comma_list(data, l)
--- a/privacyidea/lib/caconnectors/localca.py
+++ b/privacyidea/lib/caconnectors/localca.py
@@ -38,8 +38,6 @@ import shlex
 import re
 import logging
 import os
-from six import string_types
-from six.moves import input
 import traceback
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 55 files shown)


privacyidea/lib/auditmodules/sqlaudit.py
```diff
@@ -196,7 +195,7 @@ def _truncate_data(self):
         for column, l in self.custom_column_length.items():
             if column in self.audit_data:
                 data = self.audit_data[column]
-                if isinstance(data, string_types):
+                if isinstance(data, str):
```
cornelinux, line 198: Is this the same? Would "string_types" also match byte strings?

privacyidea/lib/caconnectors/localca.py
```diff
@@ -475,7 +473,7 @@ def revoke_cert(self, certificate, request_id=None, reason=CRL_REASONS[0]):
         :return: Returns the serial number of the revoked certificate. Otherwise
             an error is raised.
         """
-        if isinstance(certificate, string_types):
+        if isinstance(certificate, str):
```
cornelinux, line 476: See my missing knowledge above.

privacyidea/lib/config.py
```diff
@@ -270,7 +269,7 @@ def get_config(self, key=None, default=None, role="admin",
                 pass
             if isinstance(r_config, int):
                 r_config = r_config > 0
-            if isinstance(r_config, string_types):
+            if isinstance(r_config, str):
```
cornelinux, line 272: See above

privacyidea/lib/tokens/u2f.py
```diff
@@ -110,12 +110,12 @@ def parse_registration_data(reg_data, verify_cert=True):
     :return: tuple
     """
     reg_data_bin = url_decode(reg_data)
-    reserved_byte = six.int2byte(six.indexbytes(reg_data_bin, 0))  # must be '\x05'
-    if reserved_byte != b'\x05':
-        raise Exception("The registration data is in a wrong format. It must"
+    reserved_byte = reg_data_bin[0]  # must be '\x05'
```
cornelinux, line 113: since we do not return the '\x05' as a byte but as an int '5', should we rename the variable?
Or should we, for conformity to line 87 use "bytes()", even if it is not necessary?
I only realized this, since you explained this 5 minutes ago.

cornelinux (commented):
Please check my comments.
If necessary adapt or otherwise this can be merged.

privacyidea/lib/auditmodules/sqlaudit.py
cornelinux: Is unicode, so this is fine.

privacyidea/lib/tokens/u2f.py
plettich: Done in [436cb0a](https://github.com/privacyidea/privacyidea/pull/3554/commits/436cb0a226ee69e7467460fd04713088e51649fc)

cornelinux (approved):

PR #3572, opened 2023-03-20
3438/catching error missing password in auth request

jona-samuel:


cornelinux (approved):

cornelinux (changes requested):
@jona-samuel  Please check for the failing tests in test_api_auth!

plettich (changes requested):
Maybe we should just check if the password variable is set:
```
if password:
    ....
```
that way we don't have to change all the tests.
Also the error message "Authentication failure. Wrong credentials" might be more appropriate to avoid disclosing information.
We could add a debug/info log entry to inform in the logs that the parameter is missing.

jona-samuel:
replaced buy https://github.com/privacyidea/privacyidea/pull/3586

PR #3586, opened 2023-04-04
3438/Error handling for auth request with no password

jona-samuel:


privacyidea/api/auth.py
```diff
@@ -301,6 +301,10 @@ def get_auth_token():
     else:
         # The user could not be identified against the admin database,
         # so we do the rest of the check
+        if password is None:
+            raise AuthError(_("Authentication failure. Wrong credentials"),
+                            id=ERROR.AUTHENTICATE_WRONG_CREDENTIALS,
+                            details=details or {})
```
plettich, line 307: We already throw this error below if a user could not be found. We can just do something like:
```
if password is None:
  details['message'] = 'Missing parameter "password"'
else:
  ...
```
This way we get the information in the Audit-Log about why the login failed.
The details will also be included in the error response. I don't think that the information about the missing password parameter is a security issue.

plettich (changes requested):

privacyidea/api/auth.py
jona-samuel: da4e3786acb7be0c6e9e9a04e382ab51d1863ae8

Issue #3436, opened 2022-12-30
Challenge Response handling at /auth endpoint not consistent


lukasmatusiewicz:
Hi,
we are trying to implement 2fa in smartcard-client which is using `POST /auth` endpoint to authenticate in privacyIDEA.

Scenario: The user is sending just a PIN without OTP and awaits, that privacyIDEA will ask for the OTP in the next step.
Response: The server triggers the challenges and asks for OTP in the message, but at the same time it encounters an error: 4031.

The PI response looks as follows:

~~~~json

 "detail": { 
       "client_mode": "interactive", 
       "loginmode": "privacyIDEA", 
        "message": "please enter otp: ", 
        "messages": [ "please enter otp: " ], 
        "multi_challenge": [ 
                      { "client_mode": "interactive", "message": "please enter otp: ", "serial": "OATH00159D24", "transaction_id": "13606...996141", "type": "hotp" } ], 
           "serial": "OATH0...9D24", 
             "threadid": 13999...8608, 
             "transaction_id": "136062...90996141", 
             "transaction_ids": [ "136062...0996141" ], 
              "type": "hotp" }, 
"id": 1, 
"jsonrpc": "2.0", 
"result": { 
       "error": { "code": 4031, "message": "Authentication failure. Wrong credentials" }, 
       "status": false },`
~~~~

https://github.com/privacyidea/smartcard-client-windows/issues/6

plettich:
Actually, challenge-response works with the `/auth` endpoint quite well. Just ignore the error and send the OTP value with a second `/auth` request and the the `transaction_id`.
The Problem is, that we throw an error if we do not have a proper authentication at the end of the `/auth` request.
There is already an old issue open: #1192

plettich:
The UI needs to be updated for the changed `/auth` response:
Most of the code in the whole error handling function 
https://github.com/privacyidea/privacyidea/blob/13565489141f81b61e8b9803df9358741d9269f0/privacyidea/static/components/login/controllers/loginControllers.js#L241
needs to go into the regular response handling here:
https://github.com/privacyidea/privacyidea/blob/13565489141f81b61e8b9803df9358741d9269f0/privacyidea/static/components/login/controllers/loginControllers.js#L236
Only the *show error message* code should remain in the error handler:
https://github.com/privacyidea/privacyidea/blob/13565489141f81b61e8b9803df9358741d9269f0/privacyidea/static/components/login/controllers/loginControllers.js#L356

[...]

PR #3600, opened 2023-04-26
3436/challenge response handling at /auth endpoint not consistent

Sebastian-Lemling:
- [x] fix challenge response handling at /auth endpoint on API level
- [x] adjust existing tests

working on #3436

### DIFF
--- a/privacyidea/api/auth.py
+++ b/privacyidea/api/auth.py
@@ -339,11 +339,11 @@ def get_auth_token():
                                     "serial": serials,
                                     "info": "{0!s}|loginmode={1!s}".format(log_used_user(user_obj),
                                                                            details.get("loginmode"))})
-            if "multi_challenge" in details:
+            if user_auth and "multi_challenge" in details:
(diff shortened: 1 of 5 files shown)


privacyidea/api/auth.py
```diff
@@ -339,6 +339,11 @@ def get_auth_token():
                                     "serial": serials,
                                     "info": "{0!s}|loginmode={1!s}".format(log_used_user(user_obj),
                                                                            details.get("loginmode"))})
+            if "multi_challenge" in details:
```
plettich, line 342: We also need to check if `user_auth` is `True`

plettich (changes requested):

privacyidea/api/auth.py
NNLemling: fixed in 2a389e89b39bb6c60c8372b339772a398ca7898a

plettich (changes requested):
See my comment from the previous review

PR #3603, opened 2023-04-27
3589/adjust_get_privacyidea_note_for_privacyidea_server_in_audit_log

### LINKED ISSUE #3589: privacyidea server in the audit log is wrong on privacyidea HA scenario
### Steps to reproduce

1. Install two privacyidea server that connect to the same external mysql server (cluster or not)
2. define the PI_NODE and PI_NODES variable
3. don't define the PI_AUDIT_SERVERNAME

### Expected outcome

From documentation i would see the PI_NODE name under the column privacyidea_server on the audit log, instead i see the "common name"of the webserver used to Load Balancing the two privacyidea nodes
 
### Actual outcome

i see the "common name"of the webserver used to Load Balancing the two privacyidea nodes

### Configuration

* **privacyIDEA version**: 3.8.1
* **Installation method**: from Ubuntu packages
* **Python version**:  3.10
* **Operating system**: Ubuntu 22.04
* **Webserver**: Apache 2.4
* **Token database**: MySQL galera cluster

### Log file

let me know if you need the log file, but nothing is there 

### Fix

I fixed the problem defining PI_AUDIT_SERVERNAME to the PI_NODE value, now the privacyidea server in the audit log is correct.

NNLemling:
closes #3589

### DIFF
--- a/privacyidea/api/before_after.py
+++ b/privacyidea/api/before_after.py
@@ -38,7 +38,7 @@ from privacyidea.lib.policy import PolicyClass
 from privacyidea.lib.event import EventConfiguration
 from privacyidea.lib.lifecycle import call_finalizers
 from privacyidea.api.auth import (user_required, admin_required, jwtauth)
-from privacyidea.lib.config import get_from_config, SYSCONF, ensure_no_config_object
+from privacyidea.lib.config import get_from_config, SYSCONF, ensure_no_config_object, get_privacyidea_node
 from privacyidea.lib.token import get_token_type, get_token_owner
 from privacyidea.api.ttype import ttype_blueprint
 from privacyidea.api.validate import validate_blueprint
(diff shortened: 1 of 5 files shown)


privacyidea/api/before_after.py
```diff
@@ -186,8 +186,7 @@ def before_request():
                                 get_from_config(SYSCONF.OVERRIDECLIENT))
     # Save the HTTP header in the localproxy object
     g.request_headers = request.headers
-    privacyidea_server = current_app.config.get("PI_AUDIT_SERVERNAME") or \
-                         request.host
+    privacyidea_server = get_privacyidea_node(request.host)
```
plettich, line 189: I would still use the `current_app.config.get("PI_AUDIT_SERVERNAME")` before using `get_privacyidea_node(request.host)`. If someone has configured `PI_AUDIT_SERVERNAME` as well as `PI_NODE`, the `PI_NODE` would have precedence which could be unexpected.

plettich (changes requested):

plettich (approved):

PR #3607, opened 2023-05-03
Add filter for options

### LINKED ISSUE #3573: Filter SSH keys for users and service IDs
## Management Requirement

We need a way to find all SSH keys assigned to a certain service_id. Maybe list service_ids. Inspect service_ids and list all SSH keys assigned to this service_id.

This could be situated in the UI under the tab "machine"? It could be a table of
`| key/token | applicationtype | service_id |  options |` . 

We could then also filter for options like the user to find all ssh keys, that are allowed to login as "root" (user-option) on "mailservers" (service id).

See #3533 

## API

Starting from the API view the machine-token-assignments are fetched via `GET https:///machine/token?serial=SSHK0000EB3E`.
We can start with enhancing the API endpoint machine/token to also allow filtering for other parameters and options, here:

https://github.com/privacyidea/privacyidea/blob/258a88b8a5ae07e233d40fc48fd644a4d50a7c70/privacyidea/api/machine.py#L232-L236

The unterlying library call `list_machine_tokens` already has a parameter *filter_params*, that is used to filter for service_id.

This way we could have the REST API return e.g.
* all machine token information for certain application types (like SSH)
* for all certain service_ids...

* [x] Enhance the REST API to accept filter_params
   * [x] Prepare test class for GET /machein/token 
* [x] Ensure that the response of GET /machine/token also returns the serial number, in case we are searching for application_type and service_id, we also need to know the attached serial/token.
* [x] Add tests for finding user-options
* [x] Add WebUI location to display the service ids/SSH assignments. This would use the API endpoint `GET https:///machine/token` with the corresponding filter options.
* [x] Have an extra UI for SSH with the corresponding options.
* [x] Have an extra UI for offline with the corresponding options. We can do this similar to the token-enrollment. Have a dropdown with the applications. Depending on the application we load different ssh-template or offline-template just like with token.enroll.hotp.html or token.enroll.totp.html.
* [x] Filtering in UI (Done in de648fbb)
* [x] Sorting in UI (Done in 44ca9f12)
* [x] Filtering options in the API
* [x] Filtering serial numbers in the API (done in 86d61804)
* [x] Sorting in the API. Add to API: `sortby`, `sortdir`, ~~page`, `pagesize~~  (done in 01e4d9a8)
* [x] API tests
* [x] Do we need a policy? Not an extra policy. We can use this one https://github.com/privacyidea/privacyidea/blob/ac9135f1c48449a8121afa30ea67b41dea215dea/privacyidea/lib/policy.py#L268 or `tokenlist`.  (Done in 9682f25c)
* [x] Some docs. Where can we add this?  (Done in 816af0fc)
   - https://privacyidea.readthedocs.io/en/latest/machines/index.html?highlight=application#ssh
   - https://github.com/privacyidea/privacyidea/blob/ed4d272cb8419eeb1095521ac7724252dfd79613/doc/configuration/serviceids.rst#L3

## UI

Thoughts on the UI. How important are service_ids? 
For SSH keys these are very important. But not for other applications.

Where and how should the machinetokens-applications be listed?

These would also work of offline - e.g. check which machines, with which attributes/options are available offline.
Which options should be visible in a table?

For SSH applications this would probably be

    | token serial | application  (SSH) | tokenowner | ssh-user | service_id |

But for offline this might be

    | token serial | application (offline) | tokenowner | count | .... | 

It could also be 

    | token serial | application (offline) | tokenowner | <options> |

But we need to be able filter for the "options" like the `service_id`.

@plettich: This might also connected to a similar design/display question in regards to #1291 
A similar problem would be with tokengroups. 
* How can we display all tokens, that are in one tokengroup?
* How can we list all tokens, that are in one nested token?
* How can we list tokens, that are assigned to a certain service_id/ssh?

If a user has lost a smartcard/nestedToken, the admin would probably go to the user listing.
If an administrator wants to know, which SSH keys are allowed to login to service_id "mailserver", the admin would probably go to the token view.
If the admin wants to know, which tokens are in a tokengroup, the admin would either go to tokengroup view or to token view.

Idea: There could be a submenu in the tokenview, that is: *Token*, *Nested Token/Container*, *Tokengroups*, *ServiceID*.

cornelinux:
The GET /machine/token call now allows to filter for all options like "service_id" or "user".

Closes #3573

I added the WebUI to view SSH applications and offline.
It displays the token serial and the application options.
Can be sorted and filterd. Currently I do no pagination, since we do not do this on the database level. 

@NNLemling please take a look at the code, which is finish. I am wondering, where I should still add some documentation.

### DIFF
--- a/privacyidea/static/components/token/views/token.applications.html
+++ b/privacyidea/static/components/token/views/token.applications.html
@@ -1,0 +1,16 @@
+<form name="formTokenApplications" role="form" validate>
+
+Application:          <select class="form-control"
+                    id="application"
+                    ng-change="changeApplication()"
+                    ng-model="currentApplication"
+                    ng-options="application as desc for (application, desc) in formInit.applications">
+            </select>
+
+        <!-- This is the token data section. It differs for every token type-->
+        <ng-include src="instanceUrl +
+        '/static/components/token/views/token.applications.'
+         + currentApplication + '.html' + fileVersionSuffix">
+        </ng-include>
+
+</form>
--- a/privacyidea/static/components/token/views/token.applications.html
+++ b/privacyidea/static/components/token/views/token.applications.html
@@ -12,5 +12,4 @@ Application:          <select class="form-control"
         '/static/components/token/views/token.applications.'
          + currentApplication + '.html' + fileVersionSuffix">
         </ng-include>
-
 </form>
--- a/tests/test_api_machines_serviceid.py
+++ b/tests/test_api_machines_serviceid.py
@@ -233,3 +233,62 @@ class APIMachinesServiceIDTestCase(MyApiTestCase):
             self.assertEqual(value[0]["application"], "ssh")
             self.assertEqual(value[0].get("options").get("service_id"), self.serviceID2)
             self.assertEqual(value[0].get("serial"), self.serial1)
+
+        # test filter and wildcards
+
+        # Find all tokens attached to a *server
+        with self.app.test_request_context(
+                '/machine/token?service_id=*server*&application=ssh',
+                method='GET',
+                headers={'Authorization': self.at}):
+            res = self.app.full_dispatch_request()
+            self.assertTrue(res.status_code == 200, res)
+            result = res.json.get("result")
+            self.assertEqual(result["status"], True)
+            value = result.get("value")
+            self.assertEqual(len(value), 4)
+            self.assertEqual(value[0]["application"], "ssh")
+
+        # Find all tokens attached to the user *root*
+        with self.app.test_request_context(
+                '/machine/token?application=ssh&user=*oot',
+                method='GET',
+                headers={'Authorization': self.at}):
+            res = self.app.full_dispatch_request()
+            self.assertTrue(res.status_code == 200, res)
+            result = res.json.get("result")
+            self.assertEqual(result["status"], True)
+            value = result.get("value")
+            # 3 tokens are attached via user "root"
+            self.assertEqual(len(value), 3)
+            self.assertEqual(value[0]["application"], "ssh")
+
+        # Find all tokens attached to the user *adm*
+        with self.app.test_request_context(
+                '/machine/token?application=ssh&user=*adm*',
+                method='GET',
+                headers={'Authorization': self.at}):
+            res = self.app.full_dispatch_request()
+            self.assertTrue(res.status_code == 200, res)
+            result = res.json.get("result")
+            self.assertEqual(result["status"], True)
+            value = result.get("value")
+            # One token is attached via user "admin"
+            self.assertEqual(len(value), 1)
+            self.assertEqual(value[0]["application"], "ssh")
+
+        # Filter vor *KEY1
+        with self.app.test_request_context(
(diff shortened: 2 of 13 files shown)


tests/test_api_machines_serviceid.py
```diff
@@ -0,0 +1,310 @@
...
+        # create two tokens
+        token_obj = init_token({"serial": self.serial1, "type": "sshkey",
+                                "sshkey": SSHKEY})
+        self.assertEqual(token_obj.type, "sshkey")
+        token_obj = init_token({"serial": self.serial2, "type": "sshkey",
+                                "sshkey": SSHKEY_ecdsa})
+        self.assertEqual(token_obj.type, "sshkey")
+
+    def test_02_attach_tokens(self):
+        # Attach token S1 to webserver and mailserver
```
NNLemling, line 53: I would only write here **_"# Attach token S1 to webserver"_** since each step is described below.

tests/test_api_machines_serviceid.py
```diff
@@ -0,0 +1,310 @@
...
+            self.assertEqual(len(value), 2)
+            self.assertEqual(value[0]["application"], "ssh")
+            self.assertEqual(value[0].get("options").get("service_id"), self.serviceID2)
+            self.assertEqual(value[0].get("serial"), self.serial1)
+            self.assertEqual(value[1]["application"], "ssh")
+            self.assertEqual(value[1].get("options").get("service_id"), self.serviceID2)
+            self.assertEqual(value[1].get("serial"), self.serial2)
+
+        with self.app.test_request_context(
+                '/machine/token?service_id={0!s}&application=ssh&user=admin'.format(self.serviceID2),
```
NNLemling, line 224: "# Get token for service_id self.serviceID2 and the application=ssh and the user=**admin**"

privacyidea/static/components/token/views/token.applications.html
```diff
@@ -0,0 +1,16 @@
...
+                    ng-options="application as desc for (application, desc) in formInit.applications">
+            </select>
+
+        <!-- This is the token data section. It differs for every token type-->
+        <ng-include src="instanceUrl +
+        '/static/components/token/views/token.applications.'
+         + currentApplication + '.html' + fileVersionSuffix">
+        </ng-include>
+
+</form>
```
NNLemling, line 16: missing newline

NNLemling (changes requested):

tests/test_api_machines_serviceid.py
cornelinux: My point was, to give an overview of the token attachements at the beginning of the test, so that it is clear what can be checked later.
I am doing a bit reformatting

cornelinux: Done in f4469ac

tests/test_api_machines_serviceid.py
cornelinux: Done in f4469ac

privacyidea/static/components/token/views/token.applications.html
cornelinux: Done in f4469ac

NNLemling (approved):

PR #3611, opened 2023-05-11
tests/testdata: recreate passwd hashes with `mkpasswd -m sha-512`

### LINKED ISSUE #3613: Tests affecting passwd resolver fail when using libxcrypt with modern hashes only enabled
For instance, `APIAuditTestCase.test_03_get_allowed_audit_realm`:

```
tests/test_api_audit.py:239: AssertionError
------------------------------ Captured log call -------------------------------
WARNING  privacyidea.lib.config:config.py:336 Request-local store already contains config object, even though it should not
INFO     privacyidea.lib.usercache:usercache.py:106 Deleted 0 entries from the user cache (resolver='resolver1', username=None, expired=None)
INFO     privacyidea.lib.resolvers.PasswdIdResolver:PasswdIdResolver.py:131 loading users from file tests/testdata/passwords from within '/build/source'
INFO     privacyidea.lib.user:user.py:262 user 'selfservice' found in resolver 'resolver1'
INFO     privacyidea.lib.user:user.py:264 userid resolved to '1004' 
INFO     privacyidea.lib.user:user.py:432 User 'selfservice' from realm 'adminrealm' tries to authenticate
INFO     privacyidea.lib.resolvers.PasswdIdResolver:PasswdIdResolver.py:197 checking password for user uid 1004
ERROR    privacyidea.lib.user:user.py:451 Error checking password within module OSError(22, 'Invalid argument')
_________________ APIAuditTestCase.test_04_get_user_audit_log __________________
```

This happens because the hashes inside the fixture data don't use a hash algorithm that's enabled when using `libxcrypt` (e.g. on the upcoming NixOS 23.05) that has only "modern" hash algorithms enabled.

Ma27:
When running the tests on a distribution using `libxcrypt`[1] with support for "modern" hashes only, like the upcoming NixOS 23.05 will do, then the tests for the passwd-resolver fail with
`OSError(22, 'Invalid argument')`. Recreating all test hashes with `mkpasswd -m sha-512` fixes the issue for me.

Closes #3613.

[1] https://github.com/besser82/libxcrypt

### DIFF
--- a/tests/testdata/passwd-duplicate-name
+++ b/tests/testdata/passwd-duplicate-name
@@ -1,2 +1,2 @@
-testadmin:H9Y6eYntkrrKo:1004:1004:::
-someuser:H9Y6eYntkrrKo:1005:1005:::
+testadmin:$6$.jIGOPJO3wzw6Aj1$aNUQyxvaNR5lAvnCwUMoEfi4Qqozc01MzApGE/X3tteuEe1M3nY4Ax3fWKDOGncTo6Lw523XagISrIEV8..Ua0:1004:1004:::
+someuser:$6$k2iAT/7meR1ymsIb$VlrqjfzdEZFb/AaTJEId4.VrXGpZ80TvlzmEAq0bAXmuXWCLKxZWtoKgkSWey3TsywZw5sXmL3PArq7RWZX550:1005:1005:::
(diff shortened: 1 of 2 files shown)


cornelinux:
Thanks for noting. Could you please open an issue, since we always link PRs to issues.

plettich (approved):

PR #3658, opened 2023-06-15
3008/first version of move/copy tokens via dicts between processes

Sebastian-Lemling:
**2DO**

- [ ] handle secret key
- [ ] final comparison test whether token was copied as desired

what if token serial already exist?


cornelinux:
For better readability I am closing this.
It is handled in https://github.com/privacyidea/privacyidea/pull/3685

PR #3666, opened 2023-06-22
3664/Using _time2counter everywhere in totp

jona-samuel:
code-cleanup: using function _time2conter in all places


cornelinux (approved):
lloks good to me

PR #3684, opened 2023-07-12
Check for password length during split_pin_pass

### LINKED ISSUE #3467: Check result of splitting PIN and OTP
In some cases (multiple tokens for a user, one is a Yubikey in AES-mode) it happens that the split of PIN and OTP returns empty values which in turn create errors when binding to the LDAP-server with an empty password.
We could check whether the split was successful and return some kind of status value here:
https://github.com/privacyidea/privacyidea/blob/3f6ad403462d8147fbfe030210907fa911ad9438/privacyidea/lib/tokenclass.py#L1383

cornelinux:
If the given password is shorter than the expected otplen, we early exist with a failed PIN.

Closes #3467

### DIFF
--- a/privacyidea/lib/tokenclass.py
+++ b/privacyidea/lib/tokenclass.py
@@ -511,7 +511,8 @@ class TokenClass(object):
 
         (res, pin, otpval) = self.split_pin_pass(passw, user=user,
                                                  options=options)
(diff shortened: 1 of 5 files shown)


plettich (approved):
Looks good

PR #3685, opened 2023-07-12
Thinking about exporting tokens to reencrypt

### LINKED ISSUE #3005: Encryption Key Rollover - reencrypt data
pi-manage or a different tool should allow to reencrypt the data. It might be difficult to find all data, that needs to be reencrypted:

* token seeds in the `token` table are obvious
* Passwords in resolvers
* encrypted PINs, if they exist
* But there can be also options in `tokeninfo` marked as "password", that would need to be reencrypted.
* ...

There are different aspects, why one would want to reencrypt:

* The existing encryption algorithm has become week or the keylength need to be increased
* We want to change the encryption scheme like switching from file based encryption ([DefaultSecurityModule](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/security/default.py#L231)) to an HSM based ([AESHardwareSecurityModule](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/security/aeshsm.py#L52)) encryption or vice versa.
* We want to change the HSM like moving from Thales to YubiHSM or vice versa.
* ...

This could be done
* during runtime - then it would be challenging to know which data is encrypted by which scheme.
* with the system being offline - then we could reencrypt everything one after another.

For the configuration of the re-encryption it would be most flexible to provide two `pi.cfg` files.
This way we have the full flexibility to define the from-scheme in one pi.cfg and the to-scheme in the second pi.cfg.

We could also use #3008 

# Todo

* [x] create a list of all encrypted data, that needs to be reencrypted.

Interesting: We have 

* **encrypted** data of tokens/tokeninfo and
* encrypted data in configuration
*  and **hashed** data and also 
* **signed** data. 

We can not "roll over" hashed data. So at least for hashed data the way would be: key old hashes (like passwords) but hash new passwords in a new way.

### Encrypted data

* [x] The secret key of a token is encrypted. https://github.com/privacyidea/privacyidea/blob/a036b713d775fece1489ad7e219b0bcd1e49115a/privacyidea/models.py#L187
* [x] Tokeninfo can be encrypted if the `Type` is set to "password". https://github.com/privacyidea/privacyidea/blob/a036b713d775fece1489ad7e219b0bcd1e49115a/privacyidea/models.py#L716

#### Encrypted Data in configuration

The encrypted values in the configuration could actually be migrated by exporting and importing the configuration like:

     pi-manage config exporter -t resolver -f yaml 

We might consider, to allow the export of the passwords.

* [ ] ResolverConfig can be encrypted, if the `Type` is set to "password". https://github.com/privacyidea/privacyidea/blob/a036b713d775fece1489ad7e219b0bcd1e49115a/privacyidea/models.py#L1098
* [ ] MachineResolverConfig could be encrypted if the `Type`is set to "password", although this config is not used anymore.
* [ ] CAConnectorConfig if the `Type` is set to "password".
* [ ] SMTPServer: Field `password`.
* [ ] RADIUSEServer: Field `secret`.
* [ ] SMSGatewayOption can be encrpyted, if the `Type` is set to "password".

So basically there is config data. This could be re-encrypted by re-saving the configuration.
So actually there is only token-keys and tokeninfo-passwords that would need to be actually reencrypted.

### Hashed data

* [ ] The PIN of a token is hashed. https://github.com/privacyidea/privacyidea/blob/a036b713d775fece1489ad7e219b0bcd1e49115a/privacyidea/models.py#L173
* [ ] The password of local admins is hashed:
   - https://github.com/privacyidea/privacyidea/blob/7ed27ada2f0bf93cca53ddbfb0df04c6faf23470/privacyidea/lib/auth.py#L67
   - https://github.com/privacyidea/privacyidea/blob/d4903c388f7aab3ebd813d0350cd58b5f28472e7/privacyidea/lib/crypto.py#L218

We have to check how we can change the hashing function.

If the user authenticates and uses the PIN or the admin uses the password, privacyIDEA can check, if the password/PIN should be rehashed. If the authentication is successful, then privacyIDEA knows the given password/PIN and can save the password/PIN with the new hash function. 

This way, we can rehash on the fly.

The passlib module seems to be capable of this, accepting old hash mechanisms and using new ones.

### Signed data

* [ ] Audit log entries. We already have a way for old signing and new signing. But we probably need to redo this, if e.g. the keysize is changed from 2048 to 4096 or ECC. https://github.com/privacyidea/privacyidea/blob/e376c5081c34e6034b3b649cb240c1ceea3ca414/privacyidea/lib/auditmodules/sqlaudit.py#L134

## Config file 

If this is done during runtime we need to identify the way the data is encrypted.

The default security module only uses one encryption key file:

~~~
PI_ENCFILE = '/etc/privacyidea/enckey'
PI_ENCFILE.ID1 = '/etc/privacyidea/supersecurekey'
~~~

in the database preceed encrypted data with "{ID1}öaskdölalksdölkasdölksal".

See the changed signig method in the auditlog.

However, things get more complext, when we use different types of security modules.
The security module that should be used by privacyIDEA is determined via "PI_HSM_MODULE".
So for reencrypting there could be the old security module and the new security module.

~~~~
PI_HSM_MODULE = "privacyidea.lib.security.encryptkey.DefaultSecurityModule"
PI_HSM_MODULE.1 = "privacyidea.lib.security.encryptkey.EncryptKeyHardwareSecurityModule"
~~~~
And the new security module could also have a lot of new module options. We need to find a way to easily parse the config file to provide e.g. the AES-Module config for the existing data and provide the options for the EncryptKey-Module. Since also the *same* Module type could be used, we need all options to also come with an identifier.

When reading the pi.cfg we would have to build a list of security modules.

We could even have a longer list of secrutiy modules and add additional parameters in pi.cfg, to tell which identifier is the old security module (however, in thery there could be more than one old one) and which is the default (old) and which is the new security module.

Either:

~~~~
PI_DEFAULT_SECURITY_IDENTIFIER = ....
PI_NEW_SECURITY_IDENTIFIER =  ...
~~~~

Or:

~~~
PI_READ_SECURITY_IDENTIFIER = [ .... ]
PI_WRITE_SECURTIY_IDENTIFIER = ...
~~~

This way we could have a list of security modules, that can *read* existing encrypted data, but only one security module, that will write new encrypted data. Similar to the audit modules several writers and one reader.

## Database

We could have old encrypted and hashed data and new encrypted data and hashed in the database.

For encrypted tokens, we could add another column in the `token` table to tell, which version of encryption is acutally used. But this would not work for some hashed data or other encrypted data like data in the tokeninfo or the passwords of resolvers, which are stored as key-value in the resolver config.

So it might be the best to preceed the encrypted or hashed data with an identifier in the database.

## Runtime

A write access to the database would always be done with the new key or security module.

A read access to the database would check, if it is the old a the new security module.

## Parallel Script

At the same time an additional script (next to the server) would *iterate through the database* (token, resolver, tokeninfo, mailserver, radiusserver,...) and reencrypt old data.

The script should have a low priority or a kind of rate limit.

# Implementation consideration

The otpkey is only available on the database level. We are hiding the key in the tokenclass level by the SecretObject

The DB model `Token` has the methods [set_otpkey](https://github.com/privacyidea/privacyidea/blob/a036b713d775fece1489ad7e219b0bcd1e49115a/privacyidea/models.py#L300) and [update_otpkey](https://github.com/privacyidea/privacyidea/blob/a036b713d775fece1489ad7e219b0bcd1e49115a/privacyidea/models.py#L682). We could add a further method for reencryption. Having this on the DB level, we could in a first step easily create an external script, that iterates over the database entries.

Later we can use the same base functions for runtime re-encryption.

To keep track of reencrypted data, we can add a tokeninfo `reencrypted` with a timestamp.

cornelinux:
Working on #3005

Very early stage of testing a concept.

### DIFF
--- a/tools/privacyidea-token-janitor
+++ b/tools/privacyidea-token-janitor
@@ -56,8 +56,9 @@ from privacyidea.app import create_app
 from flask_script import Manager
 import re
 import sys
-from base64 import b32encode
-from binascii import unhexlify
+from yaml import safe_dump as yaml_safe_dump
+from yaml import safe_load as yaml_safe_load
+
 
 
 __version__ = "0.1"
@@ -451,9 +452,13 @@ def export_user_data(token_list, attributes=None):
                               'by the user.',
                 dest='sum_tokens', action='store_true')
 @manager.option('--action', help='Which action should be performed on the '
-                                 'found tokens. {0!s}'.format("|".join(ALLOWED_ACTIONS)))
+                                 'found tokens. {0!s}. Exporting to PSKC only supports '
+                                 'HOTP, TOTP and PW tokens!'.format("|".join(ALLOWED_ACTIONS)))
 @manager.option('--csv', dest='csv', action='store_true',
                 help='In case of a simple find, the output is written as CSV instead of the '
+                     'formatted output.')
+@manager.option('--yaml', dest='yaml', action='store_true',
+                help='In case of a simple find, the output is written as YAML instead of the '
                      'formatted output.')
 @manager.option('--last_auth', help='Can be something like 10h, 7d, or 2y')
 @manager.option('--assigned', help='True|False|None')
@@ -471,7 +476,7 @@ def find(last_auth, assigned, active, tokeninfo_key, tokeninfo_value,
          set_tokeninfo_key, set_tokeninfo_value, sum_tokens, tokenrealms, csv,
          chunksize, attributes, b32, has_not_tokeninfo_key, has_tokeninfo_key,
          tokenattribute, tokenattribute_value, tokenattribute_value_greater_than,
-         tokenattribute_value_less_than):
+         tokenattribute_value_less_than, yaml):
     """
     finds all tokens which match the conditions
     """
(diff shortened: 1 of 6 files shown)


tools/privacyidea-token-janitor
```diff
@@ -601,6 +616,28 @@ def find(last_auth, assigned, active, tokeninfo_key, tokeninfo_value,
...
+                help='Specify the YAML file with the previously exported tokens.')
+def updatetokens(yaml):
+    """
+    This can update existing tokens in the privacyIDEA system. You can specify a yaml file with the tokendata.
+    Can be used to reencrypt data, when changing the encryption key.
+    """
+    print("Loading YAML data. This may take a while.")
+    token_list = yaml_safe_load(open(yaml, 'r').read())
+    for tok in token_list:
+        del(tok["owner"])
```
plettich, line 629: Should we also check if the assigned user matches? or is this not necessary here?

plettich (changes requested):

cornelinux:
Add to the docs, that e.g. resolvers can be exported and imported.

tools/privacyidea-token-janitor
```diff
@@ -547,19 +552,29 @@ def find(last_auth, assigned, active, tokeninfo_key, tokeninfo_value,
...
+                                                                    token_dict.get("type"),
+                                                                    token_dict.get("otplen")))
+            elif yaml:
+                token_list = []
+                for tokenobj in tlist:
+                    try:
+                        token_dict = tokenobj._to_dict(b32=b32)
+                        token_dict["owner"] = "{!s}@{!s}".format(tokenobj.user.login, tokenobj.user.realm) if tokenobj.user else "n/a"
+                        token_list.append(token_dict)
+                        print(yaml_safe_dump(token_list))
```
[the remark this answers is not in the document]
cornelinux: Done in 6703023b

tools/privacyidea-token-janitor
cornelinux: No, this should probably not change.

cornelinux:
In a4960640 I added some docs.

cornelinux:
@plettich Will you check this? Otherwise I will merge it on wednesday.

plettich (approved):

PR #3686, opened 2023-07-12
Ignore PIN policies during rollover and verify

### LINKED ISSUE #2886: No PIN field for Admin Rollover
If an admin does a token-rollover for a user, no PIN field will be shown. The admin only ever sees a PIN field when enrolling tokens (admin PIN policies) or assigning users to tokens (which PIN policies apply then?).
A user PIN policy will be ignored for an admin rollover.
A rollover will fail if an admin PIN policy is set (even though the admin won't see a PIN field).

plettich:
The "rollover" and "verify" actions both call the `/token/init` endpoint which also checks if the given token PIN matches the PIN policies of the logged in user.
However during "rollover" or "verify" there is no PIN setting necessary so we exclude these `/token/init` variants from the PIN checking.

This PR also removes the PIN entry field in case a user initiates a token rollover.

Closes #2886

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -320,10 +320,14 @@ def check_otp_pin(request=None, action=None):
     raised.
     """
     params = request.all_data
-    realm = params.get("realm")
     pin = params.get("otppin", "") or params.get("pin", "")
     serial = params.get("serial")
     tokentype = params.get("type")
+    rollover = params.get("rollover")
+    verify = params.get("verify")
+    if rollover or verify:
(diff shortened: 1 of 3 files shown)


cornelinux (approved):

cornelinux:
@plettich  Hm, it will never be possible that everyone is happy!

We fix the rollover for admins, but we actually change the behaviour for users. Previously a user when doing the token rollover would **need** to change his PIN according to the PIN policies. I am not sure, if this was the desired behaviour.

Now, a user can rollover a token without being able to change the PIN. So if a user wants to change his PIN, he would actually have to do a 2nd step.
On the other hand: The admin can *not* force the user to rollover a token - I think. So this could not be the scenario for changing or updating the PIN.

PR #3687, opened 2023-07-14
Update translations

plettich:
- Fix some issues with translations (spelling, marks)
- update translation templates for UI and server

### DIFF
--- a/privacyidea/static/components/token/views/token.import.html
+++ b/privacyidea/static/components/token/views/token.import.html
@@ -49,7 +49,7 @@
             The file is supposed to contain one token per line.
         </p>
         <p class="help-block" translate>
-            If you enrolling the Yubikey in HOTP mode, you should use the
+            If you are enrolling the Yubikey in HOTP mode, You should use the
             PSKC file format in the newer version of the Yubikey GUI.
         </p>
     </div>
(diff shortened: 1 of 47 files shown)


plettich:
@cornelinux the `*.po` and `*.pot` file are probably not important here but i fixed some spelling and how the  translation was tagged in the code.

plettich:
@cornelinux the translation from the server for the token types works just fine.

privacyidea/static/components/token/views/token.import.html
```diff
@@ -49,7 +49,7 @@ <h3 translate>Import tokens from a file</h3>
             The file is supposed to contain one token per line.
         </p>
         <p class="help-block" translate>
-            If you enrolling the Yubikey in HOTP mode, you should use the
+            If you are enrolling the Yubikey in HOTP mode, You should use the
```
cornelinux: "You should use" shouldn't it be a lower case "y"?

cornelinux (approved):

privacyidea/static/components/token/views/token.import.html
cornelinux: @plettich I took the liberty to change this.

plettich: yeah, i guess i am old fashioned, i always capitalize the form of address :-)

PR #3696, opened 2023-07-18
Update translations

plettich:


cornelinux:
Why do we always have thousands of line break changes?

plettich:
> Why do we always have thousands of line break changes?

I guess that msgmerge and weblate don't like each other. There is an option in weblate to do the msgmerge part completely (update each language file from the template). I'll check this out.

plettich:
I activated the plugin on weblate. I'll create another PR with just the changed template. We'll see if weblate works as expected :-)

Issue #3604, opened 2023-04-28
Push Token: Set private key checking as default disabled in Sign class (lib/crypto.py)


kkalev:
In `lib/crypto.py` the init function for the `Sign` class uses the function `load_pem_private_key()` to load the private key. As per issue #3590 this function is quite heavy in execution time as per the following profiling run:
```
PATH: '/validate/polltransaction'
         17974 function calls (17838 primitive calls) in 0.462 seconds

   Ordered by: internal time, call count
   List reduced from 1307 to 30 due to restriction <30>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        3    0.396    0.132    0.396    0.132 {built-in method cryptography.hazmat.bindings._openssl.RSA_check_key}
       15    0.017    0.001    0.017    0.001 {method 'recv_into' of '_socket.socket' objects}
        4    0.006    0.002    0.006    0.002 {built-in method cryptography.hazmat.bindings._openssl.EVP_PKEY_sign}
       15    0.004    0.000    0.004    0.000 {method 'sendall' of '_socket.socket' objects}
        3    0.003    0.001    0.003    0.001 {built-in method cryptography.hazmat.bindings._openssl.PEM_read_bio_PrivateKey}
        2    0.002    0.001    0.002    0.001 {built-in method cryptography.hazmat.bindings._openssl.RSA_blinding_on}
        2    0.002    0.001    0.002    0.001 {built-in method cryptography.hazmat.bindings._openssl.PEM_read_bio_PUBKEY}
```

Private key checking can be disabled with a strong drop in performance:
```
PATH: '/validate/polltransaction'
         16821 function calls (16704 primitive calls) in 0.076 seconds

   Ordered by: internal time, call count
   List reduced from 1285 to 30 due to restriction <30>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       14    0.034    0.002    0.034    0.002 {method 'recv_into' of '_socket.socket' objects}
        4    0.006    0.002    0.006    0.002 {built-in method cryptography.hazmat.bindings._openssl.EVP_PKEY_sign}
        3    0.003    0.001    0.003    0.001 {built-in method cryptography.hazmat.bindings._openssl.PEM_read_bio_PrivateKey}
        2    0.002    0.001    0.002    0.001 {built-in method cryptography.hazmat.bindings._openssl.RSA_blinding_on}
[...]

PR #3705, opened 2023-07-20
Make RSA key check configurable

### LINKED ISSUE #3590: Add a policy action to disable private key checking
We are using push tokens to authenticate  users. The `/triggerchallenge` endpoint calls the `create_challenge()` function in `lib/tokens/pushtoken.py`. This function calls `_build_smartphone_data()` which among other things uses `load_pem_private_key()` to load a per token private key.

Profiling `create_challenge()` for execution time showed that more than 200 msecs are consumed inside `_build_smartphone_data()` with mostly all of them inside `load_pem_private_key()`. Based on this [issue](https://github.com/pyca/cryptography/issues/7236) we tested the following code (a loop of 1000 calls to the function):
```
from cryptography.hazmat.primitives import serialization

for _ in range(1000):
    with open("key.pem", "rb") as key_file:
        private_key = serialization.load_pem_private_key(key_file.read(), password=None)
```
with a total execution time of `29.089s`.

Based on suggestion in the above issue we also added the following on the above code:
```
from cryptography.hazmat.backends.openssl import backend
backend._rsa_skip_check_key = True
```
which resulted in a reduction of total execution time to **0.5 secs**. The same was true if disabling key checking in `_build_smartphone_data()` with a drop of execution time to a few msecs.

Benchmarking the endpoint with and without key checking resulted in a throughtput of 3-4 reqs/sec vs 20-30 reqs/sec depending on concurrent requests. CPU usage was 100% in the first case after 16 concurrent request while it typically remained **below** 100% when key check was disabled. We are talking about an order of magnitude improvement.

Since the private key is created by privacyIDEA itself when a token is enrolled, we would like to suggest that a policy action is added to allow for the key checking to be disabled if desired by the administrator.

Thank you for your time

plettich:
Since v39.0 `pyca/cryptography` support the parameter `unsafe_skip_rsa_key_validation` when loading RSA keys which greatly speeds up the loading time but is unsafe if there is no control over the private key.
The PUSH-token saves the generated private keys encrypted in the database so we can be fairly sure that they were not tampered with.

However for signing the Audit-log as well as signing the response we load the private key from the filesystem.

The added configuration entries `PI_RESPONSE_NO_PRIVATE_KEY_CHECK` and `PI_AUDIT_NO_PRIVATE_KEY_CHECK` (which default to `False`) can be used to disable the costly checking when loading the private keys. These entries should only be enabled if the security of the private keys is assured.

Closes #3590 #3604

**TODO:**
- [x] update documentation


cornelinux:
@plettich Is this supposed to be draft? Due to TODO-docs?

cornelinux (commented):
Looks good to me.
Please also add the docs in 
https://github.com/privacyidea/privacyidea/blob/master/doc/installation/system/inifile.rst

doc/faq/crypto-considerations.rst
```diff
@@ -55,21 +55,25 @@ This way a database administrator is not able to inject rogue password hashes.
...
+
 Audit Signing
 ~~~~~~~~~~~~~
 
-The audit log is digitally signed. (see :ref:`Audit` and :ref:`cfgfile`).
+The audit log is digitally signed. (see :ref:`audit` and :ref:`audit_parameters`).
 
-The audit log can be handled by different modules. privacyIDEA comes with an
-SQL Audit Module.
+The audit log can be handled by different modules. privacyIDEA comes with a
```
cornelinux, line 65: @plettich  "an" is correct!

cornelinux: I see. It depends on if SQL is pronounced "esS-Queue-elL" or "sequel".

plettich: Done in 1ce2fd99b

doc/faq/performance.rst
```diff
@@ -36,7 +36,41 @@ You can configure how often the timestamp should be read using the pi.cfg
...
+safety of the private key in ``PI_AUDIT_KEY_PRIVATE`` can be assured, the config
+entry ``PI_AUDIT_NO_PRIVATE_KEY_CHECK = True`` in :ref:`cfgfile` can improve the performance.
+
+With the config entry ``PI_AUDIT_NO_SIGN = True`` the signing of the Audit-log
+can be deactivated completely.
+
+The privacyIDEA Response
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+By default, privacyIDEA signs every JSON-Response with the private key in ``PI_AUDIT_KEY_PRIVATE``. To improve the performance when loading the private key the config entry ``PI_RESPONSE_NO_PRIVATE_KEY_CHECK`` can be set to ``True``.
```
cornelinux, line 71: @plettich Please put this key in the inifile.rst, this is where we will search for it.

doc/faq/crypto-considerations.rst
```diff
@@ -55,21 +55,25 @@ This way a database administrator is not able to inject rogue password hashes.
...
+SQL Audit Module which is enabled by default.
 
 For signing the audit log the SQL Audit Module uses the RSA keys specified
-with the values *PI_AUDIT_KEY_PUBLIC* and *PI_AUDIT_KEY_PRIVATE* in
+with the values ``PI_AUDIT_KEY_PUBLIC`` and ``PI_AUDIT_KEY_PRIVATE`` in
 :ref:`cfgfile`.
 
 By default the installer generates 2048bit RSA keys.
 
+If the validity of the private key can be assured, the config entry
```
cornelinux, line 74: "If the validity of the private key can be assured" and first I did not understand, that you mean, "if you can be sure that the private key is legit or something like this". Maybe you can think of a rephrase. Might not be that necessary.

doc/faq/performance.rst
```diff
@@ -36,7 +36,41 @@ You can configure how often the timestamp should be read using the pi.cfg
...
+
+Please also read :ref:`crypto_considerations` to understand the implications.
+
+.. _faq_perf_crypto_audit:
+
+The Audit-log
+^^^^^^^^^^^^^
+
+Each entry in the :ref:`audit` log is signed to detect tampering. When the
+safety of the private key in ``PI_AUDIT_KEY_PRIVATE`` can be assured, the config
```
cornelinux, line 62: Again. Maybe "if you can be sure, that the private key originates from a trusted source".

cornelinux (changes requested):
My only request is, that you put the two new keys PI_AUDIT_NO_PRIVATE_KEY_CHECK and the other one also in the description for pi.cfg, in inifile.rst. Otherwise we will search endlessly for these keys in a year, when we try to look them up in the docs.

doc/faq/performance.rst
plettich: Done in 91bcff523

doc/faq/crypto-considerations.rst
plettich: Done in 91bcff523

doc/faq/performance.rst
plettich: Done in 91bcff523

cornelinux (approved):

PR #3714, opened 2023-07-26
Try to fix failing applspec password test

plettich:
The test might fail sporadically due to incorrectly quoted data.

### DIFF
--- a/tests/test_api_validate.py
+++ b/tests/test_api_validate.py
@@ -3308,22 +3308,26 @@ class RegistrationAndPasswordToken(MyApiTestCase):
         with self.app.test_request_context('/validate/check',
                                            method='POST',
                                            data={"user": "cornelius",
-                                                 "pass": "test{0!s}".format(password)}):
+                                                 "pass": quote("test{0!s}".format(password))}):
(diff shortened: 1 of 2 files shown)


tests/test_api_validate.py
```diff
@@ -3330,11 +3330,11 @@ def test_02_application_specific_password_token(self):
                                            method='POST',
                                            data={"user": "cornelius",
                                                  "service_id": "thunderbird",
-                                                 "pass": "test{0!s}".format(password)}):
```
cornelinux, line 3333: If the password has same evil chars this would probably also happen in line 3322.

cornelinux (approved):
I think there are more locations, where this can happen. 
See before, line 3322. So please add the quote also there.
Can be merged then.

tests/test_api_validate.py
plettich: This probably slipped through since we just check if the authentication was `REJECT`ed, but not why.
Will fix this

plettich: Done in 216198383

cornelinux (approved):

PR #3735, opened 2023-08-24
Fix calculation of TOTP values

### LINKED ISSUE #3734: TOTP not calculated correctly
The algorithm we use for calculating TOTP values differs slightly from the standard:
https://github.com/privacyidea/privacyidea/blob/6f446f67a3c1395f4bdb2ad7f6e490a57371e04c/privacyidea/lib/tokens/totptoken.py#L266
We round to the nearest integer while the RFC explicitly states that `floor` should be used:
https://datatracker.ietf.org/doc/html/rfc6238#section-4.2
This wasn't noticed because of the default time-window used during checks.
 @frankmer pointed out the discrepancy while implementing the *DayPasswordToken* which doesn't use a time-window.

plettich:
Instead of using the `floor` function for calculating the counter of the *TOTP* algorithm, the counter was rounded to the nearest integer value. Thus the validity of a single *OTP* value was always shifted `timestep / 2` into the future.
For the regular *TOTP* algorithm this wasn't a problem since it also defines a time-drift and a time-window for checking the value. The *DayPassword* token which is derived from the *TOTP* algorithm however does not use a time-window.

This fix should not severely affect existing *TOTP* tokens since the algorithm already accounts for time-drifts.

Closes #3734

### DIFF
--- a/privacyidea/lib/tokens/totptoken.py
+++ b/privacyidea/lib/tokens/totptoken.py
@@ -239,7 +239,7 @@ class TotpTokenClass(HotpTokenClass):
         checks if the given OTP value is/are values of this very token at all.
         This is used to autoassign and to determine the serial number of
         a token.
-        In fact it is a check_otp with an enhanced window.
+        In fact, it is a check_otp with an enhanced window.
 
         :param otp: the to be verified otp value
         :type otp: string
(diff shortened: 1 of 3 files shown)


tests/test_lib_tokens_daypassword.py
```diff
@@ -414,97 +398,83 @@ def test_18_challenges(self):
         self.assertTrue(resp, resp)
 
         # test if challenge is valid
-        chal.is_valid()
+        self.assertTrue(chal.is_valid())
+        chal.delete()
 
-    @mock.patch('time.time', mock.MagicMock(return_value=1686902767))
+#    @mock.patch('time.time', mock.MagicMock(return_value=1417549521))
```
jona-samuel, line 404: maybe we should remove this

privacyidea/lib/tokens/totptoken.py
```diff
@@ -262,42 +262,37 @@ def check_otp_exist(self, otp, window=None, options=None, symetric=True,
...
         :rtype: float
         """
-        if type(curtime) == datetime.datetime:
-            dt = curtime
-        else:
-            dt = datetime.datetime.now()
-
-        td = (dt - datetime.datetime(1970, 1, 1))
-        return td.total_seconds()
+        if curtime:
```
jona-samuel, line 287: Is this testable?

jona-samuel (changes requested):
https://github.com/privacyidea/privacyidea/blob/6f446f67a3c1395f4bdb2ad7f6e490a57371e04c/privacyidea/lib/tokens/totptoken.py#L516

I think we should remove the __+ 0.5__ hir too

privacyidea/lib/tokens/totptoken.py
plettich: I'll address this in a different issue

tests/test_lib_tokens_daypassword.py
plettich: Done in [107a4f3](https://github.com/privacyidea/privacyidea/pull/3735/commits/107a4f35ffefc12471b6d7c7798c6c8345f5cbf0)
(force-pushed)

jona-samuel (approved):

PR #3743, opened 2023-08-30
Change prefix for the serial numbers from DayPassword to DYPW

### LINKED ISSUE #3742: 4 chars ID of day password token.
The ID of the day password token should only be 4 chars. Like "DYPW" oder "PIDP".

Currently it is "DayPassword", which leads to a very long serial number.

image

jona-samuel:
closes #3742


cornelinux (approved):
Looks good to me

PR #3745, opened 2023-08-30
Better label for day password policy

### LINKED ISSUE #3744: Better label for day password policy
The label in the policy for the `daypassword_timestep` should be more helpful.

Please replace "text..." with something like "24h".

image

jona-samuel:
closes #3744


cornelinux (approved):
Looks good to me

PR #3755, opened 2023-09-08
Add missing sequences

### LINKED ISSUE #3754: Missing sequences when upgrading to 3.9 /SQLALchemy 1.4 and mariadb
When upgrading from 3.8 to 3.9 with a mariadb we are missing sequences.
It looks like SQLAlchemy 1.4 needs sequences.

A fresh install with 3.9 and mariadb creates sequences.

We need a migration script, that creates the sequences.

Problem: We need to create the sequences with the next id, otherwise we fail with dublicate entries.
I am adding a mockup for  a migrationscript that basically works, 

* [ ] we need to add all old tables 
* [ ] and we need to verify the next id and create the sequence accordingly.

cornelinux:
Working on #3754

### DIFF
--- a/READ_BEFORE_UPDATE.md
+++ b/READ_BEFORE_UPDATE.md
@@ -20,8 +20,11 @@
   ```
   Unless you are using this API call directly, this is not relevant for normal operation.
 
-The database table `serviceid` is added, there is no data migration
-necessary.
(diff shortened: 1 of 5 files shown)


cornelinux:
Talked with @pablo-knight, last commit looks good.

PR #3756, opened 2023-09-09
Add multiple choice realms and resolvers

DesSolo:


cornelinux:
@DesSolo This sounds really interesting, if this works.

However, 

* [ ] we need an issue for this, which we can reference in the PR and later in the changelog
* [ ] Please also add a test
* [ ] Check if the docs need to be updated. See https://github.com/privacyidea/privacyidea/blob/master/doc/eventhandler/index.rst#conditions

cornelinux:
@jona-samuel Please write an issue that describes, what this should do and then take the PR.

jona-samuel:
Is replaced by #3794

Issue #3227, opened 2022-07-19
Variables in PUSH Notifications


adamboutcher:
**Is your feature request related to a problem? Please describe.**
Show the clients remote IP address trying to access resources in the PUSH notification.

**Describe the solution you'd like**
The variable that allows you to include custom text to also access certain variables.

**Describe alternatives you've considered**
I don't think there is an alternative.

**Additional context**
Discussion - https://community.privacyidea.org/t/variables-in-push-notification-text/

# Implementation considerations

Variables would be used in the policy settings MOBILE_TEXT and MOBILE_TITLE.
See https://github.com/privacyidea/privacyidea/blob/900b7250e41ff5a5eecfe89cd3f274a8f9df7595/privacyidea/lib/tokens/pushtoken.py#L349

Currently variables are also used e.g. in *smstext*. See https://privacyidea.readthedocs.io/en/v3.9/policies/authentication.html#smstext

The text is created this way:
https://github.com/privacyidea/privacyidea/blob/9c2bcdb7080460152841eb241d415bfd1e43389c/privacyidea/lib/tokens/smstoken.py#L429

The function `create_tag_dict` also knows the client_ip. But the client_ip usually is the IP of the requesting application talking to privacyIDEA.

This code
https://github.com/privacyidea/privacyidea/blob/900b7250e41ff5a5eecfe89cd3f274a8f9df7595/privacyidea/lib/tokens/pushtoken.py#L189

could be probably simply enhanced like this:

~~~~Python
    message_on_mobile = get_action_values_from_options(SCOPE.AUTH,
                                                       PUSH_ACTION.MOBILE_TEXT,
                                                       options) or DEFAULT_MOBILE_TEXT

    tags = create_tag_dict(serial=serial,
                           tokenowner=User,
                           tokentype="sms",
                           recipient={"givenname": User.info.get("givenname") if User else "",
                                      "surname": User.info.get("surname") if User else ""},
                           challenge=options.get("challenge"))
    message_on_mobile = message_on_mobile.format(**tags)
~~~~

PR #3761, opened 2023-09-18
Working on #3227

cornelinux:


nilsbehlen:
please add the user agent of the request that triggered the token as an option aswell!
Or some kind of mapping so the push message can contain the application that is being logged into.

cornelinux:
> please add the user agent of the request that triggered the token as an option aswell! Or some kind of mapping so the push message can contain the application that is being logged into.

Would be great and easy, if we had the `request` available [here](https://github.com/privacyidea/privacyidea/blob/53da741d95a8115e84812b4471197ce7bfde201a/privacyidea/lib/tokens/pushtoken.py#L193), which we do not have to my knowledge.

I will look into it.

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -348,7 +373,8 @@ def get_class_info(key=None, ret='all'):
                    SCOPE.AUTH: {
                        PUSH_ACTION.MOBILE_TEXT: {
                            'type': 'str',
-                           'desc': _('The question the user sees on his mobile phone.'),
+                           'desc': _('The question the user sees on his mobile phone. Several tags like {serial} and '
```
nilsbehlen, line 376: Why not list all tags that are working? Are there too many?

nilsbehlen (approved):

privacyidea/lib/tokens/pushtoken.py
cornelinux: These are imho too many. 
Also we provide the tags in the documentation so I would rather not raise the impression, that there is a complete list here. If further tags are added in the future, this would mean to add these at three locations:
a) the code
b) the descripiont (here)
c) the docs.

So I would rather only update 2 locations (code and docs)

nilsbehlen (approved):
don't know if failing checks should be looked at. code looks fine

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -189,6 +189,31 @@ def _build_smartphone_data(serial, challenge, registration_url, pem_privkey, opt
     message_on_mobile = get_action_values_from_options(SCOPE.AUTH,
                                                        PUSH_ACTION.MOBILE_TEXT,
                                                        options) or DEFAULT_MOBILE_TEXT
+    # Get the request object
+    req_headers = options.get("g", {}).get("request_headers", {})
+    req_environment = req_headers.environ
```
plettich, line 194: In case there are no `request_headers` in `g`-object, this fails.
Maybe we should just set `request = None` in this case.

plettich (changes requested):

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -189,6 +189,33 @@ def _build_smartphone_data(serial, challenge, registration_url, pem_privkey, opt
...
+    request = None
+    if req_headers:
+        req_environment = req_headers.environ
+        request = req_environment.get("werkzeug.request")
+    if request:
+        user_object = request.User
+    else:
+        # Get owner from token object
+        try:
+            tok = get_one_token(serial=serial)
```
plettich, line 203: To avoid another db lookup, we could just pass the token object itself to the function and get serial and assigned user from it.

plettich (changes requested):

privacyidea/lib/tokens/pushtoken.py
cornelinux: Will check and do.

cornelinux: Done in 6c5aed4c

PR #3768, opened 2023-09-25
handle policy key error in hotp

### LINKED ISSUE #2849: check policies for valid keys
At the moment there are some policies which allow the usage of keys like `{serial}` or `{realm}`.

Typos in policy keys happen quickly.

When the key is used by the code, privacyIDEA throws an according exception, like here for an /token/init with a "tokenlabel="{real}".

```
[2021-08-03 17:40:48,049][6113][140010240476928][ERROR][privacyidea.lib.tokens.hotptoken:292] Traceback (most recent call last):
  File "/opt/privacyidea/lib/python3.6/site-packages/privacyidea/lib/tokens/hotptoken.py", line 268, in get_init_detail
    extra_data=extra_data)
  File "/opt/privacyidea/lib/python3.6/site-packages/privacyidea/lib/log.py", line 155, in log_wrapper
    return func(*args, **kwds)
  File "/opt/privacyidea/lib/python3.6/site-packages/privacyidea/lib/apps.py", line 140, in create_google_authenticator_url
    surname=user_obj.info.get("surname", ""))
KeyError: 'real'

[2021-08-03 17:40:48,049][6113][140010240476928][ERROR][privacyidea.lib.tokens.hotptoken:293] failed to set oath or google url: KeyError('real',)
```

To the user this does not provide much information about the incorrect policy. In this case, the frontend user only notices, that the QR-Code of the HOTP token is not displayed during rollout.

It would be nice to catch such errors or to check policies for the supported key words.

jona-samuel:
closes #2849

### DIFF
--- a/privacyidea/lib/tokens/hotptoken.py
+++ b/privacyidea/lib/tokens/hotptoken.py
@@ -300,7 +300,7 @@ class HotpTokenClass(TokenClass):
                                                   }
                 except KeyError as ex:
                     log.debug("{0!s}".format((traceback.format_exc())))
-                    log.error('Unknown Tag {0!r} in one of your policy definition'
+                    log.error('Unknown Tag {0!s} in one of your policy definition'
                               .format(ex))
                 except Exception as ex:  # pragma: no cover
                     log.debug("{0!s}".format((traceback.format_exc())))
--- a/privacyidea/lib/tokens/hotptoken.py
+++ b/privacyidea/lib/tokens/hotptoken.py
@@ -298,6 +298,11 @@ class HotpTokenClass(TokenClass):
                                                   "value": oath_url,
                                                   "img": create_img(oath_url)
                                                   }
+                except KeyError as ex:
+                    audit_entry = self.find_most_recent_audit_entry(action= "POST /auth")
+                    log.error("{0!s}".format((traceback.format_exc())))
+                    log.error('Unknown Tag {0!r} in one of your policy definition: {1!r}'
+                              .format(ex, audit_entry['policies']))
                 except Exception as ex:  # pragma: no cover
                     log.error("{0!s}".format((traceback.format_exc())))
                     log.error('failed to set oath or google url: {0!r}'.format(ex))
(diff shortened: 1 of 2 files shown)


privacyidea/lib/tokens/hotptoken.py
```diff
@@ -298,6 +298,11 @@ def get_init_detail(self, params=None, user=None):
                                                   "value": oath_url,
                                                   "img": create_img(oath_url)
                                                   }
+                except KeyError as ex:
+                    audit_entry = self.find_most_recent_audit_entry(action="POST /auth")
```
cornelinux, line 302: I think you are mixing s.th. up?
The find_most_recent_audit_entry is a member function of the tests.

Please wirte a test, to cover this Exception.

privacyidea/lib/tokens/hotptoken.py
```diff
@@ -298,6 +298,11 @@ def get_init_detail(self, params=None, user=None):
...
+                    log.error("{0!s}".format((traceback.format_exc())))
```
cornelinux, line 303: Please make the traceback a debug level. Usually we are doing tracebacks in debug.

privacyidea/lib/tokens/hotptoken.py
```diff
@@ -298,6 +298,11 @@ def get_init_detail(self, params=None, user=None):
...
+                    log.error('Unknown Tag {0!r} in one of your policy definition: {1!r}'
```
cornelinux, line 304: You can remove {1|r}, since we can not know the name of the policy at this point.
I think it is fine, so simply have the wrong tag!

cornelinux (changes requested):
Please remove the non-existing member function and add a test.

cornelinux (approved):

cornelinux (changes requested):
Thanks for fixing these comments.
However, the tests fail with python 3.6. I think it is due to some different handling of strings and quotes.
Maybe you can mange to avoid these, so that the tests will also run with 3.6.

PR #3774, opened 2023-10-02
Fix wrong start counter for new sequences

plettich:
The initial migration script for adding missing sequences had an issue where new sequences started with 2 instead of 1.

Working on #3778

### DIFF
--- a/migrations/versions/5cb310101a1f_.py
+++ b/migrations/versions/5cb310101a1f_.py
@@ -48,7 +48,7 @@ def upgrade():
                     continue
 
                 # Create the sequence with the correct next_id!
-                current_id = session.query(func.max(tbl.c.id)).one()[0] or 1
+                current_id = session.query(func.max(tbl.c.id)).one()[0] or 0
                 print(f"CurrentID in Table {tbl.name}: {current_id}")
                 try:
                     seq = Sequence(seq_name, start=(current_id + 1), increment=inc_value)


plettich:
@cornelinux we should merge this before we branch of `branch-3.9` from master.

cornelinux:
@lets talk when we meet again :-)

cornelinux:
I would merge this into branch-3.9 and when 3.9.1 is released, we merge the 3.9 branche into the master.
Just as we always do it.

cornelinux:
It worked!

migrations/versions/5cb310101a1f_.py
```diff
@@ -48,7 +48,7 @@ def upgrade():
                     continue
 
                 # Create the sequence with the correct next_id!
-                current_id = session.query(func.max(tbl.c.id)).one()[0] or 1
+                current_id = session.query(func.max(tbl.c.id)).one()[0] or 0
```
cornelinux, line 51: Are there any databases that might start with a 1?
So we saw this problem with postgres. Would the 0 pose problems with other DBs?

plettich: There is a `+ 1` when the next id counter gets set:
https://github.com/privacyidea/privacyidea/blob/83c3c6d20e1a878c9ca9e4c738acab1064dc86e7/migrations/versions/5cb310101a1f_.py#L54

cornelinux (approved):

Issue #3764, opened 2023-09-21
WebAuthn Token for offline use


cornelinux:
We want to use webauthn tokens like Yubikeys for offline authentication with the privacyIDEA Credenital Provider.

The privacyIDEA server needs to provide the public key, the credential ID and the relying party (RP) to the credential provider.

This can be done after a successful authentication.

To do so, we need to enhance `lib/applications/offline.py`.

* [x] Hm, methods like `get_options` now need to be tokentype dependent.
   * [x] libs and APIs. Done in 4627069d
   * [x] WebUI. Done in fb7b30ad
* [x] The method `get_authentication_items` return the offline information and needs to add another tokentype
https://github.com/privacyidea/privacyidea/blob/c744d78c68de2ae36d16076381207d8bb2b348c1/privacyidea/lib/applications/offline.py#L134
* [x] We need to check `get_refill`, what needs to be done for a webauthn token, here.
* [x] The refill will check, if the token still exist and it should return a sensible return value, if the token does not exist anymore or is disabled. This probably needs to happen on the upper layer.

For more information during development talk to @nilsbehlen 

# Implementation

*check-refill* see below.

In regards to `get_options` this seems a bit far fetched, since the options dict is created in advance.
Currently the offline application returns the count and rounds. We will see, if this is somehow counter productive.

However, *offline* --> tokentype: hotp and webauthn, *ssh* -> tokentype: sshkey, *luks* -> TOTP (as yubikeys).

So there seems to be an easy mapping here between application type and token type. This could also help with the webui.

## API

For WebAuthn-Tokens we return:

~~~~JSON
{... 
  "user":  <username>,
  "response": { "pubkey": <the pubkey of the webauthn token>,
                          "credential_id": <the credential ID of the token> },
  "refilltoken":
}
~~~~

In contrast to the HOTP token we do not return OTPs but the pub key.
@nilsbehlen  We also return a refilltoken, even if there is no "refill" necessary. But during the refill the client can check if the Webauthn token is still active. The refill endpoint will return the correspdonding result.

cornelinux:
# Check refill

It is already checked if the token does still exist.

https://github.com/privacyidea/privacyidea/blob/17f9adac819fcd1b97f0b2283736001fcf8885ea/privacyidea/api/validate.py#L180

However, we need to verify, if the token is disabled or revoked.

PR #3776, opened 2023-10-04
Offline Webauthn 

cornelinux:
In case the offline token is locked or inactive,
the refill will fail.

Working on #3764


nilsbehlen:
To use webauthn token offline, the application requires the public key of the token from privacyidea, which will be sent after a sucessful authentication.
To generate challenges, there are some more parameters that can be set. The application can either copy them from the challenge that was answered before, or privacyidea can send them along the private key.
The request looks like this:
grafik
where the challenge will then be generated by the application and the other parameters will be stored with the public key.

plettich (approved):

Issue #3763, opened 2023-09-20
Use of token description in user notification event handler


maxstaubach:
Hello,

I would like to access the token description when sending an email to the token owner.
Currently, only the following attributes are available in the email body: [documentation](https://privacyidea.readthedocs.io/en/latest/eventhandler/usernotification.html#body-for-all-actions)

This requires changes to be made in two files:
- lib/eventhandler/usernotification.py
- privacyidea/lib/utils/__init__.py 

Following is a proposal for the change (save as .patch file for syntax highlighting).

``` python

diff --git a/privacyidea/lib/eventhandler/usernotification.py b/privacyidea/lib/eventhandler/usernotification.py
index c9c26151..8dc18d02 100644
--- a/privacyidea/lib/eventhandler/usernotification.py
+++ b/privacyidea/lib/eventhandler/usernotification.py
@@ -393,10 +393,12 @@ class UserNotificationEventHandler(BaseEventHandler):
             googleurl_img = content.get("detail", {}).get("googleurl",
                                                           {}).get("img")
             tokentype = None
+            tokendescription = None
             if serial:
                 tokens = get_tokens(serial=serial)
                 if tokens:
                     tokentype = tokens[0].get_tokentype()
+                    tokendescription = tokens[0].token.description
             else:
                 token_objects = get_tokens(user=tokenowner)
                 serial = ','.join([tok.get_serial() for tok in token_objects])
@@ -410,6 +412,7 @@ class UserNotificationEventHandler(BaseEventHandler):
                                    tokenowner=tokenowner,
                                    serial=serial,
                                    tokentype=tokentype,
+                                   description=tokendescription,
                                    registrationcode=registrationcode,
                                    escape_html=action.lower() == "sendmail" and
                                                handler_options.get("mimetype", "").lower() == "html")
diff --git a/privacyidea/lib/utils/__init__.py b/privacyidea/lib/utils/__init__.py
index 6116471e..62a28819 100644
--- a/privacyidea/lib/utils/__init__.py
+++ b/privacyidea/lib/utils/__init__.py
@@ -1303,6 +1303,7 @@ def create_tag_dict(logged_in_user=None,
                     serial=None,
                     tokenowner=None,
                     tokentype=None,
+                    description=None,
                     recipient=None,
                     registrationcode=None,
                     googleurl_value=None,
@@ -1320,6 +1321,7 @@ def create_tag_dict(logged_in_user=None,
     :param tokenowner: The owner of the token
[...]

PR #3767, opened 2023-09-25
added token description in user notification event handler

maxstaubach:
Closes #3763

PR #3783, opened 2023-10-16
added token description in user notification event handler

jona-samuel:
Closes https://github.com/privacyidea/privacyidea/issues/3763

Replaces: https://github.com/privacyidea/privacyidea/pull/3767

### DIFF
--- a/privacyidea/lib/eventhandler/usernotification.py
+++ b/privacyidea/lib/eventhandler/usernotification.py
@@ -39,6 +39,7 @@ It can be bound to each event and can perform the action:
 
 The module is tested in tests/test_lib_eventhandler_usernotification.py
 """
+
 from privacyidea.lib.eventhandler.base import BaseEventHandler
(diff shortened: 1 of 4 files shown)


plettich (changes requested):

privacyidea/lib/eventhandler/usernotification.py
```diff
@@ -407,6 +411,7 @@ def do(self, action, options=None):
                                    tokenowner=tokenowner,
                                    serial=serial,
                                    tokentype=tokentype,
+                                   tokendescription=tokendescription,
```
plettich, line 414: Please also add the `{tokendescription}` tag to the list of possible tags in the documentation here:
https://github.com/privacyidea/privacyidea/blob/ade6a2c85d34691ff5f1fa5062ac6c87d84aff8c/doc/eventhandler/usernotification.rst?plain=1#L152

plettich (approved):

PR #3799, opened 2023-10-30
Update notes about subscriptions

cornelinux:
Add Shibboleth Plugin


cornelinux:
@sepp67 @pablo-knight 
Would you please check from an admin and support perspective, if the new text is understandable?

privacyidea/lib/subscriptions.py
```diff
@@ -67,6 +67,8 @@
                 "privacyidea-ldap-proxy": 50,
                 "privacyidea-cp": 50,
                 "privacyidea-pam": 10000,
+                "privacyidea-shibboleth": 10000,
+                "shibbolethidp": 10000,
```
pablo-knight: This may not be necessary.

pablo-knight (approved):

PR #3802, opened 2023-10-30
Token description requirement works with verify enroll

### LINKED ISSUE #3798: Token description requirement prohibits verify_enrollment
### Top-level intent

When description is reqired for a token type the description is also required in the verification state. There is no Field to enter the description there so we can not enrol a token

### Steps to reproduce

1. Set `verify_enrollment` and `require_description` to totp
2. Start enrolling a totp token and provide a description in the first step
3. Provide the correct OTP in the verification step

### Expected outcome

The token should be enrolled

### Actual outcome

I can not enroll the token. I get an error that description is required for totp.

### Configuration

* **privacyIDEA version**: 3.9
* **Installation method**: Pypi
* **Python version**: 3.6.8
* **Operating system**: Red Hat Enterprise Linux 8
* **Webserver**: Apache
* **Token database**: Mariadb

### Fix

Adding the following code after [prepolicy.py#L2234](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/api/lib/prepolicy.py#L2234) seems to fix the problem for me:

```
serial = getParam(request.all_data, "serial", optional)
    if serial:
        type_value = get_token_type(serial)
        # Only now, we check if we need to verify
        tokenobj_list = get_tokens(serial=serial)
        if len(tokenobj_list) == 1:
            tokenobj = tokenobj_list[0]
            if tokenobj.rollout_state in [ ROLLOUTSTATE.CLIENTWAIT, ROLLOUTSTATE.VERIFYPENDING ] :
                return;
```

cornelinux:
Enrollment with verification (enter OTP) breaks the token description, since the 2nd /token/init call does not contain a description.

We identify the 2nd call and do not require the description.

Fixes #3798

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -87,7 +87,6 @@ from privacyidea.lib.clientapplication import save_clientapplication
 from privacyidea.lib.config import (get_token_class)
 from privacyidea.lib.tokenclass import ROLLOUTSTATE
 from privacyidea.lib.tokens.certificatetoken import ACTION as CERTIFICATE_ACTION
-from privacyidea.lib.tokenclass import ROLLOUTSTATE
 from privacyidea.lib.token import get_one_token
 import functools
 import jwt
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -69,7 +69,7 @@ import logging
 
 from OpenSSL import crypto
 from privacyidea.lib import _
-from privacyidea.lib.error import PolicyError, RegistrationError, TokenAdminError, ResourceNotFoundError
+from privacyidea.lib.error import PolicyError, RegistrationError, TokenAdminError, ResourceNotFoundError, ParameterError
 from flask import g, current_app
 from privacyidea.lib.policy import SCOPE, ACTION, REMOTE_USER
 from privacyidea.lib.policy import Match, check_pin
@@ -87,6 +87,8 @@ from privacyidea.lib.clientapplication import save_clientapplication
 from privacyidea.lib.config import (get_token_class)
 from privacyidea.lib.tokenclass import ROLLOUTSTATE
 from privacyidea.lib.tokens.certificatetoken import ACTION as CERTIFICATE_ACTION
+from privacyidea.lib.tokenclass import ROLLOUTSTATE
+from privacyidea.lib.token import get_one_token
 import functools
 import jwt
 import re
@@ -2233,6 +2235,12 @@ def require_description(request=None, action=None):
     token_types = list(action_values.keys())
     type_value = request.all_data.get("type") or 'hotp'
     if type_value in token_types:
-        if not request.all_data.get("description"):
+        tok = None
+        serial = getParam(params, "serial")
+        if serial:
+            tok = get_one_token(serial=serial, rollout_state=ROLLOUTSTATE.VERIFYPENDING, silent_fail=True) or \
+                  get_one_token(serial=serial, rollout_state=ROLLOUTSTATE.CLIENTWAIT, silent_fail=True)
+        # only if no token exists, yet, we need to check the description
+        if not tok and not request.all_data.get("description"):
             log.warning(_("Missing description for {} token.".format(type_value)))
             raise PolicyError(_("Description required for {} token.".format(type_value)))
--- a/privacyidea/lib/token.py
+++ b/privacyidea/lib/token.py
@@ -554,16 +554,21 @@ def get_tokens_paginate(tokentype=None, realm=None, assigned=None, user=None,
     return ret
 
 
-def get_one_token(*args, **kwargs):
+def get_one_token(*args, silent_fail=False, **kwargs):
     """
     Fetch exactly one token according to the given filter arguments, which are passed to
     ``get_tokens``. Raise ``ResourceNotFoundError`` if no token was found. Raise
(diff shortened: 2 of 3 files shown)


privacyidea/lib/token.py
```diff
@@ -554,16 +554,21 @@ def get_tokens_paginate(tokentype=None, realm=None, assigned=None, user=None,
...
+
+    :param silent_fail: Instead of raising an exception we return None silently
+    :returns: Token object
     """
     result = get_tokens(*args, **kwargs)
     if not result:
+        if silent_fail: return
         raise ResourceNotFoundError(_("The requested token could not be found."))
     elif len(result) > 1:
+        if silent_fail: return
```
plettich, line 571: Not suer if we want to allow this here when there are more than one token. After all the function is called `get_one_token()`. We should at least output something to the logs.

privacyidea/api/lib/prepolicy.py
```diff
@@ -2233,6 +2235,12 @@ def require_description(request=None, action=None):
     token_types = list(action_values.keys())
     type_value = request.all_data.get("type") or 'hotp'
     if type_value in token_types:
-        if not request.all_data.get("description"):
+        tok = None
+        serial = getParam(params, "serial")
+        if serial:
+            tok = get_one_token(serial=serial, rollout_state=ROLLOUTSTATE.VERIFYPENDING, silent_fail=True) or \
+                  get_one_token(serial=serial, rollout_state=ROLLOUTSTATE.CLIENTWAIT, silent_fail=True)
```
plettich, line 2242: I think it would be more pythonic to use:
```
if serial:
    for rs in (ROLLOUTSTATE.VERIFYPENDING, ROLLOUTSTATE.CLIENTWAIT):
        try:
            get_one_token(serial=serial, rollout_state=rs)
        except ResourceNotFoundError as _e:
            pass
```
To avoid multiple database request we could add an else block:
```
...
try:
    tok = get_one_token(serial=serial, rollout_state=rs)
except ResourceNotFoundError as _e:
    pass
else:
    break
```
This way we avoid augmenting the `get_one_token()` function.

plettich (changes requested):

privacyidea/lib/token.py
```diff
@@ -554,16 +554,21 @@ def get_tokens_paginate(tokentype=None, realm=None, assigned=None, user=None,
...
     Fetch exactly one token according to the given filter arguments, which are passed to
     ``get_tokens``. Raise ``ResourceNotFoundError`` if no token was found. Raise
     ``ParameterError`` if more than one token was found.
+
+    :param silent_fail: Instead of raising an exception we return None silently
+    :returns: Token object
     """
     result = get_tokens(*args, **kwargs)
     if not result:
+        if silent_fail: return
```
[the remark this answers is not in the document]
cornelinux: I think I tried one line for coverage reasons. But you are right. Will do both.

privacyidea/lib/token.py
cornelinux: In theory there should only be one token. So I would output a warning.
Will do.

privacyidea/api/lib/prepolicy.py
cornelinux: `get_one_token` does not allow a or list of states. It will only filter for *one* state.
This means, we need two calls!

I first started with fetching the exceptions. But as we need two calls, this is evil.

Yes, we need two database calls but we **only** need them if

- we are doing 2step enrollment in the 2nd step!
- and if it is not a verify-enrollment but a client-wait enrollment.

So I would like to stick with the current way in the PR.

privacyidea/api/lib/prepolicy.py
```diff
@@ -87,6 +87,8 @@
 from privacyidea.lib.config import (get_token_class)
 from privacyidea.lib.tokenclass import ROLLOUTSTATE
 from privacyidea.lib.tokens.certificatetoken import ACTION as CERTIFICATE_ACTION
+from privacyidea.lib.tokenclass import ROLLOUTSTATE
```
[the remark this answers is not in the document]
cornelinux: Done in 295db716

privacyidea/lib/token.py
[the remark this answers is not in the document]
cornelinux: Done in 295db716

privacyidea/lib/token.py
cornelinux: Done in 295db716

privacyidea/lib/token.py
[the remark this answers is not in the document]
cornelinux: :-/ You killed my coverage. I will NOT add a test to do this.

privacyidea/api/lib/prepolicy.py
plettich: Ok, the complete code would be:
```
if serial:
    for rs in (ROLLOUTSTATE.VERIFYPENDING, ROLLOUTSTATE.CLIENTWAIT):
        try:
            get_one_token(serial=serial, rollout_state=rs)
        except ResourceNotFoundError as _e:
            pass
        else:
            break
```
So both states are covered and if a token is found with the state `ROLLOUTSTATE.VERIFYPENDING` the other isn't tested anymore.

cornelinux: I prefer my one line

cornelinux: This makes it also fucking complicated, since you also need to test for the existence of the description.

Issue #3784, opened 2023-10-16
Convert pi-manage to click


plettich:
Drop `Flask-Script` and rewrite `pi-manage` to use `click`

TODO:
Missing commands:
- [x] `hsm`
- [x] `token`
- [ ] ~`validate`??~

Tests:
- [ ] #3814

PR #3803, opened 2023-11-03
Migrate pi-manage to click

plettich:
Working on #3784


jona-samuel (approved):

pi-manage
```diff
@@ -29,1714 +29,27 @@
 #            Add sub command for policies
 # 2014-12-15 Cornelius Kölbel, info@privacyidea.org
 #            Initial creation
+
+# SPDX-FileCopyrightText: (C) 2023 Paul Lettich
```
cornelinux, line 33: Is this correct from a copy right perspective?

cornelinux (approved):

pi-manage
plettich: @cornelinux This is based on the [Reuse-Project](https://reuse.software/) for machine-readable license information.
Probably there should be a line for each author and year but i was lazy, that is why i kept changelog.

PR #3808, opened 2023-11-08
better error message for webauth token enroll fail

### LINKED ISSUE #3807: Better error message for users - enrolling the same webauth token twice

When you try to enroll the same token for webauth twice on the same account, different error is shown in Firefox and Chrome. The Chrome one is clear and easy to understand(for users) but the Firefox one is not. The Firefox message: 'The object is in an invalid state. / An attempt was made to use an object that is not, or is no longer, usable'.

Unified and easy to understand error message so that users know the most probable source of the error. Optimally with the possibility of translation. 
From what I understand this error is generated in browser / something browser related so this would basically mean to not display the error message from browser but replace it with some static one - so it is unified and can be translated.

**Additional idea**

When this happens the 'new' rejected token is not removed but remains in clientwait state. So for example when you try to enroll the same token 10 times, the 1st passes and the remaining 9 requests stay in clientwait state.
Maybe it wouln't be a bad idea to remove that request token when it is rejected in enrolling.

**Additional context**
Firefox error message:
Screenshot from 2023-11-08 10-09-00

Hanging enroll requests after multiple attempts to enroll the same token:
Screenshot from 2023-11-08 10-10-02

*If sensible use a checklist to check, which requirements have been covered by your implementation!*

* [x] error message change
Optional: 
* [ ] delete the token if possible

# Implementation

*Describe your implementation plans - what you are exactly going to implement. Use references/link to the existing code*
https://github.com/privacyidea/privacyidea/blob/a0336fc4559d93925ea8fe03abdf578035e2bf22/privacyidea/static/components/login/factories/webauthn.js#L58C37-L58C37

On line 58 change e.message with something like: 'Error when enrolling - you are probably trying to enroll the same token twice.'

this would replace the 'e.message':
~~~~Python
gettextCatalog.getString("Error when enrolling - you are probably trying to enroll the same token twice!")
~~~~

JiriProkop:
closes #3807

### DIFF
--- a/privacyidea/static/components/login/factories/webauthn.js
+++ b/privacyidea/static/components/login/factories/webauthn.js
@@ -55,7 +55,10 @@ angular
                             .then(callback)
                             .catch(function(e) {
                                 inform.add(
-                                    domExceptionErrorMessage[e.name] + " / " + e.message,
+                                    domExceptionErrorMessage[e.name] + " / " +
+                                    gettextCatalog.getString(
+                                        "Error when enrolling - you are probably trying to enroll the same token twice!"
+                                    ),
                                     {
                                         type: "danger",
                                         ttl: 10000


cornelinux:
We are not sure, if this error messages is really always due to enrolling twice.
We are too hesitant to add this specific error message. We will leave the issue #3807 open to be able to look into this further.

PR #3813, opened 2023-11-20
migrate token janitor to click

jona-samuel:
This is the first attempt of the new tokenjanitor

### DIFF
--- a/privacyidea/cli/__init__.py
+++ b/privacyidea/cli/__init__.py
@@ -1,0 +1,22 @@
+# SPDX-FileCopyrightText: (C) 2023 Paul Lettich <paul.lettich@netknights.it>
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+#
+# Info: https://privacyidea.org
+#
+# This code is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License
+# as published by the Free Software Foundation, either
+# version 3 of the License, or any later version.
+#
+# This code is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public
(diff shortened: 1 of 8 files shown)


plettich:
Please also add something like
```python
    ...
    scripts=["pi-manage"] + get_scripts("tools"),
    entry_points={
        'console_scripts': [
            'privacyidea-token-janitor = privacyidea.cli:pi_token_janitor',
            ]},
    extras_require={
    ...
```
to [`setup.py`](https://github.com/privacyidea/privacyidea/blob/master/setup.py) to make the command available in installations.

privacyidea/cli/privacyideatokenjanitor/findtokens.py
```diff
@@ -87,46 +86,38 @@
...
        export PYTHONIOENCODING=UTF-8
    Here is a good read about it:
    https://stackoverflow.com/questions/4545661/unicodedecodeerror-when-redirecting-to-file
 
-""".format("|".join(ALLOWED_ACTIONS))
+"""
+
+ALLOWED_ACTIONS = ["disable", "delete", "unassign", "mark", "export", "listuser", "tokenrealms"]
 
 app = create_app(config_name='production', silent=True)
```
plettich, line 113: we don't need to create the `app` here. This should be done by the FlaskGroup automatically.

privacyidea/cli/privacyideatokenjanitor/findtokens.py
```diff
@@ -41,34 +43,31 @@
...
 import sys
 from yaml import safe_dump as yaml_safe_dump
-from yaml import safe_load as yaml_safe_load
-
-
 
-__version__ = "0.1"
-
-ALLOWED_ACTIONS = ["disable", "delete", "unassign", "mark", "export", "listuser", "tokenrealms"]
+findtokens_cli = AppGroup("find", help="finds tokens and give the option to export, change or delete them.")
```
plettich, line 65: We can drop the `AppGroup` here and just add the `find` command in `main.py` to the cli. This gets rid of the double use of `find`.

plettich (changes requested):

plettich (approved):

PR #3827, opened 2023-11-30
Add node names to db

### LINKED ISSUE #3757: Add pi_node names to database
The list of the PI_NODES can be stored in the database.

Each privacyIDEA node holds his own node, e.g. read from pi.cfg. 
At startup the privacyIDEA server can write this node to the table

`pinodes` with columns 
* uuid, (Foreign key)
* name,
* lastseen.

The uuid needs to be generated during installation or setup and is stored in the pi.cfg or somewhere else in the file system.

If the PI_NODE name is changed, then the table will be updated based on the uuid.

TODO:
- [x] add database table and migration script
- [x] query node table for node names
- [x] adjust documentation (remove `PI_NODES`)

plettich:
Closes #3757


doc/installation/system/inifile.rst
```diff
@@ -226,13 +226,17 @@ privacyIDEA Nodes
 -----------------
 
 privacyIDEA can run in a redundant setup. For statistics and monitoring purposes you
-can give these different nodes, dedicated names.
+can give these different nodes dedicated names.
 
-``PI_NODE`` is a string with the name of this very node. ``PI_NODES`` is a list of
-all available nodes in the cluster.
+``PI_NODE`` is a string with the name of this very node. At the startup of
+the application, an installation specific unique ID will be used to tie the
```
cornelinux, line 232: For me it would be more clear, that it is the startup of "privacyIDEA"; not the application.

doc/installation/system/inifile.rst
```diff
@@ -226,13 +226,17 @@ privacyIDEA Nodes
...
+node name to an installation. A distinct ID can be given to an application with
```
cornelinux, line 233: "A distinct ID can be given".

passive sentences are unclear. Who and how?
It think this means: instead of autogenerating the ID, the administrator (you) can manually define an ID by using PI_UUID.

doc/installation/system/inifile.rst
```diff
@@ -226,13 +226,17 @@ privacyIDEA Nodes
...
+the ``PI_UUID`` value (should conform to
+`RFC 4122 <https://datatracker.ietf.org/doc/html/rfc4122.html>`_).
+The names of all nodes in a redundant setup will be available through the database.
```
cornelinux, line 236: While elaborating on PI_NODE; you totally dropped the description of PI_NODES!

doc/installation/system/inifile.rst
```diff
@@ -226,13 +226,17 @@ privacyIDEA Nodes
 -----------------
 
 privacyIDEA can run in a redundant setup. For statistics and monitoring purposes you
```
cornelinux, line 228: Not only for statistics and monitoring purposes, but please also add the purpose of node specific configuration.

cornelinux (changes requested):
Improve docs.

privacyidea/models.py
```diff
@@ -3267,3 +3267,12 @@ def save(self):
         else:
             ret = self.id
         return ret
+
+
+class NodeName(db.Model, TimestampMethodsMixin):
+    __tablename__ = "nodename"
+    # TODO: we can use the UUID type here when switching to SQLAlchemy 2.0
```
cornelinux, line 3274: Just think about my other comment about TODOs. Thanks.

doc/installation/system/inifile.rst
cornelinux: Ah, ok, this is only in the future pull requests. So at this stage this is actually correct.

privacyidea/models.py
```diff
@@ -3267,3 +3267,12 @@ def save(self):
...
+    #  <https://docs.sqlalchemy.org/en/20/core/custom_types.html#backend-agnostic-guid-type>
+    id = db.Column(db.Unicode(36), primary_key=True)
+    name = db.Column(db.Unicode(100), index=True)
+    lastseen = db.Column(db.DateTime(False), index=True, default=datetime.now(tz=tzutc()))
```
cornelinux, line 3278: In other occasions we have

    timestamp = db.Column(db.DateTime, default=datetime.utcnow(), index=True)

Why does this have to be different?

doc/installation/system/inifile.rst
cornelinux: Well you see. I stepped into this trap.
So maybe we should add a sentence like

"In preveious versions you could use the key PI_NODES; which is not necessary anymore."

We might also need to add a note about decommissioning a node.
If one node is removed from a redundant setup. Looks like the admin needs to delete the entry from the database?

Also: A setup might start with "localnode" and the name of the node could be changed later. Hm, I will see if it is updated via the uuid --- most probably ;-)

privacyidea/lib/config.py
```diff
@@ -997,13 +998,19 @@ def get_privacyidea_node(default='localnode'):
...
-    own_node_name = get_privacyidea_node()
-    nodes = get_app_config_value("PI_NODES", [])[:]
-    if own_node_name not in nodes:
-        nodes.append(own_node_name)
-    return nodes
+    node_names = []
+    nodes = db.session.query(NodeName).all()
+    for node in nodes:
+        node_names.append(node.name)
+    # if there are no nodes in the database we add at least the local node to the list
```
cornelinux, line 1009: Should the app.py have written the local nodename to the database anyways?

privacyidea/app.py
```diff
@@ -218,18 +232,53 @@ def create_app(config_name="development",
...
+            except Exception as e:  # pragma: no cover
+                logging.getLogger(__name__).info(f"Could not determine the machine id: {e}")
+            else:
+                # we generate a random UUID which will change on every startup
+                pi_uuid = uuid.uuid4()
+                logging.getLogger(__name__).warning("Generating a random UUID! "
+                                                    "If persisting the UUID fails, "
+                                                    "it will change on every application start")
+            app.config["PI_UUID"] = str(pi_uuid)
+            # safe UUID to config file
```
cornelinux, line 264: "save UUID to config file"

privacyidea/app.py
```diff
@@ -218,18 +232,53 @@ def create_app(config_name="development",
...
+            try:
+                with open(config_file, 'a') as f:  # pragma: no cover
+                    f.write(f"PI_UUID = \"{str(pi_uuid)}\"\n")
```
cornelinux, line 267: It totally makes sense to save the uuid to a file, so that we do not need to run the maybe time consuming getting-uuid-from-machine-id at each app start.

However, I most probably would not recommend to make the config_file writeable. 
Could we save the uuid to a special uuid file? So that the privacyIDEA application only needs write access to this file?

privacyidea/app.py
```diff
@@ -218,18 +232,53 @@ def create_app(config_name="development",
...
+            except IOError as exx:
+                logging.getLogger(__name__).warning(f"Could not add UUID to config "
+                                                    f" file '{config_file}': {exx}")
+
+        pi_node_name = app.config.get("PI_NODE") or app.config.get("PI_AUDIT_SERVERNAME", "localnode")
+        insp = sa.inspect(db.get_engine())
```
cornelinux, line 273: Ah, I understand we need to do this, since the ORM is not ready at this stage, right?

cornelinux (changes requested):
I like it.

However, I have some comments about the docs and also about updating the config file.

privacyidea/app.py
cornelinux: if it is only for non-run migrations, please remove this.

doc/installation/system/inifile.rst
plettich: Yeah, these are the tricky parts...
1. We do not now if a node becomes/is offline, we don't have any heartbeat functionality. The only possible solution would be to discard entries based on the timestamp. However the timestamp only gets updated when privacyIDEA is (re-)started, not during every request (i would like to avoid that overhead).
2. Currently it is possible to have nodes with the same node name (not a unique requirement in the db). We could log a warning and/or attach part of the UUID to the name if the same node name is detected. If it bothers the admin, they can change it.
3. We have no "deprecation" or "migration" of `PI_NODES` (since each node fills the table themselves, i don't see any possible migration). We could check for the configuration entry during startup and log a warning.
4. In the future we could add some management functionality for the nodes which could also check for connectivity of the nodes on demand.

privacyidea/models.py
plettich: See https://blog.miguelgrinberg.com/post/it-s-time-for-a-change-datetime-utcnow-is-now-deprecated

But as i was reading the article further, we should probably stick to using naive UTC timestamps with the type decorator shown here: https://docs.sqlalchemy.org/en/20/core/custom_types.html#store-timezone-aware-timestamps-as-timezone-naive-utc

privacyidea/models.py
plettich: I do think that manually converting the UUID from/to strings is a hacky workaround and should be revised when switching to SQLAlchemy 2.0

cornelinux:
Why does it take over one month to close such an issue.
I am closing this PR since I am totally fed up with looking at non-done things.

doc/installation/system/inifile.rst
plettich: Done in c0c5f2d5e

doc/installation/system/inifile.rst
plettich: Done in c0c5f2d5e

privacyidea/models.py
plettich: Done in c0c5f2d5e

privacyidea/lib/config.py
plettich: Yes, it should have. Done in c0c5f2d5e

privacyidea/app.py
plettich: Done in c0c5f2d5e

privacyidea/app.py
plettich: Since the `machine-id` is just a file, there shouldn't be much overhead. Anyway i have added the `PI_UUID_FILE` configuration which defaults to `/etc/privacyidea/uuid.txt`, where the UUID is saved when a new one is generated.
Done in c0c5f2d5e

PR #3828, opened 2023-12-01
Enable enrollment verification for TAN lists

### LINKED ISSUE #3809: Verify enroll is generating a second TAN list
privacyidea 3.8.1
ubuntu 20
Customer: 3858

If we roll out a TAN or a PPR list via the PI GUI, the following happens:

- A TAN list is displayed including the dialog box for entering the verification TAN.
- I save the TAN list and enter the first TAN in the dialog box.
- The TAN is accepted, the token is also successfully set to: enrolled_state = enrolled.
- 
- BUT: now I am shown a completely new TAN list that does not match the first.
This new TAN list is then also the valid list that I have to use for authentication.

BBA2

plettich:
Fixes #3809

### DIFF
--- a/privacyidea/lib/tokens/papertoken.py
+++ b/privacyidea/lib/tokens/papertoken.py
@@ -132,11 +132,15 @@ class PaperTokenClass(HotpTokenClass):
         return ret
 
     def update(self, param, reset_failcount=True):
-        if "otpkey" not in param:
+        if "otpkey" not in param and "verify" not in param:
             param["genkey"] = 1
         HotpTokenClass.update(self, param, reset_failcount=reset_failcount)
         papertoken_count = int(param.get("papertoken_count", DEFAULT_COUNT))
         # Now we calculate all the OTP values and add them to the
-        # init_details. Thus they will be returned by token/init.
-        otps = self.get_multi_otp(count=papertoken_count)
-        self.add_init_details("otps", otps[2].get("otp", {}))
+        # init_details. Thus, they will be returned by token/init.
+        # But only if this is not in verify state.
+        # TODO: check if can finish the request after verifying the token in the
+        #  prepolicy. Thus we would not need to go through all the token updates
+        if "verify" not in param:
+            otps = self.get_multi_otp(count=papertoken_count)
+            self.add_init_details("otps", otps[2].get("otp", {}))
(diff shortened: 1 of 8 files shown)


privacyidea/static/components/token/views/token.enrolled.paper.html
```diff
@@ -56,10 +56,10 @@
                 </div>
             </div>
         </uib-accordion>
+        <button class="btn-default btn"
```
cornelinux, line 59: Why are you moving the button to print in the closed accordeon?
This way users will not see, that they can print the list.

If there is no necessary reason, please move it back.

privacyidea/static/components/token/views/token.enrolled.tan.html
```diff
@@ -56,10 +56,10 @@
                 </div>
             </div>
         </uib-accordion>
+        <button class="btn-default btn"
```
cornelinux, line 59: See above.
Please move the button back, outside of the accordeon.

privacyidea/lib/tokens/papertoken.py
```diff
@@ -132,11 +132,15 @@ def get_class_info(key=None, ret='all'):
...
             param["genkey"] = 1
         HotpTokenClass.update(self, param, reset_failcount=reset_failcount)
         papertoken_count = int(param.get("papertoken_count", DEFAULT_COUNT))
         # Now we calculate all the OTP values and add them to the
-        # init_details. Thus they will be returned by token/init.
-        otps = self.get_multi_otp(count=papertoken_count)
-        self.add_init_details("otps", otps[2].get("otp", {}))
+        # init_details. Thus, they will be returned by token/init.
+        # But only if this is not in verify state.
+        # TODO: check if can finish the request after verifying the token in the
```
cornelinux, line 142: Just one note: You are adding todos in the code a lot for aspects, that might be nice one day. I would not add todos for things, which we most probably will not/never do. ;-)

cornelinux (changes requested):
Looks good.
But please check, if the button for printing OTPs might need to be moved back to the original position to avoid the button being hidden.

privacyidea/static/components/token/views/token.enrolled.paper.html
plettich: The button is outside the accordion and still visible if the accordion is closed.
I just moved the button inside the `div` with the attribute `ng-show="otp_rows` to disable showing the accordion and button if there are no TAN values.

privacyidea/lib/tokens/papertoken.py
plettich: Yeah, i kind of like TODOs :-)
I think for notes that something is missing (like specific tests) this can be useful to have it in the code.

cornelinux (approved):

PR #3842, opened 2024-01-09
Add node names to db

### LINKED ISSUE #3757: Add pi_node names to database

plettich:
This PR is based on #3827 

See the review notes there for more information.

Closes #3757


tests/base.py
```diff
@@ -65,9 +68,7 @@ def setUpClass(cls):
         cls.app = create_app('testing', "")
         cls.app_context = cls.app.app_context()
         cls.app_context.push()
-        db.create_all()
```
cornelinux, line 68: Don't we need to create tables anmore?

plettich: I moved it into the `create_app()` factory in case the configuration is `"testing"`:
https://github.com/privacyidea/privacyidea/blob/c0c5f2d5eaab75880d05ba1b1716afbcc41f80c5/privacyidea/app.py#L243
But since i have re-introduced the check for the existence of the table, i will test if this is still necessary.

plettich: I tested it and it s not necessary anymore, i'll revert the changes.

plettich: Done in 5c5d6fd79

privacyidea/app.py
```diff
@@ -218,18 +234,62 @@ def create_app(config_name="development",
...
 
-    if init_hsm:
+    if initialize_hsm:
         with app.app_context():
             init_hsm()
 
+    # check that we have a correct node_name -> UUID relation
+    with app.app_context():
+        # first check if we have a UUID in the config file which takes precedence
+        pi_uuid = app.config.get("PI_UUID")
```
cornelinux, line 246: If an admin enters a UUID in the config file, we read this without logic check. When reading from the PI_UUID_FILE you do a `uuid.UUID()`.

Would it make sense to also add this here:

    pi_uuid = uuid.UUID(app.config.get("PI_UUID"))

otherwise an admin could enter an non-uuid string in pi.cfg and it would be written to the database.

privacyidea/app.py
```diff
@@ -218,18 +234,62 @@ def create_app(config_name="development",
...
+                                                  f"from file '{pi_uuid_file}': {e}")
+
+                # we try to get the unique installation id (See <https://0pointer.de/blog/projects/ids.html>)
+                try:
+                    with open("/etc/machine-id", 'r') as f:
+                        pi_uuid = uuid.UUID(f.read().strip())
+                except Exception as e:  # pragma: no cover
+                    logging.getLogger(__name__).debug(f"Could not determine the machine "
+                                                      f"id: {e}")
+                    # we generate a random UUID which will change on every startup
```
cornelinux, line 264: ...unless it is persisted to the pi_uuid_file in line 271.

cornelinux (commented):
Looks good to me.
Please check my minor comments, if they make sense to you.

privacyidea/app.py
plettich: correct, will do

privacyidea/app.py
plettich: will do

privacyidea/app.py
plettich: Done in 4dc04918e

privacyidea/app.py
plettich: Done in 4dc04918e

cornelinux (approved):

PR #3850, opened 2024-01-23
add rpid to returned offline data

nilsbehlen:


nilsbehlen:
the naming of the returned values is very inconsistent with `pubkey`, `credential_id` and `rpId`.
For reference, when triggering a webauthn challenge, it is: `id` (for the credential id) and `rpId`, which is why i chose `rpId`.

privacyidea/lib/applications/offline.py
```diff
@@ -171,7 +171,9 @@ def get_authentication_item(token_type,
             if token_type.lower() == "webauthn":
                 # return the pubkey and the credential_id (contained in the otpkey)
                 ret["response"] = {"pubkey": token_obj.get_tokeninfo("pubKey"),
-                                   "credential_id": token_obj.decrypt_otpkey()}
+                                   "credential_id": token_obj.decrypt_otpkey(),
+                                   "rpId": token_obj.get_tokeninfo("relying_party_id")}
```
plettich, line 175: why not call it `rp_id` to be more consistent with `credential_id`?

nilsbehlen: see my comment in the pr

plettich: Ok, i didn't get the comment at first :-)

plettich (approved):

Issue #3758, opened 2023-09-13
node specific realms


cornelinux:
**Is your feature request related to a problem? Please describe.**

In redundant setups with node in different geo-locations it can be useful to have slightly different configurations.
E.g. the priority of resolvers in a realm could be in a different order for different nodes.

We should find a way to allow some kind of node specific  configuration.

# Requirements

There are different scenarios that could result in the same requirement.

**Req1)** (necessary)

The different nodes use the same AD but should use different domain controllers due to the geo location.

"REALM A" on node 1 (Geolocation A):  Res A (Geolocation A), Res B (Geoloation B)

"REALM A" on node 2 (Geolocation B): Res B (Geolocation B), Res A (Geolocation A)

This would simply result in a different ordering. This means, there is one Realm that contains all users everywhere. Only users usually located in Geolocation A are resolved quicker in Geolocation A - users from Geoloc B will still be resolved in Geoloc A, but simply a bit slower.

This case works, if a user usually authenticates in his default Geoloc.

**Req2)**  (This is a generalization of Req1)

The different nodes could use different users, which originate even maybe from different LDAPs.

"REALM A" on node 1: Res A, Res B

"REALM A" on node2: Res B, Res C, Res D
This could also be used, e.g. if you have an Admin-Realm, to define the Admin-Realm on a DMZ node to be empty (without any resolvers), thus again preventing admins from login.

# Implementation thoughts

## Questions
- should every *new* resolver->realm relation have a node attached?

### Database

For Req1 and Req2 we could add a node to the ResolverRealm definition like this:

~~~~Python
class ResolverRealm(TimestampMethodsMixin, db.Model):
    """
    This table stores which Resolver is located in which realm
    This is a N:M relation
    """
    __tablename__ = 'resolverrealm'
    id = db.Column(db.Integer, Sequence("resolverrealm_seq"), primary_key=True)
    resolver_id = db.Column(db.Integer, db.ForeignKey("resolver.id"))
    realm_id = db.Column(db.Integer, db.ForeignKey("realm.id"))
    # If there are several resolvers in a realm, the priority is used the
    # find a user first in a resolver with a higher priority (i.e. lower number)
    priority = db.Column(db.Integer)
    node_uuid = db.Columd(db.UUID, db.ForeignKey("pinode.uuid"))
    resolver = db.relationship(Resolver,
                               lazy="joined",
                               back_populates="realm_list")
    realm = db.relationship(Realm,
                            lazy="joined",
[...]

PR #3851, opened 2024-01-25
add node specific realms on library level

plettich:
Working on #3758


privacyidea/lib/realm.py
```diff
@@ -175,61 +177,57 @@ def delete_realm(realmname):
...
     It takes a list of resolvers and adds these to the realm.
     If the realm does not exist, it is created.
     If the realm exists, the old resolvers are removed and the new ones
     are added.
 
-    :param realm: an existing or a new realm
-    :param resolvers: names of resolvers
+    :param realm: name of an existing or a new realm
+    :type realm: str
+    :param resolvers: list with names and options of resolvers
```
cornelinux, line 189: We should define in the docstring, how the resolvers list dict looks like

[ {'name': <resolvername1>,   
   'node': <uuid of the node/optional>,
    'priority': <priority of the resolver in the realm>},
  {'name': <resolvername2>,
  }
]

cornelinux (changes requested):
Small fixos

privacyidea/lib/realm.py
cornelinux: Done in 120dba6c3

migrations/versions/d0e7144947d0_.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('nodename', 'id',
+                    existing_type=sa.VARCHAR(length=40),
+                    type_=sa.Unicode(length=36),
+                    existing_nullable=False)
+    op.add_column('resolverrealm', sa.Column('node_uuid', sa.Unicode(length=36), nullable=True))
+    op.create_foreign_key(None, 'resolverrealm', 'nodename', ['node_uuid'], ['id'])
```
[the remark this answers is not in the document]
cornelinux: Done in 8b9cb785e

cornelinux (approved):

PR #3864, opened 2024-02-16
use same convention for names, typos

nilsbehlen:
Use CamelCase for webauthn related field names, as it already done when triggering webauthn challenges, eventhough the other naming uses underscores.

### DIFF
--- a/tests/test_api_validate.py
+++ b/tests/test_api_validate.py
@@ -6137,9 +6137,9 @@ class WebAuthnOfflineTestCase(MyApiTestCase):
                  'username': 'cornelius',
                  'refilltoken': '79906cc20567c6ca1e4b452500bb0662e107ce0fa14742c95b7e5c6a1417a519f829318622c1d569',
                  'repsonse': {
-                     'pubKey': 'a50102032620012158203f98500ddcedc3aa16d34ae9b7c12...ea3e37193f8f3d', 
+                     'pubKey': 'a50102032620012158203f98500ddcedc3aa16d34ae9b7c12...ea3e37193f8f3d',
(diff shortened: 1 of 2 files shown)


tests/test_api_validate.py
```diff
@@ -6137,14 +6137,14 @@ def test_02_autenticate(self):
                  'username': 'cornelius',
                  'refilltoken': '79906cc20567c6ca1e4b452500bb0662e107ce0fa14742c95b7e5c6a1417a519f829318622c1d569',
                  'repsonse': {
-                     'pubkey': 'a50102032620012158203f98500ddcedc3aa16d34ae9b7c12...ea3e37193f8f3d', 
-                     'credential_id': 'RuBlEInU7ycsILST7u6AoT7rdqN...4xM2SHltwbtBwApFYnbQXO8g5bgrb4kFh1NErnzsT6xA',
-                     'rpId': 'netknights.it'},
```
cornelinux, line 6142: You are changing the API here. 
This might break installations in the field, that update the privacyIDEA server but do not update clients.
I see no necessity to do this.

Also, please ensure to follow our development guideline:
https://github.com/privacyidea/privacyidea/blob/master/CONTRIBUTING.md#develop
I.e. please create an issue before adding/changing code.

cornelinux (changes requested):

tests/test_api_validate.py
nilsbehlen: This has not been released yet, so it does not break anything. It changes the names so that it uses the same convention as the already released API which uses camelcase for webauthn things:
 
grafik

cornelinux (approved):

PR #3887, opened 2024-03-25
migrate privacyidea-usercache-cleanup to click

jona-samuel:
Working on #2498


jona-samuel:
replaced by https://github.com/privacyidea/privacyidea/pull/3907

PR #3888, opened 2024-03-25
Add result->authentication to the event handler conditions

### LINKED ISSUE #3886: Additional Event Condition: Result->Authentication
Add a new condition for event handlers:

    result->authentication

This helps to distinguish between SUCCESS, FAILED, CHALLENGE

cornelinux:
Using the condition result->value was often misleading, since the result->value is false for failed authentications but also for the first step of the challenge response authentication. This way we add more flexibility in the conditions.

Closes #3886

### DIFF
--- a/privacyidea/lib/eventhandler/base.py
+++ b/privacyidea/lib/eventhandler/base.py
@@ -454,7 +454,7 @@ class BaseEventHandler(object):
         if CONDITION.RESULT_AUTHENTICATION in conditions:
             condition_value = conditions.get(CONDITION.RESULT_AUTHENTICATION)
             result_auth = content.get("result", {}).get("authentication")
-            if is_true(condition_value) != is_true(result_auth):
+            if condition_value != result_auth:
                 return False
 
         # checking of max-failcounter state of the token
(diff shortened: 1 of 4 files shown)


privacyidea/lib/eventhandler/base.py
```diff
@@ -444,6 +451,12 @@ def check_condition(self, options):
             if is_true(condition_value) != is_true(result_status):
                 return False
 
+        if CONDITION.RESULT_AUTHENTICATION in conditions:
+            condition_value = conditions.get(CONDITION.RESULT_AUTHENTICATION)
+            result_auth = content.get("result", {}).get("authentication")
+            if is_true(condition_value) != is_true(result_auth):
+                return False
```
nilsbehlen, line 458: can you add a testcase to cover this condition?

cornelinux: Oups, thank you for asking.
I think this is actually *wrong code*. Since the contents of result->authentication is not kind of true, but it is one of the four defined strings. So I actually have to fix the code.

cornelinux: fixed it in 9c4e630f3

nilsbehlen (approved):

PR #3889, opened 2024-03-25
Add tokenlabel and tokenissuer to PUSH enrollment

### LINKED ISSUE #3883: Label policies for PUSH token enroll via validate/check
When enrolling via validate/check the token_label etc policies are not applied to the PUSH token.

It works for HOTP and TOTP, since these policies are added here:

https://github.com/privacyidea/privacyidea/blob/662774047d37fdfe335fcf523892eff11ca5f881/privacyidea/lib/tokens/hotptoken.py#L839

Thus, we need to add `get_init_tokenlabel_parameters`  also to the `enroll_via_validate` method of the Push-Token:
https://github.com/privacyidea/privacyidea/blob/662774047d37fdfe335fcf523892eff11ca5f881/privacyidea/lib/tokens/pushtoken.py#L1056

cornelinux:
When enrolling PUSH tokens via_multichallenge the tokenlabel and tokenissuer were not processed. Fixed this.

Closes #3883

### DIFF
--- a/tests/test_api_push_validate.py
+++ b/tests/test_api_push_validate.py
@@ -380,6 +380,8 @@ class PushAPITestCase(MyApiTestCase):
         # 1. set policies.
         set_policy("pol_passthru", scope=SCOPE.AUTH, action=ACTION.PASSTHRU)
 
+        set_policy("pol_tokenlabel", scope=SCOPE.ENROLL, action="{0!s}=Pushy".format(ACTION.TOKENLABEL))
+
         # 2. authenticate user via passthru
         with self.app.test_request_context('/validate/check',
                                            method='POST',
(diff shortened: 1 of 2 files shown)


tests/test_api_push_validate.py
```diff
@@ -380,6 +380,8 @@ def test_10_enroll_push(self):
         # 1. set policies.
         set_policy("pol_passthru", scope=SCOPE.AUTH, action=ACTION.PASSTHRU)
 
+        set_policy("pol_tokenlabel", scope=SCOPE.ENROLL, action="{0!s}=Pushy".format(ACTION.TOKENLABEL))
```
nilsbehlen, line 383: i don't see an assert to verify the modified label, am i missing something?

cornelinux: I do not see a way to verify this besides debugging.

nilsbehlen (approved):

PR #3890, opened 2024-03-26
Add policy for alternative enrollment text

### LINKED ISSUE #3884: Allow to change the text during multichallenge enrollment
When enrolling a token via validate/check it should be possible to set the Text.
The text is currently hard coded

for HOTP and TOTP:

https://github.com/privacyidea/privacyidea/blob/662774047d37fdfe335fcf523892eff11ca5f881/privacyidea/lib/tokens/hotptoken.py#L829

For Push:

https://github.com/privacyidea/privacyidea/blob/662774047d37fdfe335fcf523892eff11ca5f881/privacyidea/lib/tokens/pushtoken.py#L1078

And for SMS / Email:

https://github.com/privacyidea/privacyidea/blob/662774047d37fdfe335fcf523892eff11ca5f881/privacyidea/lib/tokens/emailtoken.py#L546
https://github.com/privacyidea/privacyidea/blob/662774047d37fdfe335fcf523892eff11ca5f881/privacyidea/lib/tokens/smstoken.py#L603

We should have an additional policy to set the text:

* scope: **auth ** (We put it in scope auth, since the "enroll_via_multichallenge" is also there).
* action: enroll_via_multichallenge_text

cornelinux:
Add a policy that allows to set the text during
the enrollment via multichallenge.
This way the admin can give individual information to the user.

Closes #3884

### DIFF
--- a/doc/policies/authentication.rst
+++ b/doc/policies/authentication.rst
@@ -421,8 +421,8 @@ there is no additional authentication step anymore during enrollment.
 
 .. _policy_enroll_via_multichallenge_text:
 
-enroll_via_multichallenge
-~~~~~~~~~~~~~~~~~~~~~~~~~
+enroll_via_multichallenge_text
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 type: string
 
--- a/doc/policies/authentication.rst
+++ b/doc/policies/authentication.rst
@@ -374,7 +374,7 @@ type: string
 
 This policy allows the rollout of tokens during the authentication via `/validate/check`.
 
-The policy action can take one of the followig token types: `hotp`, `totp`, `push`, `email`, `sms`.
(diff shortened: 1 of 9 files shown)


doc/policies/authentication.rst
```diff
@@ -419,6 +419,17 @@ there is no additional authentication step anymore during enrollment.
    enroll only one token type.
 
 
+.. _policy_enroll_via_multichallenge_text:
+
+enroll_via_multichallenge
```
plettich, line 424: this should probably be `enroll_via_multichallenge_text`

plettich: Fixed in 788c40c

nilsbehlen (approved):

doc/policies/authentication.rst
cornelinux: Thanks for the "allow_white_spaces".

PR #3892, opened 2024-03-27
Fix broken Galera-Cluster installation

### LINKED ISSUE #3863: Error when installing privacyIDEA 3.9.2 with an MariaDB Galera Cluster
### Top-level intent

I've tried installing privacyIDEA 3.9.2, which comes with SQLAlchemy 1.4, with a MariaDB Galera cluster as a token database. An error, which I've described below, occurred and the installation did not succeed.

### Steps to reproduce

1. Set up a database with the current long-term versions MariaDB 10.11 and Galera 26.4.
2. Install privacyIDEA 3.9.2 from PyPI on a fresh Ubuntu 22.04 server.
3. Run the command "pi-manage create_tables" inside the virtual environment.

### Expected outcome

The tables should be initialized.

### Actual outcome

You'll get the error message from the "Error" section and the table creation will fail.

### Further scenarios

I've also tested the following additional scenarios during error handling:

privacyIDEA 3.8.1. + MariaDB Single Node: Succeeded
privacyIDEA 3.8.1. + MariaDB Galera Cluster: Succeeded

privacyIDEA 3.9.2. + MariaDB Single Node: Succeeded
privacyIDEA 3.9.2. + MariaDB Galera Cluster: Failed

### Configuration

* **privacyIDEA version**: 3.9.2
* **Python version**: 3.10
* **Operating system**: Ubuntu 22.04
* **Webserver**: Apache HTTP
* **Token database**: MariaDB Galera Cluster

### Error

pymysql.err.NotSupportedError:(1235, "This version of MariaDB doesn't yet support 'CACHE without INCREMENT BY 0 in Galera Cluster'")

### Assumption

It appears that SQLAlchemy 1.4 reports that the database supports sequences because of the "mariadb" dialect. This is true in theory, but Galera does not yet support this. It seems as if the "dialect_supports_sequences()" function is missing a check as to whether replication with Galera is being used.

Here's a search for the function I mentioned:
https://github.com/search?q=repo%3Aprivacyidea%2Fprivacyidea%20dialect_supports_sequences&type=code

At this point in the code, it appears that this has already been thought of:
https://github.com/privacyidea/privacyidea/blob/9d8b5e5fff5bfe47f5e0b15c465802126220bc5d/migrations/versions/5cb310101a1f_.py#L29

plettich:
Galera cluster based on MariaDB >= 10.6 was unable to create the initial tables since it needed sequences with an increment of 0. Unfortunately PostgreSQL does not work with an increment of 0 in sequences.
Compiling the SQL statements for creating a sequence differently for MySQL/MariaDB and PostgreSQL fixes this.
Database upgrades to this version still work as expected since the increment was already used in the migration script (now removed).

Fixes #3863

### DIFF
--- a/migrations/versions/5cb310101a1f_.py
+++ b/migrations/versions/5cb310101a1f_.py
@@ -26,12 +26,6 @@ Session = sessionmaker()
 def upgrade():
     migration_context = context.get_context()
     if migration_context.dialect.supports_sequences:
-        if migration_context.dialect.name in ['mariadb', 'mysql']:
-            # setting "increment" to 0 works like auto-increment but also supports replication
-            # See <https://mariadb.com/kb/en/sequence-overview/#replication>
-            inc_value = 0
-        else:
-            inc_value = 1
         bind = op.get_bind()
         # We only need a read session, so we do not need a commit
         session = Session(bind=bind)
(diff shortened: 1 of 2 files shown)


plettich:
wrong branch, sorry

PR #3894, opened 2024-03-27
3856/more audit columns

cornelinux:


privacyidea/api/ttype.py
```diff
@@ -74,7 +74,8 @@ def before_request():
     g.audit_object.log({"success": False,
                         "action_detail": "",
                         "client": g.client_ip,
-                        "client_user_agent": request.user_agent.browser,
+                        "user_agent": request.user_agent.browser,
```
nilsbehlen, line 77: is it the browser that we want? According to this
https://tedboy.github.io/flask/generated/generated/werkzeug.UserAgent.html
there is a list of recognized strings, which would not be what we want, since we probably need the user agents of our plugins?

cornelinux: Please check, what you like better.

One line less code but one more function call or only one function call but one more line of code.
check 9153e57f8

nilsbehlen (approved):

PR #3903, opened 2024-04-10
Fix policy description

### LINKED ISSUE #3895: Fix policy description
- [x] It looks like the policy description is a required field.
  I would not put it this way.
  image
- [ ] Also the layout is off.
- [x] The text "please read the ..." is wrong.
- [x] The description could be a text field (multiple lines), not a line edit (one line)
- [x] The first line of the migration scripts needs the `3.10` version info:
  https://github.com/privacyidea/privacyidea/blob/4f38e5e34eacbbc34e272d5eef358863ebbb53ca/migrations/versions/5741e5dac477_.py#L1

jona-samuel:
closes #3895


plettich:
@jona-samuel please take a look at the changes i made.

jona-samuel:
Lucks good

plettich (approved):

PR #3924, opened 2024-04-30
Fix wrong otplen of indexed secret

### LINKED ISSUE #3869: Verify enrollment does not work with indexed secret token
customer care: 4114

Like in #3809 the indexed secret is created new when using the "verify enrollment" policy.
Enforcing a user-attribute as the secret with the user policy `indexedsecret_force_attribute` should prevent this. This won't work for user-provided secrets.

We should add integration tests for all tokens with `can_verify_enrollment = True` if the verification **and** a following `/validate/check` succeeds.

The specific observation is, that the indexed secret token that is created via verify enroll has a different length than the requested positions: Like: During enrollment the user enters 10 characters but the length is 40.

Could this be sth with the hexlify and encoding?

cornelinux:
During verify enroll of the indexed secret token the otpkey gets regenerated in the second step.

Closes #3869

### DIFF
--- a/privacyidea/lib/tokens/indexedsecrettoken.py
+++ b/privacyidea/lib/tokens/indexedsecrettoken.py
@@ -175,9 +175,6 @@ class IndexedSecretTokenClass(TokenClass):
 
         TokenClass.update(self, param, reset_failcount)
 
-        # finally we set the otplen of the token, no matter if the otpkey was generated or not
-        key = self.token.get_otpkey().getKey()
-        self.set_otplen(len(key))
         return
 
     @log_with(log)
@@ -295,7 +292,7 @@ class IndexedSecretTokenClass(TokenClass):
                             # increase the received_count
                             challengeobject.set_otp_status()
                     else:
-                        log.debug("Length of password does not match the requested number of positions.")
+                        log.info("Length of password does not match the requested number of positions.")
                         # increase the received_count
                         challengeobject.set_otp_status()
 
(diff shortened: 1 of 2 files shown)


privacyidea/lib/tokens/indexedsecrettoken.py
```diff
@@ -170,10 +170,14 @@ def update(self, param, reset_failcount=True):
         :return: nothing
 
         """
-        if 'genkey' not in param and 'otpkey' not in param:
+        if 'verify' not in param and 'genkey' not in param and 'otpkey' not in param:
             param['genkey'] = 1
 
         TokenClass.update(self, param, reset_failcount)
+
+        # finally we set the otplen of the token, no matter if the otpkey was generated or not
```
plettich, line 178: why do we need to set the `otplen`? Technically the `otplen` for the indexed-secret token is defined in the policy `indexedsecret_count` or defaults to 2.

plettich (changes requested):

privacyidea/lib/tokens/indexedsecrettoken.py
cornelinux: You are right. I was mislead and confused by the logs from the ticket.

I confused the otplen (the count of characters input by the user) and the length of the secret.
I need to rethink this.

cornelinux: Done in 0de5ec9e2

PR #3929, opened 2024-05-06
Check Yubikey OTP length before validating

### LINKED ISSUE #3746: Check Yubikey OTP length before validating
According to the docs the Yubikey OTP is always 44 chars long (https://developers.yubico.com/OTP/OTPs_Explained.html).
We can check the length of the given OTP before  doing any further validation (this is also true for HOTP/TOTP/Email/SMS token).
That way we can identify possible broken input before getting cryptic errors like "CRC checksum failed".

jona-samuel:
closes #3746

### DIFF
--- a/privacyidea/lib/decorators.py
+++ b/privacyidea/lib/decorators.py
@@ -48,22 +48,21 @@ def check_token_locked(func):
 
 def check_token_otp_length(func):
     """
-    Decorator to check if a token otp has the correct length.
+    Decorator to check if a given otp value has the correct length.
     The decorator is to be used in token class method check_otp.
 
-    If the otp has a wrong length it returns -1 and logs the info.
+    If the given otp value has a wrong length it returns -1 and logs the info.
     """
     @functools.wraps(func)
     def token_otp_lenght_wrapper(*args, **kwds):
         token = args[0]
-        tokentype = token.type
         if len(args[1]) != token.token.otplen:
-            log.info(f"{tokentype} OPT value has a wrong length")
+            log.info(f"OTP value for token {token.token.serial} (type: {token.type}) "
+                     f"has wrong length ({len(args[1])} != {token.token.otplen})")
             return -1
         f_result = func(*args, **kwds)
         return f_result
     return token_otp_lenght_wrapper
-
 
 
 def check_user_or_serial(func):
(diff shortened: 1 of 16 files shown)


jona-samuel:
should we implement this for totp, hotp, SMS and email too?

privacyidea/lib/decorators.py
```diff
@@ -46,6 +46,26 @@ def token_locked_wrapper(*args, **kwds):
     return token_locked_wrapper
 
 
+def check_token_otp_lenght(func):
```
plettich, line 49: please correct to `check_token_otp_length()`

privacyidea/lib/tokens/emailtoken.py
```diff
@@ -328,6 +328,7 @@ def create_challenge(self, transactionid=None, options=None):
 
     @log_with(log)
     @check_token_locked
+    @check_token_otp_lenght
```
plettich, line 331: please move this decorator one line up. Decorators are applied from bottom to top (innermost first). We should first check if the token is locked and then if the otp length matches.

privacyidea/lib/tokens/hotptoken.py
```diff
@@ -400,6 +400,7 @@ def _calc_otp(self, counter):
 
     @log_with(log)
     @check_token_locked
+    @check_token_otp_lenght
```
plettich, line 403: see above

privacyidea/lib/tokens/smstoken.py
```diff
@@ -368,6 +368,7 @@ def create_challenge(self, transactionid=None, options=None):
 
     @log_with(log)
     @check_token_locked
+    @check_token_otp_lenght
```
plettich, line 371: and all others...

privacyidea/lib/tokens/yubikeytoken.py
```diff
@@ -280,6 +283,7 @@ def check_otp(self, anOtpVal, counter=None, window=None, options=None):
             otp_bin = modhex_decode(yubi_otp)
         except KeyError:
             # The OTP value is no yubikey aes otp value and can not be decoded
+            log.debug("OTP could not be decode")
```
plettich, line 286: please change to `could not be decoded`

plettich (changes requested):

plettich (approved):

PR #3935, opened 2024-05-08
Add email verification to enrollment

### LINKED ISSUE #3918: Enroll via multichallenge (EMAIL)
Please add the correctness cheking of email address format when PI asks "Please enter your new email addrees!" for @ sign and domain name presence. If the address is empty, a token should be created with the dynamic_email property.

cornelinux:
via valiadte

Working on #3918


nilsbehlen:
The same thing is happening with token when you do `get_token`, so if you are going to cache it, we might add it there as well.

nilsbehlen (approved):

PR #3939, opened 2024-05-14
Added directive to create or select a container

nilsbehlen:
* Added the directive to token enrollment and token info
* Updated some js code


privacyidea/api/container.py
```diff
@@ -140,7 +141,7 @@ def remove_token(container_serial):
     token = get_one_token(serial=serial)
     res = False
     if token:
-        container.remove_token(token)
+        count = container.remove_token(token.get_serial())
```
jelinaunger, line 144: Unused variable `count` could be removed.

nilsbehlen: 683a31b

PR #3941, opened 2024-05-15
Container list and container details

jelinaunger:
- Container list: added expandable view to show tokens of the containers and added sorting by container serial
- Initiated the container details view
- Added container to the user details view if a container is assigned

### DIFF
--- a/privacyidea/static/components/token/factories/container.js
+++ b/privacyidea/static/components/token/factories/container.js
@@ -1,4 +1,3 @@
-//angular.module("ContainerModule", ["privacyideaAuth"])
 myApp.factory("ContainerFactory", ['AuthFactory', '$http', 'containerUrl', '$q', '$state', '$rootScope',
     function (AuthFactory, $http, containerUrl, $q, $state, $rootScope) {
         let canceller = $q.defer();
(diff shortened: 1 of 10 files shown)


privacyidea/static/components/token/factories/container.js
```diff
@@ -1,19 +1,26 @@
+//angular.module("ContainerModule", ["privacyideaAuth"])
```
nilsbehlen, line 1: comment can be removed

privacyidea/static/components/token/factories/container.js
```diff
@@ -1,19 +1,26 @@
...
                 }).then(function (response) {
                     callback(response.data);
                 }, function (error) {
                     AuthFactory.authError(error.data);
                 });
             },
+            getContainerForSerial: function (serial, callback) {
+                $http.get(containerUrl + "/?serial=" + serial, {
+                    headers: {'PI-Authorization': AuthFactory.getAuthToken()}
+                }).then(function (response) { callback(response.data) }, function(error) { AuthFactory.authError(error.data) });
```
nilsbehlen, line 22: this should be formatted like the other functions

Issue #3966, opened 2024-06-04
User unassign button still visible


maxbeth:
### Top-level intent

Found a UI bug while i config the helpdesk policy

### Steps to reproduce

1.  create a policy and remove the right to unassign tokens
2.  set the flag to hide all buttons without rights

### Expected outcome

The token -> user unassign  button should not be visible

### Actual outcome

The button is still visible

### Configuration

* **privacyIDEA version**: 
* **Installation method**:  enterprise repository
* **Python version**:  Python 3.12.3
* **Operating system**: Ubuntu 
* **Webserver**:  Apache
* **Token database**:  Mysql

PR #3967, opened 2024-06-06
Hide the unassign user button

maxbeth:
solve the issue #3966


privacyidea/static/components/token/views/token.details.html
```diff
@@ -515,7 +515,7 @@ <h4 ng-show="token.user_id" translate>Assigned User</h4>
                     {{ token.username }}
                 </a></td>
                 <td>
-                    <button ng-hide="token.locked"
+                    <button ng-hide="token.locked || !checkRight('unassign')"
```
plettich, line 518: We are using the webui policy `hide_buttons` to, well, hide buttons...
Could You change this to:
```
      <button ng-hide="token.locked || !checkRight('unassign') && hide_buttons"
              ng-disabled="!checkRight('unassign')"
              ...
```
That way we can disable and (if the policy is set) hide the button.

plettich (changes requested):

privacyidea/static/components/token/views/token.details.html
maxbeth: OK, the changes are commited

plettich (approved):

PR #3973, opened 2024-06-12
Add Eventhandler for push decline

### LINKED ISSUE #3632: Push fatigue mitigation
Issue from the app with vague ideas: https://github.com/privacyidea/pi-authenticator/issues/289
Original Text:
If an attacker can trigger push notifications for a user, the static password is probably already compromised. In that case, it is not sufficient to just suppress the push notifications - the admin has to take action. However, it should be up to the admin what will be done.

General question:
How can we transfer the information of an unauthorized login attempt from the phone to privacyIDEA?

We could make the push buttons somewhat configurable:
- Optionally include a third button to indicate that the push request was not triggered by the owner of the phone -> what kind of request will be made?
- Texts

See: https://github.com/privacyidea/privacyidea/wiki/concept%3A-PushToken#decline
- We could also make use of "decline".

# Concept

We could use the "decline" information to allow the admin to trigger some notification.

A user receives a push notification. Why would he decline this message if it was not him, who triggered the login process. We could have a policy or an event handler, that the admin can set up to create such notification.

"Dear helpdesk team, please take care, since there are push notifications, that have beed declined by the user".

Maybe this could simple be a **condition** in an event handler?

We would then need to add an event definition at the endpoint `/ttype/<type>` here:
https://github.com/privacyidea/privacyidea/blob/9153e57f8830737661ca2287a1f8d28bf617652d/privacyidea/api/ttype.py#L87

We would also add more action at the declining of the challenge here:
https://github.com/privacyidea/privacyidea/blob/6e80ba271ab55bcd4f67f91207d8a7f8d8628d7a/privacyidea/lib/tokens/pushtoken.py#L641

The request in question would be

~~~
/ttype/push
serial
challenge
signature
decline=1
~~~
By the serial we would identify the token and thus the owner, e.g. to notifiy the owner or to deactivate the token.
conditions for endpoint /ttype/push would be: option decline=1.

## Todos

* [x] Add event decorator
* [x] Add token event handler action "increase tokeninfo"
* [x] Eventhandler needs additional condition:
        * ~~Either: for http parameter "decline=1" (could this done generic). Could work in *pre* and *post* event handling.~~
        * Alternative: Fitler in the *postevent* for the challenge state "declined", since this is written to the Challenge-Table. (Could also be a generic challenge state) Store it in the session.
* [x] Test: handler-action: increase_tokeninfo
* [x] Test: push-token: setting the challenge session to `challenge_declined`.
* [x] Test: Handler conditions `challenge_declined` and `expired`.
* [x] Docs

cornelinux:
The ttype endpoint gets an eventhandler decorator. The token handler now can increase tokeninfo.
Add docs for this.

Working on #3632

### DIFF
--- a/doc/eventhandler/index.rst
+++ b/doc/eventhandler/index.rst
@@ -368,6 +368,24 @@ value. Valid compares are::
 
 .. note:: A non-existing counter value will compare as 0 (zero).
 
+**challenge_session**
+
+The *challenge_session* condition can compare the value of the session attribute of
+a challenge against a regular expression. Usual values of the session are:
+
+*enrollment* during a multi challenge enrollment process and
+
+*challenge_declined* if the challenge of a PUSH token was declined by the user.
+
+This way the administrator can check for declined PUSH authentications and take
+according actions.
+
+**challenge_expired**
+
+This is a boolean check if the challenge has expired. Each challenge has an expiration
+date. If this exceeded this condition evaluates to *True*.
+
+
 Managing Events
 ---------------
 
--- a/doc/eventhandler/index.rst
+++ b/doc/eventhandler/index.rst
@@ -378,7 +378,9 @@ a challenge against a regular expression. Usual values of the session are:
 *challenge_declined* if the challenge of a PUSH token was declined by the user.
 
 This way the administrator can check for declined PUSH authentications and take
-according actions.
+according actions to implement PUSH fatigue mitigations like
+notifying the helpdesk (see :ref:`usernotification`),
+disabling the token or increasing a counter in the tokeninfo (see :ref:`tokenhandler`).
 
 **challenge_expired**
 
--- a/doc/eventhandler/index.rst
+++ b/doc/eventhandler/index.rst
@@ -377,7 +377,7 @@ a challenge against a regular expression. Usual values of the session are:
 
 *challenge_declined* if the challenge of a PUSH token was declined by the user.
 
-This way the administrator can check for declined PUSH authentications and take
+This way, the administrator can check for declined PUSH authentications and take
 according actions to implement PUSH fatigue mitigations like
 notifying the helpdesk (see :ref:`usernotification`),
 disabling the token or increasing a counter in the tokeninfo (see :ref:`tokenhandler`).
--- a/doc/eventhandler/tokenhandler.rst
+++ b/doc/eventhandler/tokenhandler.rst
@@ -171,9 +171,9 @@ and ``{ua_string}`` for information on the user agent and ``{username}`` and
 increase tokeninfo
 ...................
 
-Using the action ``increase tokeninfo`` you can increase the value of a tokeninfo key.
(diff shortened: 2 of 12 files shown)


cornelinux:
This should happen here:

https://github.com/privacyidea/privacyidea/blob/3608c7bf8623685c0e00596b87c0623dd1b81d4f/privacyidea/lib/tokens/hotptoken.py#L841

But it looks like that the decorator would add the corresponding parameters:

https://github.com/privacyidea/privacyidea/blob/662774047d37fdfe335fcf523892eff11ca5f881/privacyidea/api/token.py#L142

So actually the parameters are missing in the request.

# Todo

* Check if we can use the existing decorator with the `enroll_via_validate` method or endpoint.

doc/eventhandler/index.rst
```diff
@@ -368,6 +368,26 @@ value. Valid compares are::
...
+**challenge_session**
+
+The *challenge_session* condition can compare the value of the session attribute of
+a challenge against a regular expression. Usual values of the session are:
+
+*enrollment* during a multi challenge enrollment process and
+
+*challenge_declined* if the challenge of a PUSH token was declined by the user.
+
+This way the administrator can check for declined PUSH authentications and take
```
nilsbehlen, line 380: this way,

doc/eventhandler/tokenhandler.rst
```diff
@@ -168,6 +168,15 @@ and ``{ua_string}`` for information on the user agent and ``{username}`` and
 .. note:: You can use this to set the ``timeWindow`` of a TOTP token for
    :ref:`faq_initial_synchronization`.
 
+increase tokeninfo
+...................
+
+Using the action ``increase tokeninfo`` you can increase the value of a tokeninfo key.
```
nilsbehlen, line 174: , after tokeninfo

doc/eventhandler/tokenhandler.rst
```diff
@@ -168,6 +168,15 @@ and ``{ua_string}`` for information on the user agent and ``{username}`` and
...
+The tokeninfo is interpreted as an integer value.
+As an *increment* you can use a positive or a negative value. An increment like "-7" will
```
nilsbehlen, line 176: , after increment

nilsbehlen:
Is it possible to cover the missing lines hightlighted by the bot?

cornelinux:
> Is it possible to cover the missing lines hightlighted by the bot?

Achtually I think I covered them in the tests after the bot complained. I am not sure, why it is still there. Look at my 20 green check marks! ;-)

doc/eventhandler/tokenhandler.rst
cornelinux: So you think it is rule number 2?
https://east.iu.edu/student-success/coursework/commas.html

Will do.

cornelinux: I see, you already did.

nilsbehlen (approved):

PR #3980, opened 2024-06-19
fixing 500 if token description is longer than 80 characters

jona-samuel:


doc/webui/token_details.rst
```diff
@@ -71,7 +71,9 @@ You can change the following token settings.
 **TokenDesc**
 
    The token description is also displayed in the tokenview. You can
-   set a description to make it easier to identify a token.
+   set a description to make it easier to identify a token. The token
+   description has a maximum of 80 characters, all additional characters
+   will be cut.
```
plettich, line 76: I think `truncated` would fit better here instead of `cut`. Maye something like

> The token description will be truncated to fit into the database column (80 characters).

plettich (changes requested):

plettich (approved):

PR #3981, opened 2024-06-20
Container authorization

nilsbehlen:


jelinaunger (changes requested):
* policy for container realms is missing
* in container details the policy 'container_add_token' is not used

jelinaunger (approved):

PR #3987, opened 2024-06-25
Change AUTHTYPE.SASL_KERBEROS behaviour if upn is present in userinfo values for multidomain support

topelrapha:
We use privacyIdea with multiple ldap resolvers, want to use Kerberos authentication and had the problem that just one domain the default domain works fine. All users not being part of default domain got an error - and in log file it seems they were matched to default domain.

This fix shouldn't break the default behaviour but works with multiple ldap resolvers / multiple domains and kerberos authentication if following the next steps:

- Required Kerberos configuration incl. keytab files on Linux machine
- ADD "upn" : "userPrincipalName" (works if ldap is ActiveDirectory) to attribute values in ldap resolver configuration
- Switch bind type to "SASL Kerberos"

The fix takes care if there is a value in upn, modify this value to uppercase and use the result instead of just the username (without realm).

### DIFF
--- a/privacyidea/lib/resolvers/LDAPIdResolver.py
+++ b/privacyidea/lib/resolvers/LDAPIdResolver.py
@@ -1,6 +1,8 @@
 #  Copyright (C) 2014 Cornelius Kölbel
 #  contact:  corny@cornelinux.de
 #
+#  2024-06-25 Raphael Topel <raphael.topel@esh.essen.de>
+#             Change AUTHTYPE.SASL_KERBEROS behaviour if upn ist present in userinfo values for multidomain support
 #  2018-12-14 Cornelius Kölbel <cornelius.koelbel@netknights.it>
 #             Add censored password functionality
 #  2017-12-22 Cornelius Kölbel <cornelius.koelbel@netknights.it>
--- a/privacyidea/lib/resolvers/LDAPIdResolver.py
+++ b/privacyidea/lib/resolvers/LDAPIdResolver.py
@@ -344,7 +344,10 @@ class IdResolver (UserIdResolver):
                 return False
             # we need to check credentials with kerberos differently since we
             # can not use bind for every user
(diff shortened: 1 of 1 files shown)


privacyidea/lib/resolvers/LDAPIdResolver.py
```diff
@@ -1,6 +1,8 @@
 #  Copyright (C) 2014 Cornelius Kölbel
 #  contact:  corny@cornelinux.de
 #
+#  2024-06-25 Raphael Topel <raphael.topel@esh.essen.de>
+#             Change AUTHTYPE.SASL_KERBEROS behaviour if upn ist present in userinfo values for multidomain support
```
nilsbehlen, line 5: `ist` should be `is`

privacyidea/lib/resolvers/LDAPIdResolver.py
```diff
@@ -344,7 +346,12 @@ def checkPass(self, uid, password):
                 return False
             # we need to check credentials with kerberos differently since we
             # can not use bind for every user
-            name = gssapi.Name(self.getUserInfo(uid).get('username'))
+            if self.getUserInfo(uid).get('upn') is not None \
```
nilsbehlen, line 349: please assign `self.getUserInfo(uid).get('upn')` to a variable, so it is only done once and improves readability

nilsbehlen (changes requested):

plettich:
The Problem with this PR is, that it uses a user defined name for the UPN.
I think it is a good idea to generally add the UPN to the attribute list for an ldap resolver (if it is available) since it is more precise than just the `sAMAccountName`. Than we can use this attribute anywhere.

nilsbehlen:
We should probably add the upn to the standard mapping of the ldapresolver.

nilsbehlen (approved):

PR #3998, opened 2024-07-03
set_realm always sets the realm of the token owner

### LINKED ISSUE #3986: Remove token realm removes the user in the token details
Removing the realm of a token also removes the user on the token details page. However, the user is not unassigned. Trying to assign a new user causes an error. In the token list, the original user is still visible.

### Steps to reproduce

1. Enroll a token with a user assigned.
2. Remove the token realm.
3. Assign a user.

### Expected outcome

Either removing the token from the owners realm is disabled or the user is unassigned properly.

### Actual outcome

The user is unassigned from the token, but only on the token details page. Assigning a new user to the token results in an error: `ERR1103: Token already assigned to user`
In the token list the user is still visible for the token.

### Configuration

**privacyIDEA version**: *3.9.3*

jona-samuel:
closes #3986

### DIFF
--- a/privacyidea/models.py
+++ b/privacyidea/models.py
@@ -372,6 +372,9 @@ class Token(MethodsMixin, db.Model):
         # add new TokenRealms
         # We must not set the same realm more than once...
         # uniquify: realms -> set(realms)
+        if self.first_owner:
+            if self.first_owner.realm.name not in realms:
+                realms.append(self.first_owner.realm.name)
         for realm in set(realms):
             # Get the id of the realm to add
             r = Realm.query.filter_by(name=realm).first()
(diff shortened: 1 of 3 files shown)


privacyidea/models.py
```diff
@@ -372,6 +372,9 @@ def set_realms(self, realms, add=False):
         # add new TokenRealms
         # We must not set the same realm more than once...
         # uniquify: realms -> set(realms)
+        if self.first_owner:
+            if self.first_owner.realm.name not in realms:
+                realms.append(self.first_owner.realm.name)
```
plettich, line 377: We had a little talk about this and we should issue an info-log message here that the realm of the assigned user was to be deleted.

plettich (changes requested):

privacyidea/models.py
```diff
@@ -372,6 +372,10 @@ def set_realms(self, realms, add=False):
...
+                log.info("The realm of an assigned user cannot be removed.")
```
plettich, line 378: Please add some more information to the log message, at least the token serial.

plettich (changes requested):

plettich (approved):

Issue #1291, opened 2018-10-26
Devices as Token Container (formerly Nested Tokenclasses)


cornelinux:
### WebUI
- [x] Container creation page
  - [x] `assignUser` directive asks for PIN, make that configurable?
- [x] Add the owned containers to the user view
- [x] Extend `GET /token` to return the container the token is in and display that info in a new column in the token list
- [x] Create a directive that lets the user select or create a container. Usually, you dont want to create a container and then create a token, but rather create the container on the fly if it does not exist yet. This directive can then be used for:
  - [x] Token enrollment
  - [x] Token info
  - [x] User info page: put all token of user in container (or make a selection)
- [x] Create a container info page, similar to token info
- [x] Create a list view of all containers, similar to the token list. Ideally, the contained token are also directly visible and linked (e.g. by serial)
- [x] Make `GET /container` use pagesize like `GET /token`
- [x] Move the container sections to a different submenu
- [ ] Idea: Icons for each type in the container list behind the serial to see the type directly
- [x] Add `checkRights` to check for authorization policy on container actions

## Model
- [x] Add `last_seen` and/or `last_updated` to container database model. Set the value(s) on the corresponding operations, might require some form of the container saying "im still the same", to just the update the `last_seen`
- [x] Add `container_info` to containers, like `token_info`
- [x] Add a realm to the container, similar to token realm

## API
- [x] Add audit entries
- [x] Check that the authorization works similar to token

## Policies for Containers
- [x] Admin policies

## Everything else

- [x] Documentation
- [x] Tests

# Older content (2023-2024)
## Discussion

We will focus on the smartphone app and yubikey for the first iteration as those are the main use-cases currently.
It should be possible to fit different other types like biometrics (Hand=multiple fingers) in these ideas.

## Containers and Templates
There will be a base class and multiple concrete devices like smartphone and yubikey.
* The base class is a container that is pretty much only used for grouping the token (see operations section)
* The device will specify the **supported token types**. 
* MAYBE specify the capabilities/options that can be modified by a template explicitly, e.g. the derived class returns the template options + some base template options that every device has (e.g. naming)
* The device can have 0 or more owners
* Some devices are capable of using a backchannel (smartphone) while others can not act on their own (yubikey)
    - In case of devices that do not act on their own, there is/will be a program provided by us to execute the operations on the device. It can also check if changes have been made to the device by another program and report the changes.
    - **A backchannel is a must**

[...]

PR #4005, opened 2024-07-09
Token container

jelinaunger:
Introduces token containers #1291 

Additional enhancements and fixes:
* Added token API endpoints to enable and disable multiple tokens
* UI: Capitalize headings
* UI: Added translation for page navigation in the user details and list pages
* UI: Fixed open and close all actions in create new policy and conditions in create new event handler


privacyidea/static/components/token/views/token.details.html
```diff
@@ -555,25 +558,53 @@ <h4 ng-show="token.user_id" translate>Assigned User</h4>
...
+                    class="btn btn-primary">
+                <span class="glyphicon glyphicon-resize-small"></span>
+                <span translate>Assign User</span>
+            </button>
+        </div>
+    </form>
 </div>
 
+<!-- ============ Container ====================== -->
+<h4>Assign Container</h4>
```
plettich, line 581: So if i understand this correctly, we don't have a "container_read" right and thus the "Assign Container" block will always be displayed in a token view (and other views)?
Also the heading should be translatable.

privacyidea/static/components/token/views/token.containerlist.html
```diff
@@ -0,0 +1,118 @@
+<div uib-pagination ng-show="containerdata.count > containersPerPage"
```
plettich, line 1: The container list page is always displayed, even if the user/admin does not have any rights on containers.
While the list is empty for users, helpdesk admins (without the container rights) can see all the containers.

privacyidea/static/contrib/css/bootstrap-theme.css
```diff
@@ -309,6 +309,32 @@ fieldset[disabled] .btn-danger.active {
   background-color: #c12e2a;
   background-image: none;
 }
+.btn-transparent{
```
plettich, line 312: The `contrib` folder contains the data from external packages (bootstrap, angular and plugins). This will be overwritten in the next update of the packages.
Could You move these changes to the `static/css` folder? Or are these official changes of a newer bootstrap version?

privacyidea/static/components/config/views/config.policies.details.html
```diff
@@ -321,13 +321,13 @@ <h3 translate>Action</h3>
                     </label>
                     <label for="toggleAccordion">
                         <button class="btn btn-sm btn-success"
-                                ng-show="!isOpen"
-                                ng-click="isOpen = true">
+                                ng-show="groupsOpen < actionGroups.length"
```
plettich, line 324: there is a little bug if You apply a filter and then change the `Scope` of the policy (which loads a different set of actions). The `close all groups` button does not work anymore.

plettich (changes requested):
These are the tings i noticed while testing the UI

doc/eventhandler/containerhandler.rst
```diff
@@ -0,0 +1,106 @@
...
+A new container will be created. This new container can be assigned to a user, which was identified in the request.
+Additionally, a token identified in the request can be added to the container.
+
+The administrator has to specify the **containertype** and can optionally specify a **description**.
+
+delete
+......
+
+The container which was identified in the request will be deleted if all
+conditions are matched.
```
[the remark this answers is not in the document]
jelinaunger: The tokens will not be deleted. I clarify this in the docs. I think this is sufficient for now. In a latter version, one could think about adding a checkbox whether the contained tokens shall also be deleted.

plettich (changes requested):
I didn't take a closer look at the UI and tests except what i noticed during testing.

privacyidea/lib/token.py
```diff
@@ -181,6 +185,10 @@ def _create_token_query(tokentype=None, realm=None, assigned=None, user=None,
             # exact match
             sql_query = sql_query.filter(func.lower(Token.tokentype) == tokentype.lower())
 
+    if token_type_list is not None and len(token_type_list) > 0:
```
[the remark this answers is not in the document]
jelinaunger: It first filters for the `token_type` and based on the found token types it filters for all types that are in the `token_types_list`. Hence, if `token_type` is also in the `token_types_list`, only tokens of `token_type` will be returned. If `token_type` is not in the list, no tokens are returned, because no token matches the search query. 
This behavior is required to filter the tokens for all supported types of a container and allow additionally the user to filter this list by passing a wildcard in the `token_type`parameter.

privacyidea/api/token.py
```diff
@@ -312,9 +314,31 @@ def init():
...
         response_details.update(init_details)
+        # Check if a containerSerial is set and assign the token to the container
+        if "container_serial" in param:
+            container_serial = param.get("container_serial")
+            # check if user is allowed to add tokens to containers
+            try:
+                container_add_token_right = check_container_action(request, action=ACTION.CONTAINER_ADD_TOKEN)
+            except PolicyError:
+                container_add_token_right = False
+                log.error(f"User {user.login} is not allowed to add token {tokenobject.get_serial()} to container "
```
plettich, line 328: i would use an info log entry here (or warning) since this is kind of expected behaviour (the user does not have the right). Warning and Error are usually used to show that something is not working correctly in the server.

privacyidea/api/token.py
```diff
@@ -312,9 +314,31 @@ def init():
...
+                # TODO should this be caught here? The enrollment should not be blocked by the error?
+                try:
+                    logged_in_user_role = g.logged_in_user.get("role")
+                    add_token_to_container(container_serial, tokenobject.get_serial(), user, logged_in_user_role)
+                    response_details.update({"container_serial": container_serial})
+                    container = find_container_by_serial(container_serial)
+                    g.audit_object.log({"container_serial": container_serial,
+                                        "container_type": container.type})
+                except ResourceNotFoundError:
+                    log.error(f"Container with serial {container_serial} not found while enrolling token "
```
plettich, line 340: Same here, a warning would be more suitable here.

privacyidea/lib/container.py
```diff
@@ -0,0 +1,782 @@
...
+        for container_type, container_class in classes.items():
+            ret[container_type] = container_class.get_container_policy_info()
+        return ret
+
+
+def init_container(params):
+    """
+    Create a new container with the given parameters. Requires at least the type.
+
+    :jsonparam type: The type of the container
```
plettich, line 307: since this is no API function, `jsonparam` doesn't really work here. We should describe these as required contents of the `params` dictionary.

privacyidea/lib/container.py
```diff
@@ -0,0 +1,782 @@
...
+
+    ret = {}
+    for token_serial in token_serials:
+        # Check if admin is allowed to add the token to the container
+        if user_role == "admin" and allowed_realms:
+            token_realms = get_realms_of_token(token_serial)
+            matching_realms = list(set(token_realms).intersection(allowed_realms))
+            if len(matching_realms) == 0:
+                ret[token_serial] = False
+                log.error(
```
plettich, line 408: as mentioned above, please use `info` or `warning` here since this is (kind of) expected behaviour. `error` should be reserved for errors which impact the functionality of the server.

plettich (changes requested):

plettich (changes requested):

privacyidea/models.py
```diff
@@ -150,6 +152,23 @@ def delete(self):
...
+#
+
+class TokenContainerToken(MethodsMixin, db.Model):
+    """
+    Association table to link tokens to containers.
+    """
+    __tablename__ = 'tokencontainertoken'
+    __table_args__ = {'mysql_row_format': 'DYNAMIC'}
+    token_id = db.Column('token_id', db.Integer, db.ForeignKey('token.id'), primary_key=True)
+    container_id = db.Column('container_id', db.Integer, db.ForeignKey('tokencontainer.id'), primary_key=True)
```
plettich, line 169: is that allowed? two primary key columns and both foreign keys?
Maybe a unique constraint makes more sense here?

privacyidea/static/components/token/views/token.html
```diff
@@ -1,9 +1,50 @@
-<div class="well" ng-controller="tokenController">
+<div class="well" ng-controller="tokenMenuController">
```
plettich, line 1: somehow the `tokenMenuController` does not get registered when logged in as a user.
I get a bunch of:
```
Error: [$controller:ctrlreg] The controller with the name 'tokenMenuController' is not registered.
```
in the browser console when logging in as a user.

plettich (changes requested):

privacyidea/models.py
nilsbehlen: Yes that is allowed and serves the same purpose as a unique constraint.
https://docs.sqlalchemy.org/en/20/orm/basic_relationships.html#many-to-many

privacyidea/lib/containerclass.py
```diff
@@ -0,0 +1,445 @@
...
+
+    @classmethod
+    def get_container_policy_info(cls):
+        """
+        Returns the policy information of the container class.
+        """
+        res = {
+            "token_count": {"type": "int",
+                            "value": "any",
+                            "desc": "The maximum number of tokens in this container"},
```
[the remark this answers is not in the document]
nilsbehlen: no, we will not invest time in unsed code. this can be removed aswell.
if there will be the need to add policies to the classes, we will do it then.

plettich (approved):

Issue #3130, opened 2022-05-06
When using a sqlresolver it would be useful to be able to specify a schema


cmbyrne:
SQLSoup looks for the table in the default schema. There is an optional parameter in the entity called schema.  The table I need for user information is not in the default schema. My organization has multiple schemas for different data usaes.  An additional field for schema when creating the sqlResolver or parsing the table name <schema>.<tablename> would be useful.

cornelinux:
@cmbyrne Thank you for your request. You could push this by
* adding a link to the docs and the parameter
* adding a link to code
* giving a hint about a test environment 
The more information you provide, the less effort it is for the developer to add this and the more likely it is, that it will happen. Thanks.

cmbyrne:
@cornelinux ,
I would like  to provide a link to the docs, but they are no longer available on readthedocs.io or bitbucket. 

- That being said the method signature is:

> entity(attr, schema=None)
> Return the named entity from this [SqlSoup](https://www.codepowered.com/manuals/SQLAlchemy-0.6.9-doc/html/orm/extensions/sqlsoup.html#sqlalchemy.ext.sqlsoup.SqlSoup), or create if not present.
> 
> For more generalized mapping, see [map_to()](https://www.codepowered.com/manuals/SQLAlchemy-0.6.9-doc/html/orm/extensions/sqlsoup.html#sqlalchemy.ext.sqlsoup.SqlSoup.map_to).

The code in two places in [SQLIdResolver](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/lib/resolvers/SQLIdResolver.py) would need to change
- Line 553  in  def loadConfig(self, config):
        self.TABLE = self.db.entity(self.table)
- Line 664 in  def testconnection(cls, param):
       TABLE = db.entity(param.get("Table"))
The simplest change would be tokenize the table parameter on the '.' character, since most/all databases follow the dot notation of _schema.table_ when writing sql. If the split returns two items then the first is the schema and the table is the second, for one item set the schema to None or don't ad that parameter, otherwise report an error.
The test environment would be having the user table in a non-default schema.  Most databases default to public or username. So creating an additional schema  and table would be needed.

cornelinux:
@cmbyrne 
Can you give an example or suggestion how you would like to configure (an thus save) the schema in the resolver configuration?

plettich:
Fixed with #4007

PR #4007, opened 2024-07-11
Feature Enhancement: SQL Resolver - Allow Schema to be inferred from Table Name

cparkins:
Allow the Schema to be set for a SQL Resolver by specifying the Table with an optional '.' to separate the Schema from the Table Name. So that the Schema can be set to a different value when initializing the SQL Alchemy engine.

### DIFF
--- a/privacyidea/lib/resolvers/SQLIdResolver.py
+++ b/privacyidea/lib/resolvers/SQLIdResolver.py
@@ -489,7 +489,15 @@ class IdResolver (UserIdResolver):
         # Session should be closed on teardown
         register_finalizer(self.session.close)
         self.session._model_changes = {}
-        self.TABLE = Table(self.table, MetaData(), autoload_with=self.engine)
+
+        metadata = MetaData()
+        table_parts = self.table.split(".")
+        schema = table_parts[0] if len(table_parts) > 1 else None
+        self.table = table_parts[-1]
+        log.debug("Loading table {0!s} from schema {1!s}".format(self.table, schema))
+        if schema is not None:
+            metadata.schema = schema
+        self.TABLE = Table(self.table, metadata, autoload_with=self.engine)
 
         return self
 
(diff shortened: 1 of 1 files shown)


cparkins:
This is intended to be a solution for #3130.  I have tested locally and determined that the solution should work for scenarios where the default Schema does not contain the desired table.

privacyidea/lib/resolvers/SQLIdResolver.py
```diff
@@ -489,7 +489,15 @@ def loadConfig(self, config):
...
         register_finalizer(self.session.close)
         self.session._model_changes = {}
-        self.TABLE = Table(self.table, MetaData(), autoload_with=self.engine)
+
+        metadata = MetaData()
+        table_parts = self.table.split(".")
+        schema = table_parts[0] if len(table_parts) > 1 else None
+        self.table = table_parts[-1]
+        log.debug("Loading table {0!s} from schema {1!s}".format(self.table, schema))
+        if schema is not None:
```
plettich, line 498: We can just pass `schema` to the `Table` constructor below:
https://docs.sqlalchemy.org/en/20/core/metadata.html#sqlalchemy.schema.Table.params.schema

plettich (changes requested):

privacyidea/lib/resolvers/SQLIdResolver.py
cparkins: @plettich I've pushed a change to set the Table schema instead of setting it on the Metadata.

plettich: The `if schema is not None` can be dropped completely since `None` is the default value for the `schema` parameter of `Table()`.

plettich (changes requested):

plettich (approved):

PR #4014, opened 2024-07-19
ERROR: MariaDB doesn't yet support 'CACHE without INCREMENT BY 0 in Galera cluster'

### LINKED ISSUE #4012: ERROR: MariaDB doesn't yet support 'CACHE without INCREMENT BY 0 in Galera cluster'
### Top-level intent

Clone database running on MariaDB 10.11 + Galera 26.4.18

### Steps to reproduce

```bash
echo "create database clone;" | mysql -BN
mysqldump --single-transaction --add-drop-table --ignore-table-data pi.pidea_audit pi | mysql -BN clone
```

### Expected outcome

A clone of database `pi` named `clone`.

### Actual outcome

Error Message:
```
ERROR 1235 (42000) at line 248: This version of MariaDB doesn't yet support 'CACHE without INCREMENT BY 0 in Galera cluster'
```
The message caused by this SQL statement:
```
CREATE SEQUENCE `serviceid_seq` start with 1 minvalue 1 maxvalue 9223372036854775806 increment by 1 cache 1000 nocycle ENGINE=InnoDB
```
In particular by `... increment by 1 cache 1000 ...`

But all other sequences are defined in this way:
```text
.. increment by 0 cache 1000 ..
```
Why is it different for serviceid_seq and can it be changed to make privacyIDEA compatible with MariaDB 10.11 + Galera 26.4.18?

### Configuration

* **privacyIDEA version**: 3.9.3
* **Installation method**: venv pip
* **Python version**: 3.12
* **Operating system**: Ubuntu 24.04 LTS
* **Webserver**: Apache2
* **Token database**: MariaDB 10.11 + Galera 26.4.18

barkingup:
Fixes #4012

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+    "githubPullRequests.remotes": [
+        "origin",
+        "upstream",
+        "ll"
+    ]
+}
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,17 @@
+{
+	"version": "2.0.0",
+	"tasks": [
+		{
+			"type": "npm",
+			"script": "test",
+			"path": "privacyidea/static",
+			"group": {
+				"kind": "build",
+				"isDefault": true
+			},
+			"problemMatcher": [],
+			"label": "npm: test - privacyidea/static",
+			"detail": "echo \"Error: no test specified\" && exit 1"
+		}
+	]
+}


nilsbehlen:
Hi,
can you please double check to push the correct files?
Currently there are only vscode files, which we do not really want in the repo and they surely do not fix any problems.

plettich:
This seems to be some kind of reputation-gathering attempt. Closing this.

PR #4023, opened 2024-07-29
Node specific realms API

plettich:
Implementation of the API-level for the node-specific realms.
- Add a new endpoint `/realm/<realm>/node/<nodeid>` to add a new realm
  configuration specific to the given node

Working on #3758


privacyidea/api/periodictask.py
```diff
@@ -72,7 +72,7 @@ def list_nodes():
     """
     Return a list of available nodes
     """
-    nodes = get_privacyidea_nodes()
+    nodes = get_privacyidea_nodes(names_only=True)
```
nilsbehlen, line 75: boolean function parameter is most of the time not a good idea because it indicates that the function has two behaviors, which, against most the time, should just be two separate functions.
Is there a reason not to make a function `get_privacyidea_node_names`?

privacyidea/api/system.py
```diff
@@ -382,3 +445,43 @@ def list_ca_connectors():
     ca_list = get_caconnector_list(return_config=False)
     g.audit_object.log({"success": True})
     return send_result(ca_list)
+
+
+@system_blueprint.route("/nodes", methods=['GET'])
```
nilsbehlen, line 450: i think this should require authorization or admin only

privacyidea/api/realm.py
```diff
@@ -375,10 +431,101 @@ def delete_realm_api(realm=None):
...
+    for res in data:
+        try:
+            resolvers.append({'name': res["name"],
+                              'priority': int(res["priority"]),
+                              "node": nodeid})
+        except ValueError:
+            log.warning(f"Could not parse priority for resolver {res}")
+            raise ParameterError(_("Could not verify data in request!"))
+
+    (added, failed) = set_realm(realm, resolvers=resolvers)
```
jelinaunger, line 526: This function overwrites an existing realm. Since only one node ID can be passed to this endpoint it is not possible to specify multiple nodes within one realm. I thought that is the idea of node-specific realms, isn't it?

privacyidea/api/realm.py
```diff
@@ -375,10 +431,101 @@ def delete_realm_api(realm=None):
...
+    if not check_node_uuid_exists(nodeid):
+        log.warning(f"Node with UUID {nodeid} does not exist in the database!")
+        raise ParameterError(_("The given node does not exist!"))
+
+    data = request.json
+    resolvers = []
+    for res in data:
+        try:
+            resolvers.append({'name': res["name"],
+                              'priority': int(res["priority"]),
```
jelinaunger, line 520: It is not evaluated whether a priority is passed. A "KeyError" is raised if no priority key is available causing a 500 Internal Server Error.

privacyidea/api/realm.py
```diff
@@ -375,10 +431,101 @@ def delete_realm_api(realm=None):
...
+
+    .. versionadded:: 3.10 Node specific realm configuration
+    """
+    if not check_node_uuid_exists(nodeid):
+        log.warning(f"Node with UUID {nodeid} does not exist in the database!")
+        raise ParameterError(_("The given node does not exist!"))
+
+    data = request.json
+    resolvers = []
+    for res in data:
```
jelinaunger, line 517: The list of resolvers is not handled correctly in `before_request()`. It assumes the resolvers are json, but since it is a list it sets {"name": "priority"}` to `request.all_data`.  Additionally, it throws a `TypeError` if only a single resolver without a priority is passed, because it fails to handle a list with a single entry somehow as json.

plettich: Done in 200d355

privacyidea/api/realm.py
plettich: Done in 200d355

privacyidea/api/system.py
plettich: Done in 200d355

privacyidea/api/periodictask.py
plettich: Done in 200d355

privacyidea/api/realm.py
plettich: Done in 658afd5

privacyidea/api/realm.py
plettich: Done in 658afd5

privacyidea/api/system.py
plettich: Done in 658afd5

privacyidea/api/periodictask.py
plettich: Done in 658afd5

privacyidea/api/realm.py
plettich: Done in 658afd5

jelinaunger (approved):

nilsbehlen (approved):

PR #4025, opened 2024-07-30
SQL Resolver: Allow a Stored Procedure to validate a User's Password

cparkins:
Allow a Stored Procedure to be defined including static and dynamic fields. This stored procedure is expected to return a single value that can be tested by setting a parameter key with an @ sign to signify the name of the returned field.  This should allow output parameters to be used as long as they are named the same as the key without the @ sign.

### DIFF
--- a/privacyidea/lib/resolvers/SQLIdResolver.py
+++ b/privacyidea/lib/resolvers/SQLIdResolver.py
@@ -36,7 +36,7 @@ import re
 from privacyidea.lib.resolvers.UserIdResolver import UserIdResolver
 
 from sqlalchemy import (Integer, cast, String, MetaData, Table, and_,
-                        create_engine, select, insert, delete)
+                        create_engine, select, insert, delete, text)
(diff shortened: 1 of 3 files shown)


cparkins:
@plettich I'd appreciate some feedback on this PR. I can explain the reasoning if needed.  I'm sure I'm missing some pieces (translation for sure), but I am planning to use this code and would prefer it be integrated if it is appropriate for others to use.

plettich:
@cparkins I think i can understand the reasoning and i also think it can be a good idea but we are currently in the process of new release and i'll take care of this PR when we are done with it.

cparkins:
@nilsbehlen - In our use case, all password validation is done through a Stored Procedure.  This includes some logic to increase the number of failed attempts and things of this nature. Every attempt to login **must** use the Stored Procedure and not directly query against the database directly.  That being said, we have moved away from this implementation and towards an Admin Driven workflow, which doesn't currently require these changes.  If you have any further questions please let me know.

PR #4036, opened 2024-08-08
Updated changelog

nilsbehlen:


nilsbehlen:
@plettich, you might want add some more text to the node-specific realm feature to describe it shortly

plettich (approved):

PR #4042, opened 2024-08-12
Update translation templates for v3.10

plettich:
- Consolidate some strings to avoid duplicates
- Remove link from policy description (we should improve the text in the future)

### DIFF
--- a/po/template.pot
+++ b/po/template.pot
@@ -10,19 +10,6 @@ msgctxt "subscription"
 msgid "Level"
 msgstr ""
 
-#: privacyidea/static/components/token/views/token.enrolled.paper.html:22
-#: privacyidea/static/components/token/views/token.enrolled.paper.html:24
-#: privacyidea/static/components/token/views/token.enrolled.paper.html:26
-#: privacyidea/static/components/token/views/token.enrolled.paper.html:28
-#: privacyidea/static/components/token/views/token.enrolled.paper.html:30
-#: privacyidea/static/components/token/views/token.enrolled.tan.html:22
-#: privacyidea/static/components/token/views/token.enrolled.tan.html:24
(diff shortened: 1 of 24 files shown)


po/template.pot
```diff
@@ -2655,11 +2658,15 @@ msgstr ""
...
+msgid "Realm Name"
+msgstr ""
+
 #: privacyidea/static/components/login/controllers/loginControllers.js:639
 msgid "Realm defrealm created."
 msgstr ""
 
-#: privacyidea/static/components/config/views/config.realms.list.html:6
+#: privacyidea/static/components/config/views/config.realms.list.html:20
 msgid "Realm name"
```
jelinaunger, line 2670: This could also be capitalized to only have a single translation "Realm Name"

privacyidea/messages.pot
```diff
@@ -2427,107 +2438,112 @@ msgid ""
...
+#: lib/tokens/pushtoken.py:438
+msgid "Require the user to confirm the login with a presence check."
+msgstr ""
+
+#: lib/tokens/pushtoken.py:443
 msgid ""
 "The options that can be presented to the user to confirm the login. "
-"ALPHABETIC: A-Z, NUMERIC: 01-99, CUSTOM: user defined."
+"<code>ALPHABETIC</code>: A-Z, <code>NUMERIC</code>: 01-99, "
+"<code>CUSTOM</code>: user defined."
```
jelinaunger, line 2517: I suppose there is no way we can get around these < code > tags in the translation, is there? 
There is a space missing after "user defined. ", because it is followed by another sentence which is not included in this translation.

po/template.pot
jelinaunger, line 2670: This could also be capitalized to only have a single translation "Realm Name"

jelinaunger (approved):

PR #4046, opened 2024-08-16
Fix resolver lookup on nodes

plettich:
Working on #3758

### DIFF
--- a/privacyidea/lib/user.py
+++ b/privacyidea/lib/user.py
@@ -756,12 +756,12 @@ def get_user_list(param=None, user=None, custom_attributes=False):
             users.extend(ulist)
 
         except KeyError as exx:  # pragma: no cover
-            log.error("{0!r}".format((exx)))
+            log.error("{0!r}".format(exx))
             log.debug("{0!s}".format(traceback.format_exc()))
             raise exx
 
         except Exception as exx:  # pragma: no cover
-            log.error("{0!r}".format((exx)))
+            log.error("{0!r}".format(exx))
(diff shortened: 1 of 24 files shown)


privacyidea/lib/user.py
```diff
@@ -717,20 +717,24 @@ def get_user_list(param=None, user=None, custom_attributes=False):
...
+
+    local_node_uuid = get_app_config_value("PI_NODE_UUID")
     for pu_realm in [param_realm, user_realm]:
         if pu_realm:
             realm_config = get_realm(pu_realm)
             for r in realm_config.get("resolver", {}):
                 if r.get("name"):
-                    resolvers.append(r.get("name"))
+                    if not r.get("node") or r["node"] == local_node_uuid:
+                        resolvers.append(r.get("name"))
```
jelinaunger, line 728: So all resolvers of the node and all resolvers without a node specified are listed, right?
For example, if there is a realm with a general (not node specific) configuration specifying resolver1 and resolver2 and a configuration for node1 specifying resolver1 with prio 10 and resolver2 with prio 20. Then resolver1 and resolver2 would be listed twice for node1. Is that the expected behavior?

plettich: Further down the list of resolvers will be reduced to a set so all duplicate resolvers should be removed:
https://github.com/privacyidea/privacyidea/blob/3bb454341618d1361a22eeaa3d22158944c75322/privacyidea/lib/user.py#L739

jelinaunger (approved):

Issue #3618, opened 2023-05-17
Support Oauth Access Token as enhancement for trusted JWTs


cornelinux:
We accept trusted JWTs as described here:
https://privacyidea.readthedocs.io/en/latest/installation/system/inifile.html#trusted-jwts

It seems like we can relatively easy enhance this in a way, that OAuth Access Tokens could be accepted in the way as we currently accept the trusted JWTs.

There seem to be actually two steps to do so:

1. Add the audience (aud) to the JWT. 
2. Allow a `Bearer` in the `Authorization` header.

### Audience

The access token may come with the `aud` claim in the JWT. Thus the `jwt.decode` [here](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/api/lib/utils.py#L355) may add the audience to verify:

~~~~Python
j = jwt.decode(auth_token,
                                   trusted_jwt.get("public_key"),
                                   audience=[trusted_jwt.get("audience")],
                                   algorithms=[trusted_jwt.get("algorithm")])
~~~~

after telling the trusted JWT in `pi.cfg` which one is accepted:

~~~~
PI_TRUSTED_JWT = [{"public_key": "-----BEGIN PUBLIC KEY-----\nMIIBIjANBgkqhkiG9w0BAQEF...",
                   "algorithm": "RS256",
                   "audience": "https://myprivacyidea.company.com",
                   "role": "user",
                   "realm": "realm1",
                   "username": "userA",
                   "resolver": "resolverX"}]
~~~~

### Bearer

In addition the redirection may contain an Authorization header with the `Bearer` prefix.
We could add a config value in pi.cfg to allow to check for the Bearer prefix or strip the bearer prefix here:

https://github.com/privacyidea/privacyidea/blob/master/privacyidea/api/lib/utils.py#L346

or already before calling `verify_auth_token` here:

https://github.com/privacyidea/privacyidea/blob/da4e3786acb7be0c6e9e9a04e382ab51d1863ae8/privacyidea/api/auth.py#L428

gryffus:
Hello, sorry to hijacking this thread but I am trying to figure out SSO login to privacyIDEA webui:
https://community.privacyidea.org/t/using-jwt-to-login-to-webui/3031

Our goal is to authenticate with keycloak/JWT and just ask for OTP or enroll new one.

Something like:
otppin - userstore
passOnNoToken - enable
login_mode - privacyidea
tokenwizard - enable

[...]

PR #4065, opened 2024-09-03
JWT Token Resolver: Impelment a Token Resolver

cparkins:
This code should most of the concerns posed by Issue #3618 by implementing the higher level functionality of allowing JWT Tokens to be validated.  Currently the ability to validate the Issuer was not part of my testing requirements, but could easily be added. I also believe that further support could be added to implement OIDC standards including the ability to list users in an Organization.


cparkins:
@plettich - Once I resolve the lint issues, I'd like to see if you could review this change?

nilsbehlen:
Hi, thanks for your contribution. We will of course review it, but it might take some more time as we currently have some hickups with the release 3.10.

nilsbehlen:
Hi again. I would like to understand your code better.
Can you elaborate why you chose to implement a resolver? In the issue you referenced, there is no notion of an implementation via a resolver. 
Have you thought of alternative implementations? 
What is the implication of using this resolver in combination with other resolvers when resolving users?

nilsbehlen:
Thanks for explaining. Another naive idea (which might be stupid down the line) would be a token type. We will need to discuss this some more, so we not only cover your use case with the implementation but offer a broad solution.
You probably provided the code already to make things work, we might just have to move it around a bit :)

cparkins:
@nilsbehlen This is intended only as a first attempt. I think full OAuth/OpenID Connect support would be nice to have in the future.  But that was a bit beyond our simple use case of validating JWT solely.

cornelinux:
@cparkins To me this sounds as if this is only about login to the selfservice portal.

How would authentication look like? You need to assign 2nd factors/token to user objects, that are currently found in a/via a resolver.
Maybe you should explain a bit more your overall scenario and we should open an issue and discuss this there?

cparkins:
@cornelinux - Essentially the idea was to only allow access to the self-service portal. In my use case I have been challenged with only being able to validate a user by a JWT Token with a public token. I am able to do that with this code but storing that users User Id seems to be problematic because the Users cannot be listed.  I believe there are other Resolvers with these same limitations but I have no experience with them.

cparkins:
@nilsbehlen I wanted to discuss your comment a bit more, is this the best way to do that or should I open an issue?

> Thanks for explaining. Another naive idea (which might be stupid down the line) would be a token type. We will need to discuss this some more, so we not only cover your use case with the implementation but offer a broad solution. You probably provided the code already to make things work, we might just have to move it around a bit :)

I like the idea of a 'Token Type' but don't know exactly what you mean by that. Can you elaborate?

@cornelinux I was wondering if you had anything to add to my idea?

PR #4078, opened 2024-09-19
4049/user token number comparators

nilsbehlen:


privacyidea/lib/eventhandler/base.py
```diff
@@ -756,7 +756,8 @@ def check_condition(self, options):
 
         if CONDITION.USER_TOKEN_NUMBER in conditions and user:
             num_tokens = get_tokens(user=user, count=True)
-            if num_tokens != int(conditions.get(CONDITION.USER_TOKEN_NUMBER)):
+            cond = (conditions.get(CONDITION.USER_TOKEN_NUMBER))
```
plettich, line 759: Could You add a test for this condition here as well?

plettich (changes requested):

plettich (approved):

PR #4079, opened 2024-09-19
add token url below QR code when token wizard is used

nilsbehlen:


plettich (approved):

plettich (approved):

cornelinux:
@nilsbehlen  can we unmerge this?

PR #4081, opened 2024-09-20
remove text below qr from tokenwizard

nilsbehlen:
remove changes since they can be achieved with customization


cornelinux (approved):
Thx!

PR #4082, opened 2024-09-20
doc: Use new tags for tokenlabel example

samir-at-netknights:
Since version 3.9 <>-style tags are [no longer](https://privacyidea.readthedocs.io/en/v3.9/policies/enrollment.html#tokenlabel) mentioned in the documentation, this updates an example in the documentation to use the new format.

Before:
grafik

After:
grafik


samir-at-netknights:
Done! :smiley:

nilsbehlen (approved):

PR #4090, opened 2024-09-25
Design of the unassign button

### LINKED ISSUE #4089: Follow design in user's token view
From a design perspect I would recommend to use the same button style also in the user-token-view:

image

I think it is a nice idea to grey out the buttons to avoid distraction. 
But I think they should have the same size. Currently the text is not centered, this looks awkward.

Also I would recommend to make the *unassign* button when mouse over the orange color, just like with the other unassignment in the token details, so that the admin is aware that unassignment correlates with orange color.

image

jelinaunger:
* Changed design of the unassign button in the user details view: Same size as other buttons, unassign icon and orange color 
* Removed size from the `btn-transparent` class and use bootstrap class `btn-sm` to reduce the button size in the container details view

closes #4089

### DIFF
--- a/privacyidea/static/components/token/views/token.containerdetails.html
+++ b/privacyidea/static/components/token/views/token.containerdetails.html
@@ -408,7 +408,7 @@ SPDX-License-Identifier: AGPL-3.0-or-later
                         <span class="glyphicon glyphicon-remove"></span>
                         <span translate>Remove</span>
                     </button>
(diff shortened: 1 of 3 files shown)


jelinaunger:
The text alignment is the same as for the other buttons. It still looks like it is not perfectly centered since the icon is a bit larger than the text. That is also the case for other buttons.

nilsbehlen:
Screenshot from 2024-09-27 10-09-17
Screenshot from 2024-09-27 10-09-28
Screenshot from 2024-09-27 10-09-43

nilsbehlen (approved):

PR #4101, opened 2024-10-07
fix typo

samir-at-netknights:


plettich:
Hi @samir-at-netknights 
we are using weblate for the translation of the server and WebUI parts:
https://hosted.weblate.org/projects/privacyidea/
So the changes must be made there or they will be overwritten by the next update.

PR #4111, opened 2024-10-18
fix container serial param and log

nilsbehlen:


privacyidea/api/token.py
```diff
@@ -293,77 +293,44 @@ def init():
...
-    #    tokenrealm = res['realms']
-
     user = request.User
-    tokenobject = init_token(param,
-                             user,
-                             tokenrealms=tokenrealms)
 
-    if tokenobject:
+    token_object = init_token(param, user)
+    log.info("hallo")
```
jelinaunger, line 300: Remove log message

jelinaunger (approved):

PR #4112, opened 2024-10-18
use the correct endpoint, remove commented code

nilsbehlen:


nilsbehlen:
also removed the other factory function because it was not used.

privacyidea/static/components/user/controllers/userControllers.js
```diff
@@ -380,10 +380,9 @@ angular.module("privacyideaApp")
                 }
                 let params = {serial: selectedSerials, container_serial: $scope.containerSerial};
                 ContainerFactory.addTokenToContainer(params, function (data) {
-
+                    // Reload the token to show the container
+                    $scope._getUserToken();
```
jelinaunger, line 384: After adding the token to the container the container serial did not appear in the table. Moved the token reload to the callback to ensure it is called after the token is added to the container.

jelinaunger (approved):

PR #4117, opened 2024-10-24
Description required when editing with the corresponding active policy

henrik-falk:
for each type individually.


doc/policies/managment.rst
```diff
@@ -0,0 +1,23 @@
...
+
+The scope *management* defines what happens when managing tokens
+either by an administrator or the user.
+
+The following actions are available in the scope
+*management*:
+
+
+require_description_on_edit
+~~~~~~~~~~~~~~~~~~~~
```
plettich: Please extend the `~` line to the length of the text above, otherwise sphinx complains.

privacyidea/api/lib/prepolicy.py
```diff
@@ -2521,6 +2521,44 @@ def require_description(request=None, action=None):
...
+    token_types = list(action_values.keys())
+    type_value = request.all_data.get("type") or 'hotp'
+    if type_value in token_types:
+        tok = None
+        serial = getParam(params, "serial")
+        if serial:
+            tok = get_one_token(serial=serial, silent_fail=True)
+        description = request.all_data.get("description", "").strip()
+        if tok and not description:
+            log.warning(_("Missing description for {} token.".format(type_value)))
```
plettich: We don't need to translate warnings, they just end up in the log. Also I *think* we always log it as an `ERROR` message when raising an error.

plettich (changes requested):

privacyidea/api/lib/prepolicy.py
```diff
@@ -2521,6 +2521,44 @@ def require_description(request=None, action=None):
...
+                                  adminuser=adminuser,
+                                  user=username,
+                                  realm=realm,
+                                  user_object=user_object).action_values(unique=False)
+
+    token_types = list(action_values.keys())
+    type_value = request.all_data.get("type") or 'hotp'
+    if type_value in token_types:
+        tok = None
+        serial = getParam(params, "serial")
```
nilsbehlen: `get_required` or `get_optional` instead of `getParam`

privacyidea/api/lib/prepolicy.py
```diff
@@ -2521,6 +2521,44 @@ def require_description(request=None, action=None):
...
+        if serial:
+            tok = get_one_token(serial=serial, silent_fail=True)
+        description = request.all_data.get("description", "").strip()
```
nilsbehlen, line 2548: indent this and the following lines one more and remove `tok=None`

privacyidea/api/lib/prepolicy.py
```diff
@@ -2521,6 +2521,44 @@ def require_description(request=None, action=None):
...
+                                  scope=SCOPE.MANAGEMENT,
+                                  adminrealm=adminrealm,
+                                  adminuser=adminuser,
+                                  user=username,
+                                  realm=realm,
+                                  user_object=user_object).action_values(unique=False)
+
+    token_types = list(action_values.keys())
+    type_value = request.all_data.get("type") or 'hotp'
+    if type_value in token_types:
```
nilsbehlen, line 2547: or check serial here, without serial this can not work anyway?

privacyidea/lib/policy.py
```diff
@@ -3024,6 +3026,17 @@ def get_static_policy_definitions(scope=None):
                           'container locally on the smartphone.'),
                 'group': GROUP.SMARTPHONE
             }
+        },
+
```
nilsbehlen: remove emtpy line

doc/policies/managment.rst
```diff
@@ -0,0 +1,23 @@
+.. _management_policies:
+
+Management policies
+-------------------
+
+.. index:: management policies
+
+The scope *management* defines what happens when managing tokens
```
jona-samuel: Is this just for tokens or also containers?

privacyidea/api/lib/prepolicy.py
```diff
@@ -2498,29 +2498,63 @@ def require_description(request=None, action=None):
...
+        serial = get_optional(params, "serial")
         if serial:
-            tok = (get_one_token(serial=serial, rollout_state=ROLLOUTSTATE.VERIFYPENDING, silent_fail=True)
+            token = (get_one_token(serial=serial, rollout_state=ROLLOUTSTATE.VERIFYPENDING, silent_fail=True)
                    or get_one_token(serial=serial, rollout_state=ROLLOUTSTATE.CLIENTWAIT, silent_fail=True))
         # only if no token exists, yet, we need to check the description
-        if not tok and not request.all_data.get("description"):
-            log.warning(_("Missing description for {} token.").format(type_value))
+        if not token and not request.all_data.get("description"):
+            log.error(("Missing description for {} token.").format(type_value))
```
jona-samuel: We should use the opportunity to change this in an f-string

privacyidea/api/lib/prepolicy.py
```diff
@@ -2498,29 +2498,63 @@ def require_description(request=None, action=None):
...
+                                  user=username,
+                                  realm=realm,
+                                  user_object=user_object).action_values(unique=False)
+
+    token_types = list(action_values.keys())
+    type_value = request.all_data.get("type") or 'hotp'
+    if type_value in token_types:
+        description = request.all_data.get("description", "").strip()
+        if not description:
+            log.error(("Missing description for {} token.".format(type_value)))
```
jona-samuel: f-string

privacyidea/api/token.py
```diff
@@ -104,13 +105,14 @@
...
                                            hide_tokeninfo, init_ca_connector, init_ca_template,
-                                           init_subject_components, require_description,
+                                           init_subject_components, require_description_on_edit,
                                            check_container_action, check_user_params, check_token_list_action)
 from privacyidea.api.lib.postpolicy import (save_pin_change, check_verify_enrollment,
                                             postpolicy)
 from privacyidea.lib.event import event
 from privacyidea.api.auth import admin_required
 from privacyidea.lib.subscriptions import CheckSubscription
+from privacyidea.api.lib.prepolicy import require_description
```
jona-samuel: Is there a reason for a separate import?

doc/policies/managment.rst
```diff
@@ -0,0 +1,23 @@
+.. _management_policies:
```
jelinaunger: Typo in the filename: manag**e**ment.rst

You need to add this file to the toctree in index.rst.

privacyidea/api/token.py
```diff
@@ -866,6 +867,9 @@ def set_description_api(serial=None):
     g.audit_object.log({"serial": serial})
     description = getParam(request.all_data, "description", optional=required)
     g.audit_object.add_to_log({'action_detail': "description={0!r}".format(description)})
+    tokenobject_list = get_tokens_from_serial_or_user(serial=serial, user=user)
+    request.all_data["type"] = tokenobject_list[0].type
```
jelinaunger: What if the token can not be found due to an invalid serial or wrong user?
Maybe use the function `get_one_token(serial=serial)` which raises a ResourceNotFoundError if the token does not exist.

privacyidea/api/token.py
```diff
@@ -866,6 +867,9 @@ def set_description_api(serial=None):
     g.audit_object.log({"serial": serial})
     description = getParam(request.all_data, "description", optional=required)
     g.audit_object.add_to_log({'action_detail': "description={0!r}".format(description)})
+    tokenobject_list = get_tokens_from_serial_or_user(serial=serial, user=user)
```
jelinaunger: I would prefer "tokens"

privacyidea/api/token.py
```diff
@@ -866,6 +867,9 @@ def set_description_api(serial=None):
     g.audit_object.log({"serial": serial})
     description = getParam(request.all_data, "description", optional=required)
     g.audit_object.add_to_log({'action_detail': "description={0!r}".format(description)})
+    token = get_one_token(serial=serial)
```
nilsbehlen: we have the token object here. passing the serial to set_description, which will get the token again is inefficient. so it would be good to have set_description take a token object as input.
Add it as an optional parameter and if both serial and token object would be set, use the token object.

nilsbehlen (approved):

PR #4132, opened 2024-10-28
Token Resolver: Add the ability to use a Signed JWT for the first factor

cparkins:
Meant to be a starting point for #4065


nilsbehlen:
we will do #3618

PR #4140, opened 2024-11-11
fix two references in the documentation

samir-at-netknights:
In [the current documentation](https://privacyidea.readthedocs.io/en/v3.10/overview/index.html), two references do not resolve. This commit sets it to the correct references:
- https://github.com/privacyidea/privacyidea/blob/master/doc/installation/system/pi-manage.rst?plain=1#L1
- https://github.com/privacyidea/privacyidea/blob/master/doc/eventhandler/index.rst?plain=1#L3

Marking it as a draft since I can't get the current master to build the documentation and test it.

### DIFF
--- a/doc/overview/index.rst
+++ b/doc/overview/index.rst
@@ -72,11 +72,11 @@ to consider the following aspects:
   of possibilities to design automatation and **processes** like
   enrollment, revocation, leaving users and more. You will identify
   processes that are mandatory and others might not be relevant for you.
-  You can use again :ref:`policies` and :ref:`event_handler`s to implement these.
+  You can use again :ref:`policies` and :ref:`eventhandler` to implement these.
 
 * Authentication data is important. Plan your **backup** and **recovery**
   of your privacyIDEA system. You can use the database level but
-  also need to take a closer look at the :ref`:`pimanage`.
+  also need to take a closer look at :ref:`the pi-manage script <pimanage>`.
 
 * Think about update- and **upgrade**-processes. Both for privacyIDEA
   and for the OS.


doc/overview/index.rst
```diff
@@ -72,11 +72,11 @@ to consider the following aspects:
   of possibilities to design automatation and **processes** like
   enrollment, revocation, leaving users and more. You will identify
   processes that are mandatory and others might not be relevant for you.
-  You can use again :ref:`policies` and :ref:`event_handler`s to implement these.
+  You can use again :ref:`policies` and :ref:`eventhandler`s to implement these.
```
plettich, line 75: please remove the trailing `s` after `eventhandler`, it breaks the link.

doc/overview/index.rst
```diff
@@ -72,11 +72,11 @@ to consider the following aspects:
...
 
 * Authentication data is important. Plan your **backup** and **recovery**
   of your privacyIDEA system. You can use the database level but
-  also need to take a closer look at the :ref`:`pimanage`.
+  also need to take a closer look at the :ref:`pimanage`.
```
plettich, line 79: please remove the `the` since the generated link text already contains a `the`.

plettich (changes requested):
We have some information about building the documentation locally here:
https://github.com/privacyidea/privacyidea/blob/15a9422b0695b40e61797c0f1965285a4441aa19/CONTRIBUTING.md?plain=1#L66
After the build is finished, just point Your browser to `doc/_build/html/index.html`

samir-at-netknights:
Applied your requested changes.
I did try the steps listed in the CONTRIBUTING file before creating this PR. I just retried with a fresh virtualenv and then even in a fresh Ubuntu 24.04 VM. Same result:
[make_html.txt](https://github.com/user-attachments/files/17777517/make_html.txt)

samir-at-netknights:
Update: The build works with `sphinxcontrib-plantuml==0.25` :thinking:.

samir-at-netknights:
It looks a bit strange but the references work now:
image

Is this fine?

plettich (changes requested):
Looks good but can You rebase it onto `branch-3.10` so we'll have it fixed in the next bug-fix release?
I can confirm the error when building the documentation but this occurs only on the current master branch. I'll look into it.

plettich:
> Tested it on a working version of `sphinxcontrib-plantuml`. It looks a bit strange with the upper-case "The" but the references work now: 
> ...
> Is this fine?

We will need to go through the docs and check and fix some links someday but for now i think this is ok.
You can also change the text of the link: https://www.sphinx-doc.org/en/master/usage/referencing.html#role-ref

samir-at-netknights:
> Looks good but can You rebase it onto branch-3.10 so we'll have it fixed in the next bug-fix release?

Done!
> You can also change the text of the link: https://www.sphinx-doc.org/en/master/usage/referencing.html#role-ref

Thanks, it looks better now:
grafik

PR #4144, opened 2024-11-11
fix privacyidea-diag helper script

weimens:
Fixes a few problems with the support helper script privacyidea-diag

- replace deprecated pi-manage commands with new counterparts
- on rhel based systems the apache access log was missing
  
---

- ~~`pi-manage resolver list -v` is also deprecated, but the new counterpart does not obscure the secret, therefore the old version is still be used~~

### DIFF
--- a/tools/privacyidea-diag
+++ b/tools/privacyidea-diag
@@ -135,21 +135,21 @@ pi_config() {
 	log_pi_config "=================================="
 	log_pi_config "Resolvers"
 	log_pi_config "---------"
-	call_pi_manage resolver list -v >> "$pi_config"
+	call_pi_manage config resolver list -v >> "$pi_config"
 	log_pi_config "Realms"
 	log_pi_config "------"
-	call_pi_manage realm list >> "$pi_config"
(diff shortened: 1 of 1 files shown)


tools/privacyidea-diag
```diff
@@ -138,18 +138,18 @@ pi_config() {
 	call_pi_manage resolver list -v >> "$pi_config"
 	log_pi_config "Realms"
 	log_pi_config "------"
-	call_pi_manage realm list >> "$pi_config"
+	call_pi_manage config export -t realm -f json >> "$pi_config"
```
plettich, line 141: The original commands were just moved like
`call_pi_manage config realm list >> "$pi_config"`
The same works for the resolver list, it still censors the secret.

plettich (changes requested):

plettich:
Oh, and can You rebase this PR on to `branch-3.10`? We'll merge it into master afterwards.

tools/privacyidea-diag
weimens: thanks for the feedback. Done.

plettich (approved):

PR #4168, opened 2024-11-29
Fixing typo and update database

jona-samuel:


nilsbehlen (changes requested):
the correct spelling would be `urlencoded

migrations/versions/7301d5130c3a_.py
```diff
@@ -0,0 +1,47 @@
...
+# revision identifiers, used by Alembic.
+revision = '7301d5130c3a'
+down_revision = '69e7817b9863'
+
+
+def upgrade():
+    bind = op.get_bind()
+    session = orm.Session(bind=bind)
+    try:
+        for row in session.query(EventHandlerOption).filter(EventHandlerOption.Value == 'urlendcode'):
```
plettich, line 21: i think we should probably go with something like this:
```
op.execute("UPDATE eventhandleroption SET Value='urlencoded' WHERE Value='urlendcode'")
```
(See https://alembic.sqlalchemy.org/en/latest/ops.html#alembic.operations.Operations.execute)
I'll have to check if this works with all currently supported databases but this is regular SQL...

plettich: We could also use a rudimentary table definition or reflect it from the actual database and use SQLAlchemy constructs here.

plettich (changes requested):

plettich:
Works now with PostgreSQL, MySQL, Oracle, SQLite and Mariadb

plettich (approved):

plettich (approved):

nilsbehlen (approved):

PR #4166, opened 2024-11-28
admin policy match multiple realms and user

jelinaunger:
closes #4011
This pull request includes several changes to enhance the pre-policy checks and improve the handling of user and token actions in the `privacyidea` API. The most important changes include adding new pre-policy checks for various container actions, updating the `add_all_tokens` and `remove_all_tokens` functions to handle unauthorized tokens, and introducing new helper functions to manage token and container user attributes and actions.

Enhancements to pre-policy checks:

* [`privacyidea/api/container.py`](diffhunk://#diff-b7ccdf2332669a3ec4092e227584a5e78683b96b05dd0b03360100ed671d9571L26-R34): Added new pre-policy checks for user parameters, token actions, and token list actions in multiple container-related routes. [[1]](diffhunk://#diff-b7ccdf2332669a3ec4092e227584a5e78683b96b05dd0b03360100ed671d9571L26-R34) [[2]](diffhunk://#diff-b7ccdf2332669a3ec4092e227584a5e78683b96b05dd0b03360100ed671d9571R137) [[3]](diffhunk://#diff-b7ccdf2332669a3ec4092e227584a5e78683b96b05dd0b03360100ed671d9571R163) [[4]](diffhunk://#diff-b7ccdf2332669a3ec4092e227584a5e78683b96b05dd0b03360100ed671d9571R189) [[5]](diffhunk://#diff-b7ccdf2332669a3ec4092e227584a5e78683b96b05dd0b03360100ed671d9571R263) [[6]](diffhunk://#diff-b7ccdf2332669a3ec4092e227584a5e78683b96b05dd0b03360100ed671d9571R296) [[7]](diffhunk://#diff-b7ccdf2332669a3ec4092e227584a5e78683b96b05dd0b03360100ed671d9571R336) [[8]](diffhunk://#diff-b7ccdf2332669a3ec4092e227584a5e78683b96b05dd0b03360100ed671d9571R369) [[9]](diffhunk://#diff-b7ccdf2332669a3ec4092e227584a5e78683b96b05dd0b03360100ed671d9571R511)

Handling unauthorized tokens:

[...]

PR #4184, opened 2024-12-11
Fix test for MariaDB and postgresql

jelinaunger:
Sort lists before comparing

Test from #4166

### DIFF
--- a/tests/test_lib_policy.py
+++ b/tests/test_lib_policy.py
@@ -1583,44 +1583,36 @@ class PolicyTestCase(MyTestCase):
 
         # get policies applicable for realm A
         policies = P.list_policies(action="read", realm="realmA")
-        policy_names = [p["name"] for p in policies]
-        policy_names.sort()
+        policy_names = {p["name"] for p in policies}
         self.assertEqual(7, len(policies))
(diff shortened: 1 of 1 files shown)


tests/test_lib_policy.py
```diff
@@ -1577,34 +1576,52 @@ def test_41_list_policies_user_realm_resolver(self):
...
+        policy_names = [p["name"] for p in policies]
+        policy_names.sort()
         self.assertEqual(7, len(policies))
         correct_policies = ["scopeA_r", "scopeA_r_realmA", "scopeA_r_realmA_userA", "scopeA_r_resolverA",
-                            "scopeA_r_resolverB", "scopeA_r_realmA_userA_resolverA", "scopeA_r_realmA_userA_resolverB"]
-        self.assertListEqual(correct_policies, [p["name"] for p in policies])
+                            "scopeA_r_resolverB", "scopeA_r_realmA_userA_resolverA",
+                            "scopeA_r_realmA_userA_resolverB"]
+        correct_policies.sort()
+        self.assertListEqual(correct_policies, policy_names)
```
plettich, line 1593: If the policy names are unique You can just use sets, should be easier to compare.

plettich (approved):

PR #4205, opened 2025-01-13
Make is_orphaned fail-safe

andreasweber-uu:
Proposed the simplest solution, which is to never orphan a token on errors.

### DIFF
--- a/privacyidea/lib/tokenclass.py
+++ b/privacyidea/lib/tokenclass.py
@@ -316,9 +316,9 @@ class TokenClass(object):
                     # The token is assigned, but the username does not resolve
                     orphaned = True
             except Exception:
-                # If any other resolving error occurs, we also assume the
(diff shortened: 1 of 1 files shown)


nilsbehlen:
i think this is good. maybe add a option to still have the old behavior, so it is still there but has to be explicitly expressed.
Documentation probably needs clarification too.
@plettich

plettich:
@nilsbehlen @andreasweber-uu 
I think this behaviour was implemented because in some cases/resolvers we would also get an error if the user does not exist in the resolver. I *think* we fixed that by now so i'll look into integrating it here (legacy) and into the new token-janitor.

nilsbehlen:
@plettich i would just add it to the new token janitor with a switch on how to handle missing users

jona-samuel:
I opened a new pull request to add a few more changes. I linked you there. Thanks for your support to improve privacyIDEA

PR #4211, opened 2025-01-19
dashboard enhanced by values of 10minutes and 60minutes

gs4711:
Authentications of last 24 hours are now enhanced by two other values (10 minutes and 60 minutes).


nilsbehlen:
Hi, ideally the options would be configurable. But i currently do not know the best way to save it, i think a policy would be a bit much.

nilsbehlen:
We will do this when we have #4220, this is good use for that. And it will only be in the new WebUI.
Feel free to use your adjustments in your setup and thanks for the idea.

PR #4256, opened 2025-02-20
add table to find fido2 token faster, constants

nilsbehlen:
* Added table tokencredentialidhash to be able to find fido2 directly. The use of already enrolled token will also populate that table for future use.
* Added more constants for FIDO2 values
* Fixed WebAuthn token usage when multiple tokens have been triggered and added test for that
* Removed AuthError from TestPasskey in WebUI because it would log you out if an error occurs


privacyidea/lib/fido2/util.py
```diff
@@ -1,52 +1,74 @@
...
+                return create_tokenclass_object(db_token)
+        else:
+            log.debug(f"TokenCredentialIdHash entry not found for credential_id {credential_id}. Trying token info...")
+            token_id = (TokenInfo.query.filter(TokenInfo.Key == "credential_id_hash")
+                        .filter(TokenInfo.Value == credential_id_hash).first().token_id)
+            db_token = Token.query.filter(Token.id == token_id).first()
+            if db_token:
+                # Create a new TokenCredentialIdHash entry for the next time
+                tcih = TokenCredentialIdHash(token_id=db_token.id, credential_id_hash=credential_id_hash)
+                tcih.save()
```
jelinaunger, line 35: And remove it from the token info?

privacyidea/models.py
```diff
@@ -304,6 +306,8 @@ def delete(self):
         db.session.query(TokenTokengroup) \
             .filter(TokenTokengroup.token_id == self.id) \
             .delete()
+        db.session.query(TokenCredentialIdHash).filter(TokenCredentialIdHash.token_id == self.id).delete()
```
jelinaunger, line 309: Why not use a relationship that automatically deletes corresponding entries?

migrations/versions/903a6ed6f6c4_.py
```diff
@@ -0,0 +1,26 @@
...
+down_revision = 'c128c01a5520'
+
+
+def upgrade():
+    op.create_table('tokencredentialidhash',
+                    sa.Column('credential_id_hash', sa.String(length=1024), nullable=False),
+                    sa.Column('token_id', sa.Integer(), nullable=False),
+                    sa.ForeignKeyConstraint(['token_id'], ['token.id'], ),
+                    sa.PrimaryKeyConstraint('credential_id_hash')
+                    )
```
jelinaunger, line 22: catch error if table already exists

migrations/versions/903a6ed6f6c4_.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def downgrade():
+    op.drop_table('tokencredentialidhash')
```
jelinaunger, line 26: catch error if table does not exist

jelinaunger (approved):

privacyidea/lib/fido2/util.py
nilsbehlen: no

jelinaunger (approved):

PR #4278, opened 2025-03-06
user details fixes

### LINKED ISSUE #4277: User Details Enhancement
* Refresh user details loads full user list
* Two delete token buttons visible
* container pagination

jelinaunger:
closes #4277 
+on the container details in the token table, fixed the link to the user page

### DIFF
--- a/privacyidea/static/components/user/views/user.details.html
+++ b/privacyidea/static/components/user/views/user.details.html
@@ -138,7 +138,7 @@
          first-text="{{ 'First'|translate }}"
          items-per-page="{{ tokensPerPage }}"
          max-size="5"
-         boundary-links="true" ng-change="pageChanged()"></div>
+         boundary-links="true" ng-change="tokenPageChanged()"></div>
 
     <div class="table-responsive form-group" ng-show="showTokenOfUser">
         <table class="table table-bordered table-striped">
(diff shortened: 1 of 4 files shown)


privacyidea/static/components/user/views/user.details.html
```diff
@@ -215,7 +215,7 @@ <h3 ng-show="!showTokenOfUser" translate>Token of {{username}} without a contain
                         </button>
                         <button class="btn btn-danger btn-sm"
                                 ng-click="deleteToken(token.serial, false)"
-                                ng-show="(showDialog[token.serial] && checkRight('delete')) || !hide_buttons"
+                                ng-show="showDialog[token.serial] && (checkRight('delete') || !hide_buttons)"
```
plettich, line 218: We have this [`deletion_confirmation`](https://privacyidea.readthedocs.io/en/v3.9.2/policies/webui.html#deletion-confirmation) policy which opens a pop-up when deleting important stuff.
In the WebUI the `confirmDelete` function can be used:
https://github.com/privacyidea/privacyidea/blob/8d41a69716d64b57159e54d7bf373bde209743c6/privacyidea/static/components/login/controllers/loginControllers.js#L100

plettich (changes requested):

privacyidea/static/components/user/views/user.details.html
jelinaunger: So far, this policy is only for deleting policies, events, mresolvers, resolvers, and periodic tasks. I agree that for consistency this dialog should also be used for tokens and containers. However, the delete confirmation was not introduced in this PR. For now, I would keep it as it is and improve it in the new webui.

nilsbehlen (approved):

PR #4282, opened 2025-03-14
Fix warning

jona-samuel:


privacyidea/lib/eventhandler/usernotification.py
```diff
@@ -475,6 +432,51 @@ def do(self, action, options=None):
...
+                                        logged_in_user.get("realm"))
+                        if user_obj:
+                            reply_to = user_obj.info.get("email") if user_obj else ""
+
+                elif reply_to_type == NOTIFY_TYPE.EMAIL:
+                    email = handler_options.get("reply_to " + NOTIFY_TYPE.EMAIL, "").split(",")
+                    reply_to = email[0]
+
+                else:
+                    log.warning("Was not able to determine the email for the reply-to "
```
plettich, line 477: Please use an f-string and just "Unable to determine..." (same above)

privacyidea/lib/eventhandler/usernotification.py
```diff
@@ -402,8 +358,9 @@ def do(self, action, options=None):
             recipient = {
                 "email": email
             }
-        else:
-            log.warning("Was not able to determine the recipient for the user "
+
+        elif not action.lower() == "savefile":
```
plettich, line 362: We can just check in `if action == "sendsms":` or `if action == "sendmail":` if the `recipient` is set and issue a warning there.

plettich (changes requested):

PR #4296, opened 2025-03-21
disable trigger with pin, webui adjustments

nilsbehlen:
* By default, passkeys can not be triggered with the PIN
* There is a new policy to enable that behavior WebUI:
* Passkey login button is transparent so there is always only one main button
* Removed token type specific rows of the token table in the user detail view (otplen, window)
* Adjusted capitalization and typos

### DIFF
--- a/privacyidea/lib/tokens/passkeytoken.py
+++ b/privacyidea/lib/tokens/passkeytoken.py
@@ -104,6 +104,12 @@ class PasskeyTokenClass(TokenClass):
                         'desc': _("Alternative challenge message to use when authenticating with a passkey."
                                   "You can also use tags for replacement, "
                                   "check the documentation for more details.")
+                    },
+                    PasskeyAction.EnableTriggerByPIN: {
+                        'type': 'bool',
+                        'desc': _("When enabled, passkey token can be triggered with the PIN. For privacyIDEA plugins, "
+                                  "this is not recommended. It is advised to use a condition, for example on a "
+                                  "user-agent, with this policy."),
                     }
                 },
                 SCOPE.ENROLL: {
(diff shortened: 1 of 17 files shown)


privacyidea/lib/tokens/passkeytoken.py
```diff
@@ -104,6 +104,12 @@ def get_class_info(key=None, ret='all'):
                         'desc': _("Alternative challenge message to use when authenticating with a passkey."
                                   "You can also use tags for replacement, "
                                   "check the documentation for more details.")
+                    },
+                    PasskeyAction.EnableTriggerByPIN: {
+                        'type': 'bool',
+                        'desc': _("When enabled, passkey token can be triggered with the PIN. For privacyIDEA plugins, "
+                                  "this is not recommended. It is advised to use a condition, for example on a "
+                                  "user-agent, with this policy."),
```
jelinaunger, line 112: Please add this policy to the documentation

privacyidea/static/components/token/controllers/tokenControllers.js
```diff
@@ -376,18 +376,17 @@ myApp.controller("tokenEnrollController", ["$scope", "TokenFactory", "$timeout",
...
-                });
+                if (size === 1) {
+                    $scope.newUser = {user: "", realm: realmname};
+                } else {
+                    angular.forEach($scope.realms, function (realm, realmname) {
+                        // if there is a default realm, preset the default realm
+                        if (realm.default && !$stateParams.realmname) {
+                            $scope.newUser = {user: "", realm: realmname};
+                        }
+                    });
```
jelinaunger, line 387: Can you move this to the assignUser directive? Then it would also be applied to the token and container details.

jelinaunger (changes requested):

jelinaunger (approved):

Issue #4306, opened 2025-03-25
YubiKey signature validation not according to Yubico spec


fxwgr:
### Top-level intent

What did you try to achieve?
YubiKey signature validation through the /ttype/yubikey API.

### Steps to reproduce

1. Send a request including the needed paramaters and also the calculated parameter h to the API
2. Get this answer: BAD_SIGNATURE

### Expected outcome

What do you think, how the system should have worked?
The signature has to be identical to privacyIDEA's calculation and then be answered as OK, so the validation process is working fine.

### Actual outcome

What did actually happen?
privacyIDEA is taking one more key (ttype) into the calculation, so it is not identical to the client's process and also not according to Yubico's specification.

### Configuration

* **privacyIDEA version**: 3.11 (but also in current master)
* **Installation method**: Ubuntu packages and GitHub
* **Python version**: 3.10
* **Operating system**: Ubuntu 22.04.5 LTS
* **Webserver**: Apache
* **Token database**: integrated

PR #4307, opened 2025-03-25
Fix YubiKey signature calculation according to Yubico specificiation

fxwgr:
As mentioned in #4306 the signature validation process is currently not working as expected.
[Yubico's specification](https://developers.yubico.com/OTP/Specifications/OTP_validation_protocol.html) is not mentioning any parameter "ttype" and it is also not sent by the requester, but because of this parameter the validation is failing as BAD_SIGNATURE at each request.

Adjusted also the description accordingly.

### DIFF
--- a/privacyidea/lib/tokens/yubikeytoken.py
+++ b/privacyidea/lib/tokens/yubikeytoken.py
@@ -90,6 +90,8 @@ def yubico_api_signature(data, api_key):
     r = dict(data)
     if 'h' in r:
         del r['h']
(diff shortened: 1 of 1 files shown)


nilsbehlen:
we could probably put this in 3.11.1

jelinaunger:
Thanks for the PR. I can approve these changes. However, I want to add a test and merge it into branch 3.11 instead of the master. Hence, I created a new PR #4319 containing these changes.

Issue #2514, opened 2020-11-16
auth policy with custom header checks break (at least) push tokens


jmozd:
### Top-level intent
After creating policies that use custom HTTP headers, for me, authentication via push tokens started to fail. I chased this down to reported exceptions within policy handling:

a policy ("policyX") is defined, relying on a custom header (active condition comparing a header against a given value). Sample purpose of policyX is to inititate a push token request for a specific PAM scenario

the client sends the "/validate/check" POST request including the header, traverses the policy chain and policyX works as expected

i.e. the push token reply is coming in ("/ttype/push" GET request) and traverses the policy chain, but does not have the custom header

when checking "policyX" against that "/ttype/push" request, the missing HTTP header is reported as an error, an exception is thrown and further processing seems to stop

This PR changes the behavior of _policy_matches_request_header_condition() to only log a warning and to return as if the comparison on the header failed.

The comparator parameter of _policy_matches_request_header_condition() was not (yet) included in the decision. Returning "true" for "inverse matches" (!in, !contains,...) would possibly create a different semantic, but of course this can be subject of discussion.

### Steps to reproduce

1. create authentication policy using custom header, asking for push token
2. issue authentication request

### Expected outcome

The push token should pick up the request and once confirmed, the authentication should happen.

### Actual outcome

The push token never asks the user for confirmation, even refreshing the push token device does not show the request.
The authentication request times out.
What did actually happen?

### Configuration

* **privacyIDEA version: 3.4.1 (it'd be cool if this was part of the "system documentation")
* **Installation method: pip install privacyidea
* **Python version: Python 3.6.10 (it'd be cool if this was part of the "system documentation")
* **Operating system: OpenSUSE Leap 15.2
* **Webserver: apache2-2.4.43-lp152.1.1.x86_64
* **Token database: MariaDB database back-end

### Log file

When the httpd server receives the request "GET /ttype/push HTTP/1.1"  778 "-" "Dart/2.10 (dart:io)", PI's log file shows:

[140479983216384][ERROR][privacyidea.lib.policy:788] Policy 'policyX' has conditions on headers 'application', but http header is not available. This should not happen.

and it seems the policy processing stops here.

I tracked this error message down and demoted it to a "warning", returning as if the header actually exists, but does not match. Once I put that change into place, processing completed and worked as expected.

Created PR https://github.com/privacyidea/privacyidea/pull/2509 with the according change.

cornelinux:
[...]

PR #4346, opened 2025-04-08
extended policy conditions: added handle missing data option

### LINKED ISSUE #4339: Extended policy conditions
Add a selection for the user how the system should behave if the information required for the condition is missing.

### Problem Description
Sometimes the information to be checked in the condition is not available
* no user object if admin is doing the request
* registration info for not registered containers
* token type specific attributes

### Possible behavior if information is missing
- error_if_missing: raise error (actual behavior)
- True_if_missing: conditions is True -> policy is applied
- False_if_missing: condition is False -> policy is not applied

Adding these three possibilities as a selection to the UI, the user can decide how the system should behave.

jelinaunger:
closes #4339, #3512, #2514


migrations/versions/5f40baab76ca_.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    try:
+        op.add_column('policycondition',
+                      sa.Column('handle_missing_data', sa.Unicode(length=255), nullable=True))
+    except (OperationalError, ProgrammingError) as exx:
+        if "already exists" in str(exx.orig).lower() or "duplicate column name" in str(exx.orig).lower():
+            print("Column 'handle_missing_data' already exist.")
```
nilsbehlen, line 26: exists

privacyidea/lib/policy.py
```diff
@@ -523,9 +523,17 @@ class CONDITION_SECTION(object):
...
 class CONDITION_CHECK(object):
     __doc__ = """The available check methods for extended conditions"""
+    # TODO: Use the same datatype for all checks
     DO_NOT_CHECK_AT_ALL = 1
     ONLY_CHECK_USERINFO = [CONDITION_SECTION.USERINFO]
-    CHECK_AND_RAISE_EXCEPTION_ON_MISSING = None
+    CHECK_AND_HANDLE_MISSING_DATA = None
+
+
+class ConditionHandleMissingData:
```
nilsbehlen, line 532: could be enum?

privacyidea/lib/policy.py
```diff
@@ -929,12 +942,100 @@ def filter_policies_by_conditions(self, policies, user_object=None, request_head
...
+        :param missing: The missing data, either equal to the object_name or the key
+        :param section: The section of the condition
+        :param object_name: The name of the object that should be evaluated in the condition
+        :param key: The key that should be evaluated in th condition
+        :param available_keys: The available keys of the object (if it is available), optional
+        :return: True if handle_missing_data is CONDITION_IS_TRUE
+                 False if handle_missing_data is CONDITION_IS_FALSE
+                 raise PolicyError if handle_missing_data is RAISE_ERROR or is None
+        """
+        if not handle_missing_data or handle_missing_data == ConditionHandleMissingData.RAISE_ERROR:
```
nilsbehlen, line 964: can this be refacored to have the check for `handle_missing_data` at the bottom? If i see it correctly it influences the log level and in one case it raises the error. So ideally the log texts would be there only once, so it looks less copy-pastey

privacyidea/lib/policy.py
```diff
@@ -943,31 +1044,27 @@ def _policy_matches_request_environ_condition(policy, key, comparator, value, re
...
-                                                       ''.join(traceback.format_stack())))
-            raise PolicyError("Policy {!r} has conditions on environment {!r}, but HTTP environment"
-                              " is not available".format(policy["name"], key))
+            return cls._do_handle_missing_data(handle_missing_data, policy["name"], missing=object_name,
+                                               section=object_name, object_name=object_name, key=key)
 
-    @staticmethod
-    def _policy_matches_request_header_condition(policy, key, comparator, value, request_headers):
+    @classmethod
+    def _policy_matches_request_header_condition(cls, policy, key, comparator, value, request_headers,
```
nilsbehlen, line 1060: it would be great if you can add type hints to functions you are editing

privacyidea/lib/policy.py
```diff
@@ -1409,7 +1489,8 @@ def set_policy(name=None, scope=None, action=None, realm=None, resolver=None,
     :type check_all_resolvers: bool
     :param user_case_insensitive: The username should be case-insensitive.
     :type user_case_insensitive: bool
-    :param conditions: A list of 5-tuples (section, key, comparator, value, active) of policy conditions
+    :param conditions: A list of 5- or 6-tuples (section, key, comparator, value, active, handle_missing_data) of
```
nilsbehlen, line 1492: makes me want to die. is it possible to introduce a class or dataclass to group this data together

privacyidea/lib/policy.py
```diff
@@ -1457,15 +1538,17 @@ def set_policy(name=None, scope=None, action=None, realm=None, resolver=None,
...
-                raise ParameterError("Conditions must be 5-tuples of four strings and one boolean: {!r}".format(
-                    condition))
+            if len(condition) not in [5, 6]:
+                raise ParameterError(f"Conditions must be 5- or 6-tuples: {condition}")
+            if (not (isinstance(condition[0], str)
+                     and isinstance(condition[1], str)
+                     and isinstance(condition[2], str)
+                     and isinstance(condition[3], str)
+                     and isinstance(condition[4], bool))
+                    or (len(condition) == 6 and condition[5] is not None and not isinstance(condition[5], str))):
```
nilsbehlen, line 1548: above comment could save us from all of this madness

tests/test_lib_policy.py
```diff
@@ -377,23 +439,23 @@ def test_09_realm_resolver_policy(self):
...
+        set_policy(name="pol1", scope=SCOPE.ADMIN, action=f"enrollHOTP, {ACTION.ENABLE}, {ACTION.DISABLE}")
+        set_policy(name="pol2", scope=SCOPE.AUTH, action=f"{ACTION.RESETALLTOKENS}, {ACTION.OTPPIN}=1")
+        set_policy(name="pol3", scope=SCOPE.ADMIN, action=f"*, -{ACTION.DISABLE}")
+        set_policy(name="pol4", scope=SCOPE.AUTH, action=f"*, -{ACTION.OTPPIN}=2")
 
         P = PolicyClass()
-        p = P.match_policies(action="enroll")
-        self.assertTrue(len(p) == 4, (len(p), p))
+        p = P.match_policies(scope=SCOPE.ADMIN, action="enrollHOTP")
+        self.assertEqual(2, len(p), (len(p), p))
```
jelinaunger, line 449: If we not add the scope here, also policy "pol4" would match even if the authentication scope does not contain the action enrollHOTP. The policy matching just checks the wildcard and accepts all actions. Is that intended?

privacyidea/lib/policy.py
```diff
@@ -1410,30 +1378,48 @@ def set_policy(name=None, scope=None, action=None, realm=None, resolver=None,
...
+        policy conditions
     :param pinode: A privacyIDEA node or a list of privacyIDEA nodes.
     :param description: A description for the policy
     :type description: str
     :return: The database ID of the policy
     :rtype: int
     """
-    active = is_true(active)
+    # validate scope
+    if scope and scope not in SCOPE.get_all_scopes():
```
nilsbehlen, line 1390: so setting a policy without a scope is ok?
otherwise it would be `if not scope or scope not in`

jelinaunger: set_policy is also used to update policies. `enable_policy` only provides the `name` and `active` parameters:
https://github.com/privacyidea/privacyidea/blob/cfdad29fa8f83f19ea39ccea7234557523386516/privacyidea/lib/policy.py#L1537
Hence, the scope is not required..

nilsbehlen: we should add a TODO then that these actions should be split in multiple functions

privacyidea/lib/policy.py
```diff
@@ -1377,6 +1252,98 @@ def ui_get_enroll_tokentypes(self, client, logged_in_user):
...
+    """
+    if values is not None:
+        if isinstance(values, str):
+            values = values.replace(" ", "").split(",")
+        elif not isinstance(values, list):
+            raise ParameterError(f"Invalid resolvers type '{type(values)}'!")
+        values = remove_wildcards_and_negations(values)
+        undefined_values = list(set(values) - set(allowed_values))
+        if len(undefined_values) > 0:
+            raise ParameterError(f"Undefined {name.capitalize()} {undefined_values}!")
```
plettich, line 1343: maybe add an extra `:` or quotation marks around the `{undefined_values}`

plettich (changes requested):

nilsbehlen (approved):

PR #4359, opened 2025-04-17
convert userhandle, add test, fix naming of classes

### LINKED ISSUE #4357: WebAuthn userHandle parameter evaluation in wrong format
the userhandle parameter of webauthn token is the token serial. however, the authenticator returns it in base64url format, so it has to be decoded and encoded as utf-8 string before the comparison.
This behavior leads to authentications being rejected if the requests contains the correct userhandle.

nilsbehlen:


privacyidea/lib/tokens/webauthn.py
```diff
@@ -1556,9 +1557,11 @@ def verify(self):
             # identified by this value is the owner of the public key credential
             # identified by credential.id.
             user_handle = self.assertion_response.get('userHandle')
+            user_handle = base64url_to_bytes(user_handle).decode("utf-8") if user_handle else None
```
nilsbehlen, line 1560: this is the fix

jelinaunger (approved):

PR #4360, opened 2025-04-22
add creation date to token info

nilsbehlen:


privacyidea/lib/token.py
```diff
@@ -1348,6 +1348,9 @@ def init_token(param, user=None, tokenrealms=None, tokenkind=None):
     if validity_period_start:
         token.set_validity_period_start(validity_period_start)
 
+    # Creation Date
+    token.add_tokeninfo("creation_date", datetime.datetime.now(datetime.timezone.utc).isoformat())
```
plettich, line 1352: We should probably use `.isoformat(timespec="seconds")` here to avoid the missing microseconds issue. The creation date is likely not as sensitive as the last auth timestamp: https://docs.python.org/3.12/library/datetime.html#datetime.datetime.isoformat

plettich (approved):

nilsbehlen:
add a test or an assert to an existing test

Issue #4208, opened 2025-01-16
Containers Part 3


jelinaunger:
### Policies

- [x] Add container and container info to the extended conditions
- [x] hide_container_info

### Registration & Synchronization
- [x] Use AD password to secure the registration

### Miscellaneous
- [x] Unassign a user from a container does not work if the user does not exist (anymore) in the resolver
- [x] Allow wildcards for the container query (also for the container_serial in the token query)
- [x] Add filter to the columns of the container list
- [x] function delete_all_tokens: single api call instead of one call for each token
   There is no clear RESTful way of batch deletions. There are two main options:
     * DELETE /token/batchdeletion?serial=0,1,2
         -> Use correct http method DELETE
         -> Number of serials that can be passed is limited (~138). At the moment this might not be a problem as probably no one has that much tokens in a container. However, this endpoint could also be used in the future to delete multiple tokens from the token list. The users than could filter for specific tokens, select them and delete them at once, e.g. orphaned tokens.
     * POST /token/batchdeletion + serials in body data
         -> allows larger number of tokens

PR #4386, opened 2025-05-09
container enhancements

jelinaunger:
Working on #4208
* Added container and container info to extended policy conditions
* Added hide_container_info policy
* Allow to secure registration with user store password
* Added filter params for the container query, allow wildcards and made search case-insensitive for most parameters
* Added filter in the WebUI in the container list for the serial, type and description
* Added endpoint to remove all tokens from a container as single request


privacyidea/api/container.py
```diff
@@ -74,39 +75,67 @@ def list_containers():
...
+    :query container_serial: Serial of a single container (case-insensitive, can contain '*' as wildcards)
+    :query type: Type of the containers to return (case-insensitive, can contain '*' as wildcards)
+    :query token_serial: Serial of a token assigned to the container (case-insensitive, can contain '*' as wildcards)
+    :query template: Name of the template the container is created from (case-sensitive, can contain '*' as wildcards)
+    :query container_realm: Name of the realm the container is assigned to (case-insensitive, can contain '*' as
+        wildcards)
+    :query description: Description of the container (case-insensitive, can contain '*' as wildcards)
+    :query resolver: Resolver of the user assigned to the container  (case-insensitive, can contain '*' as wildcards)
+    :query assigned: Filter for assigned or unassigned containers (True or False)
+    :query info_key: Key of the container info (case-sensitive, can contain '*' as wildcards)
```
jelinaunger, line 87: `template` and `info_key` are the only case-sensitive parameters as e.g. a template with the name "test" and another one "Test" could exist. To get an exact match we need to be case-sensitive. Or should we be consistent and make all parameters case-insensitive?

PR #4404, opened 2025-05-22
use proxy for authz, naming

### LINKED ISSUE #4403: push notification google oauth2 dont use the https proxy
I setup the instance in a enviroment where only https Proxys are allowed for outgoing communication.
The application dont use httpsproxy for communicate with oauth2.googleapis.com.
I think

the  reason is in the class FirebaseProvider(ISMSProvider)

you are calling authed_session = AuthorizedSession(credentials) without 
session.proxies.update(proxies)

nilsbehlen:


privacyidea/lib/smsprovider/FirebaseProvider.py
```diff
@@ -103,85 +107,88 @@ def submit_message(self, firebase_token, data):
...
-        authed_session = AuthorizedSession(credentials)
+        proxies = {}
+        auth_request = None
+        if self.smsgateway.option_dict.get(FirebaseConfig.HTTPS_PROXY):
+            proxies["https"] = self.smsgateway.option_dict.get(FirebaseConfig.HTTPS_PROXY)
+            session = Session()
+            session.proxies.update(proxies)
+            auth_request = requests.Request(session=session)
+
+        if auth_request:
```
plettich, line 120: No need to check if `auth_request` is set, just pass it to `AuthorizedSession`, default is `None`

plettich (changes requested):
Just a minor tweak

plettich (approved):

Issue #4351, opened 2025-04-10
Display Token Initialization in Token-Info


katgirl:
**Is your request related to a problem? Please describe it.**
We recently had a case where we had to reallocate accounts. To identify the previously assigned tokens for these accounts, we had to navigate through the account management to the token information and then retrieve the initialization information via the audit log. This process was time-consuming and inefficient.
**What do you want to achieve?**
I want privacyIDEA to be enhanced so that the token initialization information and the time of the last logon with the token are displayed directly in the account information.
**Describe the desired solution**
A clear and concise description of what you want to achieve.
The desired solution is to extend privacyIDEA so that the following data is also directly visible when viewing user accounts:
Information about token initialization (e.g., when and how the token was assigned to the account)
The time of the last logon of the account with the respective token

nilsbehlen:
Hi, do you mean you want this info to be visible on the user details page, in the table of the users tokens?
Parts of this via #4220
We should be adding the creation date to the token info ASAP

katgirl:
Exactly, that's the point, whereby the creation is a different date than the assignment.
For example, we initialize the YubiKeys month/years in advance, but the assignment takes place much later. 
It would be nice to display not only in the WebUI, but also via API.

PR #4412, opened 2025-05-27
add assignment date

nilsbehlen:
#4351

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,8 +1,11 @@
 Version 3.11.3 TBD
 
+Enhancements:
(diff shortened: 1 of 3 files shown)


jelinaunger (commented):
Should we delete the assignment date when the user is unassigned?

nilsbehlen:
nooo, because that is more work and this is meant as a simple enhancement. also, the information ("last" assignment_date in case of token without user) *might* be useful.

jelinaunger (approved):

PR #4439, opened 2025-06-10
4293/renaming a policy

henrik-falk:
Front-end solution with two buttons, so that cloning a policy is still possible.
Maybe the button should be next to the delete button.

### DIFF
--- a/privacyidea/static/components/config/controllers/configControllers.js
+++ b/privacyidea/static/components/config/controllers/configControllers.js
@@ -320,95 +320,91 @@ myApp.controller("policyDetailsController", ["$scope", "$stateParams",
             pinode: []
         };
 
-        $scope.createPolicy = function () {
-            // This is called to save the policy
-            // get scope
-            var scope = $scope.selectedScope[0].name;
-            var realms = [];
-            var resolvers = [];
-            var adminRealms = [];
-            var pinodes = [];
-            var actions = [];
-            $scope.params.scope = scope;
-            $scope.params.action = actions;
-            $scope.params.adminrealm = adminRealms;
-            // get actions
+        function _buildParamsForSave(targetName) {
+            const p = angular.copy($scope.params);
+            p.scope    = $scope.selectedScope[0].name;
+            p.name     = targetName;
+            p.action   = [];
+            p.realm    = [];
+            p.resolver = [];
+            p.adminrealm = [];
+            p.pinode   = [];
 
             if ($scope.isActionValues) {
-                // we need to process the value-actions
-                // iterate through the checkboxes
-                angular.forEach($scope.actionCheckBox, function (value, key) {
-                    if (value) {
-                        // The action is checked. So try to get an action value.
-                        // The type is given in the $scope.actions array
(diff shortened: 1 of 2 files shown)


privacyidea/api/policy.py
```diff
@@ -97,6 +97,24 @@ def disable_policy_api(name):
     g.audit_object.log({"success": True})
     return send_result(p)
 
+@policy_blueprint.route('/rename/<name>', methods=['POST'])
```
nilsbehlen: i think there are better options for the route, either:
1) `PATCH /<name>`  with body `{"name" : <new_name>}`
2) "move" the resource by creating a new one from it:
`POST/PUT /<new_name>` with body `{"from": <previous_name> }`

privacyidea/lib/policy.py
```diff
@@ -1351,6 +1352,30 @@ def validate_values(values: Union[str, list, None], allowed_values: list, name:
     return True
 
 
+@log_with(log)
+def rename_policy(name, newname):
```
nilsbehlen: type hints

privacyidea/api/policy.py
```diff
@@ -97,6 +97,24 @@ def disable_policy_api(name):
...
+def rename_policy_api(name=None):
+    """
+    Rename a given policy by its name.
+
+    :jsonparam name: The name of the policy
+    :jsonparam newname: The new name of the policy
+    :return: ID in the database
+    """
+    param = request.all_data
+    newname = get_required(param, "newname")
```
nilsbehlen: param should be new_name

privacyidea/lib/policy.py
```diff
@@ -1351,6 +1352,30 @@ def validate_values(values: Union[str, list, None], allowed_values: list, name:
...
+
+    :param name: The name of the policy to be renamed
+    :param newname: The new name of the policy
+    :return: The database ID of the renamed policy
+    :rtype: int
+    """
+    check_policy_name(newname)
+    p1 = Policy.query.filter_by(name=name).first()
+    if not p1:
+        raise ParameterError(f"Policy '{name}' does not exist!")
```
nilsbehlen: localize error texts with `_()` like so `raise ParameterError(_("Invalid policy name:") + f" {name}")`, but be aware of separating the variable and the fixed text which will be translated

privacyidea/api/policy.py
```diff
@@ -97,6 +97,25 @@ def disable_policy_api(name):
...
+    Rename an existing policy.
+
+    Only the policy’s name is modified; all other attributes remain unchanged.
+
+    :param name: Current name of the policy (from the URL).
+    :jsonparam name: New name to assign to the policy (in the JSON body).
+    :return: Database ID of the renamed policy.
+    """
+    new_name = get_required(request.all_data, "name")
+    check_policy_name(new_name)
```
nilsbehlen: `rename_policy` already does `check_policy_name` so no need to do it here aswell

nilsbehlen (approved):

PR #4458, opened 2025-06-26
4371/policy_to_disable_token_types

henrik-falk:
l remove the specified token types from the authentication in check_token_list

### DIFF
--- a/tests/test_api_validate.py
+++ b/tests/test_api_validate.py
@@ -3252,8 +3252,7 @@ class ValidateAPITestCase(MyApiTestCase):
         with self.app.test_request_context(
                 "/validate/check",
                 method="POST",
-                data={"user": "cornelius", "realm": self.realm1, "pass": "1"},
-        ):
+                data={"user": "cornelius", "realm": self.realm1, "pass": "1"}):
             res = self.app.full_dispatch_request()
             self.assertEqual(res.status_code, 200, res)
             result = res.json["result"]


privacyidea/api/auth.py
```diff
@@ -133,6 +133,7 @@ def before_request():
 @prepolicy(pushtoken_disable_wait, request)
 @prepolicy(webauthntoken_request, request=request)
 @prepolicy(webauthntoken_authz, request=request)
+@prepolicy(disabled_token_types, request=request)
```
plettich, line 136: I think it makes more sense to decorate the [`check_user_pass()`](https://github.com/privacyidea/privacyidea/blob/1e891b71a080c28ca7137e3c78c72f52e418af43/privacyidea/lib/token.py#L2310) functions since this is the one called by `/auth` and `/validate/check`

privacyidea/api/lib/prepolicy.py
```diff
@@ -2688,3 +2688,21 @@ def rss_age(request, action):
...
+
+    :param request: The request object
+    :param action: The action parameter is not used in this decorator
+    :return: True
+    """
+    disabled = Match.user(g, scope=SCOPE.AUTH, action=ACTION.DISABLED_TOKEN_TYPES,
+                             user_object=request.User if hasattr(request, 'User') else None).action_values(unique=False)
+
+    if disabled:
+        request.all_data[ACTION.DISABLED_TOKEN_TYPES] = list(disabled)
```
plettich, line 2706: Maybe we should also *remove* a given entry from `all_data` if the policy isn't set.

privacyidea/api/auth.py
nilsbehlen: validate check can also all check_serial_pass so you would also use it on multiple functions

privacyidea/api/lib/prepolicy.py
nilsbehlen: what?

henrik-falk: set empty list or pop (check one commit earlier)

nilsbehlen: just set an emtpy list

jona-samuel (approved):

tests/test_api_validate.py
```diff
@@ -3229,6 +3229,95 @@ def test_37_challenge_response_hotp_with_container(self):
...
+            res = self.app.full_dispatch_request()
+            self.assertEqual(res.status_code, 200, res)
+            result = res.json["result"]
+            self.assertEqual(result["authentication"], "ACCEPT")
+
+        # Set a policy to trigger challenge response for HOTP
+        set_policy(
+            name="challenge_response",
+            scope=SCOPE.AUTH,
+            action="{0!s}=hotp".format(ACTION.CHALLENGERESPONSE)
```
nilsbehlen: f-string, this could probably be on a single line then

Issue #4370, opened 2025-04-24
Duplicated comparison implementation


jelinaunger:
There are two implementations for comparison functions
* lib/utils/compare: used for policy conditions
* lib/utils/\_\_init\_\_: used for event handler conditions

It should be possible to use one implementation for both, policy and event handler conditions. Especially, when adding extended (policy) conditions also to the event handlers #4201

Additionally, the error handling is not consistent. In lib/utils/compare` _compare_contains()` would raise a `CompareError` if the left value is not a list. But `_compare_bigger()` / `compare_smaller()` would raise a `ValueError` if the given values can not be converted to integers. Actually this is not a problem, since all errors are catched from the policy functions anyway. But catching the error already in the compare function would enable to write more concise error messages to the log.

PR #4470, opened 2025-07-08
Added policy comparators

### LINKED ISSUE #4380: Add/Extend policy comparators
**Compare Dates**
Some attributes in the token / container (info) are dates (e.g. last_authentication, creation_date, last_synchronization). It might be useful to allow policy conditions on these fields. There are already event handler conditions on dates, e.g. `last_auth`, `container_last_synchronization`. A function to compare times is already available here:
https://github.com/privacyidea/privacyidea/blob/f45652b22d47e20022c793012a665a106c9e422f/privacyidea/lib/utils/__init__.py#L917-L936

**String contains**
The actual comparators `contains` and `in` only allow lists. It would be nice to also allow strings or add a new comparator `string_contains`. This comparator should evaluate whether a string contains a particular substring.

jelinaunger:
closes #4380

### DIFF
--- a/privacyidea/lib/utils/compare.py
+++ b/privacyidea/lib/utils/compare.py
@@ -33,6 +33,8 @@ import re
 from functools import wraps
 from typing import Union
 
+from dateutil.parser import parse
+
 from privacyidea.lib.framework import _
 from privacyidea.lib.utils import parse_timedelta
 
@@ -185,7 +187,10 @@ def _get_datetime(date_time: Union[str, datetime.datetime]) -> datetime.datetime
     """
     if isinstance(date_time, str):
         try:
-            date_time = datetime.datetime.fromisoformat(date_time)
+            # Parse the string into a datetime object (accepts also non-iso formats)
+            # TODO : use datetime.datetime.fromisoformat() would be better, but this can not handle the last_auth
+            #  datetime format in the tokeninfo for python 3.9 and 3.10
+            date_time = parse(date_time)
         except ValueError as error:
             log.error(f"Invalid date format '{date_time}': {error}")
             raise CompareError(f"Invalid date format: {date_time!r}. Expected ISO format.")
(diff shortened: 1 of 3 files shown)


privacyidea/lib/utils/compare.py
```diff
@@ -40,14 +47,52 @@ class CompareError(Exception):
...
 
     def __repr__(self):
-        return "CompareError({!r})".format(self.message)
+        return f"CompareError({self.message!r})"
+
+
+#: This class enumerates all available comparators.
+#: In order to add a comparator to this module, add a suitable member to COMPARATORS
+#: and suitable entries to COMPARATOR_FUNCTIONS and COMPARATOR_DESCRIPTIONS.
+class Comparators():
```
nilsbehlen: no `()`

privacyidea/lib/utils/compare.py
```diff
@@ -40,14 +47,52 @@ class CompareError(Exception):
...
+    STRING_CONTAINS = "string_contains"
+    STRING_NOT_CONTAINS = "!string_contains"
 
-def parse_comma_separated_string(input_string):
+    @classmethod
+    def get_all_comparators(cls) -> list[str]:
+        """
+        Return a list of all comparators.
+        """
+        return [cls.EQUALS, cls.NOT_EQUALS, cls.CONTAINS, cls.NOT_CONTAINS, cls.MATCHES, cls.NOT_MATCHES,
```
nilsbehlen, line 90: did you try an automated approach like `vars()` or `inspect.getmembers()`?

jelinaunger: Yes, I'm actually working on it for the issue #4370. Here I just moved the existing class to the top for some reason which turned out to be pointless.

nilsbehlen (approved):

PR #4474, opened 2025-07-10
comparison duplicated implementations

### LINKED ISSUE #4370: Duplicated comparison implementation

jelinaunger:
closes #4370


privacyidea/lib/utils/compare.py
```diff
@@ -280,99 +272,354 @@ def _compare_string_contains(text: str, comparator: str, substring: str) -> bool
...
+    """
+    condition = parse_condition(condition)
+    key = condition.left_value
+    comparator = condition.comparator
+    right_value = condition.right_value
+
+    if right_value is None or not key or not comparator:
+        # There is a condition, but we do not know it!
+        log.warning(warning.format(condition))
+        raise CompareError("Condition not parsable.")
```
jelinaunger: In the original code we raise an Exception here
https://github.com/privacyidea/privacyidea/blob/cebe997cb0a9765e45224b5ad8b456de60927295/privacyidea/lib/utils/__init__.py#L917

This function is used by the event handler. Usually, we do  not want to abort a request if something in the event handler went wrong. So I'm not sure whether it is really a good idea to throw the exception here.

nilsbehlen: is this not covered by handle_missing_data?

jelinaunger: handle_missing_data is only for policy conditions

plettich: When using the `CompareError` we can catch it explicitly in the event handler and continue.

PR #4481, opened 2025-07-14
token import and export rebase

jona-samuel:
New im-/export for tokens
- [x] 4Eyes
- [x] ApplicationSpecificPassword
- [x] Certificate
- [x] DayPassword
- [x] EMail
- [x] HOTP
- [x] Indexed Secret
~mOTP~
- [x] OCRA
- [x] Paper
- [x] Push
- [x] Password
- [x] Questionnaire
- [x] Registration
~RADIUS~
~Remote~ 
- [x] SMS
- [x] Spass
- [x] SSH
- [x] TAN
- [x] TiQR
- [x] TOTP
~U2F~
~VASCO~
- [x] WebAuthn
- [x] Yubikey
~Yubico~

### DIFF
--- a/privacyidea/lib/token.py
+++ b/privacyidea/lib/token.py
@@ -60,12 +60,14 @@ This is the middleware/glue between the HTTP API and the database
 """
 import base64
 import datetime
+import json
 import logging
 import os
 import random
 import string
 import traceback
 from collections import defaultdict
+from dataclasses import dataclass
 from typing import Union
 
 from dateutil.tz import tzlocal
@@ -138,6 +140,13 @@ class clob_to_varchar(FunctionElement):
 @compiles(clob_to_varchar)
 def fn_clob_to_varchar_default(element, compiler, **kw):
     return compiler.process(element.clauses, **kw)
+
+
+@dataclass(frozen=True)
+class TokenImportResult:
+    successful_tokens: list[str]
+    updated_tokens: list[str]
+    failed_tokens: list[str]
 
 
 @compiles(clob_to_varchar, 'oracle')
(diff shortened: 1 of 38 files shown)


privacyidea/lib/token.py
```diff
@@ -3044,3 +2975,53 @@ def regenerate_enroll_url(serial: str, request: Request, g) -> Union[str, None]:
...
 
     return enroll_url
+
+def export_tokens(tokens: list[TokenClass]) -> str:
+    """
+    Takes a list of tokens and returns an exportable JSON string.
+    :param tokens: list of token objects
+    :return: JSON string representing a list of token dictionaries
+    """
+    exported_tokens = []
```
nilsbehlen: can be done with list comprehension:
`exported_token = [token.export_token() for token in tokens]`

privacyidea/lib/token.py
```diff
@@ -3044,3 +2975,53 @@ def regenerate_enroll_url(serial: str, request: Request, g) -> Union[str, None]:
...
+    """
+    successful_tokens = []
+    updated_tokens = []
+    failed_tokens = []
+    try:
+        tokens = json.loads(tokens)
+        for token_info_dict in tokens:
+            serial = token_info_dict.get("serial")
+            try:
+                serial_not_exists, a = check_serial(serial)  # replace check_serial
```
nilsbehlen: so should `check_serial` be replaced? if so, with what?

privacyidea/lib/tokenclass.py
```diff
@@ -2008,3 +1997,35 @@ def get_enroll_url(self, user: User, params: dict):
...
+        Create a dictionary with the token information that can be exported.
+        """
+        token_dict = {
+            "type": self.type.lower(),
+            "issuer": "privacyIDEA",
+            "description": self.token.description,
+            "serial": self.token.serial,
+            "otpkey": self.token.get_otpkey().getKey().decode("utf-8"),
+            "otplen": self.token.otplen,
+            "_hashed_pin": self.token.pin_hash,
```
nilsbehlen: why _ prefix?

privacyidea/lib/tokenclass.py
```diff
@@ -2008,3 +1997,35 @@ def get_enroll_url(self, user: User, params: dict):
...
+        """
+        try:
+            self.token.set_otpkey(token_information.setdefault("otpkey", ''))
+            self.token.otplen = int(token_information.setdefault("otplen", 6))
+            self.token.description = token_information.setdefault("description", '')
+            self.token.pin_hash = token_information.setdefault("_hashed_pin", None)
+            self.add_tokeninfo_dict(token_information.setdefault("tokeninfo", {}))
+            self.add_tokeninfo("import_date", datetime.now(timezone.utc).isoformat(timespec="seconds"))
+            self.save()
+        except Exception as exx:
```
nilsbehlen: how does the caller know the import was successful if the exception is caught here and the function does not return anything?
Either do not catch the exception here and let the caller catch it or return a boolean to indicate failure or success.
the try-catch indicates something can go wrong, in contrast to export_token which is assumed to always work, so the function needs to pass this information to the caller in some form.

privacyidea/lib/token.py
```diff
@@ -3044,3 +2975,53 @@ def regenerate_enroll_url(serial: str, request: Request, g) -> Union[str, None]:
...
+                    db_token = Token(serial, tokentype=token_type.lower())
+                    token = create_tokenclass_object(db_token)
+                    token.import_token(token_info_dict)
+                    successful_tokens.append(serial)
+                elif update_existing_tokens:
+                    token = get_one_token(serial=serial)
+                    token.import_token(token_info_dict)
+                    updated_tokens.append(serial)
+                else:
+                    raise TokenAdminError(f"Token with serial {serial} already exists. "
```
nilsbehlen: reaching this will cause the function to return. what if the next token could be processed properly?
The tokens that were processed up to this point are saved already, so we would get a "broken" result.
For this function to produce a clear result, i think it would be better to handle this like the exception below this: log and append to failed_tokens.

tests/test_lib_token.py
```diff
@@ -1751,6 +1752,90 @@ def test_60_get_enroll_url(self):
...
+        self.assertIn('"type": "hotp"', exported_tokens)
+        self.assertIn('"serial": "OATH12345678"', exported_tokens)
+        self.assertIn('"type": "totp"', exported_tokens)
+        self.assertIn('"serial": "TOTP12345678"', exported_tokens)
+
+        # Remove the tokens
+        hotptoken.delete_token()
+        totptoken.delete_token()
+
+        # Impoert tokens
```
nilsbehlen: import

tests/test_lib_token.py
```diff
@@ -1751,6 +1752,90 @@ def test_60_get_enroll_url(self):
...
+        self.assertEqual(len(imported_tokens.failed_tokens), 0)
+        self.assertEqual(len(imported_tokens.updated_tokens), 0)
+        self.assertIn('OATH12345678', imported_tokens.successful_tokens)
+        self.assertIn('TOTP12345678', imported_tokens.successful_tokens)
+        self.assertEqual(hotptoken.token.serial, "OATH12345678")
+        self.assertEqual(hotptoken.token.tokentype, "hotp")
+        self.assertEqual(totptoken.token.serial, "TOTP12345678")
+        self.assertEqual(totptoken.token.tokentype, "totp")
+
+        hotptoken.set_description("this will be replaced by the import")
```
nilsbehlen, line 1811: i think a comment here would be nice, explaining that the import with update enabled will overwrite tokeninfo of existing token. but maybe i'm just slow and need some time understand the code

nilsbehlen (approved):

PR #4494, opened 2025-07-17
force server generate

### LINKED ISSUE #4410: Add force_server_generate for totp/hotp token in policies (admin scope)
I would like to force the generation of a TOTP/HOTP key with a policy in the admin scope when enrolling a TOTP/HOTP token. 

There is currently no option in the admin scope as far as I can see. For example, the daypassword token has this option (daypassword_ force_server_generate).

jelinaunger:
closes #4410

if `otpkey` and `genkey` are not provided, `genkey` is set to `True` instead of raising a `ParameterError`

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -1605,7 +1605,7 @@ def check_token_init(request=None, action=None):
     return True
 
 
-def force_server_generate_key(request=None, action=None):
+def force_server_generate_key(request: Request, action=None):
     """
     Checks if for the given token type a policy to force the server to generate the key is set.
 
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -1617,8 +1617,7 @@ def force_server_generate_key(request: Request, action=None):
     tokentype = params.get("type", "HOTP")
     action = f"{tokentype.lower()}_{ACTION.FORCE_SERVER_GENERATE}"
     force_genkey = Match.admin_or_user(g, action=action, user_obj=request.User).allowed()
-    if force_genkey:
-        g.policies[action] = True
+    g.policies[action] = force_genkey
 
     return True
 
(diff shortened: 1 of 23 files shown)


privacyidea/api/lib/prepolicy.py
```diff
@@ -1605,6 +1605,24 @@ def check_token_init(request=None, action=None):
     return True
 
 
+def force_server_generate_key(request=None, action=None):
```
plettich: The `request` should not be `None` by default

privacyidea/lib/tokens/hotptoken.py
```diff
@@ -351,14 +353,16 @@ def update(self, param, reset_failcount=True):
 
         # check if the key_size is provided
         # if not, we could derive it from the hashlib
-        key_size = getParam(upd_param, 'key_size', optional) \
-                   or getParam(upd_param, 'keysize', optional)
+        key_size = upd_param.get('key_size') or upd_param.get('keysize')
```
plettich: I think we can get rid of `key`size` here, the parameter name is `keysize` as described in the token API docs

privacyidea/lib/tokens/motptoken.py
```diff
@@ -168,7 +179,11 @@ def update(self, param, reset_failcount=True):
         :return: nothing
         """
         if self.hKeyRequired is True:
```
plettich: This condition is actually always true since `hKeyRequired` is set to `True` in the constructor.

privacyidea/lib/tokens/passwordtoken.py
```diff
@@ -49,6 +50,7 @@ class PasswordTokenClass(TokenClass):
     password_detail_key = "password"  # nosec B105 # key name
     default_length = DEFAULT_LENGTH
     default_contents = DEFAULT_CONTENTS
+    desc_key_gen = lazy_gettext("Force the key to be generated on the server.")
```
plettich: We should probably define this as a module level variable in `hotptoken.py` and import it here or in the ApplSpec token module

plettich (dismissed):

nilsbehlen (approved):

PR #4508, opened 2025-07-23
Allow to unassign non-existing users from containers

jelinaunger:
working on #4208

### DIFF
--- a/privacyidea/api/container.py
+++ b/privacyidea/api/container.py
@@ -195,8 +195,8 @@ def unassign(container_serial):
     if user_id:
         user.uid = str(user_id)
 
-    # Check required parameters
-    get_required_one_of(request.all_data, ["user", "user_id"])
+    # Check if required parameter is present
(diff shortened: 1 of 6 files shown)


privacyidea/api/container.py
```diff
@@ -180,12 +180,26 @@ def assign(container_serial):
...
-    user = get_user_from_param(request.all_data, required)
+    # Get user
+    user = request.User
+    # The user id is not set in before_request, but is required to remove users that do not exist (anymore)
+    user_id = request.all_data.get("user_id", None)
+    if user_id:
+        user.uid = str(user_id)
+
+    # Check required parameters
+    get_required_one_of(request.all_data, ["user", "user_id"])
```
nilsbehlen: what does this line mean here?

jelinaunger: We either need the user name or the user id. This is checked here and an error is raised if both are not provided. We can also write `if not user.login and not user_id: raise ParameterError()` if this is more clear

nilsbehlen (approved):

PR #4517, opened 2025-07-28
Add policy to use the 'pia' scheme for push tokens

### LINKED ISSUE #2882: Different URL scheme for push tokens
When trying to invoke a push token url (i.e. the URL that's embeded in the QR code) on an iPhone, Safari opens the url in Microsoft AUthenticator, which is unable to use the push token.

Have a different specific URL scheme for privacy IDEA, e.g. pia://

jelinaunger:
closes #2882

### DIFF
--- a/privacyidea/api/lib/policyhelper.py
+++ b/privacyidea/api/lib/policyhelper.py
@@ -106,7 +106,7 @@ def get_pushtoken_add_config(g, params=None, user_obj=None):
     if len(firebase_config) == 1:
         params[PUSH_ACTION.FIREBASE_CONFIG] = list(firebase_config)[0]
     else:
-        raise PolicyError("Missing enrollment policy for push token: {0!s}".format(PUSH_ACTION.FIREBASE_CONFIG))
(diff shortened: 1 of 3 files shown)


privacyidea/lib/container.py
```diff
@@ -669,11 +672,13 @@ def create_container_tokens_from_template(container_serial: str, template_tokens
             tantoken_count(request, None)
             pushtoken_add_config(request, None)
             indexedsecret_force_attribute(request, None)
+            force_server_generate_key(request, None)
         except Exception as ex:
             log.warning(f"Error checking pre-policies for token {token_info} created from template: {ex}")
             continue
 
         init_params = request.all_data
+        init_params["policies"] = g.policies
```
plettich, line 681: is there a reason to use a `deepcopy()` above and not here?

jelinaunger: Above the idea was to not permanently write the policies in `request.all_data`. Here request.all_data will be reset to the original data from the container request anyway and `init_params` is initialized every iteration. So there is no need for a deep copy.

PR #4525, opened 2025-07-31
Close ldap connection after each request

### LINKED ISSUE #4148: LDAP resolver does not close connections
https://community.privacyidea.org/t/ldapidresolver-seems-to-keep-ldap-connections-open-indefinetly/2984/8

jelinaunger:
closes #4148

### DIFF
--- a/privacyidea/api/before_after.py
+++ b/privacyidea/api/before_after.py
@@ -30,7 +30,7 @@ import copy
 from .lib.utils import (send_error, get_all_params, verify_auth_token, get_optional)
 from .container import container_blueprint
 from ..lib.container import find_container_for_token, find_container_by_serial
-from ..lib.framework import get_app_config_value, get_request_local_store
+from ..lib.framework import get_app_config_value
 from ..lib.user import get_user_from_param
 import logging
 from flask import request, g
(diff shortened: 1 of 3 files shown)


privacyidea/api/before_after.py
```diff
@@ -469,6 +469,11 @@ def after_request(response):
     This function is called after a request
     :return: The response
     """
+    # close any open ldap connection
+    store = get_request_local_store()
+    resolvers = store.get("resolver_objects", {})
+    for resolver in resolvers.values():
+        resolver.unbind()
```
plettich: In `lifecycle.py` we implemented finalizers for this task:
https://github.com/privacyidea/privacyidea/blob/8c6813775f9b3c4f140c55892f61ffd920f00257/privacyidea/lib/lifecycle.py#L25
They are used in the `SQLIdResolver.py` to close the db session:
https://github.com/privacyidea/privacyidea/blob/db3f1ace8ce0a59918e3440719556bfbaae857f0/privacyidea/lib/resolvers/SQLIdResolver.py#L497
The finalizers are called when the request is torn down:
https://github.com/privacyidea/privacyidea/blob/c2158e37e5fa8397c02a7aae1d14527951d0817f/privacyidea/api/before_after.py#L104
We could set the finalizer to `self.connection.unbind`.

plettich (changes requested):

plettich (approved):

PR #4534, opened 2025-08-05
Fix line numbers in logging decorator

plettich:
The logging decorator did not print the correct line numbers of the wrapped functions, instead the line number of the decorator functions was used.
By inspecting the wrapped function and adding extra data to the log record, we can provide the correct values in the formatter.

This PR also suppresses the output of the compete policies when checking for a match (only the policy name is given.

The SecureFormatter class now works with printable characters (Unicode).

### DIFF
--- a/privacyidea/lib/log.py
+++ b/privacyidea/lib/log.py
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: (C) 2014 NetKnights GmbH <https://netknights.it>
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+#
 #  privacyIDEA is a fork of LinOTP
 #  May 08, 2014 Cornelius Kölbel
 #  License:  AGPLv3
@@ -23,10 +27,10 @@
 # License along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 from logging import Formatter
-import string
 import logging
 import functools
 from copy import deepcopy
+
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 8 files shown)


privacyidea/lib/log.py
```diff
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: (C) 2014 NetKnights GmbH <https://netknights.it>
```
jelinaunger: When adapting the license header anyway, this should be the new one for consistency (e.g., as in passkeytoken.py)

privacyidea/lib/policy.py
```diff
@@ -846,6 +851,7 @@ def match_policies(self, name: Optional[str] = None, scope: Optional[str] = None
         :param adminuser: see ``list_policies``
         :param time: return only policies that are valid at the specified time.
             Defaults to the current time.
+        :type time: datetime or None
```
jelinaunger, line 853: is already contained in the type hints. I think we do not need both?

tests/test_lib_tokens_remote.py
```diff
@@ -171,37 +171,55 @@ def test_08_authenticate_local_pin(self):
         self.assertFalse(r[0], r)
         self.assertTrue(r[1] == -1, r)
         self.assertTrue(r[2].get("message") == "Wrong PIN", r)
+        # right PIN
+        r = token.authenticate(self.otppin+"123456")
+        self.assertTrue(r[0], r)
+        self.assertTrue(r[1] >= 0, r)
+        self.assertTrue(r[2].get("message") == "matching 1 tokens", r)
```
jelinaunger: assertGraterEqual and assertEqual

privacyidea/lib/log.py
plettich: Have we already decided on a consistent license text? Is it written down somewhere?

nilsbehlen: yes in passkeytoken.py :D

privacyidea/lib/policy.py
plettich: Somehow my PyCharm complained when calling the function...
However when the code documentation is generated, the `type` is added to the parameter description. In my opinion this is more readable than the type hints at the function declaration.
grafik
grafik

nilsbehlen: we should work towards removing the parameters and types from the comments as they are redundant and useless in contrast to type hints. If the documentation requires them, we need to change how the documentation is built.

plettich: Can be done with https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints
grafik

jelinaunger (approved):

PR #4536, opened 2025-08-06
unassign users from deleted realms

### LINKED ISSUE #4530: Unassign user from deleted / non-existing realm
Deleting a realm while users are still assigned to tokens/containers causes some issues:

- [x] list containers throws an error as we try to set the users realm, but the realm object is None
- [x] unassignment does not work when passing the username but no realm as the realm will be set to the default realm (only pass user_id and resolver solves the problem)
- [ ] list tokens does not contain tokens with an owner in a non existing realm
       In the token query the tokens are filtered by the resolvers contained in valid realms of this node:
       https://github.com/privacyidea/privacyidea/blob/0bd6ff73b07682919bc45d43facb61d5338a98f2/privacyidea/lib/token.py#L368-L389
       Since the deleted realm will  not appear in the list, also the contained resolvers are not in the list and are not added to the list of valid resolvers.

jelinaunger:
closes #4530


privacyidea/lib/policydecorators.py
```diff
@@ -248,11 +248,13 @@ def auth_user_does_not_exist(wrapped_function, user_object, passw, options=None)
     if g:
         pass_no_user = Match.user(g, scope=SCOPE.AUTH, action=PolicyAction.PASSNOUSER,
                                   user_object=user_object).policies(write_to_audit_log=False)
-        if pass_no_user:
-            # Check if user object exists
-            if not user_object.exist():
+        if not user_object.exist():
+            if pass_no_user:
+                # Check if user object exists
```
plettich: The comment seems off, the check was done before.
Does this check have something to do with the resolver?

privacyidea/lib/tokenclass.py
```diff
@@ -274,8 +274,9 @@ def owners(self):
         """
         user_objects = []
         for tokenowner in self.token.all_owners:
+            realm_name = tokenowner.realm.name if tokenowner.realm else None
```
plettich, line 277: This is the fallback if a realm has already been deleted? Otherwise a TokenOwner entry without a realm should not exist, correct?

tests/test_api_auth.py
```diff
@@ -771,6 +772,57 @@ def test_09_auth_disabled_token_types(self):
...
+            self.assertEqual(401, res.status_code, res.json)
+            error = res.json.get("result").get("error")
+            self.assertEqual(4031, error.get("code"), error)
+            self.assertEqual(f"Authentication failure. Unknown realm: {user.realm}.", error.get("message"), error)
+
+        with self.app.test_request_context('/auth', method="POST",
+                                           data={"username": user.login, "resolver": user.resolver,
+                                                 "password": "1234"}):
+            res = self.app.full_dispatch_request()
+            self.assertEqual(401, res.status_code, res.json)
```
plettich, line 809: So this fails because the realm does not exist? Probably correct behaviour and the error message comes from the check in the authentication flow? Do we have a log message that the realm could not be found?

tests/test_api_auth.py
```diff
@@ -771,6 +772,57 @@ def test_09_auth_disabled_token_types(self):
         delete_policy("disable_spass_token")
         delete_policy("piLogin")
 
+    def test_10_auth_with_deleted_realm(self):
+        self.setUp_user_realms()
+        self.setUp_user_realm3()
```
plettich, line 777: do we need realm3 here? Is this to check that the user can not authenticate when the other realm is removed?

tests/test_api_lib_utils.py
```diff
@@ -149,6 +149,7 @@ def test_04_check_jwt_username_in_audit(self):
         # also used in the correct way for policy handling.
         with open("tests/testdata/jwt_sign.key", "r") as f:
             key = f.read()
+        self.setUp_user_realms()
```
plettich, line 152: why do we need to set up a realm here? `userA` does not exist in this realm.

tests/test_api_auth.py
```diff
@@ -771,6 +772,57 @@ def test_09_auth_disabled_token_types(self):
...
+        set_default_realm(self.realm3)
+        # User exist in realm1 (default realm) and realm3
+        user = User("cornelius", self.realm1)
+        token = init_token({"type": "spass", "pin": "1234"}, user=user)
+        user_realm1 = User("hans", self.realm1)
```
plettich, line 782: user_realm1 and token_realm1 are not used

privacyidea/lib/tokenclass.py
jelinaunger: Yes

tests/test_api_auth.py
jelinaunger: Yes, realm1 is deleted and there should be a default realm (realm3), cause I think most people still have a default realm defined. In before_request the default realm will be set if no realm is given. This should also be covered by the tests.

tests/test_api_auth.py
jelinaunger: If we do not pass a realm, it will use the default realm. In this case a user with the same name exist in the default realm, but without a token. Hence, authentication fails. (The resolver parameter has no effect on the auth endpoint, thats only for validate/check. my fault)

plettich (approved):

PR #4564, opened 2025-08-11
Batch-delete and batch-unassign selected tokens

henrik-falk:
Use existing batch-deletion endpoint,
Implement batch-unassign endpoint,
Add unassign selected tokens UI button,
Backend Tests

### DIFF
--- a/privacyidea/api/token.py
+++ b/privacyidea/api/token.py
@@ -574,6 +574,36 @@ def unassign_api():
     g.audit_object.log({"success": True})
     return send_result(res)
 
+@token_blueprint.route('/batchunassign', methods=['POST'])
+@prepolicy(check_token_list_action, request, action=PolicyAction.UNASSIGN)
+@event("token_unassign", request, g)
+@log_with(log)
+def batch_unassign_api():
+    """
+    Unassign all passed tokens, e.g. all tokens of a container
+    All errors during the unassignment of a token are fetched to be able to unassign the remaining tokens.
+
+    :jsonparam serial: A comma separated list of token serials to unassign
+    :return: Dictionary with the serials as keys and the success status of the unassignment as values
+    """
+    serial_list = getParam(request.all_data, "serial", required)
+    serial_list = serial_list.replace(" ", "").split(",")
+    g.audit_object.log({"serial": serial_list})
+    ret = {}
+    for serial in serial_list:
+        try:
+            success = unassign_token(serial)
+        except Exception as ex:
+            # We are catching the exception here to be able to unassign the remaining tokens
+            log.error(f"Error unassigning token {serial}: {ex}")
+            success = False
+        ret[serial] = success
+
+    not_authorized_serials = getParam(request.all_data, "not_authorized_serials", default=[])
+    res = add_not_authorized_tokens_result(ret, not_authorized_serials)
+
+    return send_result(res)
+
 
 @token_blueprint.route('/revoke', methods=['POST'])
 @token_blueprint.route('/revoke/<serial>', methods=['POST'])
@@ -686,7 +716,7 @@ def delete_api(serial):
 @prepolicy(check_token_list_action, request, action=PolicyAction.DELETE)
 @event("token_delete", request, g)
 @log_with(log)
-def batch_deletion():
+def batch_deletion_api():
     """
     Delete all passed tokens, e.g. all tokens of a container
     All errors during the deletion of a token are fetched to be able to delete the remaining tokens.
--- a/privacyidea/static_new/src/app/components/token/token-card/token-tab/token-tab.component.ts
+++ b/privacyidea/static_new/src/app/components/token/token-card/token-tab/token-tab.component.ts
@@ -10,7 +10,6 @@ import {
   concatMap,
   EMPTY,
   filter,
-  forkJoin,
   from,
   reduce,
   switchMap,
@@ -160,11 +159,7 @@ export class TokenTabComponent {
       .subscribe({
         next: (result) => {
           if (result) {
-            forkJoin(
-              selectedTokens.map((token) =>
-                this.tokenService.deleteToken(token.serial),
-              ),
-            ).subscribe({
+            this.tokenService.batchDeleteTokens(selectedTokens).subscribe({
               next: () => {
                 this.tokenService.tokenResource.reload();
               },
@@ -207,9 +202,6 @@ export class TokenTabComponent {
                 : assign$;
             }),
             reduce(() => null, null),
-            switchMap(() =>
-              this.tokenService.getTokenDetails(this.tokenSerial()),
-            ),
           ),
         ),
         tap(() => this.tokenService.tokenResource.reload()),
@@ -219,5 +211,34 @@ export class TokenTabComponent {
         }),
       )
       .subscribe();
+  }
+
+  unassignSelectedTokens() {
+    const selectedTokens = this.tokenSelection();
+    this.dialog
+      .open(ConfirmationDialogComponent, {
+        data: {
+          serial_list: selectedTokens.map((token) => token.serial),
+          title: 'Unassign Tokens',
+          type: 'token',
+          action: 'unassign',
+          numberOfTokens: selectedTokens.length,
+        },
+      })
+      .afterClosed()
+      .subscribe({
+        next: (result) => {
+          if (result) {
+            this.tokenService.batchUnassignTokens(selectedTokens).subscribe({
+              next: () => {
+                this.tokenService.tokenResource.reload();
+              },
+              error: (err) => {
+                console.error('Error unassigning tokens:', err);
+              },
+            });
+          }
+        },
+      });
   }
 }
--- a/tests/test_api_token.py
+++ b/tests/test_api_token.py
@@ -17,43 +17,45 @@
 # You should have received a copy of the GNU Affero General Public
 # License along with this program. If not, see <http://www.gnu.org/licenses/>.
 
+import codecs
+import datetime
+import json
+import unittest
+from urllib.parse import urlencode, quote
+
+import pytest
+import requests
+from dateutil.tz import tzlocal
+from mock import mock
+
+from privacyidea.lib import _
+from privacyidea.lib.caconnector import save_caconnector
+from privacyidea.lib.caconnectors.baseca import AvailableCAConnectors
+from privacyidea.lib.caconnectors.msca import ATTR as MS_ATTR
+from privacyidea.lib.caconnectors.msca import MSCAConnector
+from privacyidea.lib.config import set_privacyidea_config, delete_privacyidea_config
 from privacyidea.lib.container import (init_container, add_token_to_container,
                                        find_container_by_serial, find_container_for_token)
 from privacyidea.lib.error import ResourceNotFoundError
-from .base import MyApiTestCase, PWFILE2
-import json
-import datetime
-import codecs
-from mock import mock
-import pytest
-import requests
+from privacyidea.lib.event import set_event, delete_event, EventConfiguration
+from privacyidea.lib.policies.actions import PolicyAction
 from privacyidea.lib.policy import (set_policy, delete_policy, SCOPE, enable_policy,
                                     PolicyClass)
-from privacyidea.lib.policies.actions import PolicyAction
+from privacyidea.lib.realm import set_realm
+from privacyidea.lib.resolver import save_resolver
+from privacyidea.lib.smsprovider.SMSProvider import (set_smsgateway,
+                                                     delete_smsgateway)
 from privacyidea.lib.token import (get_tokens, remove_token, get_one_token,
                                    get_tokens_from_serial_or_user, enable_token,
                                    check_serial_pass, unassign_token, init_token,
                                    assign_token, token_exist, add_tokeninfo)
-from privacyidea.lib.resolver import save_resolver
-from privacyidea.lib.realm import set_realm
-from privacyidea.lib.user import User
-from privacyidea.lib.event import set_event, delete_event, EventConfiguration
-from privacyidea.lib.caconnector import save_caconnector
-from urllib.parse import urlencode, quote
 from privacyidea.lib.tokenclass import DATE_FORMAT
 from privacyidea.lib.tokenclass import ROLLOUTSTATE
 from privacyidea.lib.tokens.hotptoken import VERIFY_ENROLLMENT_MESSAGE
 from privacyidea.lib.tokens.smstoken import SMSACTION
-from privacyidea.lib.config import set_privacyidea_config, delete_privacyidea_config
-from dateutil.tz import tzlocal
-from privacyidea.lib import _
-import unittest
-from privacyidea.lib.caconnectors.baseca import AvailableCAConnectors
-from privacyidea.lib.caconnectors.msca import MSCAConnector
+from privacyidea.lib.user import User
+from .base import MyApiTestCase, PWFILE2
 from .mscamock import CAServiceMock
-from privacyidea.lib.caconnectors.msca import ATTR as MS_ATTR
-from privacyidea.lib.smsprovider.SMSProvider import (set_smsgateway,
-                                                     delete_smsgateway)
 from .test_lib_tokens_certificate import REQUEST, CERTIFICATE
 
 # Mock for certificate from MSCA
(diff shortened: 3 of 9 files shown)


privacyidea/api/token.py
```diff
@@ -574,6 +574,36 @@ def unassign_api():
...
+@log_with(log)
+def batch_unassign_api():
+    """
+    Unassign all passed tokens, e.g. all tokens of a container
+    All errors during the unassignment of a token are fetched to be able to unassign the remaining tokens.
+
+    :jsonparam serial: A comma separated list of token serials to unassign
+    :return: Dictionary with the serials as keys and the success status of the unassignment as values
+    """
+    serial_list = getParam(request.all_data, "serial", required)
```
jelinaunger: use `get_required()`

privacyidea/api/token.py
```diff
@@ -686,7 +716,7 @@ def delete_api(serial):
 @prepolicy(check_token_list_action, request, action=PolicyAction.DELETE)
 @event("token_delete", request, g)
 @log_with(log)
-def batch_deletion():
+def batch_deletion_api():
```
jelinaunger: I think there is no need to include 'api' in the function name.

tests/test_api_token.py
```diff
@@ -532,6 +534,25 @@ def test_05_helpdesk_delete_batch(self):
 
         delete_policy("policy")
 
+    def test_06_unassign_batch(self):
+        set_policy(name="policy", scope=SCOPE.ADMIN, action=PolicyAction.UNASSIGN, realm=self.realm1)
+
+        # create tokens
+        token1 = init_token({"type": "hotp", "genkey": True, "realm": self.realm1},
```
jelinaunger: "genkey" is set to True by default (we changed this recently) so you do not need to set this explicitly. And realm1 is also set automatically since the user is from this realm. fyi

tests/test_api_token.py
```diff
@@ -532,6 +534,25 @@ def test_05_helpdesk_delete_batch(self):
...
+                            user=User("cornelius", self.realm1))
+        token2 = init_token({"type": "hotp", "genkey": True, "realm": self.realm2},
+                            user=User("hans", self.realm2))
+
+        token_serials = ",".join([token1.get_serial(), token2.get_serial()])
+
+        self.assertTrue(token1.user is not None)
```
jelinaunger: Please use self.assertIsNotNone()

tests/test_api_token.py
```diff
@@ -532,6 +534,25 @@ def test_05_helpdesk_delete_batch(self):
...
+        self.assertTrue(token2.user is not None)
+        # Try to unassign all tokens will only unassign token1 from realm1
+        self.request_assert_200("/token/batchunassign", {"serial": token_serials}, self.at, "POST")
+        self.assertTrue(token1.user is None)
```
jelinaunger: and self.assertIsNone()

tests/test_api_token.py
```diff
@@ -532,6 +534,25 @@ def test_05_helpdesk_delete_batch(self):
...
+        self.assertTrue(token2.user is not None)
+
+        delete_policy("policy")
```
jelinaunger, line 631: These are only the tests to check helpdesk rights are applied correctly. Please add a test in `APITokenTestCase` for general functionality.

privacyidea/api/token.py
```diff
@@ -574,6 +574,36 @@ def unassign_api():
...
+    :jsonparam serial: A comma separated list of token serials to unassign
+    :return: Dictionary with the serials as keys and the success status of the unassignment as values
+    """
+    serial_list = getParam(request.all_data, "serial", required)
+    serial_list = serial_list.replace(" ", "").split(",")
+    g.audit_object.log({"serial": serial_list})
+    ret = {}
+    for serial in serial_list:
+        try:
+            success = unassign_token(serial)
```
jelinaunger: This function returns 1 on success. I think we should return consistently bool values. (I know I also did this not in the batch deletion, and we should change this there as well)

privacyidea/api/token.py
```diff
@@ -574,6 +574,36 @@ def unassign_api():
...
+        except Exception as ex:
+            # We are catching the exception here to be able to unassign the remaining tokens
+            log.error(f"Error unassigning token {serial}: {ex}")
+            success = False
+        ret[serial] = success
+
+    not_authorized_serials = getParam(request.all_data, "not_authorized_serials", default=[])
```
nilsbehlen: `get_optional or []` or `request.all_data.get("...", [])  instead of `getParam`

privacyidea/static_new/src/app/components/token/token-card/token-tab/token-tab.component.ts
```diff
@@ -220,4 +212,33 @@ export class TokenTabComponent {
       )
       .subscribe();
   }
+
+  unassignSelectedTokens() {
+    const selectedTokens = this.tokenSelection();
+    this.dialog
+      .open(ConfirmationDialogComponent, {
+        data: {
+          serial_list: selectedTokens.map((token) => token.serial),
```
nilsbehlen: whats the naming convention for angular? CamelCase or snake_case? There is also numberOfTokens in data

tests/test_api_token.py
```diff
@@ -3353,6 +3354,32 @@ def test_62_init_token_with_force_genkey(self):
...
+
+        # create tokens
+        token1 = init_token({"type": "hotp"},
+                            user=User("cornelius", self.realm1))
+        token2 = init_token({"type": "hotp"},
+                            user=User("hans", self.realm2))
+
+        token_serials = ",".join([token1.get_serial(), token2.get_serial()])
+
+        self.assertIsNotNone(token1.user is not None)
```
nilsbehlen: `is not None` is redundant

tests/test_api_token.py
```diff
@@ -3374,6 +3354,32 @@ def test_62_init_token_with_force_genkey(self):
...
+                                           method='POST',
+                                           data={"serial": token_serials},
+                                           headers={'Authorization': self.at}):
+            res = self.app.full_dispatch_request()
+            self.assertEqual(200, res.status_code, res)
+        self.assertIsNone(token1.user)
+        self.assertIsNotNone(token2.user)
+
+        delete_policy("policy")
+
```
jelinaunger, line 3493: I would keep this test in the API000TokenAdminRealmList class. And here I would unassign at least three tokens: one without having a user assigned, one with an invalid serial and one valid token with a user assigned. And also check the result returned for each token.

privacyidea/api/lib/prepolicy.py
```diff
@@ -1308,6 +1308,42 @@ def check_token_action(request: Request = None, action: str = None):
...
+    serial_list = params.get("serials")
+
+    # Old API calls might provide the serials as a comma-separated list in the serial parameter
+    if serial is not None and ',' in serial:
+        serial_list = [s.strip() for s in serial.split(',') if s.strip()]
+        serial = None
+
+    if serial_list is not None:
+        if serial is not None and serial not in serial_list:
+            serial_list = serial_list + [serial]
```
nilsbehlen: `serial_list.append(serial)`

privacyidea/api/lib/prepolicy.py
```diff
@@ -1308,6 +1308,42 @@ def check_token_action(request: Request = None, action: str = None):
...
     serial = params.get("serial")
+    serial_list = params.get("serials")
+
+    # Old API calls might provide the serials as a comma-separated list in the serial parameter
+    if serial is not None and ',' in serial:
+        serial_list = [s.strip() for s in serial.split(',') if s.strip()]
+        serial = None
+
+    if serial_list is not None:
+        if serial is not None and serial not in serial_list:
```
nilsbehlen: `is not None` can be removed, just test for truthness, which includes checking for empty string

privacyidea/api/lib/prepolicy.py
```diff
@@ -1308,6 +1308,42 @@ def check_token_action(request: Request = None, action: str = None):
     user_attributes = UserAttributes(role, username, realm, resolver, adminuser, adminrealm)
     user_attributes.user = user if user else None
     serial = params.get("serial")
+    serial_list = params.get("serials")
+
+    # Old API calls might provide the serials as a comma-separated list in the serial parameter
+    if serial is not None and ',' in serial:
```
nilsbehlen: remove is not none

privacyidea/api/lib/prepolicy.py
```diff
@@ -1308,6 +1308,42 @@ def check_token_action(request: Request = None, action: str = None):
...
+
+            if allowed:
+                new_serials.append(serial)
+            else:
+                not_authorized_serials.append(serial)
+                log.info(f"User {g.logged_in_user} is not allowed to manage token {serial}. "
+                         f"It is removed from the token list and will not further be processed in"
+                         f" the request.")
+
+        request.all_data["serials"] = new_serials
```
nilsbehlen: why is it written to two different keys

privacyidea/api/token.py
```diff
@@ -310,7 +313,8 @@ def init():
 
         # If the token is a fido2 token, find all enrolled fido2 token for the user
         # to avoid registering the same authenticator multiple times
-        if (token.get_type().lower() in [PasskeyTokenClass.get_class_type(), WebAuthnTokenClass.get_class_type()]
+        if (token.get_type().lower() in [PasskeyTokenClass.get_class_type(),
```
nilsbehlen: should not be formatted this way

privacyidea/api/token.py
```diff
@@ -328,22 +332,26 @@ def init():
         if container_serial:
             # Check if user is allowed to add tokens to containers
             try:
-                container_add_token_right = check_container_action(request, action=PolicyAction.CONTAINER_ADD_TOKEN)
+                container_add_token_right = check_container_action(request,
```
nilsbehlen: should not be formatted this way

privacyidea/api/token.py
```diff
@@ -328,22 +332,26 @@ def init():
...
+                                                                   action=PolicyAction.CONTAINER_ADD_TOKEN)
             except PolicyError:
                 container_add_token_right = False
-                log.info(f"User {user.login} is not allowed to add token {token.get_serial()} to container "
-                         f"{container_serial}.")
+                log.info(
```
nilsbehlen: should not be formatted this way

privacyidea/api/token.py
```diff
@@ -328,22 +332,26 @@ def init():
...
+                    f"{container_serial}.")
             if container_add_token_right:
                 # The enrollment will not be blocked if there is problem adding the new token to a container
                 # there will just be a warning in the log
                 try:
                     add_token_to_container(container_serial, token.get_serial())
                     response_details.update({"container_serial": container_serial})
                     container = find_container_by_serial(container_serial)
-                    g.audit_object.log({"container_serial": container_serial, "container_type": container.type})
+                    g.audit_object.log(
```
nilsbehlen: should not be formatted this way

privacyidea/api/token.py
```diff
@@ -328,22 +332,26 @@ def init():
...
+                        {"container_serial": container_serial, "container_type": container.type})
                 except ResourceNotFoundError:
-                    log.warning(f"Container with serial {container_serial} not found while enrolling token "
-                                f"{token.get_serial()}.")
+                    log.warning(
```
nilsbehlen: should not be formatted this way

privacyidea/api/token.py
```diff
@@ -559,18 +570,54 @@ def assign_api():
 @log_with(log)
 def unassign_api():
     """
-    Unassign a token from a user.
-    You can either provide "serial" as an argument to unassign this very
```
nilsbehlen, line 563: i think we should not remove the sentence: "... or you can provide user and realm, to unassign all tokens of a user. Unassign token(s) from a user."
Also this is probably bad design, but its there currently, so please add a todo this should be split.

privacyidea/api/token.py
```diff
@@ -559,18 +570,54 @@ def assign_api():
...
+    if single is None and many is None:
+        res = unassign_token(serial=None, user=user)
+        g.audit_object.log({"serial": None, "success": True})
+        return send_result(res)
+
+    serial_list = []
+    if isinstance(single, str):
+        serial_list.append(single)
+    if many is not None:
+        if not isinstance(many, list):
```
nilsbehlen: there is no reason to nest this. check for `isinstance` first, then `if many: serial_list.extend(many)`

privacyidea/api/token.py
```diff
@@ -559,18 +570,54 @@ def assign_api():
...
+            raise ParameterError("Parameter 'serials' must be a JSON array of strings.")
+        serial_list.extend(many)
+
+    if not serial_list:
+        raise ParameterError("Provide a non-empty 'serial' or 'serials'.")
+
+    g.audit_object.log({"serial": serial_list if len(serial_list) != 1 else serial_list[0]})
+
+    if len(serial_list) == 1 and many is None:
```
nilsbehlen: please add a comment on why the following code is a bit strange: we need to keep the old behavior (return format) while adding the functionality for a list of serials.

privacyidea/api/token.py
```diff
@@ -660,56 +707,60 @@ def disable_api(serial=None):
...
-
-@token_blueprint.route('/batchdeletion', methods=['POST'])
-@prepolicy(check_token_list_action, request, action=PolicyAction.DELETE)
-@event("token_delete", request, g)
-@log_with(log)
-def batch_deletion():
-    """
-    Delete all passed tokens, e.g. all tokens of a container
-    All errors during the deletion of a token are fetched to be able to delete the remaining tokens.
+    single_body = get_optional(request.all_data, "serial")
```
nilsbehlen: just `single` like in unassign

privacyidea/api/token.py
```diff
@@ -660,56 +707,60 @@ def disable_api(serial=None):
...
+    many = get_optional(request.all_data, "serials")
```
nilsbehlen: check if the comments of def unassign_api() apply here too since the structure is similar.

nilsbehlen (dismissed):

privacyidea/api/token.py
```diff
@@ -660,42 +690,64 @@ def disable_api(serial=None):
...
 
 
 @token_blueprint.route('/batchdeletion', methods=['POST'])
-@prepolicy(check_token_list_action, request, action=PolicyAction.DELETE)
+@prepolicy(check_token_action, request, action=PolicyAction.DELETE)
 @event("token_delete", request, g)
 @log_with(log)
 def batch_deletion():
     """
+    Not maintained anymore, marked for removal. Functionality is now in DELETE /.
```
jelinaunger: You can remove this endpoint. I've added it for 3.12, but it is not yet released in any version.

privacyidea/api/token.py
```diff
@@ -559,18 +560,47 @@ def assign_api():
...
-
-    :return: In case of success it returns the number of unassigned tokens in "value".
+    Unassign token(s) from a user.
+    You can either provide "serial" or "serials" as an argument to unassign token(s), or you can provide user and
+    realm, to unassign all tokens of a user. (old API behavior, TODO this should be split)
+    All errors during the unassignment of multiple tokens are fetched to be able to unassign the remaining tokens.
+
+    :jsonparam serial: The serial number of a single token, or comma-separated list of serials.
+    :jsonparam serials: A list of serial numbers of multiple tokens.
+    :jsonparam not_authorized_serials: A list of serial numbers of tokens which the user is not
```
jelinaunger: this is a parameter internally set by the prepolicy. It should not be set by the user and hence not appear here. Maybe we should write this to g instead of the request_data.

privacyidea/api/token.py
```diff
@@ -660,42 +690,64 @@ def disable_api(serial=None):
...
+def delete_api(serial=None):
     """
-    Delete a token by its serial number.
+    Delete token(s) by their serial number.
+    All errors during the deletion of multiple tokens are fetched to be able to delete the remaining tokens.
 
-    :jsonparam serial: The serial number of a single token.
+    :jsonparam serial: The serial number of a single token, or comma-separated list of serials.
+    :jsonparam serials: A list of serial numbers of multiple tokens.
+    :jsonparam not_authorized_serials: A list of serial numbers of tokens which the user is not
```
jelinaunger: same here

tests/test_api_token.py
```diff
@@ -514,6 +516,30 @@ def test_05_helpdesk_delete_batch(self):
...
         set_policy(name="policy", scope=SCOPE.ADMIN, action=PolicyAction.DELETE, realm=self.realm1)
 
+        # create tokens
+        token1 = init_token({"type": "hotp", "genkey": True, "realm": self.realm1})
+        token2 = init_token({"type": "hotp", "genkey": True, "realm": self.realm2})
+        token3 = init_token({"type": "hotp", "genkey": True, "realm": self.realm1})
+        token_serials = [token1.get_serial(), token2.get_serial(), token3.get_serial()]
+
+        csv_serials = ",".join(token_serials)
+        result = self.request_assert_200(f"/token/?serial={csv_serials}", {}, self.at, "DELETE")
```
jelinaunger, line 525: not serial**s**?

henrik-falk: comma separated serials need to be put in the serial parameter. just old api behavior anyway, multiple serials should be set as json array in the serials parameter from now on.

jelinaunger (approved):

PR #4571, opened 2025-08-12
Sticky header improvements

### LINKED ISSUE #4557: Sticky header improvement
Currently, it looks like this:
Image

Suggested improvements: 
* Remove the "re-open dialog" button if there is no dialog to open.
* Add the token selection to the sticky header
* Remove the token type from the enrollment button, it is implicated by the selector then. The label can just be "Enroll".
So on the left would be selector and on the right the enroll button.

jelinaunger:
closes #4557

### DIFF
--- a/privacyidea/static_new/src/app/components/token/token-enrollment/token-enrollment.component.ts
+++ b/privacyidea/static_new/src/app/components/token/token-enrollment/token-enrollment.component.ts
@@ -239,6 +239,7 @@ export class TokenEnrollmentComponent implements AfterViewInit, OnDestroy {
   protected readonly dialogService: DialogServiceInterface =
     inject(DialogService);
   private observer!: IntersectionObserver;
+  protected readonly renderer: Renderer2 = inject(Renderer2);
   timezoneOptions = TIMEZONE_OFFSETS;
   pollResponse: WritableSignal<any> = linkedSignal({
     source: this.tokenService.selectedTokenType,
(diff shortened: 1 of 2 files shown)


privacyidea/static_new/src/app/components/token/token-enrollment/token-enrollment.self-service.component.html
```diff
@@ -1,160 +1,160 @@
...
-                        [value]="type"
-                        [matTooltip]="type.info.split(':')[1]">{{ type.info.split(":")[0] }}
-            </mat-option>
-          }
-          <button (click)="versioningService.openDocumentation('tokentypes')"
-                  class="card-button width-100"
-                  mat-flat-button>
-            <mat-icon>info_outline</mat-icon>
-            Help About Tokentypes
+    <div #stickyHeader class="sticky-header">
```
henrik-falk: use @include mat.elevation(3) instead of box-shadow

privacyidea/static_new/src/app/components/token/token-enrollment/token-enrollment.self-service.component.html
```diff
@@ -1,160 +1,160 @@
...
+                        [ngModelOptions]="{ standalone: true }"
+                        [panelClass]="'multi-column-panel'">
+              @for (type of tokenService.tokenTypeOptions(); track type) {
+                <mat-option class="token-type-option type-option"
+                            [value]="type"
+                            [matTooltip]="type.info.split(':')[1]">{{ type.info.split(":")[0] }}
+                </mat-option>
+              }
+              <button (click)="versioningService.openDocumentation('tokentypes')"
+                      class="card-button height-auto width-100"
```
henrik-falk: i would prefer a uniform width for these buttons and not width 100% here

PR #4580, opened 2025-08-14
User view navigation menu drawer in audit

henrik-falk:


privacyidea/static_new/src/app/self-service.routes.ts
jelinaunger, line 1: Seems like the formatting is not applied correctly here

PR #4581, opened 2025-08-14
Set PIN is not visible and Resync Token is now hidden for passkey.

frankmer:
Set PIN is not visible and Resync Token is now hidden for passkey.


privacyidea/static_new/src/locale/messages.xlf
```diff
@@ -0,0 +1,95 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
jelinaunger: What is this for a file? Seems like it is not related to the issue

PR #4592, opened 2025-08-15
The session timer now uses the actual logout time stored in the AuthData object.

frankmer:
The session timer now uses the actual logout time stored in the AuthData object. To access this information, the AuthServiceInterface had to be refactored. Consequently, I made all AuthData variables accessible from the AuthService.


privacyidea/static_new/src/app/components/layout/header/header.component.html
```diff
@@ -73,17 +73,17 @@ <h4 class="preview"> Preview</h4>
...
         ]"
         class="session-timer">
-        @if (sessionTimerService.remainingTime() < 60000) {
-          {{ sessionTimerService.remainingTime() | date: "mm:ss" }}
-        } @else if (sessionTimerService.remainingTime() >= 60000 &&
-        sessionTimerService.remainingTime() < 600000) {
+        @if (sessionTimerService.remainingTime()! < 599500) {
+          {{ sessionTimerService.remainingTime() | date: "m:ss" }}
+        } @else if (sessionTimerService.remainingTime()! >= 599500 &&
+        sessionTimerService.remainingTime()! < 6000000) {
```
henrik-falk: same for self service please. and while you are at it i just noticed a typo in header.serlf-service.component file name

privacyidea/static_new/src/app/services/auth/auth.service.ts
```diff
@@ -73,16 +128,76 @@ export interface AuthServiceInterface {
...
+  jwtData = signal<JwtData | null>(null);
+
+  authenticationAccepted = signal<boolean>(false);
+
+  isAuthenticated = computed(() => this.authenticationAccepted() && !!this.authData());
+
+  logLevel = computed(() => this.authData()?.log_level || 0);
+  menus = computed(() => this.authData()?.menus || []);
+  realm = computed(() => this.jwtData()?.realm || this.authData()?.realm || "");
+  rights = computed(() => this.authData()?.rights || []);
```
jelinaunger: should be taken from the jwtData

privacyidea/static_new/src/app/services/auth/auth.service.ts
```diff
@@ -73,16 +128,76 @@ export interface AuthServiceInterface {
...
+  role = computed(() => this.jwtData()?.role || this.authData()?.role || "");
+  token = computed(() => this.authData()?.token || "");
+  username = computed(() => this.jwtData()?.username || this.authData()?.username || "");
+  logoutTimeSeconds = computed(() => {
+    const jwtExpDate = this.jwtExpDate()!;
+    var jwtLogoutTime: number | null = null;
```
jelinaunger: use let or const (var is deprecated)

jelinaunger (approved):

PR #4596, opened 2025-08-18
4000/user assignment for token export

jona-samuel:


privacyidea/lib/tokenclass.py
```diff
@@ -2026,3 +2042,19 @@ def import_token(self, token_information: dict):
         self.add_tokeninfo_dict(token_information.setdefault("tokeninfo", {}))
         self.add_tokeninfo("import_date", datetime.now(timezone.utc).isoformat(timespec="seconds"))
         self.save()
+
+    def assign_user_during_token_import(self, user: dict):
```
plettich: We already have `assign_token()` and `unassign_token()` in `token.py`
Please use/adjust existing functions to avoid duplicating code.

privacyidea/lib/user.py
```diff
@@ -592,6 +592,21 @@ def delete(self):
 
         return success
 
+    def user_export_dict(self) -> dict:
```
plettich, line 597: please check if we can use the `info()` property, it should return a `dict` as well.

plettich (changes requested):

privacyidea/lib/tokenclass.py
```diff
@@ -1998,7 +2012,7 @@ def get_enroll_url(self, user: User, params: dict):
         """
         return None
 
-    def export_token(self) -> dict:
+    def export_token(self, export_user: bool = False) -> dict:
```
plettich, line 2021: Could we use the `_to_dict()` function here as well?

privacyidea/lib/user.py
jona-samuel: This will not work because we get no realm or resolver from this

privacyidea/lib/tokens/passkeytoken.py
```diff
@@ -470,3 +470,12 @@ def authenticate(self, passw, user=None, options=None):
             return False, -1, None
         otp_match = self.check_otp(None, 0, None, options)
         return pin_match, otp_match, None
+
+    def import_token(self, token_information: dict):
+        """
+        Import a hotp token.
```
plettich: passkey token

privacyidea/lib/tokens/passkeytoken.py
```diff
@@ -470,3 +470,12 @@ def authenticate(self, passw, user=None, options=None):
...
+        """
+        TokenClass.import_token(self, token_information)
```
plettich: Better use `super()` here: https://docs.python.org/3/library/functions.html#super

tests/test_lib_token.py
```diff
@@ -1884,22 +1885,6 @@ def test_63_token_export_with_user(self):
         self.assertEqual(hotptoken.owners[0].login, "cornelius")
         self.assertEqual(totptoken.owners[0].login, "cornelius")
 
-        # check custom user attributes are exported and imported correctly
```
plettich: Why did You remove this test?

privacyidea/lib/token.py
```diff
@@ -3007,19 +3009,32 @@ def import_tokens(tokens: list[dict], update_existing_tokens: bool = True, assig
...
+                                 resolver=token_info_dict.get("user").get("resolver"),
+                                 realm=token_info_dict.get("user").get("realm"),
+                                 uid=token_info_dict.get("user").get("uid"))
+                    token.add_user(owner, override=True)
+                token.import_token(token_info_dict)
+
+                if not existing_token:
+                    successful_tokens.append(serial)
+                else:
+                    updated_tokens.append(serial)
```
plettich: The info about the parameter in the log message below can be removed. Just the info, that the token exists and maybe that it was not updated.

privacyidea/lib/tokenclass.py
```diff
@@ -233,9 +233,13 @@ def add_user(self, user, report=None):
...
         # FIXME: We need to remove this, if we one day want to assign several users to one token
         if self.user and self.user != user:
-            log.info("The token with serial {0!s} is already assigned "
-                     "to user {1!s}. Can not assign to {2!s}.".format(self.token.serial, self.user, user))
-            raise TokenAdminError("This token is already assigned to another user.")
+            if override:
+                # We remove the old user
+                self.remove_user()
+            else:
+                log.info("The token with serial {0!s} is already assigned "
```
plettich: Since we throw an error here, this should be at least `log.warning()`

plettich (changes requested):

privacyidea/lib/token.py
```diff
@@ -2980,22 +2980,26 @@ def regenerate_enroll_url(serial: str, request: Request, g) -> Union[str, None]:
...
+def export_tokens(tokens: list[TokenClass], export_user: bool = True) -> list[dict]:
     """
     Takes a list of tokens and returns an dict with all infos.
 
     :param tokens: list of token objects
+    :param export_user: bool
     :return: list of dict with token information
     """
-    exported_tokens = [token.export_token() for token in tokens]
+    exported_tokens = [token.export_token(export_user=export_user) for token in tokens]
```
plettich: This fails if the export of one of the token fails (unsupported token type).
We should do this in a loop and catch exceptions for each token. We could return a dictionary or a dataclass the information that the export failed for x token.

plettich (changes requested):

privacyidea/lib/token.py
```diff
@@ -3006,18 +3010,34 @@ def import_tokens(tokens: list[dict], update_existing_tokens: bool = True) -> To
...
+                else:
+                    token = existing_token
+
+                # Assign the user, if wanted and if there is a user in the token info dict
+                if assign_to_user and token_info_dict.get("user"):
+                    owner = User(login=token_info_dict.get("user").get("login"),
+                                 resolver=token_info_dict.get("user").get("resolver"),
+                                 realm=token_info_dict.get("user").get("realm"),
+                                 uid=token_info_dict.get("user").get("uid"))
+                    token.add_user(owner, override=True)
```
plettich, line 3046: Same here, this can throw an exception when the token is assigned to another user.

plettich (changes requested):

privacyidea/lib/token.py
nilsbehlen: it can not since override=True

PR #4597, opened 2025-08-18
add 120 character line width to editorconfig

henrik-falk:
quick update

### DIFF
--- a/privacyidea/static_new/src/app/components/token/container-details/container-details.self-service.component.html
+++ b/privacyidea/static_new/src/app/components/token/container-details/container-details.self-service.component.html
@@ -249,7 +249,9 @@
               [ngClass]="{ 'pad-top-10': showOnlyTokenNotInContainer() }">
               <span> Only display tokens not in a container </span>
               @if (!showOnlyTokenNotInContainer()) {
-                <br/><span> Note: Adding a token removes it from</span><br/><span style="margin-left: 39px"> its previous container. </span>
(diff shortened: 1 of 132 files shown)


privacyidea/static_new/src/app/components/token/container-details/container-details.self-service.component.html
```diff
@@ -275,8 +274,7 @@ <h3 style="color: var(--mat-sys-primary)"> {{ containerSerial() }} </h3>
               <mat-option
                 class="token-option"
                 (click)="addTokenToContainer(option)">
-                <span class="token-option-span"> {{ option["serial"] }} </span>
-                <span class="token-option-span">
+                <span class="token-option-span"> {{ option["serial"] }} </span> <span class="token-option-span">
```
jelinaunger: I think this should not happen?

henrik-falk: I cannot get the ide reformat or prettierrc to fix this. Only solution that i see is that we manually add a line break everytime a multiclass span follows another span element.

jelinaunger: Add span to `ij_html_add_new_line_before_tags` and remove from `ij_html_inline_tags` works.
Or is there a reason to keep span as inline tag?

jelinaunger (approved):

PR #4605, opened 2025-08-20
rework token enrollment last step dialog

henrik-falk:
- [ ] spass
- [ ] tan
- [ ] paper
- [ ] registration
- [ ] tiqu


jelinaunger (commented):
Remove QR code for indexed secret, sms, and I think also for applspec. For self-service remove the QR code text for paper and tan token and in the admin view the regenerate button for these tokens is not visible.

jelinaunger (approved):

PR #4606, opened 2025-08-21
Write policies only once to the audit

jelinaunger:
In the audit column `policies`, each policy should appear only once, even if the policy is applied multiple times in the request

### DIFF
--- a/privacyidea/lib/policy.py
+++ b/privacyidea/lib/policy.py
@@ -706,8 +706,10 @@ class PolicyClass(object):
                 [p.get('name') for p in reduced_policies]))
 
         if audit_data is not None:
+            audit_data.setdefault("policies", [])
             for p in reduced_policies:
-                audit_data.setdefault("policies", []).append(p.get("name"))
+                if p.get("name") not in audit_data["policies"]:
+                    audit_data["policies"].append(p.get("name"))
 
         return reduced_policies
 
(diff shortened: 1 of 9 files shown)


privacyidea/lib/policy.py
```diff
@@ -706,8 +706,10 @@ def match_policies(self, name: Optional[str] = None, scope: Optional[str] = None
                 [p.get('name') for p in reduced_policies]))
 
         if audit_data is not None:
+            audit_policies = set(audit_data.get("policies", []))
```
plettich, line 709: we have this `add_policy()` function in the Audit class, maybe we can use that:
https://github.com/privacyidea/privacyidea/blob/5510737af4ace5b2f6405a06fd4a4c3b90b88d07/privacyidea/lib/auditmodules/base.py#L198

plettich (changes requested):

privacyidea/lib/policy.py
jelinaunger: At this point there is no audit object, hence we can not use this function

nilsbehlen (approved):

privacyidea/lib/policy.py
plettich: Would it make sense to pass the audit object instead of just the `audit_data` dictionary?

plettich (changes requested):

PR #4609, opened 2025-08-22
Implemented app-clearable-input on some inputs where it was missing

frankmer:
Implemented app-clearable-input on some inputs where it was missing


jelinaunger:
Bildschirmfoto vom 2025-08-22 13-23-15

Button should be round

PR #4611, opened 2025-08-22
Update webui login

nilsbehlen:


nilsbehlen:
second step looks like this now:
image

frankmer (approved):

PR #4615, opened 2025-08-25
set random pin

jelinaunger:
* open dialog to show randomly generated pin
* after set pin, clear pin input fields
* implement canSetRandomPin()


privacyidea/static_new/src/app/components/shared/simple-dialog/simple-dialog.component.html
henrik-falk, line 1: scss is missing with import of style.scss with dialog.scss so that
.mat-mdc-dialog-surface {
  border-radius: var(--border-radius);
}

PR #4618, opened 2025-08-25
I have added i18n to every element that I have found that needs internationalisation.

frankmer:
I have added i18n to every element that I have found that needs internationalisation.


plettich (approved):
The setup of the translation looks correct, have not looked at the actual translations/sources.
We should merge and set the translation up with weblate to do the translations there.

PR #4628, opened 2025-08-27
Added and used the model class 'FilterValue' to remove some effects and make it easier to use.

frankmer:
Added and used the model class 'FilterValue' to remove some effects and make it easier to use.


jelinaunger (commented):
The token query for "type: hotp" still does not work (or may be got broken during merge)

PR #4629, opened 2025-08-27
Fix layout and disable links

henrik-falk:
just some small fixes


privacyidea/static_new/src/app/components/token/container-table/container-table.component.html
```diff
@@ -34,149 +34,151 @@
...
+                }
+                @case ("realms") {
+                  <div [ngClass]="(element[column.key]?.length || 0) > 1 ? 'realm-list' : 'realm'">
+                    @for (realm of element[column.key]; track realm; let idx = $index) {
+                      <span
+                        attr.aria-label="{{ column.label }} Link">
+                        {{ realm }}
+                      </span>
+                      <br style="margin-top: 3px" />
+                      <br style="margin-top: 3px" />
```
jelinaunger: This is too much space. Anyway \<br/> should not be used for styling and css code should also not be used within html. What about simply wrapping the span in a div which will also add a line break after each realm?

nilsbehlen (approved):

PR #4634, opened 2025-08-27
fix assign user to all tokens

henrik-falk:
~fixed but still only show pin~,
removed dialog,
fix icon


privacyidea/static_new/src/app/components/token/container-details/user-assignment-dialog/user-assignment-dialog.component.html
```diff
@@ -21,7 +21,6 @@
       matSuffix
       type="button">
       <mat-icon>{{ hidePin() ? "visibility_off" : "visibility" }}</mat-icon>
-      Show PIN
```
jelinaunger, line 24: This user-assignment-dialog could be removed completely. Or is it used somewhere else?

nilsbehlen (approved):

Issue #4684, opened 2025-09-15
auth-endpoint's response details allow fishing for existing users


commel:
### Top-level intent

When a login fails for whatever reason the possible attacker should not gain any knowledge from the internal data. At the moment when using TOTP and using `/auth` endpoint and providing a correct user and a wrong password, the attempt is returned with details for the auth failure:

```json
{"detail": {"loginmode": "privacyIDEA", "message": "wrong otp value. previous otp used again", "otplen": 6, "serial": "OATH00009A6C", "threadid": 140573918873280, "type": "hotp"},...
```

or 
```json
{"detail": {"loginmode": "privacyIDEA", "message": "wrong otp value", "otplen": 6, "serial": "OATH00009A6C", "threadid": 140573918873280, "type": "hotp"}, "id": 1, "jsonrpc": "2.0", "result": {"error": {"code": 4031, "message": "Authentication failure. Wrong credentials"}, "status": false}
```

When running this request with a non existing user it states:
```json
{"detail": {}, "id": 1, "jsonrpc": "2.0", "result": {"error": {"code": 4031, "message": "Authentication failure. Wrong credentials"}, "status": false}
```

### Steps to reproduce

1. Bind a TOTP token to an existing user, enabled login_mode = "userstore"
2. Use /auth to login with existing user and wrong password -> "wrong otp value"
3. Use /auth to login with non-existing user -> "auth failed"

### Expected outcome

No details of any failed login attempt should be returned but logged internally. In any case if a user exists or not exists or a pin was given wrongfully, the returned dataset should only provide:

```json
{"detail": {}, "id": 1, "jsonrpc": "2.0", "result": {"error": {"code": 4031, "message": "Authentication failure. Wrong credentials"}, "status": false}
```

### Actual outcome

Current details returned enable a potential attacker to gain internal knowledge of the 

### Configuration

* **privacyIDEA version**: 3.11.4, master
* **Installation method**: github, PyPI
* **Python version**: 3.10.18
* **Operating system**: openSUSE Tumbleweed
* **Webserver**: none
* **Token database**: sqlite3

nilsbehlen:
Hi, i have described in #4622 that this problem can not be fixed by simply removing the message. 
If a user does not exist, the server responds with a 403 code, which allows to clearly distinguish between existing and not-existing users, regardless of the message.
[...]

PR #4685, opened 2025-09-15
move auth details to log on failed attempts

commel:
Moving details to log for analysis instead of the endpoint result object. By this every attempt on fishing for existing or non existing users will result in the exact same response message.

refs #4684


nilsbehlen:
superceded by #4709

Issue #3330, opened 2022-10-11
Tags in the text of the email token


sepp67:
The customization of the e-mail token is not working.

The documentation says that the following tags can be [used](https://privacyidea.readthedocs.io/en/latest/policies/authentication.html#emailtext).
       {otp} or <otp> the One-Time-Password
        {serial} or <serial> the serial number of the token.
        {user} the given name of the token owner.
        {givenname} the given name of the token owner.
        {surname} the surname of the token owner.
        {username} the loginname of the token owner.
        {userrealm} the realm of the token owner.
        {tokentype} the type of the token.
        {recipient_givenname} the given name of the recipient.
        {recipient_surname} the surname of the recipient.
        {time} the current server time in the format HH:MM:SS.
        {date} the current server date in the format YYYY-MM-DD

But only `<otp>` and `<serial>` are working.

I tested all the tags with both forms  {tag} and `<tag>`.
{tag} is generating an error message during the login
        Error message on the WebGui: 
        Authentication failed. Authentication failure. Wrong credentials

        Error message in the logs: 
        [2022-10-11 14:21:10,797][11094][140002954303232][WARNING][privacyidea.lib.tokens.emailtoken:315] The PIN was correct, but the EMail could not be sent! ('tokentype')

`<tag>` is not interpreted and is displayed in the email as `<tag>`

plettich:
When we drop Python 2 we can use [`format_map()`](https://docs.python.org/3.8/library/stdtypes.html?highlight=format_map#str.format_map).
Until then we can catch the `KeyError` or check if all given keys exist in the dictionary.

cornelinux:
@sepp67 the tags are interpreted. But, as @plettich pointed out, if the tag may not exist, it raises a KeyError.
the curly tags are known to work in /validate/check. I assume that during `/auth` (login to the webui) some information do not exist and thus some values in tags can not be used.
However, this is s.th. we have at several points, that the admin needs to know, what he is configuring and which information is available. (See extended policies --- otherwise it might also contradict his expected outcome).

E.g. if a user logs in without a 2nd factor, there is no tokentype.

So, works as intended. So I am moving this to known issue.

cornelinux:
It looks like this line causes all the problems:
https://github.com/privacyidea/privacyidea/blob/754f49b32760097a34af7d8b08b4424e32057080/privacyidea/lib/tokens/emailtoken.py#L386

cornelinux:
[...]

PR #4701, opened 2025-09-19
 Adding a warning about the use of commas in the email and sms text.

sepp67:
Some users add ¨,"in the email or sms text. Or as commas separated different policy actions, the code interprets them.  This is why a comma in a policy action-value cannot be used.


sepp67:
Actually, the issue was fixed in (issue #3330)

PR #4708, opened 2025-09-25
Disable require_presence when push_wait is active

### LINKED ISSUE #4702: PUSH-wait and require presence do not work together
Using `push_wait` policy together with `push_require_presence` currently fails with a 500
Since they are both mutually exclusive (you could not get the `require_presence` information back to the client since there is no response, the thread just waits) we should catch this behaviour, issue a warning in the logs/audit and disable `push_require_presence` in this case.

plettich:
Since there is no possibility to pass the valid presence question to the
user, the 'require_presence' policy will be disabled in this case and a
warning will be written to the log.

- Also fixed the timestamp check since it was not working as advertised.
- Add documentation entry regarding the require_presence policy together
  with push_wait.
- Change the numbers in the presence options list to integer.

Closes #4702

### DIFF
--- a/privacyidea/lib/tokens/pushtoken.py
+++ b/privacyidea/lib/tokens/pushtoken.py
@@ -1,4 +1,6 @@
-#  License:  AGPLv3
+# SPDX-FileCopyrightText: 2019 NetKnights GmbH <https://netknights.it>
+# SPDX-License-Identifier: AGPL-3.0-or-later
+#
 #  contact:  http://www.privacyidea.org
 #
 #  2019-02-08   Cornelius Kölbel <cornelius.koelbel@netknights.it>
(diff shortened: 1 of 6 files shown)


privacyidea/lib/tokens/pushtoken.py
```diff
@@ -661,15 +664,17 @@ def _check_timestamp_in_range(timestamp, window):
...
             log.debug(f'{traceback.format_exc()}')
-            raise privacyIDEAError(f'Could not parse timestamp {timestamp}. '
-                                   'ISO-Format required.')
+            raise privacyIDEAError(f'Could not parse timestamp {timestamp}. ISO-Format required.')
         td = timedelta(minutes=window)
         # We don't know if the passed timestamp is timezone aware. If no
         # timezone is passed, we assume UTC
         if ts.tzinfo:
-            now = datetime.now(utc)
+            # We honor the timezone of the timestamp, ideally it should be UTC
```
nilsbehlen: "honor" seems like a big word here, "consider" or "take into account"

PR #4731, opened 2025-10-02
passthru verify fallback

jelinaunger:
* token event handler can do any action on the users token
* added event handler context


privacyidea/lib/eventhandler/base.py
```diff
@@ -108,6 +110,18 @@ class GROUP(object):
...
+    """
+
+    def __init__(self, user: Optional[User] = None, token_serials: Optional[list[str]] = None,
+                 container: Optional[TokenContainer] = None):
+        self.user = user
+        self.token_serials = token_serials
+        self.container = container
+
+
 class BaseEventHandler(object):
```
nilsbehlen, line 125: inheriting from `object` can be removed if you come across

Issue #4766, opened 2025-10-24
More tags for challenge texts of SMS and Email


nilsbehlen:
Add tags to have the email address or phone number in the text in redacted form. As i could not find a real standard for this (NIST de-identification does not really provide a guideline if i understood it correctly). I think of something like this:
Phone Number: **** - *******81 with fixed number of *, independently of the actual length, just showing the last two digits.
Email:`n******n@n****.it` first and last character of username/local-part and the first character of the domain. Also with fixed number of *.

This way, challenge texts like:
"An authentication code has been sent to your phone with the number ****-*****81!"
will be possible

PR #4771, opened 2025-10-27
More tags for challenge texts of SMS and Email

jona-samuel:
add tags phone_redacted and email_redacted to challenge_text

fixing #4766

### DIFF
--- a/privacyidea/lib/utils/__init__.py
+++ b/privacyidea/lib/utils/__init__.py
@@ -1495,7 +1495,7 @@ def redacted_email(email: str) -> str:
     """
     name, domain = email.split('@') if '@' in email else ("**", email)
     name_redacted = name[:2] + "*" * (10 - len(name[:2]))
-    domain_parts = domain.split('.') if '.' in domain else ("**", "***")
+    domain_parts = domain.split('.') if '.' in domain else ["**", "***"]
     return f'{name_redacted}@{domain_parts[0][0]}****.{domain_parts[-1]}'
 
 


doc/policies/authentication.rst
```diff
@@ -815,7 +815,9 @@ tag.
     * {time} the current server time in the format HH:MM:SS.
     * {date} the current server date in the format YYYY-MM-DD.
     * {phone} the phone number from the challenge in case of sms token.
+    * {phone_redected} the phone number from the challenge in case of sms token in redacted form.
```
jelinaunger: phone_red*a*cted (typo)

privacyidea/lib/utils/__init__.py
```diff
@@ -1486,6 +1484,28 @@ def get_computer_name_from_user_agent(user_agent: str) -> Union[str, None]:
...
+
+
+def redacted_email(email: str) -> str:
+    """
+    Censors an email address from 'example.mail@test.com' to 'exa*********@t***.com'
+    """
+    try:
+        name, domain = email.split('@')
+        domain_parts = domain.split('.')
+        return f'{name[:2]}********@{domain_parts[0][0]}****.{domain_parts[1]}'
```
jelinaunger: As the domain could contain multiple subdomains use domain_parts[-1] instead of index 1

privacyidea/lib/utils/__init__.py
```diff
@@ -1486,6 +1484,28 @@ def get_computer_name_from_user_agent(user_agent: str) -> Union[str, None]:
...
+    except:
+        return '********@****.***'
```
jelinaunger: Instead of wrapping everything in a try except, I would prefer something like

`name_redacted = name[:2] or "**"`

This has also the advantage if one part fails, the other ones could be included correctly.

privacyidea/lib/utils/__init__.py
```diff
@@ -1486,6 +1484,28 @@ def get_computer_name_from_user_agent(user_agent: str) -> Union[str, None]:
...
                 # This exception is likely to happen, because words/parts like "Mac" are common
-                #log.debug(f"Could not extract computer name from user agent: {ex} with key {key}")
+                # log.debug(f"Could not extract computer name from user agent: {ex} with key {key}")
                 pass
     return None
+
+
+def redacted_email(email: str) -> str:
+    """
+    Censors an email address from 'example.mail@test.com' to 'exa*********@t***.com'
```
jelinaunger: 'ex**** ...' only the first two letters are shown

privacyidea/lib/utils/__init__.py
```diff
@@ -1486,6 +1484,28 @@ def get_computer_name_from_user_agent(user_agent: str) -> Union[str, None]:
...
+
+
+def redacted_phone_number(number: str) -> str:
+    """
+    Censors a phone number from 01234567890 to *********90
+    """
+    try:
+        return f'****-******{number[-2:]}'
+    except:
+        return '****-********'
```
jelinaunger: same here

privacyidea/lib/token.py
```diff
@@ -2929,6 +2931,7 @@ def challenge_text_replace(message, user, token_obj, additional_tags: dict = Non
             email = TokenClass.get_tokeninfo(token_obj, token_obj.EMAIL_ADDRESS_KEY)
         if email:
             tags["email"] = email
+            tags["email_redacted"] = redacted_email(email)
```
jelinaunger, line 2934: Seems like there is only an API test rarely testing this function. Could you add a test only for the challenge_text_replace() function? At least for the new tags, but ideally with complete coverage.

tests/test_lib_utils.py
```diff
@@ -909,6 +911,17 @@ def test_37_useragent_split(self):
             res = get_plugin_info_from_useragent(val[0])
             self.assertEqual(res, val[1:], res)
 
+    def test_38_redacted_email(self):
+        self.assertEqual(redacted_email("example.mail@test.com"), "ex********@t****.com")
```
jelinaunger: Please change the order of the arguments:
self.assertEqual(expected, actual)

jelinaunger (approved):

Issue #4739, opened 2025-10-06
API responses are too specific


eisenmann-b1:
### Top-level intent

Some of the API responses are too specific, to the extent that they could be used to guess privacyIDEA internals, like token serials or container serials.

Here is a list of those endpoints, together with the error condition, HTTP status, error message, and the type of action this enables:

| Endpoint | Error condition | HTTP status | Error message | Action |
| --- | --- | --- | --- | --- |
| `/register` | user already exists | 400 | `{"code": 402, "message": "ERR402: The username is already registered!"}` | Guess user name |
| `/healthz/resolversz` | - | 200 | Shows name and status resolvers | Exposes LDAP and SQL resolver names |
| `/validate/offlinerefill` | token does not exist | 400 | `{"code": 905, "message": "ERR905: The token does not exist"}` | Guess any token serial |
| `/ttype/tiqr` | TiQR token does not exist | 404 | `{"code": 601, "message": "The requested token could not be found."}` | Guess TiQR token serial |
| <ul><li>`/container/register/finalize`</li><li>`/container/register/terminate/client`</li><li>`/container/challenge`</li><li>`/container/synchronize`</li><li>`/container/rollover` (if `container_client_rollover` policy is set)</li></ul> | container does not exist | 404 | `{"code": 601, "message": "Unable to find container with serial xxx."}` | Guess container serial |

While access to these endpoints could be restricted by other means (e.g. the web server), it would be nice to see a solution directly in privacyIDEA.

### Steps to reproduce

1. Configure a single TOTP token (assume its serial is `TOTP0000XXXX`).
2. Run `curl $API_BASE/validate/offlinerefill -d refilltoken= -d pass= -d serial=abc`.
   Will respond with an error, saying `"message": "ERR905: The token does not exist"`
3. Run `curl $API_BASE/validate/offlinerefill -d refilltoken= -d pass= -d serial=TOTP0000XXXX`.
   Will respond with an error, saying `"message": "ERR905: Token is not an offline token or refill token is incorrect"`
4. Notice the different error messages, where step 3 implies the the existence of a token with serial `TOTP0000XXXX`.

### Expected outcome

Error messages should be generic, such that container/token serials and/or other such internal data can be guessed.

### Actual outcome

Error messages are (overly) specific.

### Configuration

* **privacyIDEA version**: v3.12
* **Installation method**: github
* **Python version**: 3.12
* **Operating system**: Debian
* **Webserver**: `pi-manage run`
[...]

PR #4776, opened 2025-10-29
Hide individual resolver status unless authenticated

eisenmann-b1:
Adds optional authentication to the `/healthz/resolversz` endpoint.

If the auth token is:
* present and valid: Include individual resolver status (as is the behavior until now).
* not present: Hide individual resolver status, i.e. only send total status.
* present but invalid: Raise an `AuthError`.

This new behavior is locked behind the new `AUTHZ`-scoped policy `REQUIRE_AUTH_FOR_RESOLVER_DETAILS`.

ref #4739

### DIFF
--- a/doc/policies/authorization.rst
+++ b/doc/policies/authorization.rst
@@ -346,3 +346,20 @@ in the attestation certificate matches accordingly the token can be enrolled.
 
 .. note:: If you configure this, you will likely also want to configure
     :ref:`policy_webauthn_enroll_req`
+
+.. _policy_require_auth_for_resolver_details:
+
+require_auth_for_resolver_details
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+type: ``bool``
+
+Usually, `/healthz/resolversz` will include in its response the name and status
+of each resolver individually, as well as the total status of all resolvers;
+without requiring any form of authentication.
+
+If this policy is set, admin credentials must be provided to receive the
+individual resolver details.  The total status will be included either way.
+
+.. note:: In order to limit the amount of information exposed to third parties,
+    it is recommended to activate this policy.


eisenmann-b1:
I mostly followed @plettich's [suggestion](https://github.com/privacyidea/privacyidea/issues/4739#issuecomment-3407105378) for making the authentication optional.

Making authentication required by default would be more secure, I agree.
But wouldn't that constitute a breaking change in the API, thus requiring a bump of the major version?
AFAIU this was a big concern in the discussion in #4684.

tests/test_api_healthcheck.py
```diff
@@ -60,78 +85,92 @@ def check_status(expected_status_code, expected_ready_status, app_ready_value):
...
+
+                result_value = result.get("value")
+                ldap_resolvers = result_value.get("ldapresolver")
+                sql_resolvers = result_value.get("sqlresolver")
+
+                if ldap_expected_status is None:
+                    assert 'ldapresolver' not in result_value, "Expected missing 'ldapresolver' in result, but is present"
+                else:
+                    assert ldap_resolvers is not None, "Expected 'ldapresolver' in result, but got None"
+                    assert all(status == ldap_expected_status for status in ldap_resolvers.values()), \
```
nilsbehlen: please dont use \ for multiline code

tests/test_api_healthcheck.py
```diff
@@ -60,78 +85,92 @@ def check_status(expected_status_code, expected_ready_status, app_ready_value):
...
+                        f"At least one LDAP resolver does not have '{ldap_expected_status}' status."
+
+                if sql_expected_status is None:
+                    assert 'sqlresolver' not in result_value, "Expected missing 'sqlresolver' in result, but is present"
+                else:
+                    assert sql_resolvers is not None, "Expected 'sqlresolver' in result, but got None"
+                    assert all(status == sql_expected_status for status in sql_resolvers.values()), \
```
nilsbehlen: \

tests/test_api_healthcheck.py
nilsbehlen: i see its copied, but can be corrected now

privacyidea/api/healthcheck.py
```diff
@@ -241,22 +245,37 @@ def resolversz():
     resolver_types = ["ldapresolver", "sqlresolver"]
     total_status = "OK"
 
+    accepts_auth = Match.action_only(g, scope=SCOPE.AUTHZ, action=PolicyAction.REQUIRE_AUTH_FOR_RESOLVER_DETAILS).any()
```
nilsbehlen: requires_auth?

eisenmann-b1: That would be more in line with the policy name.

The reason I named it `accepts_auth` was that authentication is *only* required to get the details.
I.e. the endpoint "accepts" the authentication header, but does not "require" it.

PR #4781, opened 2025-10-31
New destination for action:sendemail for notification handler

### LINKED ISSUE #3963: New action for notification handler

A user may have several different email addresses mapped, that can be used to send notifications to using the event handler.
These email addresses can not be reached by the destination "To Email" of the event handler.

We need a "To" mechanism like "To: Attribute", since the other email address is mapped to a different user attribute.
Then the additional user attribute needs to be specified.

**Additional context**

See 
https://community.privacyidea.org/t/multiple-email-addresses-eventhandlerframework/3395/9

# Top level requirements and scenarios

TBD

*Describe what needs to be achieved and how the scenario looks like*

*If sensible use a checklist to check, which requirements have been covered by your implementation!*

* [ ] We need to ensure that...
* [ ] Administrators must be able to...

# Implementation

*Describe your implementation plans - what you are exactly going to implement. Use references/link to the existing code*

*Cover the following aspects:*

## Database

Describe the database table that will be added. Define the `name` of the table and the `names of the columns`.

Define, if you need to add a DB migration script.

## Library and API

Describe what needs to be implemented, preferably also in which files.

Use example code of what you plan to implement:

~~~~Python
new_feature = [ x.name for x in some_return_values if x.id > 0 ]
~~~~

# TODOs

*In the TODO section you can add all steps that need to be implemented*

**USE CHECKLISTS, THESE WILL BE VISIBLE AS TASKS!**

* [ ] add DB table `name` in `models.py`
* [ ] create above mentioned migration script.
* ~~[ ] there might be an aspect that is not relevant anymore~~
* [ ] e.g. Add util function to...
* [ ] e.g. add REST API...
* [ ] e.g. we need to add policy and implement it in e.g. `prepolicy`
  - scope: admin
  - action: new_action_name
  - ...    

Use further checkboxes for

* [ ] add documentation
* [ ] implement tests

jona-samuel:
Closes #3963

### DIFF
--- a/doc/eventhandler/usernotification.rst
+++ b/doc/eventhandler/usernotification.rst
@@ -41,8 +41,9 @@ Possible recipient types are:
 Depending on the recipient type you can enter additional information. The
 recipient type *email* takes a comma separated list of email addresses.
 
-.. note:: If you use token owner or logged in user, you can use 'to tokenowner' or 'to logged_in_user' to choose an
-user attribute from with the email address should be taken.
+.. note:: If you select token owner or logged_in_user, you can enter a user
+    attribute from which the email address should be taken in the
+    input field 'To tokenowner' / 'To logged_in_user'.
 
 **reply_to**
 
--- a/tests/test_lib_eventhandler_usernotification.py
+++ b/tests/test_lib_eventhandler_usernotification.py
@@ -853,6 +853,83 @@ class UserNotificationTestCase(MyTestCase):
         un_handler = UserNotificationEventHandler()
         res = un_handler.do("sendmail", options=options)
         self.assertTrue(res)
+        msg = smtpmock.get_sent_message()
+        assert 'Subject: An action was performed on your token.' in msg
+        assert 'To: user@localhost.localdomain' in msg
+
+    @smtpmock.activate
+    def test_12_send_with_diffent_attribute(self):
+        # setup realms
+        self.setUp_user_realms()
+
+        r = add_smtpserver(identifier="myserver", server="1.2.3.4", tls=False)
+        self.assertTrue(r > 0)
+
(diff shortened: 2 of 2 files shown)


doc/eventhandler/usernotification.rst
```diff
@@ -41,6 +41,9 @@ Possible recipient types are:
 Depending on the recipient type you can enter additional information. The
 recipient type *email* takes a comma separated list of email addresses.
 
+.. note:: If you use token owner or logged in user, you can use 'to tokenowner' or 'to logged_in_user' to choose an
+user attribute from with the email address should be taken.
```
jelinaunger: If you select *token owner* or *logged_in_user*, you can enter a user attribute from which the email address should be taken in the input field 'To tokenowner' / 'To logged_in_user'.

tests/test_lib_eventhandler_usernotification.py
```diff
@@ -849,11 +848,7 @@ def test_12_send_to_tokenowner(self):
                        "options": {"body": "your {registrationcode}",
                                    "emailconfig": "myserver",
                                    "To": NOTIFY_TYPE.TOKENOWNER,
-                                   "To " + NOTIFY_TYPE.TOKENOWNER:
-                                       "recp@example.com",
-                                   "reply_to": NOTIFY_TYPE.TOKENOWNER,
-                                   "reply_to" + NOTIFY_TYPE.TOKENOWNER:
-                                       "recp@example.com"}}}
+                                   "reply_to": NOTIFY_TYPE.TOKENOWNER}}}
```
jelinaunger, line 851: This test should also check the correct email address (e.g. see test_02)
Additionally, there should be a test which defines a different user attribute for the e-mail  (create user with different e-mail addresses and test that the correct one is used for different event handler settings)

PR #4799, opened 2025-11-12
add disabled token type check for fido

nilsbehlen:


privacyidea/api/auth.py
```diff
@@ -274,7 +274,8 @@ def get_auth_token():
                             id=ERROR.AUTHENTICATE_MISSING_USERNAME)
         if token.get_type() in request.all_data.get("disabled_token_types", []):
             raise AuthError(
-                _("Authentication failure. The token type {token_type} is disabled.").format(token.get_type()),
+                _("Authentication failure. The token type {token_type} is disabled.").format(
+                    token_type=token.get_type()),
```
nilsbehlen, line 278: format() was missing the named parameter "token_type"

PR #4803, opened 2025-11-13
4717/user sources realms

henrik-falk:


privacyidea/static_new/src/app/components/user/realm-table/realm-table.component.scss
```diff
@@ -0,0 +1,150 @@
...
+  margin-left: 1rem;
+}
+
+.icon-true {
+  color: var(--mdc-theme-success, var(--green-base));
+}
+
+.icon-false {
+  color: var(--mdc-theme-error, var(--red-base));
+}
```
jelinaunger, line 26: Is this used anywhere?

privacyidea/static_new/src/app/components/user/realm-table/realm-table.component.html
```diff
@@ -0,0 +1,354 @@
...
+        class="token-table"
+        mat-table
+        matSort>
+        @for (column of columnKeysMap; track column) {
+          <ng-container [matColumnDef]="column.key">
+            <th
+              *matHeaderCellDef
+              attr.aria-label="{{ column.label }} Header"
+              mat-header-cell
+              mat-sort-header>
```
jelinaunger, line 66: Sorting does not make sens for all columns. I think realm and maybe default would be enough.

privacyidea/static_new/src/app/components/user/realm-table/realm-table.component.html
```diff
@@ -0,0 +1,354 @@
+<div
```
jelinaunger, line 1: check for rights if the user is allowed to edit and delete the realms (resolverwrite + resolverdelete)

PR #4804, opened 2025-11-13
wizard doc

jelinaunger:
Added documentation for customizing the token / container wizard in the new webui.
Sanitized customization html files.

related to #4569


privacyidea/static_new/src/app/components/token/container-create/container-create.wizard.component.ts
```diff
@@ -81,21 +81,24 @@ export class ContainerCreateWizardComponent extends ContainerCreateComponent {
     this.description.set("");
   };
 
+  // TODO: Get custom path from pi.cfg
+  customizationPath = "/static/public/customize/";
```
jelinaunger, line 85: We could already do this. However, if no customization path is defined, the API always return "static/customize". Hence, the default case would be broken.

henrik-falk: yep, I see no other option than to solve the problem in the backend.

PR #4811, opened 2025-11-18
clients

### LINKED ISSUE #4718: Audit: Clients
From the old webui: Components->Client Application Type
Lists the last seen User-Agents with date and IP.
So it is a just a simple list, we probably dont need two columns like in the old webui, just one.

Possible improvements:
* Searchbar
* Copy button for User-Agent and IP
* Quick link to search in audit -> will go to audit page with preset filter for the User-Agent/IP
* Sorting by last seen

jelinaunger:
closes #4718

### DIFF
--- a/privacyidea/static_new/src/app/services/clients/clients.service.spec.ts
+++ b/privacyidea/static_new/src/app/services/clients/clients.service.spec.ts
@@ -36,6 +36,7 @@ import { MockAuthService } from "../../../testing/mock-services/mock-auth-servic
 describe("ClientsService", () => {
   let clientService: ClientsService;
   let contentService: MockContentService;
+  let mockAuthService: MockAuthService;
 
   beforeEach(() => {
     TestBed.resetTestingModule();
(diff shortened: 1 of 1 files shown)


privacyidea/static_new/src/app/services/clients/clients.service.ts
```diff
@@ -0,0 +1,60 @@
...
+  providedIn: "root"
+})
+export class ClientsService implements ClientsServiceInterface {
+  private readonly authService: AuthServiceInterface = inject(AuthService);
+  private readonly contentService: ContentServiceInterface = inject(ContentService);
+  private clientsBaseUrl = environment.proxyUrl + "/client/";
+
+  clientsResource = httpResource<PiResponse<ClientsDict>>(() => {
+    if (this.contentService.routeUrl() !== ROUTE_PATHS.CLIENTS) {
+      return undefined;
```
frankmer, line 50: Return 'undefined' when the user is not logged in or is not permitted to receive the resource.

frankmer (approved):

PR #4840, opened 2025-11-27
policies menu panel

henrik-falk:


privacyidea/static_new/src/app/components/layout/navigation/navigation.component.html
```diff
@@ -357,15 +355,21 @@ <h4>{{ profileText }}</h4>
     </mat-expansion-panel>
     @if (authService.actionAllowed("policyread")) {
       <mat-expansion-panel
-        class="nav-expension-panel policy-header"
```
frankmer, line 360: Is there a reason why policy has its own css class? Shouldn't it look like other expension-panels with no children? Maybe we can rename it to something more generic.

PR #4869, opened 2025-12-05
flask talisman

### LINKED ISSUE #4802: Use flask-talisman for security headers
Use the [flask-talisman](https://github.com/GoogleCloudPlatform/flask-talisman) extension (with a feature switch!), which will add security headers to the server responses.
We will then have to check if the frontend works properly, because the CSP will be 'self' by default.

jona-samuel:


nilsbehlen (changes requested):
Please add a feature switch, so that `PI_DISABLE_TALISMAN=true` in the config can disable this feature. sorry for not mentioning that in the original issue.

nilsbehlen (approved):

PR #4938, opened 2026-02-04
configuration subscription

### LINKED ISSUE #4711: Configuration: Subscription
Can be similar to the old web for starters, maybe a bit nicer layout.
If a subscription is present, the "delete" and "upload subscription file" buttons should be next to each other and not split across the screen.

henrik-falk:


nilsbehlen (approved):
please disable the expiry reminder dialog for now, and make hide welcome policy work without having a subscritpion.
otherwise its good

PR #4939, opened 2026-02-04
add option to get offline data doing evm, slight refactor

### LINKED ISSUE #4259: Offline with enroll_via_multichallenge for fido2 token
It would be cool if fido2 token could be marked for offline use directly when enrolled via multichallenge, so that the enrolling client would not have to do another authentication to get the offline data.
Is also interesting for this (standalone capability part): 
https://github.com/privacyidea/privacyidea-credential-provider/issues/180

nilsbehlen:
closes #4259

### DIFF
--- a/privacyidea/api/validate.py
+++ b/privacyidea/api/validate.py
@@ -84,7 +84,7 @@ from privacyidea.api.lib.postpolicy import (postpolicy,
                                             add_user_detail_to_response, construct_radius_response,
                                             mangle_challenge_response, is_authorized,
                                             multichallenge_enroll_via_validate, preferred_client_mode,
-                                            hide_specific_error_message, get_passkey_enroll_offline_data)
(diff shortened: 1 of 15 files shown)


privacyidea/api/validate.py
```diff
@@ -448,198 +448,291 @@ def check():
...
-    # If only the credential ID is given, try to use it to identify the token
-    if credential_id:
-        # Find the token that responded to the challenge
-        transaction_id: str = get_required(request.all_data, "transaction_id")
-        if serial:
-            token = get_one_token(serial=serial)
-        else:
-            token = get_fido2_token_by_credential_id(credential_id)
+
+def _handle_enrollment_cancellation(data):
```
jelinaunger: type hints would be nice

jelinaunger (approved):

PR #4954, opened 2026-02-13
hide passkey login

### LINKED ISSUE #4949: Make passkey button configurable
The passkey login button on the webui should be configurable.

Maybe 90% of installations currently do not use passkeys, so in most cases, this button could be irritating.

There should be a policy to configure this button:

* Scope: Webui
* Action: passkey_login
* Values:
    - disable (do not show the passkey button)
    - enable (show the passkey button)
    - preferred (nice to have, first show the passkey button and then switch to normal login)
    - depending (nice to have, check if actually passkeys in privacyidea exist and only then show the button)

We also need this for the current webui.

jelinaunger:
closes #4949

### DIFF
--- a/privacyidea/lib/policy.py
+++ b/privacyidea/lib/policy.py
@@ -2581,9 +2581,8 @@ def get_static_policy_definitions(scope=None):
             },
             PolicyAction.ENROLL_VIA_MULTICHALLENGE_PASSKEY_OFFLINE: {
                 'type': 'bool',
-                'desc': _(
-                    'If enabled, a successful registration with the enroll_via_multichallenge action for passkeys '
-                    'will directly mark the token for offline use and send the offline data.')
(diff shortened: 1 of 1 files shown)


privacyidea/lib/policy.py
```diff
@@ -2581,8 +2581,9 @@ def get_static_policy_definitions(scope=None):
             },
             PolicyAction.ENROLL_VIA_MULTICHALLENGE_PASSKEY_OFFLINE: {
                 'type': 'bool',
-                'desc': _('If enabled, a successful registration with the enroll_via_multichallenge action for passkeys '
-                          'will directly mark the token for offline use and send the offline data.')
+                'desc': _(
```
henrik-falk: why new line?

henrik-falk (approved):

PR #4973, opened 2026-02-20
Hide empty policy accordion

nilsbehlen:
- if a policy section is empty (because of search or "show only selected") it is hidden
- selecting policy template will not reset the priority
- buttons other than the save button don't trigger form evaluation (and there dont cause a jump to the policyname field if its empty and don't trigger the "please fill this field" hint)
- adjusted input sizes to be consistent
- selecting a policy template will close the templates
- added info that no scope is selected and therefore no actions are shown
- enable/disable/delete are only shown when the policy exists, not for creating a new one

### DIFF
--- a/privacyidea/static/components/config/views/config.policies.details.html
+++ b/privacyidea/static/components/config/views/config.policies.details.html
@@ -24,15 +24,17 @@
                 If you change the name of the policy, it will create a new policy
                 with the new name!
             </p>
-        </div>
-        <div class="col-sm-3">
-            <button class="btn btn-primary"
+            <button type="button"
+                    class="btn btn-primary"
+                    style="margin-top: 10px;"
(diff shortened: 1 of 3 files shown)


privacyidea/static/components/config/views/config.policies.details.html
```diff
@@ -24,15 +24,17 @@
                 If you change the name of the policy, it will create a new policy
                 with the new name!
             </p>
-        </div>
-        <div class="col-sm-3">
-            <button class="btn btn-primary"
+            <button type="button"
+                    class="btn btn-primary"
+                    style="margin-top: 10px;"
```
jelinaunger: inline styles should be avoided

jelinaunger (approved):

Issue #4947, opened 2026-02-11
pi-tokenjanitor should use tokenclass method


cornelinux:
the pi-tokenjanitor should use the tokenclass method `is_orphaned`.

If necessary, the method can be enhanced.

PR #4980, opened 2026-02-23
Pi tokenjanitor

Lucas-Scholz:
Adresses Issue #4947

### DIFF
--- a/tests/cli/test_cli_pitokenjanitor.py
+++ b/tests/cli/test_cli_pitokenjanitor.py
@@ -406,11 +406,11 @@ class TestPiTokenJanitorFind:
         assert "HOTP0001" in result.output
         assert "TOTP0001" not in result.output
 
-    def test_find_orphaned(self, app, tokens, orphaned_tokens):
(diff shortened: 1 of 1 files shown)


privacyidea/cli/pitokenjanitor/utils/findtokens.py
```diff
@@ -409,17 +409,15 @@ def _get_token_list(assigned: Union[bool, None], active: Union[bool, None], rang
             if orphaned is not None:
                 if is_true(orphaned):
                     try:
-                        if token_obj.user is None or token_obj.user.exist():
-                            # Either the token has no user assigned or the
-                            # assigned user exists in the resolver
+                        if token_obj.is_orphaned(orphaned_on_error) is False:
                             add = False
                     except ResolverError:
```
frankmer: errors are already caught in is_orphaned at tokenclass.py:362
Do we still need it here?
When we don't need to catch errors here we can simplify the code to three lines:
```
            if orphaned is not None:
                if token_obj.is_orphaned(orphaned_on_error=orphaned_on_error) != is_true(orphaned):
                    add = False
```

PR #4981, opened 2026-02-23
fix webhook content type

nilsbehlen:
* set Content-Type to application/json
* test
* added a mapping dict so old database entries also work


privacyidea/lib/eventhandler/webhookeventhandler.py
```diff
@@ -86,8 +97,8 @@ def actions(self):
                 "required": True,
                 "description": _("The encoding that is sent to the WebHook, for example json"),
```
plettich: We probably should change the description to at least "MIME type" or "Content type".

PR #4982, opened 2026-02-23
4975/enrollpin policy behavior

nilsbehlen:
* If a pin is set in the request to enroll a token, but enrollpin right is not there, a PolicyError is raised and no enrollment takes places as a policy is violated.
* Disable the PIN input in the UI if enrollpin right is not set
* Test

### DIFF
--- a/privacyidea/api/lib/prepolicy.py
+++ b/privacyidea/api/lib/prepolicy.py
@@ -506,7 +506,7 @@ def enroll_pin(request=None, action=None):
     if not allowed_action:
         # Not allowed to set a PIN during enrollment!
         if "pin" in request.all_data:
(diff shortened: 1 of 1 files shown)


privacyidea/api/lib/prepolicy.py
```diff
@@ -505,7 +506,7 @@ def enroll_pin(request=None, action=None):
     if not allowed_action:
         # Not allowed to set a PIN during enrollment!
         if "pin" in request.all_data:
-            del request.all_data["pin"]
+            raise PolicyError("Setting a PIN is not allowed due to missing 'enrollpin' right.")
```
jelinaunger: use PolicyAction.ENROLLPIN

Issue #4969, opened 2026-02-19
Challenge-Response does not work with Oracle database


plettich:
Since switching the `challenge` column in the `Challenge` table to `db.Text` (in v3.12) Oracle dbs throw an error when checking querying challenges: "ORA-00932: inconsistent datatypes: expected - got CLOB"
`db.Text` gets converted to `CLOB` in Oracle but it can not be used in a *where* clause: https://stackoverflow.com/a/12980560

Maybe we can use the `clob_to_varchar` function defined in `lib/token.py`:
https://github.com/privacyidea/privacyidea/blob/c81459578bce9a903bebc9f36aa864195025fb86/privacyidea/lib/token.py#L135

PR #4983, opened 2026-02-23
Cast Text column to VARCHAR before comparing

plettich:
Oracle DB does not allow to compare with a column of type Text (CLOB). We need to cast it before comparing.

Fixes: #4969

### DIFF
--- a/privacyidea/lib/challenge.py
+++ b/privacyidea/lib/challenge.py
@@ -116,9 +116,10 @@ def get_challenges_paginate(serial=None, transaction_id=None,
 
 def _create_challenge_query(serial: str = None, transaction_id: str = None) -> Select:
     """
-    This function create the sql query for fetching transaction_ids. It is
+    This function creates the SQL query for fetching transaction_ids. It is
     used by get_challenge_paginate.
-    :return: An SQLAlchemy sql query
+
+    :return: An SQLAlchemy SQL query
     """
     stmt = select(Challenge)
     if serial is not None and serial.strip("*"):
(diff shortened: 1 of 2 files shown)


privacyidea/lib/challenge.py
```diff
@@ -165,7 +170,7 @@ def delete_challenges(serial: str = None, transaction_id: str = None) -> int:
     return result.rowcount
 
 
-def _build_challenge_criterion(age: int = None) -> 'sqlalchemy.sql.expression.BinaryExpression':
+def _build_challenge_criterion(age: int = None) -> 'sqlalchemy.sql.expression.ColumnElement':
```
jelinaunger, line 175: Are you sure it is a ColumnElement? Cause it looks like a binary expression to me

privacyidea/models/utils.py
```diff
@@ -15,16 +15,39 @@
...
     """
     Return the current UTC time as a naive datetime object.
     """
     return datetime.now(timezone.utc).replace(tzinfo=None)
 
+
+# Define a function to convert Oracle CLOBs to VARCHAR before using them in a
+# compare operation.
+# By using <https://docs.sqlalchemy.org/en/13/core/compiler.html> we can
+# differentiate between different dialects.
```
jelinaunger: I think we do not need this documentation link here, do we? Otherwise, it should be of sqlalchemy 2

privacyidea/lib/challenge.py
plettich: According to https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.ColumnElement [`BinaryExpression`](https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.BinaryExpression) is a subclass of ColumnElement and it looks like `BinaryExpression` does not apply since the second element is not an SQLAlchemy object.
But more importantly my Editor doesn't complain ;-)

privacyidea/models/utils.py
plettich: Done in 93cd351

PR #4984, opened 2026-02-23
force updated tar version

nilsbehlen:
* pacote was pulling tar 6.2.1


privacyidea/static_new/package.json
```diff
@@ -28,6 +28,9 @@
     "tslib": "^2.3.0",
     "uuid": "^11.1.0"
   },
+  "overrides": {
+    "tar": "^7.5.8"
```
frankmer, line 32: The latest version is 7.5.9, which closed a critical security vulnerability regarding path traversal.

PR #4992, opened 2026-02-26
Remove default values for validity period

### LINKED ISSUE #4978: Token Enrollment default end time
Image
Image
Image

The Tokens only are useable for the rest of the day when not changing the default values. Witch are only visible after scrolling down and then expanding the Extended Attributes

jelinaunger:
closes #4978

### DIFF
--- a/privacyidea/static_new/src/app/components/token/token-enrollment/token-enrollment.component.html
+++ b/privacyidea/static_new/src/app/components/token/token-enrollment/token-enrollment.component.html
@@ -339,7 +339,7 @@
             </mat-form-field>
 
             <mat-form-field class="time-picker">
-              <mat-label>Validity End Time</mat-label>
+              <mat-label i18n>Validity End Time</mat-label>
               <input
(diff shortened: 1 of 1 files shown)


privacyidea/static_new/src/app/components/token/token-enrollment/token-enrollment.component.html
```diff
@@ -326,21 +325,21 @@ <h4 class="flex-column enroll-h4">Add Token to a Container</h4>
...
+              <mat-hint i18n>Date Format: YYYY-MM-DD</mat-hint>
               <mat-datepicker-toggle
                 [for]="endPicker"
                 matIconSuffix></mat-datepicker-toggle>
               <mat-datepicker #endPicker></mat-datepicker>
             </mat-form-field>
 
             <mat-form-field class="time-picker">
-              <mat-label>Enter an End Time</mat-label>
+              <mat-label>Validity End Time</mat-label>
```
frankmer: i18n

PR #4994, opened 2026-02-27
1991/change webauthn library

### LINKED ISSUE #1991: PyOpenSSL suggest using cryptography
According to https://pypi.org/project/pyOpenSSL/ and https://www.pyopenssl.org/en/stable/api/crypto.html, the `PyOpenSSL` authors suggest using [`pyca/cryptography](https://github.com/pyca/cryptography) for anything other than SSL-connections.

- [x] #4212
- [x] [`localca.py`](https://github.com/privacyidea/privacyidea/blob/d5eb8fe722351218ce90ddef1d488a465bbf3cf7/privacyidea/lib/caconnectors/localca.py#L397)
- [ ] [`webauthn.py`](https://github.com/privacyidea/privacyidea/blob/d5eb8fe722351218ce90ddef1d488a465bbf3cf7/privacyidea/lib/tokens/webauthn.py#L1877) #4820 
- [x] ~[`u2f.py`](https://github.com/privacyidea/privacyidea/blob/d5eb8fe722351218ce90ddef1d488a465bbf3cf7/privacyidea/lib/tokens/u2f.py#L121)~ (this one is deprecated anyway, won't change it)

jona-samuel:
closes #1991

### DIFF
--- a/dev/token-deprecation-strategy.md
+++ b/dev/token-deprecation-strategy.md
@@ -60,16 +60,21 @@ is a second line of defence: any code path that bypasses the class
 (e.g. raw SQL reporting, future refactors) still sees the token as
 disabled. It costs one extra `UPDATE` in the migration.
 
-### Why `is_challenge_request` returns False instead of raising
+### Why authentication methods return failure values instead of raising
 
 A user with a deprecated token *and* another working token should not
-be locked out. Raising during challenge evaluation would abort the
-whole auth flow; returning False makes the deprecated token silently
-skipped, and the user's other tokens are still evaluated. The failure
-mode for a user with *only* a deprecated token is "no working token"
-rather than "clear error", which is worse for debugging — but the
-prominent migration log and the admin's cleanup workflow are the
-primary signal, not a runtime error.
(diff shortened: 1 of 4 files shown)


dev/token-deprecation-strategy.md
```diff
@@ -0,0 +1,190 @@
...
+A future UI improvement could surface `original_tokentype` as a hint
+column, but it is not required for the deprecation to be usable.
+
+### Which methods are refused
+
+`DeprecatedTokenClass` overrides the following methods from
+`TokenClass` to raise `NoLongerSupportedError`:
+
+- **Authentication:** `check_otp`, `check_challenge_response`,
+  `authenticate`, `create_challenge`
```
jelinaunger: Wouldn't this also break the authentication flow? At least create_challenge at trigger_challenge endpoint I guess. The others are maybe never reached due to filtering out disabled token types.

jelinaunger: A test for this might be good anyway (user with working and deprecated token can authenticate with valid token but not with deprecated one)

PR #4990, opened 2026-02-25
4958/standardize style and sizes

henrik-falk:

PR #5002, opened 2026-03-02
4950/SMTP CA and Resolver Fixes

### LINKED ISSUE #4950: SMTP CA and Resolver Fixes
- [x] Changes from #4288  are missing in the new WebUI 
- [x] When editing an existing Resolver the save or discard dialog will always pop up 
- [x] discard changes for edit resolver not possible
- [x] expression has changed after it has been checked error in the console after saving resolvers
- [x] ldap resolver: base dn input field for recursive group config
- [x] Users overview: when changing the realm, but the request fails for some reason, the users from the previous realm are still shown while the realm selection staxs on the new but failed realm.

HTTP resolvers: 
   - [x] Method selection should be case insensitive
   - [x] get dfaults (for pre-selection) from the server instead of hardcoding them in the web ui
   - [x] After editing EntraID resolver the save button is still disabled

henrik-falk:


privacyidea/static_new/src/app/components/external-services/smtp-servers/new-smtp-server/new-smtp-server.component.html
```diff
@@ -136,16 +136,63 @@
...
+
+      <mat-form-field
+        appearance="fill"
+        class="width-100">
+        <mat-label i18n>Password for the private key</mat-label>
+        <input
+          formControlName="private_key_password"
+          matInput
+          placeholder="topsecret"
+          type="password" />
```
jelinaunger, line 188: For some reason this field is already pre filled which should not be the case

jelinaunger (changes requested):
* create / edit resolvers: Discard changes is not possible (also the text in the dialog is weird)
* edit passwd resolver: file is not pre-filled with existing value
* request to get http resolver defaults does not happen (I think cause the ui defaults are still passed into the dialog)

privacyidea/static_new/src/app/components/external-services/smtp-servers/new-smtp-server/new-smtp-server.component.html
henrik-falk: For me its not prefilled. Whats in it? There should only be the placeholder

henrik-falk: Discard changes was resolved in #4990

henrik-falk: passwd resolver is prefilled with either etc/passwd or the existing value for me.

jelinaunger: for the passwd resolver: the old webUI used the key "filename" while the new one uses "fileName". The API knows both

jelinaunger: regarding the password of the smtp server, I can also not reproduce this anymore

privacyidea/static_new/src/app/services/token/token.service.ts
```diff
@@ -31,7 +31,7 @@ import {
   throwError,
   timer
 } from "rxjs";
-import { environment } from "../../../environments/environment";
+import { environment } from "@env/environment";
```
frankmer: add "^@env/(.*)$": "<rootDir>/src/environments/$1" to moduleNameMapper in jest.config.ts so jest can find the import

henrik-falk: oh i just reverted to the old import. I will do this in #4990

privacyidea/static_new/src/app/components/external-services/smtp-servers/new-smtp-server/new-smtp-server.component.ts
frankmer: Use async save(): Promise<void> because it ensures a consistent return type and allows for cleaner, linear asynchronous logic using await. Then you also don't need to call it with Promise.resolve() anymore.

privacyidea/static_new/src/app/components/external-services/smtp-servers/new-smtp-server/new-smtp-server.component.spec.ts
```diff
@@ -79,6 +79,25 @@ describe("NewSmtpServerComponent", () => {
...
+  });
+
+  it("should show TLS when server does not start with smtps:", () => {
+    component.smtpForm.patchValue({ server: "smtp.example.com" });
+    expect(component.showTLS).toBe(true);
+
+    component.smtpForm.patchValue({ server: "smtps://smtp.example.com" });
+    expect(component.showTLS).toBe(false);
+  });
+
```
frankmer, line 159: I would implement 4 additional tests:
 - should return true for hasChanges if form is dirty
 - should return false for hasChanges if form is pristine
 - should disable identifier control when data is provided
 - should not call smtpService.postSmtpServer if the form is invalid

privacyidea/static_new/src/app/components/user/user-new-resolver/http-resolver/http-groups-attribute/http-groups-attribute.component.ts
```diff
@@ -66,9 +66,8 @@ export class HttpGroupsAttributeComponent implements OnInit, OnDestroy {
       this.activeSubscription?.unsubscribe();
       const activeControl = this.userGroupsControl.get("active");
       if (activeControl) {
-        // Initialize signal
         this.activeSignal.set(!!activeControl.value);
-        // Subscribe to changes in the "active" control to enable/disable other controls
+
```
frankmer, line 66: DRY:
```
const controls = ["user_groups_attribute", "method", "endpoint"];

const updateControls = (active: boolean) => {
  this.activeSignal.set(active);
  controls.forEach(ctrl => {
    const control = this.userGroupsControl.get(ctrl);
    if (active) {
      control?.enable({ emitEvent: false });
    } else {
      control?.disable({ emitEvent: false });
    }
  });
};
this.activeSubscription = activeControl.valueChanges.subscribe(updateControls); 
updateControls(!!activeControl.value);
```

jelinaunger (approved):

privacyidea/static_new/src/app/components/user/user-new-resolver/http-resolver/http-resolver.component.spec.ts
```diff
@@ -42,6 +50,30 @@ describe("HttpResolverComponent", () => {
...
+    expect(mockResolverService.getDefaultResolverConfig).toHaveBeenCalledWith("httpresolver");
+  });
+
+  it("should apply defaults from server", () => {
+    const defaults = {
+      endpoint: "http://default-endpoint",
+      method: "POST",
+      verify_tls: false
+    };
+    mockResolverService.getDefaultResolverConfig.mockReturnValue(of({
```
frankmer: I get this error:
Argument of type 'Observable<{ result: { status: true; value: { endpoint: string; method: string; verify_tls: boolean; }; }; }>' is not assignable to parameter of type 'Observable<PiResponse<any, any>>'.

This should work:
of(MockPiResponse.fromValue(defaults))

privacyidea/static_new/src/app/components/user/user-new-resolver/http-resolver/http-resolver.component.ts
```diff
@@ -90,8 +91,17 @@ export class HttpResolverComponent {
   ];
   protected readonly displayedColumns: string[] = ["privacyideaAttr", "userStoreAttr", "actions"];
   protected readonly CUSTOM_ATTR_VALUE = "__custom__";
+  private resolverService = inject(ResolverService);
   data = input<any>({});
   type = input<string>("httpresolver");
+  serverDefaults = signal<any>({});
```
frankmer, line 98: Is 'any' really necessary here?

privacyidea/static_new/src/app/components/user/user-new-resolver/http-resolver/http-resolver.component.ts
```diff
@@ -367,8 +387,27 @@ export class HttpResolverComponent {
 
   protected syncGroup(group: FormGroup, config: any) {
```
frankmer, line 389: 'any' is here too. We have no idea what kind of data we are dealing with here.

privacyidea/static_new/src/app/services/ca-connector/ca-connector.service.ts
```diff
@@ -38,7 +39,9 @@ export interface CaConnectorServiceInterface {
   caConnectors: WritableSignal<CaConnectors>;
 
   postCaConnector(connector: CaConnector): Promise<void>;
+
   deleteCaConnector(connectorname: string): Promise<void>;
+
   getCaSpecificOptions(catype: string, params: any): Promise<any>;
```
frankmer: Here too. When we can know what data we have, we can avoid using 'any'.

privacyidea/static_new/src/styles/styles.scss
```diff
@@ -61,32 +61,32 @@ body {
   --background-green-brand: linear-gradient(145deg, var(--green-brand-light), var(--green-brand-base));
 
   --box-shadow: var(--shadow-distance) var(--shadow-distance) var(--blur) var(--mat-sys-shadow),
-    var(--light-distance) var(--light-distance) var(--blur) var(--mat-sys-light);
+  var(--light-distance) var(--light-distance) var(--blur) var(--mat-sys-light);
```
frankmer: Please keep the indentation for multi-line values. It follows standard CSS/SCSS style guides (like Google/Airbnb) to ensure that continuation lines aren't mistaken for new property declarations during scanning.

privacyidea/static_new/src/app/components/user/user-new-resolver/http-resolver/http-resolver.component.ts
henrik-falk: any is fine if its not implicit. but i updated where we know the type

privacyidea/static_new/src/styles/styles.scss
henrik-falk: i will see that i can do in #4990

PR #5013, opened 2026-03-04
fix old migration: do not use ORM model but raw sql

nilsbehlen:
* ORM model reflects the current state, which would be the one from master, not at the point in time of the migration

### DIFF
--- a/privacyidea/migrations/versions/fa07bd604a75.py
+++ b/privacyidea/migrations/versions/fa07bd604a75.py
@@ -5,79 +5,172 @@ Revises: 3ba618f6b820
 Create Date: 2021-05-18 23:29:59.273457
 
 """
-from alembic import op
 import sqlalchemy as sa
+from alembic import op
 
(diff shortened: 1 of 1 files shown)


privacyidea/migrations/versions/fa07bd604a75.py
```diff
@@ -5,79 +5,172 @@
...
+                                 sa.column("Value", sa.UnicodeText()))
+
+_policy_table = sa.table("policy",
+                          sa.column("id", sa.Integer),
+                          sa.column("name", sa.Unicode(64)),
+                          sa.column("scope", sa.Unicode(32)),
+                          sa.column("active", sa.Boolean),
+                          sa.column("check_all_resolvers", sa.Boolean))
+
+_policy_action_table = sa.table("policyaction",
```
plettich, line 36: There is no table `policyaction` in our code, not even in v3.5.

plettich (changes requested):

PR #5015, opened 2026-03-04
add migration test and action, fix downgrade in some migrations

nilsbehlen:
Add a migration test that checks:
* Structural Integrity: Ensures the migration graph has a single head, valid parent 
  links, and descriptive messages.
* Schema Equivalence: Upgrades the database to head and uses Alembic's `compare_metadata` 
  to guarantee the resulting database schema strictly matches the SQLAlchemy ORM models.
* Granular Idempotency (Round-Trip): Marches linearly through the migration history, 
  testing the exact `upgrade -> downgrade -> upgrade` lifecycle for every single 
  revision. It compares deep schema snapshots to ensure no columns are orphaned or 
  dropped incorrectly during downgrades.
* Downgrade Safety: Verifies that walking the migration graph backward does not crash 
  and does not unintentionally destroy data in tables that survive the downgrade.

Other points:
* Fixed migration scripts where guards where missing

### DIFF
--- a/tests/test_migrations.py
+++ b/tests/test_migrations.py
@@ -1,26 +1,7 @@
 """
-Alembic Database Migration Test Suite
-
-This module validates the structural integrity, idempotency, and operational safety 
-of all database migrations from a pinned starting revision (START_REVISION) to head.
-It relies exclusively on dynamic database inspection and SQLAlchemy Core/ORM reflection 
-rather than hardcoded schema dictionaries.
-
-Key Validations:
-* Structural Integrity: Ensures the migration graph has a single head, valid parent 
-  links, and descriptive messages.
-* Schema Equivalence: Upgrades the database to head and uses Alembic's `compare_metadata` 
-  to guarantee the resulting database schema strictly matches the SQLAlchemy ORM models.
-* Granular Idempotency (Round-Trip): Marches linearly through the migration history, 
-  testing the exact `upgrade -> downgrade -> upgrade` lifecycle for every single 
-  revision. It compares deep schema snapshots to ensure no columns are orphaned or 
-  dropped incorrectly during downgrades.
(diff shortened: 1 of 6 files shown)


tests/test_migrations.py
```diff
@@ -0,0 +1,534 @@
...
+    _setup_db_at_start_revision(flask_app, START_REVISION)
+    command.upgrade(_get_alembic_cfg(), "head")
+
+    script = ScriptDirectory.from_config(_get_alembic_cfg())
+    # Walk from head down to (but not including) START_REVISION
+    revisions_in_window = list(script.iterate_revisions("head", START_REVISION))
+
+    for rev in revisions_in_window:
+        if rev.down_revision is None:
+            break
```
jelinaunger: Shouldn't all migrations script have a down_revision? I would expect that if there is no down revision the test also fails cause the downgrade is not possible.

privacyidea/migrations/versions/b637a1eeeeeb_.py
```diff
@@ -1,4 +1,4 @@
-"""empty message
+"""v3.14: Add S/MIME columns to smtpserver table
```
jelinaunger: 3.13

tests/test_migrations.py
```diff
@@ -0,0 +1,534 @@
...
+
+    This avoids running migrations from 10 years ago that expect tables to
+    already exist (ALTER TABLE etc.) from an empty database.
+    """
+    from alembic import command
+    from privacyidea.models import db
+
+    db.create_all()
+    command.stamp(_get_alembic_cfg(), "head")
+    command.downgrade(_get_alembic_cfg(), start_revision)
```
jelinaunger: It might be better to directly create the table at start revision. Cause in this way we already assume that the downgrade works properly (even if we might notice that it is not in your other tests)

PR #5031, opened 2026-03-09
Create, edit, and delete user

### LINKED ISSUE #4979: Edit user
On the user details page it should be possible to edit the user attributes (not only custom), if this is configured for the resolver.
The returned user also has an editable attribute if this option should be allowed or not.

old UI:

Image

\+ create 
\+ delete

jelinaunger:
close #4979

### DIFF
--- a/privacyidea/static_new/src/app/components/user/user-details-edit/user-details-edit.component.html
+++ b/privacyidea/static_new/src/app/components/user/user-details-edit/user-details-edit.component.html
@@ -17,19 +17,15 @@
  SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 
-<div class="flex-row gap-16">
-  @for (column of [0, 1]; track column) {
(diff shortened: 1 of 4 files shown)


privacyidea/static_new/src/app/components/user/user-details-edit/user-details-edit.component.html
```diff
@@ -0,0 +1,35 @@
...
+ GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+
+ You should have received a copy of the GNU Affero General Public
+ License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+ SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<div class="flex-row gap-16">
+  @for (column of [0, 1]; track column) {
```
henrik-falk: attribute grid instead of for loop is maybe a bit cleaner?

PR #5068, opened 2026-03-16
Keycloak set request parameter exact for non-wildcard search

### LINKED ISSUE #5057: Keycloak resolver always does a wildcard user matching
### Top-level intent

I have a realm with an ldap resolver and a keycloak resolver.

When I call

    GET /user/ username=*cornelius*

I get a list of

* user cornelius from LDAP resolver
* user cornelius2 from KC resolver
which is correct.

But when I call

   GET /user username=cornelius

I also get both the users, since the KC resolver always does a wildcard matching or an un-exact match.

The other resolvers actually explicitly use the wildcards or expect wildcards to find unexpected users.

See e.g. the SQL resolver, which replaces the "*" with the SQL search pattern:

https://github.com/privacyidea/privacyidea/blob/0caa296276ee5b1f0eff02ea9b8150cd706dbce5/privacyidea/lib/resolvers/SQLIdResolver.py#L414

### Expected outcome

The KC resolver should, if the searchdict does not contain a wildcard
https://github.com/privacyidea/privacyidea/blob/9e8927976712ebeb2f428417902bbe749965c3a4/privacyidea/lib/resolvers/HTTPResolver.py#L438
do an exact match.

jelinaunger:
closes #5057

### DIFF
--- a/privacyidea/lib/user.py
+++ b/privacyidea/lib/user.py
@@ -834,7 +834,7 @@ def get_user_list(param: dict = None, user: User = None, include_custom_attribut
             if not resolver:
                 continue
             log.debug("With this search dictionary: %r", search_dict)
-            requested_pi_user_attributes = list({"resolver", "editable"}.intersection(requested_attributes))
+            requested_pi_user_attributes = list({"resolver", "editable"}.intersection(requested_attributes or []))
             requested_user_store_attributes = list(set(requested_attributes or []) - set(requested_pi_user_attributes))
             user_list = resolver.getUserList(search_dict, requested_user_store_attributes)
             # Add resolvername to the list
--- a/privacyidea/lib/resolvers/HTTPResolver.py
+++ b/privacyidea/lib/resolvers/HTTPResolver.py
@@ -847,7 +847,8 @@ class HTTPResolver(UserIdResolver):
         Maps the attributes from the user store to the attributes used in privacyidea.
 
         :param user: Dictionary containing user attributes from the user store
-        :param attributes: List of attributes to be included in the returned dictionary. If None, all attributes are included.
+        :param attributes: List of attributes to be included in the returned dictionary. If None or an empty list, all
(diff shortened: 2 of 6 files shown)


privacyidea/lib/resolvers/KeycloakResolver.py
```diff
@@ -160,6 +160,51 @@ def _replace_resolver_specific_tags(self, config: RequestConfig):
...
+        """
+        Fetches a list of users from the user store.
+
+        :param search_dict: Dictionary containing search parameters that are added as query to the endpoint url
+        :param config: Configuration contains all information of the api endpoint to fetch the users.
+        :param attributes: List of attributes that should be included in the response. If None, all attributes are included.
+        :return: List of dictionaries containing pi conform user attributes
+        """
+
+        request_params = config.request_mapping if config.request_mapping else {}
```
nilsbehlen: `request_params = config.request_mapping or {}`

privacyidea/lib/user.py
```diff
@@ -834,13 +834,15 @@ def get_user_list(param: dict = None, user: User = None, include_custom_attribut
             if not resolver:
                 continue
             log.debug("With this search dictionary: %r", search_dict)
-            user_list = resolver.getUserList(search_dict, requested_attributes)
+            requested_pi_user_attributes = list({"resolver", "editable"}.intersection(requested_attributes)) if requested_attributes else []
```
nilsbehlen: use `or`

privacyidea/lib/user.py
```diff
@@ -834,13 +834,15 @@ def get_user_list(param: dict = None, user: User = None, include_custom_attribut
...
+            requested_user_store_attributes = list(set(requested_attributes) - set(requested_pi_user_attributes)) if requested_attributes else None
```
nilsbehlen: or

PR #5078, opened 2026-03-17
keycloak group key

### LINKED ISSUE #5058: The keyname for user groups should be configurable in Keycloak resolver

A realm has several resolvers. We are using user groups.

The LDAP resolver allows the mapping of *anything* -> *memberOf*, this way have a configurable privacyIDEA user attribute key "anything", that contains the group memberships.

The KC resolver currently has the groups in the fixed attribute *groups*.

In this case in a multi resolver realm some users have their groups in "anything" and other users in "groups".

The KC resolver should also offer the possibility to use an arbitrary key-name for the group memberships.

jelinaunger:
closes #5058

### DIFF
--- a/tests/test_lib_resolver_httpresolver.py
+++ b/tests/test_lib_resolver_httpresolver.py
@@ -1088,11 +1088,12 @@ class HTTPResolverTestCase(MyTestCase):
 
         # advanced settings with groups
         config = dict(self.advanced_config)
+        custom_groups_key = "custom-groups"
         config[CONFIG_GET_USER_GROUPS] = {ACTIVE: True, METHOD: "GET", ENDPOINT: "/users/{userid}/groups",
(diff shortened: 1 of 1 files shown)


tests/test_lib_resolver_httpresolver.py
```diff
@@ -1353,6 +1377,18 @@ def test_06_getUserList_success(self):
             else:
                 self.assertListEqual([], user["groups"])
 
+        # with custom groups key
+        resolver.config_get_user_groups = {ACTIVE: True, USER_GROUPS_ATTRIBUTE: "displayName", PI_USER_GROUPS_KEY: "custom-groups"}
```
nilsbehlen: the custom group key "custom-groups" should be a variable that is reused throughout this test

PR #5083, opened 2026-03-18
token config fixes

### LINKED ISSUE #5042: Token Type Config Fixes
* Fix selection for SMS gateways & SMTP server (also in system config)
* use clearable selection instead of None option

jelinaunger:
closes #5042

### DIFF
--- a/privacyidea/static_new/src/app/components/configuration/token-type-config/token-types/yubikey-config/yubikey-config.component.spec.ts
+++ b/privacyidea/static_new/src/app/components/configuration/token-type-config/token-types/yubikey-config/yubikey-config.component.spec.ts
@@ -63,9 +63,8 @@ describe("YubikeyConfigComponent", () => {
     jest.spyOn(component.onYubikeyCreateNewKey, "emit");
     const newId = "newClientId";
     component.newYubikeyApiId.set(newId);
-    component.newYubikeyApiKey.set("newApiKey");
     component.createNewKey();
-    expect(component.onYubikeyCreateNewKey.emit).toHaveBeenCalledWith(newId);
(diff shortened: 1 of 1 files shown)


privacyidea/static_new/src/app/components/configuration/token-type-config/token-type-config.component.ts
```diff
@@ -194,7 +203,12 @@ export class TokenTypeConfigComponent implements OnInit, AfterViewInit {
       next: (response) => {
         if (response?.result?.status) {
           this.notificationService.openSnackBar($localize`System entry deleted.`);
-          this.systemService.systemConfigResource.reload();
+          // Update entries in the formData but not reload the whole config to prevent losing unsaved changes
```
frankmer, line 208: 1. The problem wasn't that it reloaded the systemConfigResource. The problem was that an actual request was made to delete it.
2. We should save the 'patch' or 'changes' independently of the systemConfigResource so that they are not affected when we get a new systemConfigResource.

yubikeyCreateNewKey does not actually create a new key. It simply gets a random key with len=20 and encode=b64 and ignores the entered key.
It seems okay that the key only gets saved when we press the save button, because everything else here also only gets saved when we click the save button. However, the key is removed instantly when klicking the remove button.

jelinaunger: > 1. The problem wasn't that it reloaded the systemConfigResource. The problem was that an actual request was made to delete it.

You can turn it in both directions. However, removing the reload was the more simple solution for now to fix overwrite pending changes. Created a new issue #5105 to improve this.

> yubikeyCreateNewKey does not actually create a new key. It simply gets a random key with len=20 and encode=b64 and ignores the entered key.

Well the UI seemed to be confusing here. Create new key actually did the right thing: It generates a random key. But you can also enter your own custom key (no create key request required), which was not yet covered.

PR #5096, opened 2026-03-20
5070/consistently use rollout state and make sure every token has one

### LINKED ISSUE #5070: Consistently use RolloutState and make sure every token has one
Currently, the RolloutState of a token is only used/set if there is a "use" for it, e.g. to indicate verify_enrollment or 2step. But token that are enrolled via one-shot, have an empty RolloutState. And e.g. after completing 2step, the token will also have an empty rolloutstate.

We need to make sure the rolloutstate is set consistently, so that every token always has a state.

jona-samuel:
closing #5070


tests/test_api_2stepinit.py
```diff
@@ -406,7 +406,7 @@ def test_04_no_2stepinit(self):
             serial = detail.get("serial")
             otpkey_url = detail.get("otpkey", {}).get("value")
             otpkey_bin = binascii.unhexlify(otpkey_url.split("/")[2])
-            self.assertEqual(detail.get("rollout_state"), "")
+            self.assertEqual(detail.get("rollout_state"), "enrolled")
```
nilsbehlen, line 409: tests should compare with the value from the enum ``RolloutState.ENROLLED``

nilsbehlen (changes requested):

nilsbehlen (approved):

PR #5101, opened 2026-03-23
5073/SMS / E-Mail rollover should override previous settings in token info

### LINKED ISSUE #5073: SMS / E-Mail rollover should override previous settings in token info
For SMS and E-Mail tokens one can either select to dynamically read the phone number / mail from the user info or set a fixed one. 
This info is stored and read from the token info.
When doing a token rollover and changing this setting either from dynamic read to a fixed value or the other way around, the previous setting stays in the token info. This can cause unexpected behavior. 
We should delete any previous enrollment info from the token info on rollover if a new setting is provided and only keep this one.

jona-samuel:
Closes #5073

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -40,6 +40,7 @@ Version 3.13
   * Fixed TLS configuration for machine resolver (#4964)
   * Verify token enrollment now throws an error for a failed verification (#5056)
   * Fixed displaying of checkbox to use userstore password for container registration (#5114)
+  * Fixed updating of phone and email settings in the token info during SMS and email token rollover (#5073)
 
 Version 3.12.2, 2025-12-04
 


privacyidea/lib/tokens/emailtoken.py
```diff
@@ -228,11 +228,13 @@ def update(self, param, reset_failcount=True):
         if not verify:
             if getParam(param, "dynamic_email", optional=True):
                 self.add_tokeninfo("dynamic_email", True)
+                self.delete_tokeninfo(self.EMAIL_ADDRESS_KEY)
             else:
                 # specific - e-mail
                 self._email_address = getParam(param,
                                                self.EMAIL_ADDRESS_KEY,
                                                optional=False)
+                self.delete_tokeninfo("dynamic_email")
```
jelinaunger: Could you please add a constant for the dynamic_email key?

privacyidea/lib/tokens/smstoken.py
```diff
@@ -276,10 +276,12 @@ def update(self, param, reset_failcount=True):
         if not verify:
             if getParam(param, "dynamic_phone", optional):
                 self.add_tokeninfo("dynamic_phone", True)
+                self.delete_tokeninfo("phone")
             else:
                 # specific - phone
                 phone = getParam(param, "phone", required)
                 self.add_tokeninfo("phone", phone)
+                self.delete_tokeninfo("dynamic_phone")
```
jelinaunger: Can you also add constants for these two keys?

tests/test_lib_tokens_email.py
```diff
@@ -597,3 +597,40 @@ def test_25_email_token_import(self):
...
+        """
+        Test that updating a token with dynamic_email removes the static email
+        address and vice versa.
+        """
+        # 1. Create a token with a static email address
+        db_token = Token("PIEM_DYN_TEST1", tokentype="email")
+        db_token.save()
+        token = EmailTokenClass(db_token)
+        token.update({"email": "static@example.com"})
+        token.save()
```
jelinaunger: use init_token library function instead

tests/test_lib_tokens_email.py
```diff
@@ -597,3 +597,40 @@ def test_25_email_token_import(self):
...
+
+        # Verify the static email is set and dynamic_email is not
+        self.assertEqual(token.get_tokeninfo("email"), "static@example.com")
```
jelinaunger: revert argument order (expected, actual)

PR #5102, opened 2026-03-23
5088 user and user realm filter on enter

henrik-falk:


privacyidea/static_new/src/app/components/token/token-table/token-table.component.html
```diff
@@ -83,14 +87,36 @@
...
-                        </mat-icon>
-                      </button>
+                      @if (apiFilterKeyMap[column.key] === 'realm' || apiFilterKeyMap[column.key] === 'tokenrealm') {
+                        <button
+                          [matMenuTriggerFor]="realmMenu"
+                          [matMenuTriggerData]="{ keyword: apiFilterKeyMap[column.key] }"
+                          class="filter-button"
+                          mat-icon-button>
+                          <mat-icon class="keyword-button-icon margin-left-0">
+                            {{ getFilterIconName(apiFilterKeyMap[column.key]) }}
```
jelinaunger: Clicking on a clear filter icon to open a realm selection isn't intuitive. Maybe we do not toggle the icon here and always keep the filter icon (I also have no better idea for now)

jelinaunger: also adding a tooltip could be helpful

privacyidea/static_new/src/app/components/token/token-table/token-table.component.ts
```diff
@@ -111,6 +118,22 @@ export class TokenTableComponent implements AfterViewInit, OnDestroy {
...
+    source: () => this.tokenService.tokenFilter().filterString,
+    computation: (v) => v
+  });
+
+  protected readonly showFilterHint = computed(() => {
+    const current = this.filterInputValue().trim().toLowerCase();
+    const applied = this.tokenService.tokenFilter().filterString.trim().toLowerCase();
+
+    if (current !== applied) {
+      return current.includes("user:") || current.includes("realm:") || current.includes("tokenrealm:");
```
jelinaunger: For tokenrealm pressing enter should not be required (it also works with wildcards)

privacyidea/static_new/src/app/components/token/token-table/token-table.component.ts
```diff
@@ -221,6 +253,14 @@ export class TokenTableComponent implements AfterViewInit, OnDestroy {
         currentValue: this.tokenService.tokenFilter()
       });
     }
+
+    if (filterKeyword === "user" && newValue.hasKey("user") && !newValue.hasKey("realm")) {
+      const defaultRealm = this.realmService.defaultRealm();
+      if (defaultRealm) {
+        newValue = newValue.addEntry("realm", defaultRealm);
```
jelinaunger, line 271: It would be nice to first have the realm and the user comes last. Cause than I can just start typing the users name without changing cursor position.

henrik-falk: fixed by changing the focus instead

PR #5103, opened 2026-03-23
Fix realm table and user resolver components, improve node handling and UI interactions

### LINKED ISSUE #5045: Resolver & Realm Fixes
### Realm
Node selection broken

#### LDAP
* Bind type not set correctly
* If STARTTLS is not selected, initially TLS version and verify TLS certificate are not disabled
* Create New: TLS inputs behave not correctly (or as in the old UI)
   Expected behaviour:
     - server URI *ldaps* -> STATTLS appears / is enabled, selecting it also enables the two TLS inputs
     - server URI *ldap:* -> only TLS inputs appear / are enabled (not STARTTLS)

#### SQL
* remove Editable from caption of last box, cause the editable user store input is in another box

frankmer:
Resolver & Realm Fixes #5045


privacyidea/static_new/src/app/components/user/user-new-resolver/ldap-resolver/ldap-resolver.component.ts
```diff
@@ -54,60 +51,100 @@ import { startWith } from "rxjs/operators";
...
   ldapBaseControl = new FormControl<string>("", { nonNullable: true, validators: [Validators.required] });
-  loginNameAttributeControl = new FormControl<string>("", { nonNullable: true, validators: [Validators.required] });
-  ldapSearchFilterControl = new FormControl<string>("", { nonNullable: true, validators: [Validators.required] });
-  userInfoControl = new FormControl<string>("", { nonNullable: true, validators: [Validators.required] });
+  // Server URI
+  ldapUriControl = new FormControl<string>("", { nonNullable: true, validators: [Validators.required] });
+  // STARTTLS
   startTlsControl = new FormControl<boolean>(true, { nonNullable: true });
-
+  // TLS Version
```
jelinaunger: these single comments above each variable are unnecessary in my opinion, cause the variable name itself should already be meaningful enough

jelinaunger (approved):

henrik-falk (approved):

PR #5160, opened 2026-04-01
fix max width

### LINKED ISSUE #5126: Support zoom levels in new WebUI (Success Criterion)
https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-scale.html

with @media https://sass-lang.com/documentation/at-rules/css/

### LINKED ISSUE #5142: Fix max width for all components
When having a small browser window, the content is not always visible. It is shifted behind the navigation bar and truncated at the right as well. due to a wrong definition of the max-width of the main component div.

This one is correct for admin views:
`max-width: calc(100vw - var(--drawer-width) - 2 * var(--double-global-padding));`
We could also define it as a global variable and the same for the self-service. Or define a mixin style for it (for the main component div generally, which is basically `table-container-styles` but should not be restricted to tables)

Additionally, we should add a horizontal scroll bar to make the full content accessible.

I think this holds for almost all components, except events list and audit log.

jelinaunger:
closes #5142


henrik-falk:
closes  #5126 too

henrik-falk (approved):

PR #5170, opened 2026-04-03
refactor: replace user.exist() with user.is_resolved() for clarity an…

### LINKED ISSUE #4788: User.exist() not really checks if the user exist
The function `User.exist()` only checks if the user ID and the realm ID are available. However, this is no prove that the user really exists. Consequently, when search for orphaned tokens some can not be correctly identified: The ones where the realm and resolver still exist, but the user does not exist in the user store anymore. Since the user id is stored in the db, the exist function returns false. 

https://github.com/privacyidea/privacyidea/blob/217a97340995a4cad078cef3d1af185db2057dac/privacyidea/lib/user.py#L303-L310

We could ask the user store to be sure if the user really exist. However, in most scenarios this might be an unnecessary call.
To fix the token janitor orphaned search it would be enough to additionally check if the login is set:

`exist = self.uid and self.realm_id and self.login`

nilsbehlen:
…d accuracy

### DIFF
--- a/privacyidea/lib/user.py
+++ b/privacyidea/lib/user.py
@@ -88,7 +88,7 @@ class User:
     # NOTE: Directly decorating the class ``User`` breaks ``isinstance`` checks,
     # which is why we have to decorate __init__
     @log_with(log)
-    def __init__(self, login="", realm="", resolver="", uid=None):
+    def __init__(self, login="", realm="", resolver="", uid=None, _resolved=False):
         self.login = login or ""
         self.used_login = self.login
         self.realm = (realm or "").lower()
@@ -108,8 +108,9 @@ class User:
         if not self.login and not self.resolver and uid is not None:
             raise UserError("Can not create a user object from a uid without a resolver!")
         # Enrich user object with information from the userstore or from the
-        # user cache
-        if login or uid is not None:
+        # user cache, unless _resolved=True signals that this is a cheap
+        # reference-only construction (no LDAP call desired).
+        if not _resolved and (login or uid is not None):
             self._get_user_from_userstore()
             # Just store the resolver type
             self.rtype = get_resolver_type(self.resolver)
@@ -144,6 +145,16 @@ class User:
             return True
         else:
             return False
+
+    def is_resolved(self) -> bool:
+        """
+        True if the user was successfully confirmed in an external store during
+        construction: a non-empty uid, a resolver, and a valid realm_id must all
+        be set.  Use this instead of ``exist()`` when you want to check whether
+        the user was found during construction rather than performing a new live
+        store lookup.
+        """
+        return bool(self.uid) and bool(self.resolver) and self.realm_id is not None
 
     def __eq__(self, other):
         """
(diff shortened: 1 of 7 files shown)


nilsbehlen:
this is just an intermediate step tho, for a larger refactoring which is already continued in #5212

privacyidea/lib/user.py
```diff
@@ -145,6 +146,16 @@ def is_empty(self) -> bool:
...
 
+    def is_resolved(self) -> bool:
+        """
+        True if the user was successfully confirmed in an external store during
+        construction: a non-empty uid, a resolver, and a valid realm_id must all
+        be set.  Use this instead of ``exist()`` when you want to check whether
+        the user was found during construction rather than performing a new live
+        store lookup.
+        """
+        return bool(self.uid) and bool(self.resolver) and self.realm_id is not None
```
jelinaunger, line 157: also check if login is defined. Cause e.g. for token owners uid, resolver and realm are stored in the db, hence all three could be available even if the user does not exist anymore

nilsbehlen:
### Follow-up note: close the empty-`User` vs `None` policy-matching ambiguity

Parking this here so it isn't lost — it's the step that connects this user-resolving rework to the class of bug fixed in #5536.

**The problem this rework sets up to fix (but doesn't yet):** `match_policies` infers meaning from the *emptiness* of the user object rather than from an explicit contract:

- `None` → user filter disabled → **all** policies match
- empty `User()` → filters on `""` → **only unscoped** policies match
- resolved `User` → unscoped + that user's scoped policies

Because `get_user_from_param` / `UserRepository.find_from_params` always returns an *empty* `User` (never `None`), the "match all" path is almost never intended, yet the `None`-vs-empty distinction is load-bearing. Decorators each have to decide which object to pass to `Match`, and getting it wrong silently skips scoped policies (see #5536: `check_otp_pin` by serial, `init_random_pin` OTPPINCONTENTS).

This transition also *widens* the state space before narrowing it: after `user_ref()`, a ref is `login/realm` set but `uid`/`realm_id` unset (`is_resolved()` is `False`), so we now have four shapes flowing into policy code — `None`, empty `User`, ref-but-unresolved, resolved.

**Proposed step (after the `UserRef` / dataclass split in "What Comes Next"):**

1. Make `match_policies` **type-dispatch** instead of inferring from `is_empty()`:
   - `None` → no user filter (explicit "match all")
   - `UserRef` (login/realm, unresolved) → scope by the *claimed* login/realm
   - `ResolvedUser` → scope + verified identity
2. Add a single `policy_user(request)` seam (natural home: the repository/factory layer) so the "`request.User` vs token owner vs `None`" decision from #5536 is made **once**, structurally, rather than re-derived in each decorator.

**Constraints any solution must satisfy:**
- `tests/test_lib_policy.py::test_06_token` — ownerless-token semantics rely on "empty → only unscoped". The broad "normalize empty → `None` at the Match layer" fix was already tried and rejected because it breaks this. So the states can't just be collapsed; the fix is to make the distinction *explicit and typed*, not to erase it.
- The `check_otp_pin` guard added in #5536 (`if token_owner and not request.User`) and its rationale (an unowned token's `.user` is `None`; passing `None` to `check_pin` matches every policy) is the kind of per-call-site decision this seam should absorb.

Refs: #5536 (the concrete bugs), `test_06_token` (the constraint).

PR #5174, opened 2026-04-08
5063 get ordered resolvers

### LINKED ISSUE #5063: Inconsistancy with multiple resolvers in a realm between /validate/check and GET /user/
### Top-level intent

Assume we have a realm that contains several resolvers.

The /validate/check endpoint creates a user object in the request. The creation of the `User() ` object usually happens via `User(login="admin",realm="someRealm")`. To find the correct user-resolver in this realm, the User class uses the method `get_ordered_resolver`.

https://github.com/privacyidea/privacyidea/blob/6540367e60d2fe2bc76204ae80ab7913f63b2268/privacyidea/lib/user.py#L205

This is intentional. Since it can happen, that two resolvers contain the same username. In this case privacyIDEA would identify the userobject, that is found with this username in the most priorized resolver.

However, in the case of `GET /users/`  I can query `GET /users/ username=admin realm=someRealm` and the API returns a non-ordered or non-predictive list of all users with the name admin found in different resolvers in that realm.

The problem arrises, that when you then click in the API to view the user details, you would see the details of an arbitrary "admin".

When you then again would do an assign-token for this user, the /token/assign endpoint would again use the User object constructor to create the same userobject, that was also used in /validate/check.

### Expected outcome

The expectation is, that the `GET /users/` endpoint should return only the first occurance (by get_ordered_resolvers) of the userobject with the given name "admin".

This way the correct details of the user will be displayed which would finally assign the token to this very user anyways.

(Actually users with the name "admin" in lower priority resolvers are intentionally masked from being accessable)

### Solution

1.  To have a common implementation for User object creation and user listing, we can convert the method `get_ordered_resolvers` to a function in `lib.users.py`, since it does not contain any user specifcs anyways. It does not need to be a class method of the User class.
2. We can then user the `get_orders_resolvers` the modify the way, we iterate over resolvers here. https://github.com/privacyidea/privacyidea/blob/6540367e60d2fe2bc76204ae80ab7913f63b2268/privacyidea/lib/user.py#L829
Maybe - if more than one realm is requested - we first need to iterate via realms and the itereate via *ordered resolvers*.
3. When appending the found users from a resolver here (https://github.com/privacyidea/privacyidea/blob/6540367e60d2fe2bc76204ae80ab7913f63b2268/privacyidea/lib/user.py#L854), we need to check
    a) Is the list, we want to append of len 1 (one specific user)
    b) then we first check, if a user with this username and realm is already contained in the list and only append, if not so.
  This way, only the first appearance of the user "admin" in realm "someRealm" will be added to the returned list.

### Steps

* [x] have a function `get_ordered_resolvers` Done in b86856f075767d8b0248c45587b381c121219f98
* [x] use the above funtion in `get_user_list`.

# Bummer / To be discussed

There is a logical problem in the lib function `get_user_list`. It returns a list of user dictionaries. These dicts contain the login name, the UID and the resolver. But they do not contain the realm.

This is no problem for the WebUI, since it only calls this function with a given realm. However, this function can be called without a realm. Thus returnind users in all realms. And returning user information without realm information. It is not clear, in which realm this user is located.

I **solved** this by changing the implementation. The old implentation simply collected all possible resolvers, thus forgetting the information in which realm a found user actually is. The new implementation iterated over all realms and thus keeping the priority of the resolvers per realm intact.

cornelinux:
Please check the connected issue for explanation what I have done and why.

After running all tests I could also rebase locally.

### DIFF
--- a/privacyidea/lib/user.py
+++ b/privacyidea/lib/user.py
@@ -779,6 +779,7 @@ def get_user_list(param: dict = None, user: User = None, include_custom_attribut
     for pu_realm in [param_realm, user_realm]:
         if pu_realm:
             realm_config = get_realm(pu_realm)
+            # Fixme: Use get_ordered resolvers
             for r in realm_config.get("resolver", {}):
                 if r.get("name"):
                     if not r.get("node") or r["node"] == local_node_uuid:
@@ -789,6 +790,7 @@ def get_user_list(param: dict = None, user: User = None, include_custom_attribut
         # in all realms
         all_realms = get_realms()
         for _name, res_list in all_realms.items():
+            # Fixme: From each realm we need to get the ordered resolvers
             for resolver_entry in res_list.get("resolver"):
                 if not resolver_entry.get("node") or resolver_entry["node"] == local_node_uuid:
                     resolvers.append(resolver_entry.get("name"))
@@ -823,6 +825,7 @@ def get_user_list(param: dict = None, user: User = None, include_custom_attribut
                     # Remove the userid if it is not requested, as it is only needed for the custom attributes
                     user_info.pop("userid", None)
             log.debug("Found this userlist: {0!r}".format(user_list))
+            # Fixme: Add the user one by one and only, if it does not exist in the list, yet.
             users.extend(user_list)
 
         except (ResolverError, ParameterError) as ex:
--- a/privacyidea/lib/user.py
+++ b/privacyidea/lib/user.py
@@ -59,6 +59,7 @@ from .framework import get_app_config_value
 from .log import log_with
 from .realm import (get_realms, realm_is_defined,
                     get_default_realm,
+                    get_ordered_resolvers,
                     get_realm, get_realm_id)
 from .resolver import (get_resolver_object,
                        get_resolver_type)
@@ -201,34 +202,6 @@ class User:
 
     __nonzero__ = __bool__
 
-    @log_with(log)
-    def get_ordered_resolvers(self) -> list[str]:
-        """
-        returns a list of resolver names ordered by priority.
-        The resolver with the lowest priority is the first.
-        If resolvers have the same priority, they are ordered alphabetically.
-
-        :return: list of resolver names
-        :rtype: list
-        """
-        resolver_tuples = []
-        realm_config = get_realms(self.realm)
-        resolvers_in_realm = realm_config.get(self.realm, {}).get("resolver", [])
-        for resolver in resolvers_in_realm:
-            # append a tuple
-            resolver_tuples.append((resolver.get("name"),
-                                    resolver.get("priority") or 1000,
-                                    resolver.get("node")))
-
-        # sort the resolvers by the 2nd entry in the tuple, the priority
-        sorted_resolvers = sorted(resolver_tuples, key=lambda res: res[1])
-        # if the resolver contains a node setting, we only add it if it is on the correct node
-        local_node_uuid = get_app_config_value("PI_NODE_UUID")
-        resolvers = [r[0] for r in sorted_resolvers if not r[2] or r[2] == local_node_uuid]
-        # remove duplicate resolver names but keeping the order
-        seen = set()
-        return [x for x in resolvers if not (x in seen or seen.add(x))]
-
     def _get_resolvers(self, all_resolvers=False) -> list[str]:
         """
         This returns the list of the resolvernames of the user.
@@ -248,7 +221,7 @@ class User:
             return [self.resolver]
 
         resolvers = []
-        for resolver_name in self.get_ordered_resolvers():
+        for resolver_name in get_ordered_resolvers(self.realm):
             # test, if the user is contained in this resolver
             if self._locate_user_in_resolver(resolver_name):
                 break
@@ -652,7 +625,6 @@ class User:
             "uid": self.uid,
             "custom_attributes": self.attributes
         }
-
 
 @log_with(log, hide_kwargs=["password"])
 def create_user(resolvername: str, attributes: dict, password: str = None) -> int or str:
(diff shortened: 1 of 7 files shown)


privacyidea/lib/user.py
```diff
@@ -758,15 +729,23 @@ def get_user_from_param(param: dict, optional_or_required: bool = optional) -> U
 def get_user_list(param: dict = None, user: User = None, include_custom_attributes: bool = False,
                   requested_attributes: list[str] = None) -> list[dict]:
     """
-    This function returns a list of user dictionaries.
+    This function returns a list of user dictionaries. The user dict contains the resolver and custom user attributes,
+    if requested.
+    If no realm is given in the param, the users from all realms are returned.
+    If users are searched in more than one realm, it also contains the realm name.
+    If users are only searched in a specific resolver, it does not contain the realm name.
```
jelinaunger, line 736: To have a consistent response, including always the realm would be nice

privacyidea/lib/user.py
```diff
@@ -793,74 +772,86 @@ def get_user_list(param: dict = None, user: User = None, include_custom_attribut
...
+            if user_resolver:
+                resolvers.append(user_resolver)
+
+        for resolver_name in resolvers:
+            try:
+                log.debug("Check for resolver class: {0!r}".format(resolver_name))
+                resolver = get_resolver_object(resolver_name)
+                # Continue if we couldn't find a resolver with the given name
+                if not resolver:
+                    continue
```
jelinaunger, line 821: A debug log message could be helpful here

privacyidea/lib/user.py
```diff
@@ -793,74 +772,86 @@ def get_user_list(param: dict = None, user: User = None, include_custom_attribut
...
+                log.debug("With this search dictionary: %r", search_dict)
+                user_list = resolver.getUserList(search_dict, requested_attributes)
+                realm_id = get_realm_id(realm)
+                for user_info in user_list:
+                    if len(realm_iteration) > 1:
+                        # Only add the realm information, if we iterate over more than one realm
+                        user_info["realm"] = realm
```
jelinaunger, line 828: We've recently introduced the requested_attributes parameter to control which user info should be returned. We should handle it similar with the realms (see for resolvers below)

privacyidea/lib/user.py
```diff
@@ -793,74 +772,86 @@ def get_user_list(param: dict = None, user: User = None, include_custom_attribut
...
-    for resolver_name in set(resolvers):
-        try:
-            log.debug("Check for resolver class: {0!r}".format(resolver_name))
-            resolver = get_resolver_object(resolver_name)
-            # Continue if we couldn't find a resolver with the given name
-            if not resolver:
+
+    for realm in realm_iteration:
+        if realm:
+            resolvers = get_ordered_resolvers(realm)
```
jelinaunger, line 806: In this case param_resolver and user_resolver are not considered.

privacyidea/lib/user.py
cornelinux: I did this initially. But it is not always possible, to include the realm, since it is not unambiguous.

The function can be called for a resolver, that is not contained in a realm. In this case: the realm would be "none".
The function can be called for a resolver but not with realm information. The resolver could be included in two realms. Which realm should be added then?

These are rare cornercases, but the realm assignment in these corner cases is either not existent or not clear.
If I am not missing anything here, I would keep it this way.

privacyidea/lib/user.py
cornelinux: You are right. 
In my understanding this was also the previous behaviour and for simplicity I kept it this way.

Otherwise again we would end up with strange cases:

What should we do, if a realm is given and a resolver, that is not included in the realm?

List all users from the realm and from the single resolver?
Raise an error?

Maybe I simply explain the behaviour in the docstring? Does this sound all right?

privacyidea/lib/user.py
cornelinux: I did not touch this. But will do.

privacyidea/lib/user.py
cornelinux: I will have do rebase and will do so.

privacyidea/lib/user.py
cornelinux: Done in feff58d00

privacyidea/lib/user.py
jelinaunger: The behavior I would expect is for all filter parameters to be combined using a logical AND. This is also the filter behavior at other endpoints. Hence, if a resolver is not contained in a realm, an empty user list would be returned.

However, you are right, the behavior was the same before. Hence, we can keep it like this for now. Adding an explanation to the docstring would be nice.

privacyidea/lib/user.py
jelinaunger: It could be a list of realms. If the resolver is in no realm the list would be empty. Just as a thought.

privacyidea/lib/user.py
cornelinux: I added a docstring in an upcomint commit in my other PR.

privacyidea/lib/user.py
cornelinux: Oh, this is even more problematic, than when filter for realm and resolver.

The resolverX could be located in two realms. 


realmA
- resolverY (including username:admin)
- resolverX  (including username:admin)

realmB
- resolverX (including username:admin)
- resolverY (including username:admin)

But the resolver could have low priority in realm A. And high priority in Realm B.
Now a user admin is found in the resolverX, but in realmA this very user is masekd by a different higher priority resolver. In Realm B the user is found with highest priorty.

=> admin@realmA  resolves to User(admin, resolverY, realmA)
=> admin@realmB resolves to User(admin, resolverX, realmB)

So what should the function list as realm, when filtered for resolverX?

A)   realm:  ["realmB"]

since, a user admin@realmA will not be found in resolverX or

B) realm: ["realmA", "realmB"]

since... why?

I would also simply add a note to the function, that currently filtering for resolvers can result in a problem anyways.

We should rather solve the issue filter for realm + resolver first.

cornelinux:
We have a new PR https://github.com/privacyidea/privacyidea/pull/5184

PR #5184, opened 2026-04-10
5063/get ordered resolvers

### LINKED ISSUE #5063: Inconsistancy with multiple resolvers in a realm between /validate/check and GET /user/

cornelinux:
Please check the connected issue for explanation what I have done and why.
I created a new PR, since the rebase or merge did not work out.
So I did also not add your comments about the parameters, yet. 

I put all the problems with resolvers in the docstring.


privacyidea/lib/user.py
```diff
@@ -794,77 +781,88 @@ def get_user_list(param: dict = None, user: User = None, include_custom_attribut
...
+                log.debug(f"Check for resolver class: {resolver_name!r}")
+                resolver = get_resolver_object(resolver_name)
+                # Continue if we couldn't find a resolver with the given name
+                if not resolver:
+                    log.info(f"Can not find a resolver with the name '{resolver_name}'")
+                    continue
+                user_list = resolver.getUserList(search_dict, requested_user_store_attributes)
+                realm_id = get_realm_id(realm)
+                for user_info in user_list:
+                    user_info["realm"] = realm
```
jelinaunger, line 832: similar as for the resolver, we need to check if realm is in the requested attributes

nilsbehlen (approved):

PR #5199, opened 2026-04-15
5198 add audit download

### LINKED ISSUE #5198: Add download button for audit table
was left out in #5141

henrik-falk:


privacyidea/static_new/src/app/components/audit/audit.self-service.component.html
```diff
@@ -46,6 +46,18 @@
       [pageSize]="auditService.pageSize()"
       showFirstLastButtons />
   </div>
+  <div class="margin-bottom-16">
```
jelinaunger: users can never download the audit (policy does not exist in scope user)

privacyidea/static_new/src/app/services/audit/audit.service.ts
```diff
@@ -199,4 +203,21 @@ export class AuditService implements AuditServiceInterface {
...
+      responseType: "text"
+    }).subscribe(data => {
+      const blob = new Blob([data], { type: "text/csv" });
+      const url = window.URL.createObjectURL(blob);
+      const a = document.createElement("a");
+      a.href = url;
+      a.download = "audit.csv";
+      a.click();
+      window.URL.revokeObjectURL(url);
+    });
```
jelinaunger, line 233: also handle potential errors here

privacyidea/static_new/src/app/services/audit/audit.service.ts
```diff
@@ -199,4 +204,39 @@ export class AuditService implements AuditServiceInterface {
...
+      next: (data) => {
+        const blob = new Blob([data], { type: "text/csv" });
+        const url = window.URL.createObjectURL(blob);
+        const a = document.createElement("a");
+        a.href = url;
+        a.download = "audit.csv";
+        a.click();
+        window.URL.revokeObjectURL(url);
+      },
+      error: (error) => {
```
frankmer: unused variabe. Use error or replace it with _

frankmer:
It takes a while to download the file. The transfer is verry slow (12.91 MB / 12.91 MB transferred | Finish: 1.40 min).
It would be helpful to receive a notification informing me that it will take a long time. The web UI gets status 200 almost instantly. However, there is no information about how big the response will be. So we can't show a progress bar. Can we guess it? We know how many entries we have. For me, one entry is about 300 bytes. But that can vary significantly from setup to setup.

Please remove the cancel button unless we can actually cancel it.

frankmer:
However, I can confirm that the transfer speed does not depend on the WebUI. It's the same speed in Postman.

privacyidea/static_new/src/app/components/audit/audit.component.ts
```diff
@@ -106,7 +106,8 @@ const columnKeysMap = [
     ClearableInputComponent,
     RouterLink,
```
frankmer: RouterLink is 2x in the imports

privacyidea/static_new/src/app/components/audit/audit.component.html
```diff
@@ -46,6 +46,28 @@
       [pageSize]="auditService.pageSize()"
       showFirstLastButtons />
   </div>
+  <div class="margin-bottom-16">
+    @if (authService.actionAllowed("auditlog_download")) {
+      <button
+        (click)="auditService.downloadCSV()"
+        [disabled]="auditService.isDownloading()"
+        class="action-button-primary button-width-m"
+        i18n
```
frankmer, line 55: angular extracts the whole content for i18n. Better remove it here and add <span i18n>Download</span>

henrik-falk: What do you mean by Better remove it here and add Download?

jelinaunger (approved):

PR #5203, opened 2026-04-16
5105/improve-token-type-config

### LINKED ISSUE #5105: Improve Usability of Token Type Config
Actually,  the config for the questionnaire token and yubikey behaves differently:
* Questionnaire: delete question request is done at save
* yubikey: delete key request is done immediately (no further save required)

-> This should be handled consistently

Adding new questions and yubikey keys also require a separate save. This should also be consistent with the delete behavior.

If we decide to allow intermediate/partial saves, we need to track the changed config independently of the config resource to avoid overwriting pending changes.
It would also be nice to add the discard changes dialog when the user tries to leave the page and there are pending changes.

Additionally, the save button should be sticky.

henrik-falk:


privacyidea/static_new/src/app/components/configuration/token-type-config/token-type-config.component.ts
```diff
@@ -80,21 +86,68 @@ import { ActivatedRoute } from "@angular/router";
...
+  @ViewChild("stickyHeader") stickyHeader!: ElementRef;
+  @ViewChild("stickySentinel") stickySentinel!: ElementRef;
+
+  private observer!: IntersectionObserver;
+
+  formData = linkedSignal<any, Record<string, any>>({
+    source: () => this.systemService.systemConfig(),
+    computation: (config) => ({ ...config })
+  });
+  nextQuestion = linkedSignal<any, number>({
```
jelinaunger: nextQuestionIndex?

jelinaunger (approved):

PR #5205, opened 2026-04-16
5112/token description length indicator

### LINKED ISSUE #5112: token description length indicator
Since the description length of a token is limited to 80 characters an indicator in the input field showing the actual and maximal number of characters such as `16/80` would be nice.
Should be added in the enrollment and on the token details page

henrik-falk:


privacyidea/static_new/src/app/components/token/token-details/token-details.component.html
```diff
@@ -182,9 +182,10 @@ <h3 i18n>{{ " (Offline)" }}</h3>
                         matInput
                         placeholder="Enter description"
                         rows="4"></textarea>
+                      <mat-hint align="end">{{ element.value?.length || 0 }} / 80</mat-hint>
```
jelinaunger: This is not completely related to your PR, but having the max length as constant in the token service might make sense

jelinaunger (approved):

PR #5213, opened 2026-04-16
5106/show api error messages

### LINKED ISSUE #5106: Show API error messages
In the following services we do not catch and / or show the error message:
* machine
* event (enable / disable -> add message to notification)
* periodic task (enable / disable -> add message to notification)
* policies (enable, disable, delete)
* realm

double check for an entire list

### LINKED ISSUE #5202: Show error messages for get requests
For most get requests we are actually not showing the error message if an error occurs. This should be added.
This issue might be split up into multiple sub-issues to avoid a PR with many project-wide changes.

henrik-falk:


privacyidea/static_new/src/app/services/audit/audit.service.ts
```diff
@@ -136,8 +137,20 @@ export interface AuditServiceInterface {
 export class AuditService implements AuditServiceInterface {
   private readonly authService: AuthServiceInterface = inject(AuthService);
   private readonly contentService: ContentServiceInterface = inject(ContentService);
+  private readonly notificationService = inject(NotificationService);
   sort = signal({ active: "serial", direction: "asc" } as Sort);
 
+  constructor() {
+    effect(() => {
+      if (this.auditResource.error()) {
```
frankmer, line 145: To reduce Boilerplate, i would do something like this:

```suggestion
constructor() {
  effect(() => {
    this.notificationService.handleResourceError(this.auditResource, "audit data");
  });
}

// NotificationService:
handleResourceError(resource: HttpResourceRef<any>, subject: string): void {
  if (resource.error()) {
    const err = resource.error() as HttpErrorResponse;
    const message = err.error?.result?.error?.message || err.message;
    this.openSnackBar(`Failed to get ${subject}. ${message}`);
  }
}
```

privacyidea/static_new/src/app/services/event/event.service.ts
```diff
@@ -349,4 +351,10 @@ export class EventService implements EventServiceInterface {
     }
     return conditions;
   });
+
+  constructor() {
+    effect(() => {
+      this.notificationService.handleResourceError(this.allEventsResource, "event handlers");
+    });
```
jelinaunger, line 358: Some resources are missing here (eventHandlerModulesResource, availableEventsResource, moduleConditionsResource)

privacyidea/static_new/src/app/services/notification/notification.service.ts
```diff
@@ -81,4 +83,12 @@ export class NotificationService implements NotificationServiceInterface {
       this.startTimer(snackBarRef);
     }
   }
+
+  handleResourceError(resource: HttpResourceRef<any>, subject: string): void {
```
jelinaunger: We could pass the error directly to this function as the full resource is not required and this would also avoid reading the error twice

privacyidea/static_new/src/app/services/periodic-task/periodic-task.service.ts
```diff
@@ -117,6 +117,12 @@ export class PeriodicTaskService implements PeriodicTaskServiceInterface {
 
   private periodicTaskBaseUrl = environment.proxyUrl + "/periodictask/";
 
+  constructor() {
+    effect(() => {
+      this.notificationService.handleResourceError(this.periodicTasksResource, "periodic tasks");
+    });
```
jelinaunger, line 123: add periodicTaskModuleResource

privacyidea/static_new/src/app/services/policies/policies.service.ts
```diff
@@ -236,6 +236,12 @@ export class PolicyService implements PolicyServiceInterface {
   readonly policyBaseUrl = environment.proxyUrl + "/policy/";
   private readonly http: HttpClient = inject(HttpClient);
   private readonly authService: AuthServiceInterface = inject(AuthService);
+
+  constructor() {
+    effect(() => {
+      this.notificationService.handleResourceError(this.allPoliciesResource, "policies");
+    });
```
jelinaunger, line 243: Here are also some resources missing (double-check for the full PR)

privacyidea/static_new/src/app/services/privacyidea-server/privacyidea-server.service.spec.ts
```diff
@@ -39,7 +39,8 @@ describe("PrivacyideaServerService", () => {
       getHeaders: jest.fn().mockReturnValue({})
     };
     const notificationServiceMock = {
-      openSnackBar: jest.fn()
+      openSnackBar: jest.fn(),
+      handleResourceError: jest.fn()
```
jelinaunger: Replacing the local mock with the global mock would be better instead of adding the new function to all local mocks (can stay like this for now)

jelinaunger (approved):

PR #5215, opened 2026-04-17
5133/ignore rollout state policy for passthru and passnotoken

### LINKED ISSUE #5133: Exclude tokens in 'verify' rollout state from passthru authentication
> We are gradually migrating more and more users to TOTP. In doing so, we are using the two-step rollout process, particularly for the initial token issuance.
> As a workaround, we have a cron job that checks when a token receives the “Verify” status and deletes it if it has not yet received the “Enrollment” status after one hour. We check the audit log for this.
> Unfortunately, we have noticed an increasing number of cases in recent days where this does not work. Sometimes there is no audit entry when the token received the status, and other times there are two or more entries (in these cases, users are likely too impatient and click the button multiple times).
> Our request is that tokens with the “verify” status not be taken into account during passthrue. 

 _Originally posted by @katgirl in [#5070](https://github.com/privacyidea/privacyidea/issues/5070#issuecomment-4140626636)_

jona-samuel:


privacyidea/lib/policydecorators.py
```diff
@@ -289,7 +301,19 @@ def auth_user_passthru(wrapped_function, user_object, passw, options=None):
         pass_thru = Match.user(g, scope=SCOPE.AUTH, action=PolicyAction.PASSTHRU,
                                user_object=user_object).policies(write_to_audit_log=False)
         # We only go to passthru, if the user has no tokens!
-        if pass_thru and get_tokens(user=user_object, count=True) == 0:
+        # If PASSTHRU_IGNORE_ROLLOUT_STATE is set, tokens with a rollout state
```
frankmer, line 317: It is almost identical to lines 224–236.
DRY:
I would create a private helper method like def _get_relevant_token_count(g, user_object, action):

This would also make it easier to read.

frankmer (approved):

privacyidea/lib/policies/actions.py
```diff
@@ -64,8 +64,10 @@ class PolicyAction:
     OTPPINMINLEN = 'otp_pin_minlength'
     OTPPINCONTENTS = 'otp_pin_contents'
     PASSNOTOKEN = "passOnNoToken"
+    PASSNOTOKEN_IGNORE_ROLLOUT_STATE = "passnotoken_ignore_rollout_state"
```
nilsbehlen, line 67: the name is of the policy action is "passOnNoToken"

privacyidea/lib/policy.py
```diff
@@ -2605,6 +2605,13 @@ def get_static_policy_definitions(scope=None):
                           'given RADIUS config,'
                           ' if the user has no tokens assigned.')
             },
+            PolicyAction.PASSTHRU_IGNORE_ROLLOUT_STATE: {
+                'type': 'str',
+                'multiple': True,
+                'desc': _(
+                    'Ignore tokens in the given rollout state. This will only work if the passthru policy is active.'),
+                'value': ['clientwait', 'pending', 'verify', 'enrolled', 'broken', 'failed', 'denied']
```
nilsbehlen, line 2613: this should be a list from the enum/object RolloutState, not the values added by hand

privacyidea/lib/policy.py
```diff
@@ -2617,6 +2624,13 @@ def get_static_policy_definitions(scope=None):
                 'desc': _('If the user has no token, the authentication '
                           'request for this user will always be true.')
             },
+            PolicyAction.PASSNOTOKEN_IGNORE_ROLLOUT_STATE: {
+                'type': 'str',
+                'multiple': True,
+                'desc': _(
+                    'Ignore tokens in the given rollout state. This will only work if passOnNoToken policy is active.'),
+                'value': ['clientwait', 'pending', 'verify', 'enrolled', 'broken', 'failed', 'denied']
```
nilsbehlen, line 2632: same as above

privacyidea/lib/policies/actions.py
jona-samuel: #5220

privacyidea/lib/policy.py
jona-samuel: #5220

privacyidea/lib/policy.py
jona-samuel: #5220

PR #5219, opened 2026-04-20
feat: reorganize test files and extract common utilities for improved…

nilsbehlen:
… maintainability

* Split into more files. Later, we want to run 1 file per worker, so if there are still 7k line files, it would bottleneck the parallel work.
* Add a shim for the test setup to rewrite the database url to a numbered one, so each worker gets its own
* Run the tests that we have in parallel


tests/test_lib_token_export.py
```diff
@@ -0,0 +1,131 @@
...
+The lib.token depends on the DB model and lib.user and
+all lib.tokenclasses
+
+This tests the token functions on an interface level
+
+We start with simple database functions:
+
+getTokens4UserOrSerial
+gettokensoftype
+getToken....
```
jelinaunger: This description is not fully true anymore

jelinaunger: Same for other test_lib_token_* files

tests/test_api_lib_policy_actions.py
```diff
@@ -0,0 +1,638 @@
+"""
+This test file tests the api.lib.policy.py
```
jelinaunger: This file does not exist (and maybe not required at all)

jelinaunger: Does we need license headers in the test files?

jelinaunger (approved):

PR #5227, opened 2026-04-20
adding event condition

### LINKED ISSUE #5023: Userinfo Condition for Event Handler + Multiple Tokeninfo

I am trying to create event handlers, which are based on the userinfo attributes.

For example I can configure policies that use: "Additional conditions" where I can configure a userinfo, as a condition for the policy to trigger.

Currently there is no such option for the event handlers. Right now it is only possible to configure event handlers based on tokeninfo.

It would be good to have the option to select userinfo and set according conditions, just like the policy conditions.

Also it is currently not possible to select multiple tokeninfo attributes, having a table like the "Additional conditions" for policies, would solve both the issues by allowing the user to select userinfo, as well as allowing the user to select multiple tokeninfo conditions.

Right now you can do this by creating multiple resolvers that filter the users based on the where condition (in sql resolver) or the search filter when using ldap resolver.

Then you could filter based on the resolver in the event handler. This solution is possible but not optimal for the user experience and can lead to other problems, because the users are now in separate resolvers.

**Additional context**

Event Handler (that needs changing):
Image

Policies (possible solution):

Image

jona-samuel:


privacyidea/lib/eventhandler/base.py
```diff
@@ -750,6 +760,19 @@ def check_condition(self, options):
             if user.realm not in conditions.get(CONDITION.REALM).split(","):
                 return False
 
+        if CONDITION.USER_INFO in conditions:
+            cond = conditions.get(CONDITION.USER_INFO)
+            # replace {now} in condition
+            cond, td = parse_time_offset_from_now(cond)
```
jelinaunger, line 766: use full words for variable names

privacyidea/lib/eventhandler/base.py
```diff
@@ -750,6 +760,19 @@ def check_condition(self, options):
...
+            s_now = (datetime.datetime.now(tzlocal()) + td).strftime(
+                DATE_FORMAT)
+            cond = cond.format(now=s_now)
+            if not compare_generic(cond,
+                                   lambda key: user.get_specific_info([key]).get(key),
+                                   "Misconfiguration in your userinfo "
+                                   "condition: {0!s}"):
```
jelinaunger, line 773: Check your line length settings (should be 120 char). At least the warning could be in one line

privacyidea/lib/eventhandler/base.py
```diff
@@ -750,6 +760,19 @@ def check_condition(self, options):
             if user.realm not in conditions.get(CONDITION.REALM).split(","):
                 return False
 
+        if CONDITION.USER_INFO in conditions:
+            cond = conditions.get(CONDITION.USER_INFO)
+            # replace {now} in condition
+            cond, td = parse_time_offset_from_now(cond)
+            s_now = (datetime.datetime.now(tzlocal()) + td).strftime(
+                DATE_FORMAT)
+            cond = cond.format(now=s_now)
```
jelinaunger, line 769: Using .replace() is better as format could raise an error in case there are unexpected tags

tests/test_lib_events.py
```diff
@@ -1272,6 +1255,62 @@ def test_29_check_condition_sets_context(self):
...
+                         user=user)
+        self.assertEqual(tok.type, "pw")
+        uhandler = BaseEventHandler()
+        req_data = {'user': "cornelius@realm1", "pass": "secret"}
+        resp_data = """{"result": {"value": true}}"""
+        options = self.setup_request(req_data=req_data, all_data=req_data, user=user, resp_data=resp_data)
+
+        # Verify the user info is what we expect from the passwd resolver
+        info = user.info
+        self.assertEqual(info.get("givenname"), "Cornelius")
```
jelinaunger, line 1275: argument order is the other way around (expected, actual)

privacyidea/lib/eventhandler/base.py
```diff
@@ -210,6 +211,15 @@ def conditions(self):
                     "value": [{"name": r} for r in resolvers],
                     "group": GROUP.USER
                 },
+            CONDITION.USER_INFO:
+                {
+                    "type": "str",
+                    "desc": _("This condition can check any arbitrary userinfo "
+                              "field. You need to enter something like "
+                              "'<fieldname> == <fieldvalue>', '<fieldname> > "
+                              "<fieldvalue>' or '<fieldname> < <fieldvalue>'."),
```
jelinaunger, line 220: Also mention here that the tag {now} etc. can be used here. And update the documentation for this new condition.

jelinaunger (approved):

PR #5231, opened 2026-04-20
fix yubico token validation

### LINKED ISSUE #5229: Signature verification for Yubico Validation Protocol v2 Responses fails
### Top-level intent

Using Yubico-Tokens results in token not being accepted. The log file shows the following entry:

[ERROR][privacyidea.lib.tokens.yubicotoken:206] The hash of the return from the yubico authentication server (http://[...]/wsapi/2.0/verify) does not match the data!

### Steps to reproduce

1. Import/create a yubico token.
2. Validate token against yubico servers (https://api.yubico.com/wsapi/2.0/verify)

### Expected outcome

The token should be successfully validated.

### Actual outcome

The token is not validated successfully and fails with the above error message in the log.

### Configuration

* **privacyIDEA version**: 3.13
* **Python version**: 3.13.5
* **Operating system**: Debian GNU/Linux 13 (trixie)
* **Webserver**: Apache/2.4.66 (Debian)

### Log file

[ERROR][privacyidea.lib.tokens.yubicotoken:206] The hash of the return from the yubico authentication server (http://[...]/wsapi/2.0/verify) does not match the data!

According to https://developers.yubico.com/OTP/Specifications/OTP_validation_protocol.html the keys used in the requests are:

 h, id, nonce, otp, sl, timeout and timestamp

These are included in _YUBICO_PROTOCOL_KEYS (except the h of course) in yubikeytoken.py, but some of the keys used in *responses* are missing:

sessioncounter, sessionuse, status, t.

These must be included, because function yubico_api_signature is used to calculate signatures for requests *and* to check the signature in yubico_check_api_signature.

Also please correct the links to the specification of the validation protocol in yubicotoken.py and yubikeytoken.py. The specification of the validation protocol can now be found at https://developers.yubico.com/OTP/Specifications/OTP_validation_protocol.html. The Yubikey Manual seems now to reside under https://downloads.yubico.com/support/yubico_YubiKey-Technical-Manual-v3.3_2014.pdf.

Maybe add tests using the test vectors from https://developers.yubico.com/OTP/Specifications/Test_vectors.html to make sure validation also succeeds in future releases.

nilsbehlen:
- fixed parameters used for validation response
- added more tests from real captures and official vectors
- added the yubico api url as default value for the configuration so you dont have to type it

### DIFF
--- a/tests/test_lib_tokens_yubico.py
+++ b/tests/test_lib_tokens_yubico.py
@@ -2,7 +2,9 @@
 This test file tests the lib.tokens.yubicotoken
 This depends on lib.tokenclass
 """
-from privacyidea.lib.token import init_token, get_tokens, import_tokens
+from unittest.mock import patch
+
+from privacyidea.lib.token import init_token, remove_token, get_tokens, import_tokens
 from .base import MyTestCase
 from privacyidea.lib.tokens.yubicotoken import (YubicoTokenClass, YUBICO_URL)
 from privacyidea.models import Token
@@ -11,7 +13,6 @@ from privacyidea.lib.config import set_privacyidea_config
 
 
 class YubicoTokenTestCase(MyTestCase):
-
     otppin = "topsecret"
     serial1 = "ser1"
     params1 = {"yubico.tokenid": "vvbgidlghkhgfbvetefnbrfibfctu"}
@@ -80,7 +81,6 @@ status=REPLAYED_OTP"""
         self.assertTrue(otpcount == -2, otpcount)
         set_privacyidea_config("yubico.do_post", False)
 
-
     @responses.activate
     def test_05_check_otp_fail(self):
         responses.add(responses.POST, YUBICO_URL,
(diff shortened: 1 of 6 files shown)


tests/test_lib_tokens_yubico.py
```diff
@@ -93,6 +93,105 @@ def test_05_check_otp_fail(self):
...
+
+    @responses.activate
+    def test_05c_check_otp_yubicloud_real_replayed(self):
+        """Real YubiCloud ``status=REPLAYED_OTP`` — signature still must verify."""
+        set_privacyidea_config("yubico.id", self.YUBICLOUD_API_ID)
+        set_privacyidea_config("yubico.secret", self.YUBICLOUD_API_KEY)
+        set_privacyidea_config("yubico.do_post", False)
+
+        responses.add(responses.GET, YUBICO_URL,
+                      body=self.YUBICLOUD_REPLAY_BODY)
```
jelinaunger, line 168: format: could be one line ( also check format below)

jelinaunger (approved):

PR #5247, opened 2026-04-23
Modified `tsconfig.json` to include a path mapping for `src/*` to facilitate absolute imports throughout the project.

frankmer:
- Updated import statements in various  files to use absolute paths instead of relative paths.


privacyidea/static_new/jest.config.ts
```diff
@@ -100,9 +100,19 @@ const config: Config = {
 
   // A map from regular expressions to module names or to arrays of module names that allow to stub out resources with a single module
   moduleNameMapper: {
-    "^src/(.*)$": "<rootDir>/src/$1",
-    "^@styles/(.*)$": "<rootDir>/src/app/styles/$1",
-    "^@components/(.*)$": "<rootDir>/src/app/components/$1"
+    // 1. Specific aliases (order here is usually irrelevant since they don't overlap)
```
henrik-falk, line 103: clean up ai comments please.

henrik-falk (approved):

privacyidea/static_new/jest.config.ts
frankmer: I left these comments so that we know the catch-all alias has to come after the specific aliases, in case we want to change it one day.

frankmer: Do i still have to remove this comment?

PR #5248, opened 2026-04-23
New webui ci

nilsbehlen:
## Frontend CI coverage — new workflows                                                     
                                                                       
  - **`webui-test.yml`** — now uploads Jest coverage to Codecov with `flags: frontend`.       
  Enabled `lcov` reporter in `jest.config.ts`.                         
  - **`webui-build.yml`** *(new)* — runs `npm run build` (= `ng build` production) so type    
  errors and template breakage fail CI.                                
  - **`webui-lint.yml`** *(new)* — runs `npm run lint` (ESLint via `ng lint`). Step-level
  `continue-on-error` → reporting-only until the 2,343-error backlog is worked down.                                                                    
                                                                       
  ## Frontend tooling setup                                                                   
                                                                       
  - **ESLint** added via `ng add @angular-eslint/schematics`: `eslint.config.js`, `lint`      
  script, `angular.json` lint target.
  - **Prettier** cleanup: removed deprecated `pluginSearchDirs` and `jsxBracketSameLine` from 
  `.prettierrc`; added `format:check` / `format:write` scripts; ran one-shot `prettier        
  --write` reformatting 368 files.
                                                                                              
  ## Security scanners — made non-blocking                             

  - **`osv-scanner.yml`** rewritten — dropped the reusable workflow, installs the osv-scanner 
  binary directly (pinned `v2.3.5`), runs with step-level `continue-on-error`, still uploads
  SARIF to the Security tab via `github/codeql-action/upload-sarif`. Added                    
  `--config=osv-scanner.toml`.                                         
  - **`privacyidea/static_new/osv-scanner.toml`** *(new)* — suppresses `GHSA-w5hq-g745-h8pq` 
  with full rationale (we only use `v4()`; transitive uuid@8.3.2 is dev-only) and advisory    
  `ignoreUntil: 2026-10-23`.
  - **`uuid` bump** attempted (11→14), then reverted because uuid@14 is ESM-only and breaks   
  Jest. Documented as out-of-scope in the toml.                                               
   
  ## Backend coverage cleanup                                                                 
                                                                                              
  - **`codecov.yml`** *(new)* — splits backend vs frontend into separate flags, components, 
  and project/patch statuses with `carryforward: true`. Stops frontend coverage from          
  "diluting" the project number.                                       
  - **`unit-tests-mariadb.yml`** — added `flags: backend` to the Codecov upload. Remains the  
  single coverage source.                                              
  - **`unit-tests-postgres.yml`** — dropped `--cov` and the Codecov upload entirely. Same test
   suite ran twice with coverage was pure waste; postgres jobs now ~10–30% faster.            
   
  ## What's explicitly *not* addressed on this branch                                         
                                                                       
  - uuid actual upgrade (deferred; suppressed in scanner instead).
  - ESLint backlog cleanup (2,343 errors remain; workflow is reporting-only).
  - `pip-audit.yml` lying in its header comment about SARIF uploads (discussed, not fixed).


nilsbehlen:
i dont know if we can get the split coverage (front/backend) to work in this pr, so the delta that causes check to fail could be ignored this one time. 
osv-scanner config seems to be required in the main branch, but its in this branch so that does not work either right here

privacyidea/static_new/src/proxy.conf.json
```diff
@@ -1,6 +1,6 @@
 {
   "/proxy": {
-    "target": "http://localhost:5000",
+    "target": "https://pi.fritz.box:5000",
```
henrik-falk: i think these changes are meant to be local only?

henrik-falk:
for prettier there is one option that i would like to discuss. In html files the settings in this pr will put a new line for some tags > symbol. I think its unnecessary and does not look clean. Tags like strong will make texts longer than needed and less readable (e.g. login) and links get chopped up (e.g. navigation) like i would never write myself.

henrik-falk (approved):

PR #5262, opened 2026-04-28
move internal user attributes to separate table

### LINKED ISSUE #5095: The privacyIDEA server misuses custom user attributes
Custom user attributes were introduced, so that users (administrators) could willingly add additional attributes to users in scenarios where those attributes are not contained and managed in the user store or where a privacyIDEA admin has no access to the user store.
https://privacyidea.readthedocs.io/en/stable/webui/user_attributes.html#user-attributes

The custom user attributes are ment for rather static information that are bound to the user and set by a human being. Your shoe size. A group membership. The IP address that should be assigned to this user on the VPN...

In some version the privacyIDEA server started to write its own, internal information to the custom user attributes like FIDO information or the last used token type.

I think this is a bad design and can lead to problems. Written attributes by the server and willingly added attributes by administrators can interfere and lead to unexpected effects.

We need to revert or change this behaviour.

The privacyIDEA server needs its own new table for such user specific information.
These information are volatile and can change automatically by the privacyIDEA server (like the last used tokentype)

# Recommendation

Add a new table `InternalUserAttributes` that is used by the privacyIDEA server internal methods.

When a realm is deleted, all these attributes can be happily deleted automatically, since these are only attributes that the privacyIDEA server uses internally to handle this user.

nilsbehlen:
* create table internaluserattribute
* update usages of fido2_user_id and last used token type per application
* migrate entries from customuserattributes to the new table
* add migration test
* add cli command to clean up orphaned internal attributes
* new api endpoint to query the attributes (only triggers on the details page then), with @admin_required


privacyidea/lib/user.py
```diff
@@ -383,6 +383,57 @@ def delete_attribute(self, attribute_key: str = None) -> int:
...
+        Delete an internal attribute. If ``key`` is None, all internal
+        attributes for this user are deleted.
+        """
+        stmt = delete(InternalUserAttribute).filter_by(user_id=self.uid, resolver=self.resolver,
+                                                       realm_id=self.realm_id)
+        if key:
+            stmt = stmt.filter_by(Key=key)
+        result = db.session.execute(stmt)
+        db.session.commit()
+        return result.rowcount
```
jelinaunger, line 472: some kind of clean up would be nice to allow removing internal custom attributes for users that do not exist anymore

tests/test_lib_user.py
```diff
@@ -876,6 +876,41 @@ def test_50_user_attributes(self):
...
+                       "fileName": PWFILE})
+        set_realm(self.realm1, [{'name': self.resolvername1}])
+        user = User(login="root", realm=self.realm1)
+
+        # Set + read JSON-typed values (dict and scalar string)
+        user.set_internal_attribute("last_used_token", {"app-a": "hotp"})
+        user.set_internal_attribute("fido2_user_id", "abc123")
+        attrs = user.internal_attributes
+        self.assertEqual(attrs.get("last_used_token"), {"app-a": "hotp"})
+        self.assertEqual(attrs.get("fido2_user_id"), "abc123")
```
jelinaunger: order should be vice versa: expected, actual

jelinaunger (commented):
Internal custom attributes are not shown in the UI (old and new)

tests/testdata/testuser-api.sqlite
jelinaunger, line 1: Double check if these changes were intentional also for the other testuser db

nilsbehlen:
added to bottom of the user details
image

privacyidea/migrations/versions/7d4e9b2c1a3f_internaluserattribute.py
```diff
@@ -0,0 +1,253 @@
...
+            )
+        )
+
+    # 3. Remove migrated rows from the old table. The DELETE is naturally
+    # idempotent; running it after a clean run is a no-op.
+    conn.execute(old.delete().where(old.c.Key == FIDO2_USER_ID_KEY))
+    conn.execute(old.delete().where(
+        old.c.Key.like(f'{LAST_USED_TOKEN_PREFIX}%'),
+        old.c.Type == INTERNAL_TYPE_MARKER,
+    ))
```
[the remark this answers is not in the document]
nilsbehlen: have to live with this

jelinaunger (approved):

PR #5265, opened 2026-04-29
add serial and token type to audit log for passkey authentication

### LINKED ISSUE #5264: passkey serial not in audit for /auth
for /auth requests (e.g. webui login), using a passkey will not log the serial to the audit

nilsbehlen:


privacyidea/api/auth.py
```diff
@@ -347,6 +347,8 @@ def get_auth_token():
             "user": user.login,
             "realm": user.realm,
             "resolver": user.resolver,
+            "serial": token.get_serial(),
+            "token_type": token.get_type(),
```
jelinaunger, line 351: Should we also add the token type for other tokens?

jelinaunger (approved):

PR #5270, opened 2026-04-30
metrics: recording and dashboard panels for some metrics

### LINKED ISSUE #5266: More dashboard data: cert expiry and ldap time
it would be good to see on the dashboard for how long the certs (pi and ldap) are valid. it would also be good to see the ldap time, to see if some resolver is slow.

nilsbehlen:
## Overview                                                

  Adds a generic in-process metrics module (`privacyidea/lib/metrics.py`) backed by a new                                                                
  pre-aggregated `metric_aggregate` table, and uses it to power two new dashboard panels —
  **Resolver Timing** and **Notification Delivery** — alongside the existing **Certificates**                                                            
  panel. Naming follows Prometheus conventions (`observe()` for histograms, `inc()` for                                                                  
  counters, prom-style `_total` / `_seconds` suffixes, cumulative `_bucket{le="…"}` columns).                                                            
                                                                                                                                                         
  Beyond the metrics work, the dashboard gains a top news banner sourced from the existing                                                               
  RSS feed (now backed by an in-memory TTL cache so dashboard + news page share one fetch),                                                              
  the news page itself is rebuilt with a panel-per-item layout, and admins get a one-click                                                               
  **Delete Old Metrics** trigger on the system config page so the new table can be pruned                                                                
  on demand without scheduling the periodic task.                                                                                                        
                                                                                                                                                         
  ## What's new                                                                                                                                          
                                                                                                                                                         
  ### Metrics module (`lib/metrics.py`)                                
  - `observe(name, value, labels=None)` — histogram: count / sum / max plus 9 cumulative
    prom-style buckets (50 ms … 5 s, plus implicit `+inf`).                                                                                              
  - `inc(name, labels=None, by=1)` — counter.                                                                                                            
  - `get_metrics(name=None, since_seconds=3600)` — aggregates across all privacyIDEA nodes                                                               
    and 5-minute windows. p50/p95 are bucket-bound approximations; avg and max are exact.                                                                
  - `cleanup_old_metrics(older_than_seconds=86400)` — periodic-task target.                                                                              
  - `track_resolver_op(op_name)` — decorator applied to public `UserIdResolver` methods of                                                               
    every resolver type (LDAP / SQL / HTTP / EntraID / Keycloak / Passwd).                                                                               
                                                                                                                                                         
  Storage: one row per `(metric_name, labels_key, node, window_start)` in 5-minute windows.                                                              
  Writes commit explicitly so read endpoints (which never call `db.session.commit()`) don't                                                              
  lose metric rows at session teardown. Concurrent-worker `INSERT` races are tolerated via                                                               
  `SELECT`-after-rollback.                                                                                                                               
  - Feature switch for metrics collection can disable it completely.                                                                                     
                                                                                                                                                         
  ### Resolver Timing dashboard panel                                  
  `GET /system/health/resolver_timing` aggregates timing across resolvers. The panel                                                                     
  sorts worst-p95 first, links each row to the corresponding resolver detail page, and                                                                   
  color-codes avg / p95 / max columns (green <100 ms / yellow <500 ms / red ≥500 ms). The                                                                
  p95 column is suppressed below 20 samples and exposes the bucket boundaries via tooltip.                                                               
                                                                                                                                                         
  ### Notification Delivery dashboard panel                                                                                                              
  `GET /system/health/notification_delivery` folds three new counters by their primary                                                                   
  label dimension:                                                                                                                                       
  
  - **Push** — `push_delivery_total{gateway, result}` from `FirebaseProvider.submit_message`,                                                            
    paired with `push_delivery_duration_seconds`. Labelled by gateway identifier (matching
    the SMS convention) so multiple Firebase configs surface as separate rows rather than                                                                
    collapsing under a single hardcoded `"firebase"` provider label.                                                                                     
  - **SMS** — `sms_send_total{gateway, result}` instrumented at the chokepoint                                                                           
    `send_sms_identifier` (covers Http / Sipgate / Smtp / Smpp / Script in one place),                                                                   
    paired with `sms_send_duration_seconds`. Push goes through `create_sms_instance`                                                                     
    directly from `pushtoken.py`, not via `send_sms_identifier`, so there is no                                                                          
    double-counting.                                                                                                                                     
  - **Email** — `email_send_total{identifier, result}` from `send_email_identifier` and
    `send_email_data` in `lib/smtpserver.py`, paired with `email_send_duration_seconds`.                                                                 
                                                                                                                                                         
  The panel renders three side-by-side sub-tables, color-codes the failed cell green                                                                     
  <1% / yellow <5% / red ≥5%, and links each target name to its configuration page —                                                                     
  SMTP identifier → SMTP edit, gateway name → SMS gateway edit (resolved via a one-shot                                                                  
  `getSMSGateways` call cached on the controller scope; non-resolvable names render as                                                                   
  plain text).                                                                                                                                           
                                                                                                                                                         
  > **Caveat:** when the SMTP job queue is enabled, the email metric records dispatch                                                                    
  > success rather than final delivery — the synchronous return value is `True` once the
  > job has been queued. Documented in `doc/webui/index.rst`.                                                                                            
                                                                                                                                                         
  ### Periodic task — `MetricsCleanupTask`
  Drops `metric_aggregate` rows older than `older_than_hours` (default 24). Recommended                                                                  
  cadence: daily. **Operators must wire this up under *Config → Tasks*** after upgrading
  or the table grows unbounded. A bullet has been added to `READ_BEFORE_UPDATE.md`.                                                                      
                                                                                                                                                         
  ### Manual metrics cleanup trigger                                                                                                                     
  `POST /system/metricscleanup` runs the same `cleanup_old_metrics` logic on demand,                                                                     
  taking `older_than_hours` (default 24, clamped to ≥1 to protect the in-progress                                                                        
  window). The system-config page (next to *Delete User Cache*) gains a **Delete Old                                                                     
  Metrics** button + numeric input. The dashboard carries a small footer pointing                                                                        
  admins at the periodic-task UI for scheduled cleanup or the system-config page for                                                                     
  one-off runs, since the table is not auto-cleaned out of the box.                                                                                      
                                                                                                                                                         
  ### Dashboard reorganization                                                                                                                           
  Four semantic rows: news banner / operational (unchanged) / health monitoring                                                                          
  (Certificates + Resolver Timing) / service delivery (Notification Delivery) /                                                                          
  configuration reference. Several panel headers gained shortcut links to compensate for
  the deeply-nested menu structure.                                                                                                                      
                                                                                                                                                         
  ### Dashboard news banner
  A panel-styled banner at the top of the dashboard surfaces the most recent RSS item                                                                    
  (first item of the first source — same ordering as the news page), with a bullhorn                                                                     
  glyph, the title, and a formatted `pub_date`. The "News" heading itself links through                                                                  
  to the existing `info.rss` page. The banner is hidden when no items are available.                                                                     
                                                                                                                                                         
  ### News page redesign (`info.rss.html`)                                                                                                               
  Rewritten layout: source titles render as `<h3>`, each item is a `panel panel-default`                                                                 
  with the headline + a small muted formatted date in the heading and the summary HTML                                                                   
  in the body. The previous markup had three issues now fixed: per-item `<uib-accordion>`                                                                
  wrappers (every visit required N clicks just to read the news), `<span ng-repeat>`                                                                     
  around block-level elements (invalid HTML), and a leftover `<div ui-view class="slide                                                                  
  panel">` outer div. The "Read more in new window" link is dropped — the summary content                                                                
  already includes an "originally published at" link. RFC-822 `pub_date` strings are                                                                     
  parsed to `Date` in the controller so Angular's `date:'medium'` filter renders human                                                                   
  text instead of the raw `Wed, 22 Apr 2026 07:02:18 +0000`.                                                                                             
                                                                                                                                                         
  ### RSS in-memory TTL cache (`lib/info/rss.py`)                                                                                                        
  `get_news()` now memoises the parsed feed response for 15 minutes, keyed by                                                                            
  `(sorted feed items, channel, days)`. Drops the dead `RSS_NEWS = {}` global. Adds                                                                      
  `invalidate_news_cache()` for tests and future manual refresh. The dashboard banner                                                                    
  and the news page now share one external fetch; admins repeatedly opening either                                                                       
  trigger no extra HTTP traffic to the upstream feed servers within the TTL window.                                                                      
  Caveat: per-worker, lost on restart, no cross-node sharing — see `notes/redis.md`                                                                      
  for the planned Redis-backed alternative once that landing PR ships.                                                                                   
                                                                                                                                                         
  ## Bug fixes shipped along the way                                                                                                                     
                                                                                                                                                         
  - **`paged_search` LDAP timing** — `paged_search(generator=True)` returns a generator;
    the previous `log.debug(... elapsed_time)` on `LDAPIdResolver.py:878` fired before the                                                               
    generator was consumed, so it captured bind time only, not the actual search. Moved                                                                  
    to after the for-loop. Affects every multi-page LDAP search (user list, group lookups).                                                              
  - **Stale `getLdapTiming()` reference** — the dashboard's `piReload` handler called a                                                                  
    function that no longer existed (left over from the LDAP→Resolver rename), so manual                                                                 
    reloads silently no-op'd the timing panel.                         
  - **Missing `$timeout` injection** — `dashboardController` used `$timeout` in the                                                                      
    `piReload` handler without declaring it in the DI list, throwing   
    `Unknown provider: $timeoutProvider` on every reload. Added to the injection list.                                                                   
  - **Periodic-task form template missing `int` input** — `config.periodictasks.details.html`                                                            
    only had `ng-if` branches for `str` / `text` / `bool`, so any task option declaring                                                                  
    `"type": "int"` rendered with no editable input. Added an `<input type="number">`                                                                    
    branch, plus a `Number()` coercion in the controller's load path so existing                                                                         
    string-stored values populate correctly when editing.              
                                                                                                                                                         
  ## Tests                                                             
                                                                                                                                                         
  - `tests/test_lib_metrics.py` — 23 tests covering label round-trips, window truncation,
    `_percentile_from_buckets`, `inc` / `observe` round-trips, bucket assignment, the
    `+inf` tail, cross-node and cross-window aggregation, `cleanup_old_metrics`,                                                                         
    `_get_or_create_row` race recovery, and the failure-safety guarantee that `inc` /                                                                    
    `observe` swallow DB exceptions instead of breaking the calling request.                                                                             
  - `tests/test_lib_health.py` — 22 tests for the certificate-status helpers (classify                                                                   
    thresholds, cert-info derivation, cache hit / refresh / invalidate, server-cert
    not-configured / error / ok branches, LDAP endpoint dispatch and probe failures,                                                                     
    resolver-list filtering, multi-URI resolver entries, and the resolver-hook
    invalidation contract).                                                                                                                              
  - `tests/test_api_system.py` — covers the three `/system/health/*` endpoints plus four
    new tests for `POST /system/metricscleanup` (default 24h, custom hours, invalid input                                                                
    fallback, sub-1h clamp).                                                                                                                             
  - `tests/test_lib_info.py` — adds `RSSCacheTestCase` covering cache hit, key separation                                                                
    (different feed dict / channel / days bypass cache), manual invalidation, and                                                                        
    TTL-based refetch.                                                                                                                                   
                                                                                                                                                         
  ## Docs                                                                                                                                                
                                                                                                                                                         
  - `doc/webui/index.rst` — three new subsections under *Dashboard*: *Resolver timing*,
    *Notification delivery* (incl. SMTP queue caveat), *Storage and cleanup*.
  - `doc/modules/api/system.rst` — no change needed; `autoflask` picks up the new                                                                        
    endpoints from their docstrings.                                                                                                                     
  - `READ_BEFORE_UPDATE.md` — operator note for the *Update from 3.13 to 3.14* section.                                                                  
  - `notes/redis.md` — appends a "Future opt-in caches to migrate to Redis" section                                                                      
    listing the in-memory TTL caches (RSS, certificate health probes, metric                                                                             
    aggregations) that should gain a Redis backend once the broader Redis work lands.                                                                    
                                                                                                                                                         
  ## Operator action after upgrade                                                                                                                       
                                                                                                                                                         
  1. Run the schema migration (`pi-manage db upgrade`) to create `metric_aggregate`.
  2. *Config → Tasks* → schedule **MetricsCleanup**, daily cadence recommended. (If you                                                                  
     prefer not to schedule it, the new **Delete Old Metrics** button on the system                                                                      
     config page lets you prune on demand.)                                                                                                              
  3. Hard-reload the browser to pick up the JS / template changes.


nilsbehlen:
image

privacyidea/api/system.py
```diff
@@ -487,6 +489,152 @@ def list_nodes():
     return send_result(nodes)
 
 
+@system_blueprint.route("/health/certificates", methods=['GET'])
+@admin_required
+@log_with(log)
+def get_health_certificates():
+    """
+    Return certificate expiry information for all configured LDAP resolvers
```
jelinaunger: Keycloak resolvers can also contain certificates

privacyidea/lib/metrics.py
```diff
@@ -0,0 +1,334 @@
...
+    """Aggregate stored metric rows across nodes and windows.
+
+    Returns a list of dicts, one per ``(metric_name, labels)`` group::
+
+        {"metric": "ldap_op_duration_seconds",
+         "labels": {"resolver": "openldap", "op": "bind"},
+         "count": 120, "avg": 0.024, "p50": 0.025, "p95": 0.1,
+         "max": 0.34, "since_seconds": 3600}
+    """
+    cutoff = datetime.datetime.utcnow() - datetime.timedelta(seconds=since_seconds)
```
jelinaunger: utcnow is deprecated as it has no timezone information

privacyidea/api/system.py
jelinaunger: Also EntraID can have one for the client authorization

privacyidea/lib/health.py
```diff
@@ -0,0 +1,346 @@
...
+    entry = {
+        "source": "keycloak-resolver",
+        "name": resolver_name,
+        "host": f"{host}:{port}",
+        "tls_mode": "https",
+    }
+    try:
+        # A Keycloak endpoint is a plain TLS server, so the generic TLS-wrap
+        # probe used for ldaps:// reads its certificate too.
+        cert = _fetch_ldaps_cert(host, port, timeout)
```
jelinaunger: If it is not ldap specific, the name is misleading. Could be e.g. _fetch_tls_cert?

privacyidea/lib/health.py
```diff
@@ -0,0 +1,346 @@
...
+        entry.update({"subject": None, "issuer": None, "not_after": None,
+                      "days_remaining": None, "status": "error", "error": str(e)})
+    return entry
+
+
+def _check_keycloak_resolvers() -> list:
+    """Probe the TLS server certificate of each Keycloak resolver's ``base_url``.
+
+    Keycloak resolvers talk to an admin-operated (often self-hosted) HTTPS
+    endpoint, so its server certificate is worth surfacing. The EntraID resolver
```
jelinaunger: not sure if we need to talk about entra id here as the function name clearly points to a keycloak specific function.

privacyidea/static/components/dashboard/controllers/dashboardControllers.js
```diff
@@ -21,18 +21,49 @@
...
                      "inactive": [], "num_inactive": 0};
     $scope.subscriptions = {};
     $scope.authentications = {"success": 0, "fail": 0};
+    $scope.latestNews = null;
+
+    $scope.getLatestNews = function () {
+        InfoFactory.getRSS(function (rss) {
+            // rss is { source: [items, ...], ... }; pick the first item of the
+            // first source - matches what the news page shows at the top.
+            for (var source in rss) {
```
jelinaunger, line 55: let / const should be used instead of var

tests/test_lib_health.py
```diff
@@ -0,0 +1,539 @@
...
+            .not_valid_after(not_after)
+            .sign(key, hashes.SHA256(), backend=default_backend()))
+
+
+class ClassifyTest(MyTestCase):
+    """Pure-function classification thresholds."""
+
+    def test_status_buckets(self):
+        # ok > 30, warning <= 30, critical <= 7, expired <= 0, error on None.
+        self.assertEqual(health._classify(31), "ok")
```
jelinaunger: argument order should be (expected, actual) to better understand failing tests

privacyidea/api/system.py
```diff
@@ -564,6 +566,184 @@ def list_nodes():
...
+    try:
+        since_seconds = int(get_optional(request.all_data, "since_seconds") or 3600)
+    except (TypeError, ValueError):
+        since_seconds = 3600
+    raw = get_metrics(name="resolver_op_duration_seconds", since_seconds=since_seconds)
+    g.audit_object.log({"success": True})
+    return send_result(raw)
+
+
+def _aggregate_delivery_channel(counter_name, duration_name, key_label, since_seconds):
```
jelinaunger: type hints

privacyidea/static/components/dashboard/controllers/dashboardControllers.js
nilsbehlen: the whole file uses var and is going to be deleted the in the future, not worth it to even comment that :D

jelinaunger (approved):

PR #5293, opened 2026-05-05
5254/improve zoom

### LINKED ISSUE #5254: Improve zoom for enroll token
To avoid having a two-line sticky header on zoom (cause the enroll button moves to the new line), we could reduce the size of the token type selections.  This might require to switch to a single column selection.

Additionally, the enroll button sometimes seems to be a bit misaligned (which is also the case for container create):

Image

henrik-falk:


.gitignore
```diff
@@ -36,3 +36,4 @@ dist/
 *.swp
 /.vscode/
 /ca/
+/privacyidea/static/public/assets/*
```
jelinaunger: hm this folder does not exist or did you mean static_new? But this might also not make sense

jelinaunger (approved):

.gitignore
henrik-falk: i created this folder for the logos.

PR #5310, opened 2026-05-07
fix all the things from the issues #5307 #5308 #5309

### LINKED ISSUE #5307: /system/random writes result to audit
The generated random bytes are written to the audit ``success``.
Instead, ``success`` should just be ``true/false``, we can additionally log the length of generated bytes, but not the bytes themselves.

### LINKED ISSUE #5308: ``DELETE /user/attribute/<attrkey>/<username>/<realm>`` does not check target rights
A user authenticated with the user-scope policy ``delete_custom_user_attributes=<attr>`` (intended to let them delete their own attribute) can delete that attribute from any user that resolves, by passing the victim's username and realm in the URL. No admin credentials are required. Confidentiality is not affected, but integrity of  user-attribute state is.

For admins, this should still be possible, just not for users.

### LINKED ISSUE #5309: multiple small things
Things that came up during an update of the API doc:
- [ ] `set_default_realm_api` currently logs the audit `success` field as an int (0/1); should log a bool (`api/realm.py`)                                              
- [ ] `get_security_module` / `set_security_module` currently log the audit `success` field as a dict (`{"is_ready": <bool>}`); should log a bool (`api/system.py`)
- [ ] `setpin` and `setrandompin` currently emit identical `action_detail="otppin, "` audit entries; `setrandompin` should emit a distinct value (e.g. `"otppin (random), "`) so generated PINs are distinguishable from explicitly set ones (`api/token.py`)                                                                           
- [ ] `tokenrealm_api` currently audits only the serial; should also audit the new realm list assigned to the token (`api/token.py`)                                    
- [ ] `init` currently reads `token.token.serial` for the audit block outside the `if token:` guard; should read it inside the guard so a falsy `init_token` return cannot raise `AttributeError` (`api/token.py`)                                                                                                                          
- [ ] `reset_password` currently audits only successful resets; should also audit failed attempts (`api/recover.py`)                                                    
- [ ] `register_post` currently compares `realm` against the function reference `get_default_realm` instead of its return value; should call `get_default_realm()` (`api/register.py`)                                                                                                                                                     
- [ ] `set_realm_node_api` currently returns 500 on a non-JSON body; should return a clean 400 (`api/realm.py`)                                                         
- [ ] `set_policy_api` currently ignores the `user_case_insensitive` flag; should wire it through to the stored policy (`api/policy.py`)                                
- [ ] subscriptions list endpoint currently ignores the `<application>` filter and carries a misleading `timedelta` comment; should honour the filter and the comment should describe what the code actually does (`api/subscriptions.py`)                                                                                                    
- [ ] container `set_states` / `set_realms` currently return 500 when the body is a JSON list (`replace` on a list); should accept either a JSON list or a  comma-separated string transparently (`api/container.py`)                                                                                                               
- [ ] `get_eventhandling` currently returns a hardcoded list of handler modules for `/event/handlermodules`; should derive the list dynamically so new handler modules  show up automatically (`api/event.py`, `lib/event.py`)                                                                                                                  
- [ ] `delete_config` currently has a `if not key: raise ParameterError(...)` branch that Flask routing makes unreachable; should be removed (`api/system.py`)
- [ ] `event.py` lookup endpoints (`/positions/<m>`, `/actions/<m>`, `/conditions/<m>`) are currently admin-only but ungated; should be gated with `EVENTHANDLINGREAD`,  matching the rest of the read surface                                                                                                                                   
- [ ] `periodictask.py` lookup endpoints (`/taskmodules/`, `/nodes/`, `/options/<m>`) are currently admin-only but ungated; should be gated with `PERIODICTASKREAD`     
- [ ] `policy.py` `check_policy_api` and `get_policy_defs` are currently admin-only but ungated; should be gated with `POLICYREAD`                                      
- [ ] `system.py` `get_config` is currently ungated for admins (only the user-side public-key filter applies); should be gated with `SYSTEMREAD` so the existing  `configread` policy actually restricts admins                                                                                                                           
- [ ] `container_list` policy action is currently undocumented in `doc/policies/admin.rst` and `doc/policies/user.rst`; should have its own section with a  `_policy_container_list:` cross-link label

nilsbehlen:


privacyidea/api/realm.py
```diff
@@ -306,7 +306,7 @@ def set_default_realm_api(realm=None):
     """
     realm = realm.lower().strip()
     r = set_default_realm(realm)
-    g.audit_object.log({"success": r,
+    g.audit_object.log({"success": bool(r),
```
jelinaunger: can we be sure the db id is never 0 on sucess?

jelinaunger (approved):

privacyidea/api/realm.py
nilsbehlen: will set true explicitly

PR #5326, opened 2026-05-07
Update self service user details

### LINKED ISSUE #4832: Fix self-service user details panel
- [x] Add something similar to the account and utils div to the self service. (Done in #4993)

In addition fix the width of the self service user details panel. With ad the panel explodes:

Image

henrik-falk:


privacyidea/static_new/src/app/components/layout/navigation-self-service/navigation-self-service.component.html
```diff
@@ -19,40 +19,20 @@
...
-        </div>
-      </mat-expansion-panel>
-    </mat-accordion>
+    @if (contentService.onUserDetailsSelfService()) {
+      <div class="username-button disabled">
+        <span class="username">{{ userData().username || "-" }}
+          &#64; {{ userService.selectedUserRealm() || "-" }}</span>
+      </div>
+    } @else {
+      <a
```
frankmer: Before klicking at the user the box is much smaller and looks wired.

privacyidea/static_new/src/app/components/user/user-details/user-details.self-service.component.html
```diff
@@ -0,0 +1,48 @@
...
+    <h3 i18n>Your Details</h3>
+  </div>
+
+  <div class="user-details">
+    @for (col of detailsColumns(); track $index) {
+      <div class="user-details-column">
+        @for (item of col; track item.key) {
+          <div class="label">{{ item.label }}:</div>
+          @if (Array.isArray(item.value) && item.value.length > 1) {
+            <div class="value">
```
frankmer: When the value does not fit here the scrollbar covers almost the whole value text

privacyidea/static_new/src/app/components/layout/navigation-self-service/navigation-self-service.component.html
```diff
@@ -19,40 +19,20 @@
 <div class="navigation-self-service-column">
 
   @if (authService.actionAllowed("userlist")) {
```
jelinaunger: I would always display the login name and only remove the link in case the user has not the right userlist

frankmer (approved):

PR #5329, opened 2026-05-07
better icons

henrik-falk:


jelinaunger (commented):
Can you also add them for container templates, resolvers, machines and events?

frankmer (approved):

PR #5338, opened 2026-05-11
test passkey add match check

### LINKED ISSUE #5263: Identify passkey in the WebUI, if user has more than one passkey

A user has more than one passkey. One passkey is lost.
Currently (3.13) it is not easily possible in the WebUI to determine, which token is lost.

a) The user has two passkey tokens A and B. Let us assume token A is lost. Admin goes to token A details und uses "TestPasskey" and successfully tests with token B --- although he is on the details of token A. 
b) The user logs in to the web ui (/auth endpoint) with token B, but the token information is not contained in the audit log.

There are only two ways to determine the token:

c) the admin disables token A and tries the "TestPasskey" with token B and it works. If the admin disables token B and tries with the token B, it fails. So to test, which token this actually is, the admin would have to disable all passkeys in a row.

d) The user logs in to an application and the validate/check audit entry hopefully contains the serial number.

It would be great, if **scenario a)** would work. If "TestPasskey" on the detail page on token A would fail, if the Test was performed with token B.

Scenario b) should also work. The /auth endpoint should write in the audit log, which passkey (serial) was used.

**Additional context**

In my tests I the user had two passkeys, a Yubikey and a Nitrokey. But usually you do not see this in the token detail. Sine the passkeys are indistingushable, the testing mechanisms should be able to identify the passkey used.

nilsbehlen:
- Show a warning if the tested passkey is not the one that is viewed
- Make the Login page not submit if passkey login is cancelled

### DIFF
--- a/privacyidea/static/components/token/controllers/tokenDetailController.js
+++ b/privacyidea/static/components/token/controllers/tokenDetailController.js
@@ -141,14 +141,14 @@ myApp.controller("tokenDetailController", ['$scope', 'TokenFactory',
                                 if ($scope.loggedInUser.role === "admin"
                                     && data.detail.serial
                                     && data.detail.serial !== $scope.tokenSerial) {
+                                    $scope.passkeyMismatch = {serial: data.detail.serial};
                                     TokenFactory.getTokenForSerial(data.detail.serial, function (tokData) {
-                                        const tokens = tokData.result.value.tokens;
-                                        const matched = tokens && tokens.length ? tokens[0] : null;
+                                        const matched = tokData?.result?.value?.tokens?.[0];
                                         $scope.passkeyMismatch = {
                                             serial: data.detail.serial,
-                                            username: matched ? matched.username : "",
-                                            realm: matched ? matched.user_realm : "",
-                                            resolver: matched ? matched.resolver : ""
+                                            username: matched?.username || "",
+                                            realm: matched?.user_realm || "",
+                                            resolver: matched?.resolver || ""
                                         };
                                     });
                                 }
--- a/privacyidea/static/components/token/controllers/tokenDetailController.js
+++ b/privacyidea/static/components/token/controllers/tokenDetailController.js
@@ -106,6 +106,8 @@ myApp.controller("tokenDetailController", ['$scope', 'TokenFactory',
         }
 
         $scope.testPasskey = function () {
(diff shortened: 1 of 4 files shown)


privacyidea/static/components/token/views/token.details.html
```diff
@@ -522,7 +522,17 @@ <h3 name="Token Details">
                     <span>Test Passkey</span>
                 </button>
                 <div ng-show="loggedinUsername">
-                    <p translate>You will be logged in as: {{ loggedinUsername }}</p>
+                    <p ng-hide="passkeyMismatch" translate>You will be logged in as: {{ loggedinUsername }}</p>
+                    <p ng-show="passkeyMismatch && loggedInUser.role === 'admin'"
+                       style="color: darkorange;">
+                        <span translate>Authentication successful, but the passkey that was used is not this, but</span>
+                        <a ui-sref="token.details({tokenSerial: passkeyMismatch.serial})">{{ passkeyMismatch.serial }}</a>
+                        <span ng-show="passkeyMismatch.username">
```
frankmer: If the Passkey token is not assigned to a user, the text will not read very naturally: "Authentication successful, but the passkey that was used is not this, but [SERIAL]."

privacyidea/static/components/token/controllers/tokenDetailController.js
```diff
@@ -136,6 +138,20 @@ myApp.controller("tokenDetailController", ['$scope', 'TokenFactory',
...
                                 $scope.loggedinUsername = data.detail.username;
+                                if ($scope.loggedInUser.role === "admin"
+                                    && data.detail.serial
+                                    && data.detail.serial !== $scope.tokenSerial) {
+                                    TokenFactory.getTokenForSerial(data.detail.serial, function (tokData) {
+                                        const tokens = tokData.result.value.tokens;
+                                        const matched = tokens && tokens.length ? tokens[0] : null;
+                                        $scope.passkeyMismatch = {
+                                            serial: data.detail.serial,
+                                            username: matched ? matched.username : "",
```
frankmer: I guess
```
username: matched?.username || "",
realm:    matched?.user_realm || "",
resolver: matched?.resolver || "",
```
would be better to catch the case if username is undefined and not an empty string.

frankmer (approved):
Please note that I left two minor comments.

PR #5343, opened 2026-05-12
Dialog width and icon update

henrik-falk:


privacyidea/static_new/src/styles/dialog.scss
```diff
@@ -35,4 +35,9 @@
 }
 .mat-mdc-dialog-content {
   max-height: 82vh;
+  font-size: 16px
```
frankmer, line 38: can we use a var() for the font size?

frankmer (approved):

PR #5349, opened 2026-05-12
5232/Consistent handeling of resources

### LINKED ISSUE #5232: Consistent handling of error and loading states in linked signals of httpResources
Currently, linkedSignal computations that derive their value from an httpResource handle the "no value" case inconsistently across services. Some return a default value, some return the previous value, and none distinguish between the two distinct states that cause hasValue() to return false: Loading and Error.

All such linkedSignal computations should be refactored to explicitly check resource.error() / resource.isLoading():
* Loading – resource is in flight; keeping the previous value is acceptable to avoid UI flicker
* Error – request failed; silently returning stale data is incorrect and potentially misleading -> return empty / default data

henrik-falk:


privacyidea/static_new/src/app/services/application/application.service.ts
```diff
@@ -91,35 +91,38 @@ export class ApplicationService implements ApplicationServiceInterface {
...
-              webauthn: {}
-            }
-          },
-          ssh: {
-            options: {
-              sshkey: {
-                service_id: { description: "", type: "", value: [] },
-                user: { description: "", type: "" }
-              }
+      const empty: Applications = {
```
frankmer: Is this the only place where an empty Applications object is usefull? Maybe we should create a emptyApplications constant?

frankmer: Currently only used here, but extracting it as a constant would slim down the computation function and avoid creating the object on every run.

frankmer (approved):

jelinaunger (approved):

PR #5380, opened 2026-05-21
container resource fix

### LINKED ISSUE #5376: containerResource reused for multiple purposes causes incorrect filtering on token details page
The containerResource in ContainerService currently serves three different contexts: the paginated container list, user-scoped containers on the user details page, and token-type-compatible containers on the token details/enrollment pages.

  This causes a bug: on the token details page, when a user is assigned to a token, selectedUser leaks into the container request, filtering results to only that user's containers — instead of showing all containers compatible with the token type.

  Proposed fix: Split into purpose-specific httpResource instances (e.g. containerListResource, compatibleContainersResource, userContainersResource), each computing only the parameters relevant to its context. To avoid duplication, extract the shared request shape (URL, method, headers) into a small builder function that each resource calls with its own params.

jelinaunger:
closes #5376


privacyidea/static_new/src/app/components/token/token-details/token-details.component.html
```diff
@@ -81,6 +81,12 @@ <h3 i18n>{{ " (Offline)" }}</h3>
...
                       </mat-autocomplete>
                     </mat-form-field>
+                    <mat-checkbox
+                      [(ngModel)]="containerService.filterContainersByTokenOwner"
+                      [disabled]="!tokenDetails().username"
+                      i18n>
+                      Filter by token owner
+                    </mat-checkbox>
                   } @else {
                     <app-copyable [copyText]="element.value">
```
frankmer, line 91: click must be accompanied by either keyup, keydown or keypress event for accessibility. (eslint@angular-eslint/template/click-events-have-key-events)
Elements with interaction handlers must be focusable. (eslint@angular-eslint/template/interactive-supports-focus)

privacyidea/static_new/src/app/components/user/user-details/user-details-container-table/user-details-container-table.component.ts
frankmer: Reloads the wrong resource. reload the used userContainersResource

privacyidea/static_new/src/app/services/container/container.service.ts
```diff
@@ -365,32 +369,31 @@ export class ContainerService implements ContainerServiceInterface {
...
-
     // Only load containers on routes with a container list or selection.
-    const onAllowedRoute =
-      this.contentService.onTokenDetails() ||
-      this.contentService.onUserDetails() ||
-      this.contentService.onContainers() ||
-      this.contentService.onTokensEnrollment() ||
-      this.contentService.onTokens();
-
+    const onAllowedRoute = this.contentService.onContainers() || this.contentService.onTokens();
```
frankmer: Do we need to load containers on Tokens route?
I think we only need to get the container serial list on TokenDetails.

frankmer (approved):

PR #5389, opened 2026-05-22
plugin notifications: basic calculation and dashboard panel

### LINKED ISSUE #4057: Notification about used privacyIDEA plugins
privacyIDEA knows subscriptions and it knows, which plugins communicate with privacyIDEA.
These are stored in the table [ClientApplication](https://github.com/privacyidea/privacyidea/blob/master/privacyidea/models.py#L2674)
including the name of the Plugin and the lastseen timestamp.

If the privacyIDEA server has a sever subscription, we can use this to correlate with the plugin subscriptions.
I.e. if we see recent requests from a certain plugin but we do not see a subscription for this plugin, we can notify the admin, that there is the possibility to get a subscription (sometime he is not even aware of this).

The notification could be done in a similar way like the subscription notification itself in dialogs.welcome.

It is handled in the loginController
https://github.com/privacyidea/privacyidea/blob/1573c7df8ce1cf6521cbf741222d95478f5c5816/privacyidea/static/components/login/controllers/loginControllers.js#L479
and prepared in the backend here https://github.com/privacyidea/privacyidea/blob/064f0954e0e70844d72e3bee93a2ec2b62a080a9/privacyidea/api/lib/postpolicy.py#L679
We could then display a corresponding dialog
https://github.com/privacyidea/privacyidea/blob/fcc9eb6191be4f492ae943041b94d23652694995/privacyidea/static/components/dialogs/views/dialog.welcome.html#L27
noting, that a certain plugin has been used during the last X days.

### LINKED ISSUE #5383: Plugin traffic light in Dashboard
The privacyIDEA WebUI should have the dashboard as default start screen.
The dashboard should contain a kind of list or traffic light, of the available plugins/client applications.

# Why

This traffic light can help to tell admin who newly installed privacyIDEA, which scenarios with privacyIDEA are possible. 

If the system has a subscription, it can be displayed here.

The admin will also know, which plugins are used or not.

The traffic light should always be there. It should not be able to be disabled.

# What

The plugin traffice light can be grouped in different types:

* SSO
	- Keycloak
	- Shib
	- simpleSAML
	- ADFS
* Computerlogin
	- Windows CP
	- Linux PAM
* FreeRADIUS
* LDAP Proxy
* CA Connector
* Windows Enrollment Client
* privacyIDEA Authenticator App

The privacyIDEA server can check via the API request, if a Plugin sends requests to the privacyIDEA server.
E.g. during the last one week or two weeks. Maybe this time is configurable. Maybe the light dimms, when the request are more in the past.

To do so, it might be a good idea, to have an extra database table to record the last request of a plugin. Similar to the clientsapplication table.

The traffic lights can have three different colors, like:

gray: This plugins is possible, but not used.
green: This plugin is used, and you have a subscription
red: This plugin is used, but there is no subscription.

### click actions

The different lights/plugins could also have a link, that is leading to the documentation or the github repository.

This way, an administrator who installs privacyIDEA anews has some links, directly on the startscreen, to get more ideas, what he can do with the privacyIDEA server.

There could be different click actions depending on the color of the light:

* gray: link to the documentation.
* red: (the plugin is used, but not subscribed --- maybe dependent on the number of users) an additional link to sales, to avoid downtimes due to missing subscriptions.

nilsbehlen:


tests/test_lib_subscriptions.py
```diff
@@ -173,3 +177,227 @@ def test_04_subscription_status(self):
...
+        overview = get_plugin_subscription_status()
+        # All plugins still report unused (no ClientApplication rows seeded).
+        for entry in overview:
+            self.assertEqual(entry["status"], "unused")
+
+
+class ServerSubscriptionStatusTestCase(MyTestCase):
+    """
+    Tests for :func:`get_server_subscription_status`. Covers the four
+    branches (no_subscription / ok / expiring / expired) plus the
```
frankmer: (no_subscription / active / expiring / expired)

privacyidea/api/subscriptions.py
```diff
@@ -71,6 +74,37 @@ def api_get(application=None):
...
+    """
+    overview = [get_server_subscription_status()] + get_plugin_subscription_status()
+    # Enrich each entry with the latest release from GitHub (version, date and
+    # release page URL), cached server-side.
+    latest_versions = get_latest_github_versions()
+    for entry in overview:
+        release = latest_versions.get(entry["application"])
+        entry["current_version"] = release.get("version") if release else None
+        entry["current_version_date"] = release.get("released") if release else None
+        entry["current_version_url"] = release.get("url") if release else None
```
jelinaunger: might make sense to use a dataclass for each release

privacyidea/lib/subscriptions.py
```diff
@@ -63,23 +65,173 @@
...
+APPLICATION_ALIASES = {user_agent.lower(): application
+                       for application, config in SUBSCRIPTIONS.items()
+                       for user_agent in config.get("user_agents", [])}
+
+
+def get_subscription_application(plugin_name):
+    """
+    Map a plugin user-agent name to the application whose subscription it
+    counts against, following :data:`APPLICATION_ALIASES`. Names that are not
+    aliases are returned lower-cased and otherwise unchanged.
```
jelinaunger: Is there a reason to lower-case the one but not the other? Cause at the end the entire list is an application name list no matter where these names are derived from. Seems to be inconsistent to me.

privacyidea/lib/subscriptions.py
```diff
@@ -63,23 +65,173 @@
...
+            data = response.json()
+            version = (data.get("tag_name") or "").lstrip("v")
+            if not version:
+                return None
+            return {"version": version,
+                    "released": (data.get("published_at") or "")[:10] or None,
+                    "url": data.get("html_url")}
+        log.info(f"GitHub returned {response.status_code} for the latest release of {repo}")
+    except (requests.RequestException, ValueError) as error:
+        log.info(f"Could not fetch the latest release for {repo}: {error}")
```
jelinaunger: I would move this to debug level. In air-gapped installations this is expected to not work and might get annoying on the info level.

jelinaunger: Or it might make sense to disable it entirely for exactly these installations it can never work, so it does not waste time on trying it.

privacyidea/lib/subscriptions.py
```diff
@@ -63,23 +65,173 @@
...
+    Return ``{application: {"version": ..., "released": ...} or None}`` for the
+    clients in :data:`GITHUB_REPOS`. Results are fetched from GitHub
+    concurrently and cached for :data:`GITHUB_VERSION_TTL`; this is best-effort,
+    so unreachable or unknown repositories map to None.
+
+    :rtype: dict
+    """
+    now = datetime.datetime.now()
+    if (_github_version_cache["fetched_at"]
+            and now - _github_version_cache["fetched_at"] < GITHUB_VERSION_TTL):
```
jelinaunger: The IDE shows type warnings as it expects a datetime object here but the datatype is unknown from the dict. A dataclass might be useful here as well.

privacyidea/lib/subscriptions.py
```diff
@@ -100,6 +252,169 @@ def get_users_with_active_tokens():
...
+
+    Possible values, with the colour the frontend maps them to:
+
+    * ``none`` (grey) — no subscription on file.
+    * ``valid`` (green) — subscription valid, not near expiry, within token limit.
+    * ``expiring`` (yellow) — subscription valid but ends within
+      :data:`EXPIRING_THRESHOLD_DAYS` days.
+    * ``exceeded`` (yellow) — subscription valid but more users with active
+      tokens than the subscription allows (``num_tokens``).
+    * ``expired`` (red) — subscription ``date_till`` is in the past.
```
jelinaunger: these could be enums

privacyidea/lib/subscriptions.py
```diff
@@ -100,6 +252,169 @@ def get_users_with_active_tokens():
...
+
+    :param subscription: the subscription dict, or None if none is on file
+    :param now: the reference "now" timestamp
+    :param token_users: number of users with active tokens (for the token check)
+    :return: ``(state, date_till, days_left)`` — ``date_till``/``days_left`` are
+        None when no subscription is on file.
+    :rtype: tuple
```
jelinaunger: more information about each value in the tuple would be useful (or replace the tuple entirely)

privacyidea/lib/subscriptions.py
```diff
@@ -100,6 +252,169 @@ def get_users_with_active_tokens():
...
+    subscription on file or was seen within :data:`USAGE_RECENT_DAYS` days,
+    otherwise ``"no"`` (blue).
+
+    :rtype: str
+    """
+    if has_subscription:
+        return "yes"
+    if last_seen is not None and (now - last_seen).days < USAGE_RECENT_DAYS:
+        return "yes"
+    return "no"
```
jelinaunger: Why not use a bool here?

tests/test_lib_subscriptions.py
```diff
@@ -173,3 +181,445 @@ def test_04_subscription_status(self):
...
+        with mock.patch("privacyidea.lib.subscriptions.requests.get") as mock_get2:
+            versions2 = get_latest_github_versions()
+            mock_get2.assert_not_called()
+        self.assertEqual(versions2["privacyidea"]["version"], "4.7.3")
+
+    def test_02_unreachable_repo_maps_to_none(self):
+        with mock.patch("privacyidea.lib.subscriptions.requests.get",
+                        side_effect=requests.RequestException("boom")):
+            versions = get_latest_github_versions()
+        self.assertIsNone(versions["privacyidea"])
```
jelinaunger, line 810: What happens here at a second call, is it also taken from the cache?

nilsbehlen:
image
image

privacyidea/static_new/src/app/components/dashboard/widgets/subscriptions-widget/subscriptions-widget.component.scss
```diff
@@ -21,20 +21,164 @@
...
+.subscription-table {
+  width: 100%;
+  border-collapse: collapse;
+  // The widget cannot grow when the admin switches to the detailed view — the dashboard
+  // layout is theirs to arrange — so both views have to live in the same box. Fixed
+  // column widths keep the extra columns inside the widget instead of widening the
+  // table, and every cell stays one line high, so the two views are the same height.
+  table-layout: fixed;
+  // Dense rows so the compact view fits every component without scrolling.
+  font-size: 0.85rem;
```
jelinaunger: Should get the same font size as the rest. It is very small and does not even solve the problem described here, scrolls and truncated values can still appear. You could use tool tips to make the complete value visible.

privacyidea/static_new/src/app/components/dashboard/widgets/subscriptions-widget/subscriptions-widget.component.scss
```diff
@@ -21,20 +21,164 @@
...
+    // tall in the detailed view as in the compact one.
+    line-height: 1.5;
+    // Opt out of the app-wide column striping (styles/table.scss): here the columns are
+    // an overview of states, and a fill on every other one only groups them by accident.
+    background: transparent;
+  }
+
+  th {
+    font-weight: 600;
+    opacity: 0.7;
```
jelinaunger: To be consistent with the other tables, headers shouldn't be transparent. Especially when a section-row (e.g. Use Cases) is right below it, it is not readable.

privacyidea/static_new/src/app/components/dashboard/widgets/subscriptions-widget/subscriptions-widget.component.scss
```diff
@@ -21,20 +21,164 @@
...
+  &.dot-good {
+    background: var(--green-light);
+  }
+
+  &.dot-warn {
+    background: var(--yellow-light-active);
+  }
+
+  &.dot-bad {
+    background: var(--red-light);
```
jelinaunger, line 203: dot-good (green) vs dot-bad (red) is the only difference between   
  "in use" and "not in use" for a sighted user, and those two are the classic red/green confusion pair — WCAG   
  1.4.1. The tooltip is the only disambiguator, and it needs a pointer. Worth a second channel: distinct glyphs 
  (✓ / ! / –) inside the dot, or a shape difference.

privacyidea/static_new/src/app/components/dashboard/widgets/subscriptions-widget/subscriptions-widget.component.html
```diff
@@ -18,18 +18,171 @@
...
+                  </span>
+                </td>
+              } @else {
+                <!-- Carries the subscription state in its tooltip, for what the date alone
+                     does not say (a valid subscription can still be exceeded). -->
+                <td
+                  class="expires-cell"
+                  [matTooltip]="subscriptionReason(status.subscription)">
+                  @if (status.date_till) {
+                    {{ status.date_till | date: "yyyy-MM-dd" }}
```
jelinaunger: `date_till` is a calendar date, not an instant — it comes from the signed subscription file as `%Y-%m-%d` (`lib/subscriptions.py:51`), and the `00:00:00 GMT`                           
  is invented by `strptime` plus Flask's serializer. Converting it to local time shifts it a day back for anyone west of UTC:                                                                  
                                                                                                                
      TZ=America/New_York npx jest .../subscriptions-widget

      Expected: "2026-06-30 (expired, 49 days ago)"
      Received: "2026-06-29 (expired, 49 days ago)"

  So the date contradicts the `days_left` note next to it (that's server-side), and it disagrees with the subscription page, which renders the same field unconverted. Three tests in this spec are timezone-dependent as a result — green in CI, red for anyone running them in the Americas.                                                                          
                                                                                                                
  Local time is right for instants like `last_seen`; there's just no instant here to convert. Simplest fix:

    {{ status.date_till | date: "yyyy-MM-dd" : "UTC" }}

Cleaner, if you want to touch the API: return `date_till` as a date-only string and drop the pipe — that's already how `current_version_date` in the next cell works.

privacyidea/lib/subscriptions.py
```diff
@@ -100,6 +428,193 @@ def get_users_with_active_tokens():
...
+    """
+    if not subscription:
+        return SubscriptionStateInfo(SubscriptionState.NONE)
+    date_till = subscription.get("date_till")
+    if not date_till:
+        # Subscription.date_till is nullable. Without an end date the record cannot be
+        # said to cover anything, so it is reported like an expired one rather than
+        # staying green forever.
+        return SubscriptionStateInfo(SubscriptionState.EXPIRED)
+    days_left = (date_till - now).days
```
jelinaunger: `timedelta.days` floors, so this reports a whole day the moment the boundary is crossed:
      
      >>> till = datetime(2026, 6, 30); now = datetime(2026, 6, 30, 0, 0, 1)
      >>> (till - now).days
      -1                                                                                                        
                                                                                                                
  One second past midnight the widget reads "expired, 1 days ago". Since `date_till` is a calendar date anyway (see the other comment), counting whole dates rather than datetimes gives the number the text claims:                                                                   
                                                                                                                
      days_left = (date_till.date() - now.date()).days

Then the first day of expiry says "0 days ago" — or "today", if you'd rather special-case it — and a subscription ending tomorrow says "1 day left" instead of "0".

nilsbehlen:
image

i put the buttons in the widget header. the widget now allow adding actions like that to the header. looks cleaner than having it inside the widget taking up another row of space

privacyidea/static_new/src/app/components/dashboard/widgets/subscriptions-widget/subscriptions-widget.component.scss
nilsbehlen: added icons inside the dots, see new screenshot

privacyidea/static_new/src/app/components/dashboard/widgets/subscriptions-widget/subscriptions-widget.component.html
nilsbehlen: didnt touch the api, added a explicit "today".

privacyidea/lib/subscriptions.py
nilsbehlen: added today

privacyidea/static_new/src/app/components/dashboard/widgets/subscriptions-widget/subscriptions-widget.component.scss
nilsbehlen: removed any font size specific to this widget

jelinaunger (approved):

PR #5399, opened 2026-05-22
Improve user details for long attribute values #5379

### LINKED ISSUE #5379: Improve user details for long attribute values
Users might have quite long attributes. They can also contain a list (groups). Actually, the width and height is fixed resulting in a horizontal and for some additionally in a horizontal scroll. This is very bad readable.

Multi value attributes could be expandable on click to avoid making the UI crowded while keeping accessibility.

Image

frankmer:


privacyidea/static_new/src/app/components/user/user-details/user-details.component.scss
```diff
@@ -53,22 +54,62 @@
...
-  margin-block: auto;
-  max-width: 240px;
-  height: 28px;
+  min-height: 28px;
+  align-self: start;
+  overflow-wrap: anywhere;
+}
+
+.value-list {
+  list-style: disc inside;
```
jelinaunger: I would prefer outside

jelinaunger (approved):

PR #5408, opened 2026-05-27
refactor and fix token type enroll components

### LINKED ISSUE #5407: Reuse enroll token type switch for rollover and in container templates
Token type switch used to render the correct enroll component for the selected token type in the token enrollment could also be used for token rollover and in container templates to remove duplicated code.

Additionally, there is a bug for container templates, where all token types are disabled, even if they should be expandable. This is cause `additionalFormFieldsChange.emit({})` empty additional form fields are emitted on init. 
Instead of this conditional disabling we could include the token type description, hence every type has a content and no further siabling is required.
AdditionalFormFieldsChange is not used anymore, hence it can be removed.

jelinaunger:
* fix disabled token types in container templates: Include token type description making disable state obsolete
* Refactor enroll token type switch and reuse it for container templates and rollover
* Remove unused `AdditionalFormFieldsChange` from enroll token type components


privacyidea/static_new/src/app/components/token/token-enrollment/enroll-asp/enroll-applspec.component.ts
```diff
@@ -95,7 +94,6 @@ export class EnrollApplspecComponent implements OnInit {
         this.otpKey.set(this.enrollmentData()!.otpKey ?? "");
       }
     }
-    this.additionalFormFieldsChange.emit({});
     this.enrollmentArgsGetterChange.emit(this.enrollmentArgsGetter);
```
henrik-falk: would it be possible to remove the enrollmentArgsGetterChange emitter too? we should replace all emitter with signals now that we have signal forms.

henrik-falk (approved):

PR #5422, opened 2026-05-28
apply ESLint suggestions

### LINKED ISSUE #5377: apply ESLint suggestions
for example no-explicit-any

frankmer:


privacyidea/static_new/src/app/app.component.ts
```diff
@@ -119,14 +118,11 @@ export class AppComponent implements OnInit {
       console.warn("User is already logged in.");
       this.notificationService.warning("User is already logged in.");
     }
-
-    const _welcomeInit = inject(WelcomeDialogService);
+    // const _welcomeInit = inject(WelcomeDialogService);
```
henrik-falk: why remove welcomeDialog? es lint suggestions should not change the logic.

henrik-falk (changes requested):

privacyidea/static_new/src/app/app.component.ts
frankmer: lint was "unused variable"

privacyidea/static_new/src/app/mappers/token-api-payload/4eyes-token-api-payload.mapper.ts
```diff
@@ -67,7 +67,7 @@ export class FourEyesApiPayloadMapper
 
   override fromApiPayload(payload: FourEyesEnrollmentPayload): FourEyesEnrollmentData {
     const requiredTokenOfRealms = Object.entries(payload["4eyes"] || {})
-      .filter(([_, value]) => value.selected)
+      .filter(([, value]) => value.selected)
```
jelinaunger, line 70: Using _ to assign returned but unused variables is a widely used pattern. We could add a rule to es lint config to allow this.

privacyidea/static_new/src/app/services/system/system.service.ts
```diff
@@ -32,35 +32,71 @@ export interface NodeInfo {
...
-
-  deleteUserCache(): Observable<PiResponse<any>>;
-
-  loadSmtpIdentifiers(): Observable<PiResponse<any>>;
-
+  saveSystemConfig(config: Record<string, unknown>): Observable<PiResponse<Record<string, "insert" | "update">>>;
+  deleteSystemConfig(key: string): Observable<PiResponse<boolean>>;
+  deleteUserCache(): Observable<PiResponse<DeleteUserCacheResult>>;
+  // No backend route exists for this and it is currently unused, so the type is unknown.
+  loadSmtpIdentifiers(): Observable<PiResponse<Record<string, string>>>;
```
jelinaunger: Then it should be removed instead of adding a random type here

privacyidea/static_new/src/app/app.component.ts
jelinaunger: The injection is required here to correctly trigger the welcome dialog (effect in the welcome dialog service). For now, just remove the constant assignment but keep the service injection to make ES lint happy.
In general, we might think about another approach to avoid this injection is removed in the future. Maybe adding this to app.config.ts using provideAppInitializer.

privacyidea/static_new/tsconfig.check.json
```diff
@@ -0,0 +1,11 @@
...
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./out-tsc/check",
+    "noEmit": true,
+    "types": ["jest", "node"]
+  },
+  "include": [
+    "src/app/components/token/token-details/token-details.component.spec.ts"
+  ]
+}
```
jelinaunger: What is this? Seems to be unused?

privacyidea/static_new/eslint.config.js
```diff
@@ -30,6 +30,15 @@ module.exports = defineConfig([
...
           style: "kebab-case"
         }
+      ],
+      "@typescript-eslint/no-unused-vars": [
+        "error",
+        {
+          argsIgnorePattern: "^_+$",
+          varsIgnorePattern: "^_+$",
+          caughtErrorsIgnorePattern: "^_+$",
+          destructuredArrayIgnorePattern: "^_+$"
```
jelinaunger, line 40: I would define this for underscore only ("^_$"), cause I think we still have some private variables prefixed with underscore which would mix this up.

jelinaunger (approved):

privacyidea/static_new/eslint.config.js
frankmer: ^\_$ would only match exact 1 \_ 
^\_+$ match any variable name that only contains \_ e.g: (\_, \_\_, \_\_\_)
Waht you describe would look like : ^\_+ or ^\_

henrik-falk (approved):

PR #5428, opened 2026-06-03
4914/conditional access

### LINKED ISSUE #4914: Conditional Access / Lockout Policies
This feature will cover several needs that were previously discussed as "Pre-Authentication", but not all of them. The previous discussions about Pre-Authentication can be divided into what this feature would do (#4092, #4087, #3741, #4128, #3708 ...), and the other part would be updating and evolving our authentication flow.
This also touches on points of #4394

For a first iteration we should focus on the main use-cases:
- Detect a compromised PIN/Password and allow administrators to define a reaction
- Detect brute force attempts and allow administrators to define a reaction
- Detect Password Spraying / Credential Stuffing attempts and allow administrators to define a reaction
- Allow administrators to configure escalation stages depending on the type of auth failure 

This system should be active before any other existing measure (token fail count, max_auth policy...) is even active.

What we need:

### Data:
- **Dedicated Authentication Log**. The current audit log is good for what its name suggests: being an _verifiable audit log of all things_ that happend (including management operations), but it is not suited to support this suggested feature with data. 
We need a log that stores data optimized for queries like: "how often in the last 5 minutes did this user try to log in?" "how often in the last hour did the user fail enter their otp/pin/password?", "is there an accumulation of attempts where the password is correct but the OTP isnt?" (or the other way around, depending on what is compromised)
-  **Precise error codes for each type of failure.** To infer information to use in policies configured with this feature, we need the precise reason of failure: was the pin wrong or the OTP or both? User has no token, User does not exist etc.

### Configuration
For a framework type approach, we need to allow administrators to set their own limits/values. This could like this (in terms of tables/data):
- **user_lockout_state**: tracks if a user locked out, why, and when it expires.
- **lockout_policy_stages**: stores the stages defined by the admins with their triggers.
- **lockout_policy_actions**: stores what actions (send mail, reset pin, deactivate token...) should be taken at the given stage.
- **block_list**: to block IPs (spraying) or devices, or anything not user-centric. (?)

### Dashboard
The data collected in the auth log should be shown on the dashboard

#### Bonus points
- Implementing a dry-run feature, so that the auth log contains what _would have_ happened if it was activated.

In further iterations, this can be extended with stuff from #4394, establishing a baseline for user behavior (known devices, login times etc), step-up authentication (force to use fido for these applications...)

henrik-falk:


privacyidea/api/validate.py
```diff
@@ -741,6 +834,55 @@ def _finalize_auth_response(context):
...
+        # context raises InvalidRequestError ("Can't operate on closed transaction
+        # inside context manager") — which this finally would silently swallow.
+        evaluate_lockout_policies(context["user"], context[AUTH_EVENT_TYPE_KEY], source_ip=g.client_ip)
+    except Exception as ex:
+        log.warning(f"Conditional-access policy evaluation failed: {ex!r}")
+        # A failure may leave the session in an aborted state; clear it so request
+        # teardown can proceed cleanly. Guard the rollback so this finally-context
+        # helper never raises (it must never break the already-completed response).
+        try:
+            db.session.rollback()
```
henrik-falk: rollback in try?

privacyidea/api/ttype.py
henrik-falk, line 183: Only where nothing has decided the outcome yet?                                                                                               
The conditional-access pre-check in PushTokenClass._api_endpoint_post logs {"success": False, "info": "Rejected: …"} for a refused push answer and hands the result back as an ordinary (False, {}) pair — so control returns here and this unconditional True overwrites it, since audit_object.log merges by key. The persisted audit row for a rejected push answer reads success=1, and an admin filtering the audit log for  
failed push answers finds none. The authentication-log row added just below is written separately and says USER_LOCKED either way, so this is the audit view only. but that is the view the WebUI shows.                                                                                                                  
if g.audit_object.audit_data.get("success") is None: around the flag would leave whoever already judged the request in charge. user has the same shape of problem: get_user_from_param finds nothing on a push answer (the smartphone sends only serial and signature), so this logs an empty user while _push_token_owner right below resolves the real one for the auth-log row.

nilsbehlen (approved):

PR #5438, opened 2026-06-08
4914/conditional access policy framework

### LINKED ISSUE #5433: Conditional Access: M1 - Policy Framework
* create db model for lockout policies
* live state tables (user lockout state)
* engine workflow

henrik-falk:


privacyidea/api/auth.py
```diff
@@ -278,6 +342,38 @@ def get_auth_token():
...
+    # Pre-auth conditional-access decision (DENY action), after the lock/block
+    # pre-checks so ALLOW cannot override them. A DENY rejects this single login
+    # with a message stating it was a conditional-access decision (the policy is
+    # not named); like the lock/block messages it is maskable via the
+    # hide_specific_error_message policy. ALLOW / CONTINUE fall through silently.
+    if evaluate_access_decision(user, g.client_ip) == AccessDecision.DENY:
+        log.info(f"Denying /auth login for {user!r} by conditional-access policy.")
+        g.audit_object.log({"info": "Rejected: denied by conditional-access policy"})
+        raise AuthError(_("Authentication failure. Access has been denied by a conditional-access policy."),
+                        id=Error.AUTHENTICATE_WRONG_CREDENTIALS)
```
jelinaunger: move this into a helper function to not further clutter this endpoint code

privacyidea/api/validate.py
```diff
@@ -469,6 +472,32 @@ def check():
...
+        return send_result(False, rid=2, details={})
+    # Pre-auth conditional-access decision (DENY action). Evaluated after the
+    # lock/block pre-checks (so ALLOW cannot override them) and before any token
+    # logic. A DENY rejects this single request generically without persisting
+    # state; ALLOW / CONTINUE fall through to the normal flow.
+    if evaluate_access_decision(request.User, g.client_ip) == AccessDecision.DENY:
+        log.info(f"Denying authentication for {request.User!r} by conditional-access policy.")
+        g.audit_object.log({"success": False,
+                            "info": "Rejected: denied by conditional-access policy"})
+        return send_result(False, rid=2, details={})
```
jelinaunger: same here, move this into a helper function

privacyidea/api/validate.py
```diff
@@ -502,8 +531,11 @@ def check():
             _handle_standard_auth(context)
         response = _finalize_auth_response(context)
     finally:
-        # Write the single authentication-log row for this request
+        # Write the single authentication-log row for this request, then let the
+        # conditional-access engine react to the classified outcome. Both run in
+        # the finally so they cover early returns and handler errors alike.
```
jelinaunger: Last sentence of the comment is not required as that is the general purpose of try finally blocks

privacyidea/api/validate.py
```diff
@@ -803,6 +835,26 @@ def _log_authentication_event(context):
     )
 
 
+def _evaluate_lockout_policies(context):
+    """
+    Run the conditional-access policy engine for this request's
+    classified outcome.
+
+    Called from check()'s finally, after the authentication-log row is written,
+    so a failure count over the log includes the just-written event. This only
```
jelinaunger, line 860: We have a special case for enroll via multichallenge where we adjust the auth log entry in a postpolicy (running after validate/check finally block) We need to ensure that this works correctly. It is not yet in this branch, so nothing to do now, just as a remark.

privacyidea/lib/conditional_access/engine.py
```diff
@@ -0,0 +1,763 @@
...
+
+def _resolved(user: "User") -> bool:
+    """
+    Return ``True`` only for a fully resolved user, i.e. one with a complete
+    ``(resolver, uid, realm)`` identity tuple. The lockout state and the
+    authentication-log count are both keyed by that tuple, so an unresolved user
+    (e.g. ``USER_UNKNOWN``, which has ``uid=None``) is never counted or locked
+    here.
+    """
+    return bool(user and user.uid and user.resolver and user.realm)
```
jelinaunger, line 124: there is an open pr also implementing this function in the user class. maybe add a todo here that we do not forget to replace it here to not have duplicated implementations

privacyidea/lib/conditional_access/engine.py
```diff
@@ -0,0 +1,763 @@
...
+        or ``None`` if permanent)
+    """
+    if not _resolved(user):
+        return None
+    state = db.session.get(UserLockoutState, (user.resolver, user.uid, user.realm))
+    if not state or not state.is_locked:
+        return None
+    if state.lock_expires_at is None:
+        # Permanent lock; only an admin reset clears it.
+        return {"permanent": True, "expires_at": None, "seconds_remaining": None}
```
jelinaunger: The return type could be a dataclass to make clear what exactly is returned  (can also be used for get_ip_block)

privacyidea/lib/conditional_access/engine.py
```diff
@@ -0,0 +1,763 @@
...
+    """
+    Build the ``{tag}`` substitution context available to EMAIL_* templates. Only
+    fields already loaded on the request are included here; the resolver-backed
+    user attributes (email, givenname, surname) are added lazily in
+    :func:`_send_lockout_email`, so a non-email action never triggers a resolver
+    lookup.
+    """
+    return {
+        "user": user.login,
+        "username": user.login,
```
jelinaunger, line 574: Why is the username here twice?

privacyidea/lib/conditional_access/engine.py
```diff
@@ -0,0 +1,763 @@
...
+        "count": count,
+        "threshold": stage.failure_threshold,
+        "event_type": event_type,
+        "stage_id": stage.id,
+        "policy": policy.name,
+        "time": now.strftime("%Y-%m-%d %H:%M:%S UTC"),
+    }
+
+
+def _resolve_admin_recipients(recipient_group) -> list[str]:
```
jelinaunger: type hint

privacyidea/lib/conditional_access/engine.py
```diff
@@ -0,0 +1,763 @@
...
+    """
+    custom = cfg.get("login_notice")
+    if custom:
+        return _safe_format(str(custom), render_tags)
+    if action_type == LockoutAction.EMAIL_USER:
+        return _("A notification email has been sent to your email address.")
+    return _("Your administrator has been notified by email.")
+
+
+def _send_lockout_email(action_type: "LockoutAction", action, user: "User", tags: dict) -> str | None:
```
jelinaunger: type hint for action is missing

privacyidea/lib/conditional_access/engine.py
```diff
@@ -0,0 +1,763 @@
...
+    optional ``login_notice`` (overrides the message surfaced on the login
+    screen) and, for EMAIL_ADMIN, an optional ``recipient_group``. EMAIL_USER
+    sends to the user's own email address. A missing field or a user without an
+    email address is logged and skipped; this runs post-response and must never
+    raise.
+
+    :return: the user-facing login-screen notice if the email was sent, else
+        ``None`` (misconfiguration, no recipient, or delivery failure).
+    """
+    cfg = action.action_value if isinstance(action.action_value, dict) else {}
```
jelinaunger: config?

jelinaunger: action seems to  hold some email configuration, hence action might not be the most suitable parameter name

privacyidea/models/lockout_policy.py
```diff
@@ -0,0 +1,182 @@
...
+    currently blocked, and a ``NULL`` value means a permanent block (only an
+    admin reset clears it). ``is_blocked`` lets an admin lift a block without
+    deleting the row, so ``last_stage_triggered`` (which records the stage that
+    produced the block, for de-duplication) is preserved. Timestamps are naive
+    UTC, see :func:`~privacyidea.models.utils.utc_now`.
+    """
+    __tablename__ = 'block_list'
+    # 50 matches authentication_log.source_ip, which is wide enough for an
+    # IPv4-mapped IPv6 address.
+    ip: Mapped[str] = mapped_column(Unicode(50), primary_key=True)
```
jelinaunger, line 172: for now we use the IP, however, in the future this could also hold some other identifier (device, API key, ..). That's why a generic id, entry type and value were suggested as columns

tests/test_api_authentication_log.py
```diff
@@ -105,7 +105,6 @@ def test_unknown_user_logs_user_unknown(self):
             res = self.app.full_dispatch_request()
             self.assertFalse(res.json["result"]["status"], res.json)
         entries = assert_authentication_log([AuthEventType.USER_UNKNOWN])
-        # TODO: What should be logged here for the user?
         assert_authentication_log_entry(entries[AuthEventType.USER_UNKNOWN], user=None,
```
jelinaunger, line 109: these were my todos to double-check this behavior. Anyway, as long as they are in my branch it's fine, but please keep them in the future as long as it is not the final PR.

tests/test_api_authentication_log.py
```diff
@@ -234,7 +230,6 @@ def test_challenge_answered_ok(self):
                                         user=User("cornelius", self.realm1), serial=self.serial)
 
     # --- Serial auth (serial provided instead of user) ---
-    # TODO: Serial should be added to logs (context) if passed as request parameter
```
jelinaunger, line 238: This is a long term todo which should stay as it might be out of scope for this issue

tests/test_api_conditional_access.py
```diff
@@ -0,0 +1,753 @@
...
+        return get_tokens(serial=self.serial)[0].token.failcount
+
+    # --- pre-check ------------------------------------------------------------
+
+    def test_locked_user_rejected_without_token_logic(self):
+        self._lock_user(utc_now() + timedelta(seconds=600))
+        self.assertEqual(0, self._failcount())
+
+        # Even valid credentials must be rejected while the user is locked.
+        body = self._check({"user": "cornelius", "pass": "pin755224"})
```
jelinaunger, line 138: just as safety check, add this also before locking the user to be sure these are the right credentials and it is the CA blocking authentication

jelinaunger: or as an additional success test or to be very precise in the setup function to run it before every test case

jelinaunger (commented):
Some parts are missing, but could also be done in a new PR if you prefer this
* lib functions and api endpoints to create / edit policies
* engine at /validate/polltransaction, /validate/triggerchallenge, and /ttype (for push)

privacyidea/lib/conditional_access/engine.py
```diff
@@ -0,0 +1,806 @@
...
+    Build the ``{tag}`` substitution context available to EMAIL_* templates. Only
+    fields already loaded on the request are included here; the resolver-backed
+    user attributes (email, givenname, surname) are added lazily in
+    :func:`_send_lockout_email`, so a non-email action never triggers a resolver
+    lookup.
+
+    Some values are exposed under two names on purpose (``user``/``username`` and
+    ``source_ip``/``client_ip``): an admin authoring a template may reach for
+    either spelling, so both resolve rather than silently leaving a ``{tag}``
+    unsubstituted.
```
jelinaunger: I get the idea. However, I would still prefer to have a single tag per value as it is more precise and avoids confusion. The problem you point here, is not fully solved by this solution. Users can still enter typos or use different wording. Instead, available tags should be displayed in the UI. Additionally, either the UI can evaluate the input and scan for invalid tags and block saving until the user fixed it (or could be done by the API on save). This would cover all invalid tags.

PR #5450, opened 2026-06-11
5431/remove sensitive data from api return

### LINKED ISSUE #5431: Remove sensitive data from api return
Remove sensitive data like AD password from api return.

jona-samuel:
fixing #5431 
replacing #5447


nilsbehlen:
we need to make the smtpservers declare which of their fields are secrets, but that is out of scope here.

nilsbehlen (approved):

PR #5456, opened 2026-06-12
Added Dashboard

### LINKED ISSUE #4696: Dashboard
Implement a dashboard. Might want to look what is displayed on the dashboard in the old webui and start from there.
- I think the dashboard would greatly benefit from #4220, maybe we can have some sort of slider to configure a timeframe (without saving it, so without having #4220).
- Look into options to create graphs?
- Additional Improvements:
      * ...

frankmer:


privacyidea/static_new/src/app/components/dashboard/widgets/authentications-widget/authentications-widget.component.html
```diff
@@ -0,0 +1,54 @@
...
+          <td>{{ counts().fail }}</td>
+        </tr>
+        <tr>
+          <td i18n>failed users / tokens</td>
+          <td class="auth-failed-cell">
+            @for (userobj of counts().users; track userobj.user + userobj.realm; let first = $first) {
+              @if (!first) {
+                <span class="auth-separator">|</span>
+              }
+              <a [routerLink]="[routePaths.USERS_DETAILS, userobj.user]">{{ userobj.user }}&#64;{{ userobj.realm }}[{{ userobj.fails }}]</a>
```
jelinaunger: the link does not work (I guess the realm is missing)

privacyidea/static_new/src/app/components/dashboard/widgets/widget-state/widget-state.component.html
```diff
@@ -0,0 +1,43 @@
...
+      <mat-progress-spinner
+        mode="indeterminate"
+        [diameter]="40"
+        aria-label="Loading widget data"
+        i18n-aria-label />
+    </div>
+  }
+  @case ("denied") {
+    <div class="widget-state-status">
+      <p i18n>You are not allowed to view this widget.</p>
```
jelinaunger, line 35: Do not show the widget at all if the user is not allowed to see it. (Same logic as for other action)

jelinaunger (commented):
reload all resources on refresh is missing

privacyidea/static_new/src/app/components/dashboard/widgets/widget-state/widget-state.component.html
frankmer: It should not be shown now. It's just a backup in case the rights check in the widget does not match the endpoint.

jelinaunger (approved):

PR #5461, opened 2026-06-15
reset failcount after fido auth, tests

### LINKED ISSUE #5340: Issue with reset_all_user_tokens
There is an issue when reset_all_user_tokens is set true there is no failcounter reset when user correctly logs in. I am using privacyidea credential provider for windows. After correctly logged in there is no reset. I tried with policy with other settings and with dedicated policy. None worked.

nilsbehlen:


privacyidea/lib/policydecorators.py
```diff
@@ -574,6 +574,40 @@ def config_lost_token(wrapped_function, *args, **kwds):
     return wrapped_function(*args, **kwds)
 
 
+def reset_all_user_tokens_if_policy(g, token_objects, user_object=None):
```
jelinaunger: type hints

privacyidea/lib/policydecorators.py
```diff
@@ -574,6 +574,40 @@ def config_lost_token(wrapped_function, *args, **kwds):
...
+        the owner of the first available token is used.
+    :return: True if a matching policy was found and the tokens were reset.
+    """
+    available_tokens = [tok for tok in token_objects if tok.get_class_type() not in ['registration']]
+    if not (g and available_tokens):
+        return False
+
+    token_owner = user_object or available_tokens[0].user
+    reset_all = Match.user(g, scope=SCOPE.AUTH, action=PolicyAction.RESETALLTOKENS,
+                           user_object=token_owner if token_owner else None).policies()
```
jelinaunger: policy is already checked before calling this function

jelinaunger (approved):

jelinaunger (approved):

PR #5474, opened 2026-06-17
5378/filter by multiple container types

### LINKED ISSUE #5378: allow to filter containers by multiple container types
The container list API should accept a list of container types as filter parameter.

This is useful if we want to get all containers a token could be added to, e.g. on token enrollment or token details page. Since some token types can be added to multiple container types we need a list.

Alternatively, we could also pass the token type to the api which will return all containers of a compatible type. This logic is actually implemented in the frontend.

henrik-falk:


privacyidea/api/container.py
```diff
@@ -128,8 +128,12 @@ def list_containers():
...
     :query container_serial: filter by container serial
         (case-insensitive, ``*`` wildcard).
-    :query type: filter by container type (case-insensitive, ``*``
-        wildcard).
+    :query type: filter by container type. Either a single type
+        (case-insensitive, ``*`` wildcard) or a list of types — passed
+        as a JSON array or a comma-separated string (e.g.
+        ``type=generic,smartphone``). List entries are matched
+        case-insensitively as exact values; wildcards are not honored
+        inside a list.
```
jelinaunger: for consistency with the token API this should be 2 separate parameters (type and type_list)

jelinaunger: you can add a todo that both parameters should be replaced with types (list only parameter) in the 4.0.0 release

privacyidea/lib/container.py
```diff
@@ -178,7 +179,9 @@ def _create_container_query(user: User = None, serial: str = None, ctype: str =
 
     :param user: container owner, optional
     :param serial: container serial (case-insensitive and allows '*' as wildcard), optional
-    :param ctype: container type (case-insensitive and allows '*' as wildcard), optional
+    :param ctype: container type filter, optional. May be either a string (case-insensitive, allows '*' as wildcard)
+        or a list of strings (case-insensitive, exact match; matches any type in the list). An empty list returns
+        no matches.
```
jelinaunger: since they also behave differently it might make sense to split the parameters here as well in something like ctype_like: str, ctype_exact: list[str].

tests/test_api_container_api.py
```diff
@@ -759,6 +759,81 @@ def test_21_get_all_containers_paginate(self):
...
+            containers = result["result"]["value"]["containers"]
+            return {c["serial"] for c in containers}, {c["type"] for c in containers}
+
+        try:
+            # The scalar ``type`` param still works (regression check)
+            serials, types = serials_and_types({"type": "smartphone"})
+            self.assertIn(smartphone_serial, serials)
+            self.assertNotIn(generic_serial, serials)
+            self.assertNotIn(yubikey_serial, serials)
+            self.assertTrue(types.issubset({"smartphone"}))
```
jelinaunger: This is already covered by the previous test

tests/test_api_container_api.py
```diff
@@ -759,6 +759,81 @@ def test_21_get_all_containers_paginate(self):
...
+
+            # ... including its ``*`` wildcard
+            serials, types = serials_and_types({"type": "smart*"})
+            self.assertIn(smartphone_serial, serials)
+            self.assertNotIn(generic_serial, serials)
+            self.assertTrue(types.issubset({"smartphone"}))
```
jelinaunger: According to the test name this does not belong in this test, but that's neat

jelinaunger (approved):

privacyidea/lib/container.py
```diff
@@ -223,8 +224,13 @@ def _create_container_query(user: User = None, serial: str = None, ctype: str |
         else:
             stmt = stmt.where(func.upper(TokenContainer.serial) == serial.upper())
 
-    if isinstance(ctype, list):
-        type_values = [str(t).strip() for t in ctype if t and str(t).strip()]
+    # Backward compatibility: callers may still pass list values through ``ctype``.
```
jelinaunger: you are introducing lists in this PR so there should no list be passed here. Keep it clean or keep the one parameter solution imo.

PR #5497, opened 2026-06-22
refactor machine resolvers

### LINKED ISSUE #5402: Refactor Machine Resolvers
Refactor to have a table as overview and dedicated create new and details / edit page

frankmer:


privacyidea/static_new/src/app/components/machine-resolver/machine-resolver.component.html
```diff
@@ -17,17 +17,120 @@
...
+            aria-label="Filter Input"
+            i18n-aria-label
+            i18n-placeholder
+            matInput
+            placeholder="Filter by name or type" />
+        </app-clearable-input>
+      </mat-form-field>
+      <mat-paginator
+        [pageSizeOptions]="tableUtilsService.pageSizeOptions()"
+        showFirstLastButtons />
```
jelinaunger: No need for pagination here

privacyidea/static_new/src/app/components/machine-resolver/machine-resolver-details/machine-resolver-details.component.html
```diff
@@ -0,0 +1,138 @@
...
+        class="margin-8"
+        i18n>
+        Edit Machine Resolver
+      </h3>
+    } @else {
+      <h3
+        class="margin-8"
+        i18n>
+        Machine Resolver Details
+      </h3>
```
jelinaunger: First landing on a readonly page where inputs are just disabled is different from everything else we have in the UI. Two suggestions:
* directly rout to edit mode if the user has write rights (same behavior as for other config pages). And only in case of read only rights route to a read only mode.
* Disabled input fields are a bit confusing imo. Add a cleaner read only view, e.g. similar to user details. (Might be too much effort for less gain if previous suggestion is realized and coming to the read only page might happen rarely)

privacyidea/static_new/src/app/admin.routes.ts
```diff
@@ -146,7 +147,14 @@ export const routes: Routes = [
     path: "configuration",
     children: [
       // { path: "", component: SystemComponent },
-      { path: "machine_resolver", component: MachineResolverComponent, canDeactivate: [pendingChangesGuard] },
+      {
+        path: "machine_resolver",
+        children: [
+          { path: "", component: MachineResolverComponent },
+          { path: "new", component: MachineResolverDetailsComponent, canDeactivate: [pendingChangesGuard] },
+          { path: "details/:name", component: MachineResolverDetailsComponent, canDeactivate: [pendingChangesGuard] }
```
jelinaunger, line 155: also add menu entries for the new routes

privacyidea/static_new/src/app/components/machine-resolver/machine-resolver-details/machine-resolver-details.component.html
frankmer: directly editable now when you have mresolverwrite; only read-only rights land on the disabled/read-only view. Went with suggestion 1; skipping the separate read-only view for now since, as you noted, that page should rarely be hit.

jelinaunger (approved):

privacyidea/static_new/src/app/components/machine-resolver/machine-resolver-details/machine-resolver-details.component.ts
```diff
@@ -0,0 +1,253 @@
...
+    effect(() => {
+      const name = this.selectedName();
+      if (!name) {
+        this._loadedName = null;
+        return;
+      }
+      if (untracked(() => this.isEdited())) {
+        return;
+      }
+      const resource = this.machineResolverService.machineResolverResource;
```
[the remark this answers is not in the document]
frankmer: There is no way to navigate from one machine resolver details page to another except by editing the browser URL, which triggers a full page reload. To prevent accidentally losing changes in that case, the pending-changes service now uses the browser's "beforeunload" hook, which shows a native browser dialog (not an Angular one). There is no other way to handle this.

PR #5501, opened 2026-06-22
Webui language switcher

nilsbehlen:
Adds a UI language switcher to the new WebUI, makes the session survive the full-page reload it requires, and cleans up what gets persisted to browser storage.

  Features

  - Language switcher in the user-utils panel (next to the theme and refresh controls), available for both admin and self-service layouts. Lists all 13 locales by their
  native name (Deutsch, Français, 繁體中文, …) with the current one marked.
  - Selecting a language stores the choice in a pi_ui_locale cookie and navigates to that locale's compiled bundle, preserving the current page (route + query string), so
  you stay where you were instead of being sent back to a landing page.
  - The choice persists across reloads and deep links: the backend resolves the UI locale as cookie → Accept-Language header → default, on the / route and the SPA
  fallback.

  Fixes (exposed by the switcher's full reload)

  - Session no longer drops on reload. Previously only the bearer token was persisted, so any full reload (language switch or a plain F5) reset the in-memory auth state
  and bounced the user to the login page. The session is now rehydrated on bootstrap while the stored JWT is still valid.
  - Authenticated users are kept off the login page. A new route guard forwards an already-authenticated user from /login to their landing page, instead of showing the
  login form with a spurious "already logged in" warning.

  Storage cleanup

  - The persisted auth data no longer duplicates anything the JWT already carries (token, rights, role, username, realm). The bearer token remains the single source of
  truth for identity and rights; only UI/policy config (menus, page sizes, wizards, defaults, …) is stored alongside it. This removes redundant copies of the (potentially
  large) rights list and the token itself.

  Notes

  - UI translations are compiled per-locale at build time, so switching languages is inherently a full page load of a different bundle — the cookie is what lets the
  language-agnostic / redirect honor the user's choice.
  - The cookie is designed as a drop-in seam for a future per-user language setting (/user/settings): the precedence would simply become stored setting → cookie → 
  Accept-Language → default, with no rework to this code.

  Testing

  - Frontend unit suite green (incl. new tests for the switcher, route preservation, session rehydration, the login guard, and the storage dedup).
  - Production and development builds green; backend syntax-checked.

### DIFF
--- a/privacyidea/static_new/src/app/components/login/login.component.ts
+++ b/privacyidea/static_new/src/app/components/login/login.component.ts
@@ -43,7 +43,6 @@ import { ClearButtonComponent } from "@components/shared/clear-button/clear-butt
 import { environment } from "@env/environment";
 import { AuthResponse, AuthService, AuthServiceInterface, PasswordLoginParams } from "@services/auth/auth.service";
 import { ConfigService } from "@services/config/config.service";
-import { LocalService, LocalServiceInterface } from "@services/local/local.service";
 import { NotificationService, NotificationServiceInterface } from "@services/notification/notification.service";
 import { SessionTimerService, SessionTimerServiceInterface } from "@services/session-timer/session-timer.service";
 import { ValidateService, ValidateServiceInterface, WebAuthnSignRequest } from "@services/validate/validate.service";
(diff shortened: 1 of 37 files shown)


privacyidea/static_new/src/app/services/auth/auth.service.ts
```diff
@@ -336,6 +336,60 @@ export class AuthService implements AuthServiceInterface {
...
+  constructor() {
+    this.restoreSession();
+  }
+
+  /**
+   * Rehydrate the session from storage on bootstrap so a full page reload (e.g. switching
+   * the UI language, which loads a different locale bundle) does not drop an active login.
+   * The token and the auth data are restored only while the JWT is still valid; an
+   * expired or corrupt session is cleared instead.
+   */
```
jelinaunger: I think this comment should be before restoreSession()?

privacyidea/static_new/src/app/components/login/login.component.ts
```diff
@@ -235,9 +227,8 @@ export class LoginComponent implements OnDestroy, AfterViewInit {
   }
 
   logout(): void {
+    // AuthService.logout() already clears stored session keys and navigates to /login.
```
jelinaunger: While these comments are nice for the review, I think they don't gain anything without the context of this PR

jelinaunger (approved):

PR #5511, opened 2026-06-25
4914/conditional access m1 lib and api

### LINKED ISSUE #5490: Conditional Access: M1 - lib functions and api endpoints
Some parts are missing, but could also be done in a new PR if you prefer this

-     lib functions and api endpoints to create / edit policies
-     engine at /validate/polltransaction, /validate/triggerchallenge, and /ttype (for push)

henrik-falk:


privacyidea/api/conditional_access.py
```diff
@@ -0,0 +1,247 @@
...
+        counter_types_to_track=_get_json_param(params, "counter_types_to_track", required=True),
+        stages=_get_json_param(params, "stages", required=True),
+        enabled=is_true(enabled) if enabled is not None else True,
+        dry_run=is_true(dry_run) if dry_run is not None else False,
+        priority=get_optional(params, "priority", default=1))
+    g.audit_object.log({"success": True, "info": f"created policy '{name}' (id {policy_id})"})
+    return send_result(policy_id)
+
+
+@conditional_access_blueprint.route('policy/<policy_id>', methods=['POST'])
```
jelinaunger: this would be PATCH (only update given values and leave remaining ones unchanged)

privacyidea/api/conditional_access.py
```diff
@@ -0,0 +1,247 @@
...
+    """
+    delete_lockout_policy(_int_policy_id(policy_id))
+    g.audit_object.log({"success": True, "info": f"deleted policy {policy_id}"})
+    return send_result(policy_id)
+
+
+@conditional_access_blueprint.route('policy/<policy_id>/enable', methods=['POST'])
+@prepolicy(check_base_action, request, PolicyAction.CONDITIONAL_ACCESS_WRITE)
+@log_with(log)
+def enable_policy(policy_id):
```
jelinaunger: enable and disable are somehow duplicates of the update endpoint. You can achieve exactly the same behaviour there

privacyidea/api/conditional_access.py
```diff
@@ -0,0 +1,247 @@
...
+    update_lockout_policy(
+        policy_id,
+        name=get_optional(params, "name"),
+        time_window_seconds=get_optional(params, "time_window_seconds"),
+        counter_types_to_track=_get_json_param(params, "counter_types_to_track"),
+        stages=_get_json_param(params, "stages"),
+        enabled=is_true(enabled) if enabled is not None else None,
+        dry_run=is_true(dry_run) if dry_run is not None else None,
+        priority=get_optional(params, "priority"))
+    g.audit_object.log({"success": True, "info": f"updated policy {policy_id}"})
```
jelinaunger: to be more precise we can add the fields which were updated here

tests/test_api_conditional_access.py
```diff
@@ -811,3 +959,181 @@ def test_lockout_message_and_email_notice_combined(self):
             self.assertTrue(is_user_locked(self.user))
         finally:
             delete_smtpserver("lockoutmail")
+
+
+class ConditionalAccessPolicyApiTestCase(MyApiTestCase):
```
jelinaunger: this can go in an own test file

jelinaunger: your testing multiple endpoints and cases within single tests. I think it is better to have an own test case for each endpoint x case. Hence, on failure it is already clear what went wrong just from the test name. It also improves readability to better get an overview which usecases are tested here for which endpoint.

privacyidea/lib/conditional_access/lockout_policy.py
```diff
@@ -0,0 +1,371 @@
...
+    its stages and actions into the plain-dict shape documented in the module
+    docstring (plus the ``id`` of each row).
+    """
+    return {
+        "id": policy.id,
+        "name": policy.name,
+        "time_window_seconds": policy.time_window_seconds,
+        "enabled": policy.enabled,
+        "dry_run": policy.dry_run,
+        "priority": policy.priority,
```
jelinaunger: up to here could be simplified to

 
`result = {name: getattr(policy, name) for name in policy.__table__.columns.keys()}`

only counter_types_to_track and stages needs to be added manually

privacyidea/lib/conditional_access/lockout_policy.py
```diff
@@ -0,0 +1,371 @@
...
+        raise ParameterError("'counter_types_to_track' must be a non-empty list of "
+                             "authentication event types.")
+    valid_types = {event_type.value for event_type in AuthEventType}
+    seen = []
+    for counter_type in counter_types:
+        if counter_type not in valid_types:
+            raise ParameterError(f"Unknown counter type '{counter_type}'. "
+                                 f"Valid types: {', '.join(sorted(valid_types))}.")
+        if counter_type in seen:
+            raise ParameterError(f"Duplicate counter type '{counter_type}'.")
```
jelinaunger: not sure if we really need to raise an error here, cause what would happen if there is a duplicated counter type in the list? I guess it does not differ at all. Only if the user entered the same value twice due to a copy paste mistake, it might be nice to inform him/her

privacyidea/lib/conditional_access/lockout_policy.py
```diff
@@ -0,0 +1,371 @@
...
+    here and a :class:`ParameterError` is raised on any invalid input before
+    anything is written.
+
+    :return: the id of the new policy
+    """
+    name = _validate_name(name)
+    time_window_seconds = _validate_positive_int(time_window_seconds, "time_window_seconds")
+    priority = _validate_positive_int(priority, "priority")
+    counter_types = _validate_counter_types(counter_types_to_track)
+    stage_dicts = _validate_stages(stages)
```
jelinaunger: you could introduce a stages dataclass returned by the validate function which is handed in the _build_stages class. Just to use more clear type hints and ensure the data structure is as expected and not any list of dicts.

privacyidea/lib/conditional_access/lockout_policy.py
```diff
@@ -0,0 +1,371 @@
...
+    if enabled is not None:
+        policy.enabled = bool(enabled)
+    if dry_run is not None:
+        policy.dry_run = bool(dry_run)
+    if counter_types_to_track is not None:
+        policy.counter_types_to_track = counter_types_to_track
+    if stages is not None:
+        policy.stages = _build_stages(stages)
+    db.session.commit()
+    log.info(f"Updated lockout policy '{policy.name}' (id {policy.id}).")
```
jelinaunger: same as for the audit, you could add the fields which were updated here

privacyidea/api/ttype.py
```diff
@@ -140,6 +161,14 @@ def token(ttype=None):
             code_to_phone_message = list(policies)[0]
         request.all_data[PushAction.PUSH_CODE_TO_PHONE_MESSAGE] = code_to_phone_message
 
+        # Conditional-access pre-check (push only): reject the smartphone's poll /
+        # answer when the token owner is locked, the source IP is blocked, or a DENY
+        # policy applies. Generic failure, reason recorded only in the audit log.
+        push_owner = _push_token_owner(g.serial)
+        rejection = conditional_access_precheck(push_owner)
+        if rejection is not None:
+            return rejection
```
jelinaunger: this must be moved into _api_endpoint_post method of the pushtoken. This endpoint is also called for enrollment. But the conditional access precheck should only run on the auth path

privacyidea/api/validate.py
```diff
@@ -477,8 +501,10 @@ def check():
     """
     # Conditional-access pre-check: reject a locked user, a blocked source IP, or a
     # policy DENY decision before any token logic (generic failure response, reason
-    # recorded only in the audit log; see the helper for details).
-    rejection = _conditional_access_precheck(request.User)
+    # recorded only in the audit log; see the helper for details). The identity is
+    # resolved up front (see _conditional_access_identity) so a username-less passkey
+    # (credential_id) or serial-only request is gated on the real token owner.
+    rejection = conditional_access_precheck(_conditional_access_identity())
```
jelinaunger: I think we should move this into a decorator which runs first before every other pre-policy checks

jelinaunger: same for the other endpoints

privacyidea/cli/pimanage/conditional_access.py
```diff
@@ -0,0 +1,124 @@
...
+
+conditional_access_cli = AppGroup("conditionalaccess",
+                                  help="Inspect and clear conditional-access locks and IP blocks")
+
+
+def _format_expiry(expires_at):
+    return expires_at.isoformat() if expires_at else "permanent"
+
+
+@conditional_access_cli.command("list-blocks", help="List the currently blocked IPs.")
```
jelinaunger: lict-blocks and list-locks is quite similar. list-blocked-ips and list-locked-users is more clear (also for clear)

privacyidea/cli/pimanage/conditional_access.py
```diff
@@ -0,0 +1,124 @@
...
+    if not row:
+        click.echo(f"No block found for IP {ip}.")
+        return
+    db.session.delete(row)
+    db.session.commit()
+    click.echo(f"Removed the block for IP {ip}.")
+
+
+@conditional_access_cli.command("clear-blocks", help="Remove ALL IP blocks.")
+@click.confirmation_option(prompt="Remove all IP blocks?")
```
jelinaunger, line 65: might it make sense to add any conditions here? Is there something useful we could add as filter here?

privacyidea/cli/pimanage/conditional_access.py
```diff
@@ -0,0 +1,124 @@
...
+    if not row:
+        click.echo(f"No lock found for (resolver={resolver}, uid={uid}, realm={realm}).")
+        return
+    db.session.delete(row)
+    db.session.commit()
+    click.echo(f"Unlocked (resolver={resolver}, uid={uid}, realm={realm}).")
+
+
+@conditional_access_cli.command("clear-locks", help="Remove ALL user locks.")
+@click.confirmation_option(prompt="Remove all user locks?")
```
jelinaunger: this could get at least realm as option

privacyidea/lib/policies/actions.py
```diff
@@ -227,6 +227,8 @@ class PolicyAction:
     PASSKEY_LOGIN = "passkey_login"
     HIDE_VERSION = "hide_version"
     AUTHENTICATION_LOG_READ = "authentication_log_read"
+    CONDITIONAL_ACCESS_READ = "conditional_access_read"
+    CONDITIONAL_ACCESS_WRITE = "conditional_access_write"
```
jelinaunger: this is only for the lockout policies right? So the name could be more explicit, e.g. conditional_access_policy_read / lockout_policy_read / ...

privacyidea/static_new/src/app/components/login/login.component.scss
```diff
@@ -22,15 +22,15 @@ $border-radius: var(--border-radius);
   }
 
   .logo {
-    width: 8%;
+    width: 150px;
```
jelinaunger, line 25: we need to ensure that this also works for custom logos. Just to keep in mind, but is out of scope here.

privacyidea/static_new/src/app/components/login/login.component.scss
```diff
@@ -100,4 +100,12 @@ $border-radius: var(--border-radius);
...
     width: 300px;
   }
+
+  // The global .margin-top-16 utility adds 16px above the action row, reset and
+  // passkey buttons; drop it on the login page only (utility stays global).
+  .login-actions.margin-top-16,
+  .reset-btn.margin-top-16,
+  .passkey-btn.margin-top-16 {
+    margin-top: 0;
+  }
```
jelinaunger, line 110: are these styles applied anywhere else than on the login page?

tests/test_api_conditional_access.py
```diff
@@ -443,6 +445,152 @@ def test_deny_with_lower_threshold_shadows_lock_policy(self):
...
+                                           data={"user": "cornelius"},
+                                           headers={"Authorization": self.at}, **kwargs):
+            response = self.app.full_dispatch_request()
+            self.assertEqual(200, response.status_code, response)
+            return response.json
+
+    def test_triggerchallenge_locked_user_rejected(self):
+        self._lock_user(utc_now() + timedelta(seconds=600))
+        body = self._trigger_challenge()
+        # Generic failure (no challenge triggered) and no token logic ran.
```
jelinaunger, line 465: to be sure, check that there is really no challenge created in the db, even if no transaction id is returned

tests/test_api_conditional_access.py
```diff
@@ -443,6 +445,152 @@ def test_deny_with_lower_threshold_shadows_lock_policy(self):
...
+
+    def test_triggerchallenge_no_token_event_feeds_engine(self):
+        # With no challenge-capable token, triggering classifies NO_TOKEN; a policy
+        # tracking NO_TOKEN locks the user via the post-eval seam.
+        remove_token(self.serial)
+        self._make_lock_policy(counter_type=AuthEventType.NO_TOKEN, threshold=1, duration=600)
+        self.assertFalse(is_user_locked(self.user))
+        body = self._trigger_challenge()
+        self.assertEqual(0, body["result"]["value"], body)
+        self.assertEqual([AuthEventType.NO_TOKEN],
```
jelinaunger: use assertListEqual for clearer fail messages

tests/test_api_conditional_access.py
```diff
@@ -443,6 +445,152 @@ def test_deny_with_lower_threshold_shadows_lock_policy(self):
...
+                         [entry.event_type for entry in get_authentication_logs()])
+        self.assertTrue(is_user_locked(self.user))
+
+    # --- /validate/polltransaction --------------------------------------------
+
+    def _poll(self, transaction_id, remote_addr=None):
```
jelinaunger: to be kind, also use type hints in tests

tests/test_api_conditional_access.py
```diff
@@ -443,6 +445,152 @@ def test_deny_with_lower_threshold_shadows_lock_policy(self):
...
+        body = self._ttype_push({"serial": self.serial}, remote_addr="203.0.113.7")
+        self.assertFalse(body["result"]["value"], body)
+        self.assertEqual(0, len(get_authentication_logs()))
+
+    # --- serial-only lock-evasion (resolve owner before the pre-check) ---------
+
+    def test_locked_user_rejected_via_serial(self):
+        # Regression: a locked user could authenticate by sending serial=... instead
+        # of user=..., because request.User was empty at the pre-check. The owner is
+        # now resolved from the serial before the pre-check, so the lock is enforced.
```
jelinaunger: remove the regression comment, just state the current behaviour

privacyidea/lib/token/auth.py
```diff
@@ -399,6 +399,7 @@ def check_token_list(token_object_list: list[TokenClass], passw: str, user: User
                 if otp_count >= 0:
                     # This is a successful authentication
                     valid_token_list.append(token_object)
+                    request_events.append(AuthEventType.LOGIN_SUCCESS)
```
jelinaunger: this is a duplicate of the final classification within valid_token_list check

privacyidea/api/conditional_access.py
henrik-falk: i thought redundant api endpoints could be helpful but nvm

privacyidea/cli/pimanage/conditional_access.py
henrik-falk: IP blocks have no realm/user grouping, and unblock-ip already handles single IPs. I cannot think of any other filter.

privacyidea/static_new/src/app/components/login/login.component.scss
henrik-falk: fixed pixel width should be more reliable. but yes check on custom logos is needed.

privacyidea/static_new/src/app/components/login/login.component.scss
henrik-falk: yes its nested under .login-container

jelinaunger (approved):

PR #5537, opened 2026-06-30
Enforce owner/param-scoped PIN policies when request.User is empty

### LINKED ISSUE #5536: PIN-quality policies silently skipped when request.User is empty (check_otp_pin by serial; init_random_pin OTPPINCONTENTS)
## Summary

In two prepolicy functions, a **PIN-quality policy can be silently skipped** because an *empty* `User` (not `None`) is passed into the policy match. The policy engine treats an empty user as "match only unscoped policies", so any realm/user-scoped PIN policy is dropped. This is a **PIN-quality enforcement gap**, not an authorization bypass (the action gates themselves remain owner-derived and fail-closed).

## Background: empty `User` vs `None`

`get_user_from_param()` never returns `None`; with no user in the request it returns an **empty** `User(login="", realm="")`. In `match_policies`:

| Passed in | Result |
|---|---|
| real user | unscoped + that user's/realm's scoped policies (normal) |
| `None` | filter disabled → **all** policies match |
| empty `User()` | filter active on `""` → **only unscoped** policies match; user/realm-scoped policies silently dropped |

So a decorator that means "no user → match all" but forwards the empty `User()` instead of `None` silently skips every scoped policy.

**Note:** the broad "normalize empty → `None` in the Match layer" fix was tried and **rejected** — it regressed ownerless-token semantics (`tests/test_lib_policy.py` `test_06_token`), which intentionally rely on empty → only-unscoped. This issue is only about the two concrete spots where the empty user is the *wrong* user to scope by.

## Issue 1 — `check_otp_pin` (admin setting a PIN by serial)

`privacyidea/api/lib/prepolicy.py` `check_otp_pin` passed `request.User` into the fail-open `check_pin`. When an admin sets a PIN on an existing token **by serial** with no `user` param, `request.User` is empty, so a realm/user-scoped **PIN-quality** policy that applies to the **token owner** is silently skipped.

**Fix:** derive the PIN-policy user from the token owner (`tokensobject_list[0].user`), falling back to `request.User` only for unowned/enrollment tokens.

## Issue 2 — `init_random_pin` (OTPPINCONTENTS during enrollment)

In the same file, `init_random_pin` resolves `OTPPINRANDOM` and `PINHANDLING` via `user_object` (from `get_user_from_param(params)`) but resolved `OTPPINCONTENTS` via `request.User`. When `request.User` diverges from the params user (e.g. a realm injected by a preceding `realmadmin` decorator, or an empty `request.User`), the user/realm-scoped contents policy is skipped and a generic random PIN is generated instead of one matching the contents requirement.

**Fix:** resolve `OTPPINCONTENTS` with the same `user_object` the other enrollment PIN lookups use, so all of an enrollment's PIN policies scope the user consistently. (Lower severity than #1 — it generates a PIN rather than validating one — but it's an inconsistency with no justification.)

## Scope / severity

Neither is an authz bypass: the `setpin` gate is owner-derived and fail-closed, and `init_random_pin` only generates a PIN. Both are **PIN-quality enforcement gaps**.

## Status (branch `empty-user-policy-matching`, off master)

Both fixes implemented with tests; **uncommitted**, ready for a PR:

- `privacyidea/api/lib/prepolicy.py` — `check_otp_pin` (owner-derived user) + `init_random_pin` (`OTPPINCONTENTS` via `user_object`).
- `tests/test_api_lib_policy_pin.py`:
  - `test_09b_pin_policies_admin_by_serial` — realm-scoped PIN policy enforced for an admin acting by serial with empty `request.User`.
  - `test_09c_random_pin_contents_uses_param_user` — `OTPPINCONTENTS` applied via the param user when `request.User` is empty (verified to fail on the old code, pass with the fix).

Tests green: pin file (8), and previously enroll (28) / token (77).

## Acceptance criteria

- An admin setting a PIN by serial (no `user` param) has the token owner's realm/user-scoped PIN-quality policy enforced.
- Random-PIN enrollment applies a realm/user-scoped `OTPPINCONTENTS` policy resolved from the enrollment user, regardless of `request.User`.
- No regression to ownerless-token semantics (`test_06_token`).

nilsbehlen:
check_otp_pin: derive the PIN-quality policy user from the token owner, so an admin setting a PIN on an existing token by serial (no user param -> empty request.User) still has the owner's realm/user-scoped PIN policy enforced instead of silently matching only unscoped policies.

init_random_pin: resolve OTPPINCONTENTS via the enrollment user_object (the same user the OTPPINRANDOM/PINHANDLING lookups already use) instead of request.User, so a realm/user-scoped contents policy is applied consistently when request.User diverges from the params user.

Neither is an authz bypass (the setpin gate is owner-derived and fail-closed, and init_random_pin only generates a PIN) - both are PIN-quality enforcement gaps. Adds test_09b_pin_policies_admin_by_serial and test_09c_random_pin_contents_uses_param_user.


privacyidea/api/lib/prepolicy.py
```diff
@@ -346,9 +351,17 @@ def check_otp_pin(request=None, action=None):
...
                 return True
             tokentype = tokensobject_list[0].token.tokentype
+            # When no user was supplied in the request (e.g. an admin acting on an existing token
+            # by serial), request.User is empty and only matches policies without a user/realm,
+            # silently skipping a user- or realm-scoped PIN policy that applies to the token owner.
+            # Fall back to the token owner in that case. If the request did carry a user, that
+            # explicit user is honored.
+            token_owner = tokensobject_list[0].user
+            if token_owner and not request.User:
+                pin_user = token_owner
```
jelinaunger, line 360: could be simplified: 

pin_user = pin_user or tokensobject_list[0].user

nilsbehlen: token.user is None for an unowned token, so pin_user or token.user would pass None into check_pin — which disables the user filter and matches all PIN policies. The if 
  token_owner and not request.User guard keeps an unowned-token/by-serial set at "only unscoped policies"

jelinaunger (approved):

PR #5563, opened 2026-07-06
Conditional Access M3 - Policy Editor UI

### LINKED ISSUE #5494: Conditional Access: M3 - Policy Editor UI
* list
* edit / create

henrik-falk:


privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/stages-list/stage-item/actions-list/action-item/conditional-access-action-item.component.html
```diff
@@ -0,0 +1,59 @@
...
+    subscriptSizing="dynamic"
+    class="input-width-xl ca-action-value">
+    <mat-label i18n>Action value (JSON)</mat-label>
+    <textarea
+      matInput
+      rows="3"
+      i18n-placeholder
+      placeholder="e.g. {{ '{' }}&quot;lock_duration_seconds&quot;: 600{{ '}' }}"
+      [value]="actionValueText()"
+      (input)="onActionValueInput($any($event.target).value)"></textarea>
```
jelinaunger: A generic json free text input field is bad ux. Input fields should adopt according to the selected action. Either hardcoded in the UI or it might also make sense to add an api endpoint returning a descriptor which input fields are available for which action of which type, etc.

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/stages-list/stage-item/actions-list/action-item/conditional-access-action-item.component.html
```diff
@@ -0,0 +1,59 @@
...
+  <mat-form-field
+    appearance="fill"
+    subscriptSizing="dynamic"
+    class="input-width-xl ca-action-value">
+    <mat-label i18n>Action value (JSON)</mat-label>
+    <textarea
+      matInput
+      rows="3"
+      i18n-placeholder
+      placeholder="e.g. {{ '{' }}&quot;lock_duration_seconds&quot;: 600{{ '}' }}"
```
jelinaunger: by the way this type hint is wrong, it is only duration_seconds or duration (why are there actually two possible keys for the same action value?)

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/stages-list/stage-item/actions-list/conditional-access-actions-list.component.scss
```diff
@@ -0,0 +1,25 @@
+@use "styles";
+
+.ca-actions-list {
+  display: flex;
+  flex-direction: column;
+  padding-left: 16px;
+  border-left: 2px solid rgba(0, 0, 0, 0.08);
```
jelinaunger: use the CSS variables e.g., --mat-sys-outline-variant for proper theming

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/stages-list/stage-item/conditional-access-stage-item.component.html
```diff
@@ -0,0 +1,62 @@
...
+ GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+
+ You should have received a copy of the GNU Affero General Public
+ License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+ SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<div class="ca-stage-item">
+  <div class="ca-stage-header">
+    <span class="ca-stage-title" i18n>Stage {{ stageNumber() }}</span>
```
jelinaunger: maybe we should allow users to define a name for th stage. Or we do not show a title at all and handle this as a list. But the numbering is not useful at all and can be confusing with the priority

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/stages-list/stage-item/conditional-access-stage-item.component.html
```diff
@@ -0,0 +1,62 @@
...
+      appearance="fill"
+      subscriptSizing="dynamic"
+      class="input-width-s">
+      <mat-label i18n>Failure threshold</mat-label>
+      <input
+        matInput
+        type="number"
+        min="1"
+        [value]="stage().failure_threshold"
+        (input)="onFailureThresholdInput($any($event.target).value)" />
```
jelinaunger, line 85: add hint, also to other input fields as it might not be fully clear what they are about

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/stages-list/stage-item/conditional-access-stage-item.component.html
```diff
@@ -0,0 +1,62 @@
...
+      appearance="fill"
+      subscriptSizing="dynamic"
+      class="input-width-s">
+      <mat-label i18n>Stage priority</mat-label>
+      <input
+        matInput
+        type="number"
+        min="1"
+        [value]="stage().priority"
+        (input)="onPriorityInput($any($event.target).value)" />
```
jelinaunger: especially here a hint is usefully on the order of priorities. (we might need to change it as in the normal policies lower numbers have higher priority, while here it is the other way around and it should be uniform, but this might be out of scope here)

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/stages-list/stage-item/conditional-access-stage-item.component.scss
```diff
@@ -0,0 +1,23 @@
+.ca-stage-item {
+  border: 1px solid rgba(0, 0, 0, 0.12);
```
jelinaunger: same here for the colour variable

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/stages-list/conditional-access-stages-list.component.html
```diff
@@ -0,0 +1,40 @@
...
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+
+ You should have received a copy of the GNU Affero General Public
+ License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+ SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<div class="ca-stages-list">
```
jelinaunger: a short explanation/hint on what a stage is might be helpful

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/stages-list/conditional-access-stages-list.component.ts
```diff
@@ -0,0 +1,56 @@
...
+import { MatButtonModule } from "@angular/material/button";
+import { MatIconModule } from "@angular/material/icon";
+import { LockoutPolicyStage } from "@services/conditional-access/conditional-access-policy.service";
+import { ConditionalAccessStageItemComponent } from "./stage-item/conditional-access-stage-item.component";
+
+const NEW_STAGE: LockoutPolicyStage = { failure_threshold: 1, priority: 1, actions: [] };
+
+// Stages (and their nested actions) are edited via a plain parent-owns-array/child-emits-partial
+// composition, not Angular Signal Forms: applyEach() (the Signal Forms array API) is
+// @experimental, has no push/remove primitive (form() does not copy the model, so the array must
```
jelinaunger: why this comment about a feature that is not used anyway?

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/conditional-access-edit-page.component.html
```diff
@@ -0,0 +1,162 @@
...
+        appearance="fill"
+        class="input-width-l"
+        subscriptSizing="dynamic">
+        <mat-label i18n>Name</mat-label>
+        <input
+          matInput
+          [formField]="policyForm.name"
+          (blur)="nameTouched.set(true)" />
+        @if (showNameError()) {
+          <mat-error i18n>Name is required (max. 255 characters).</mat-error>
```
jelinaunger: max length could also be set on the input with a small hint <num_entered>/255 as we have it already for the token description I think

jelinaunger: layout jumps if error appears

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/conditional-access-edit-page.component.html
```diff
@@ -0,0 +1,162 @@
...
+        }
+      </mat-form-field>
+
+      <mat-form-field
+        appearance="fill"
+        class="input-width-s"
+        subscriptSizing="dynamic">
+        <mat-label i18n>Time window (seconds)</mat-label>
```
jelinaunger: for a better user experience it might be nice to allow also other units (at least minutes)

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/conditional-access-edit-page.component.html
```diff
@@ -0,0 +1,162 @@
...
+          [value]="editPolicy().priority"
+          (input)="onPriorityInput($any($event.target).value)" />
+        @if (!priorityValid()) {
+          <mat-error i18n>Must be at least 1.</mat-error>
+        }
+      </mat-form-field>
+
+      <mat-form-field
+        appearance="fill"
+        class="input-width-l"
```
jelinaunger: this is too small, one option is truncated

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/stages-list/stage-item/actions-list/action-item/conditional-access-action-item.component.html
```diff
@@ -0,0 +1,59 @@
...
+ You should have received a copy of the GNU Affero General Public
+ License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+ SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<div class="ca-action-item">
+  <mat-form-field
+    appearance="fill"
+    subscriptSizing="dynamic"
+    class="input-width-s">
```
jelinaunger: it is too small some values are truncated

privacyidea/static_new/src/app/components/conditional-access/conditional-access.component.html
```diff
@@ -0,0 +1,328 @@
...
+  appScrollToTop
+  aria-label="Conditional Access Policies"
+  class="conditional-access-container"
+  i18n-aria-label>
+  <div class="filter-paginator-container">
+    <mat-form-field
+      appearance="fill"
+      class="conditional-access-table-filter"
+      subscriptSizing="dynamic">
+      <mat-label i18n>Filter Policies...</mat-label>
```
jelinaunger: the menu is named conditional access and here you name it policies. We should be consistent in the wording

privacyidea/static_new/src/app/components/conditional-access/conditional-access.component.html
```diff
@@ -0,0 +1,328 @@
...
+        i18n
+        mat-header-cell>
+        Stages
+      </th>
+      <td
+        *matCellDef="let policy"
+        class="text-center"
+        mat-cell>
+        {{ policy.stages.length }}
+      </td>
```
jelinaunger, line 196: I think the number of stages does not gain us anything. Also thresholds without the action context is useless. Suggestion: we have an optional detailed view (similar as for event handlers) there we can list for each stage the threshold priority and actions (without action values). Stages could be visually separated by horizontal line. Or we remove it entirely from the table and provide a detailed read only mode (we need to ensure admins with only read rights can access the full information)

privacyidea/static_new/src/app/components/conditional-access/conditional-access.component.html
```diff
@@ -0,0 +1,328 @@
...
+        @if (isEditing(policy) && editIsSingleStage()) {
+          <input
+            (click)="$event.stopPropagation()"
+            (input)="setEditThreshold($any($event.target).value)"
+            [value]="editThreshold()"
+            aria-label="Failure threshold"
+            class="ca-edit-input ca-edit-input-num"
+            i18n-aria-label
+            min="1"
+            type="number" />
```
jelinaunger: no need to edit thresholds in place here (it is even not clear which threshold one is editing)

privacyidea/static_new/src/app/components/conditional-access/conditional-access.component.html
```diff
@@ -0,0 +1,328 @@
...
+          matTooltip="Toggle Dry Run (evaluate but do not enforce)" />
+      </td>
+    </ng-container>
+
+    <ng-container matColumnDef="actions">
+      <th
+        *matHeaderCellDef
+        i18n
+        mat-header-cell>
+        Actions
```
jelinaunger: hide this column entirely if no write rights (otherwise it is confusing, cause I first expected the stage actions to appear there)

privacyidea/static_new/src/app/components/conditional-access/conditional-access.component.html
jelinaunger: priority might not even be required to show if we sort stages already accordingly

privacyidea/static_new/src/app/components/conditional-access/conditional-access.component.html
jelinaunger: I would not allow in-place editing for any column at all

privacyidea/static_new/src/app/components/conditional-access/conditional-access.component.scss
```diff
@@ -0,0 +1,86 @@
...
+// space, so the multi-select looks like the cell it replaces. The subscript
+// wrapper and the bottom line-ripple are hidden so the field is exactly the
+// container height (52px) and does not add a pixel that grows the row.
+.ca-edit-select {
+  display: block;
+  width: 100%;
+  height: 52px;
+  --mat-form-field-container-height: 52px;
+
+  ::ng-deep .mat-mdc-form-field-subscript-wrapper {
```
jelinaunger: no ng-deep (deprecated)

privacyidea/static_new/src/app/components/layout/navigation/navigation.component.html
```diff
@@ -442,6 +442,30 @@
       </span>
       <span i18n>Create</span>
     </a>
+    <a
```
jelinaunger: hide these if there are no read rights

privacyidea/static_new/src/app/services/auth/policy-actions.ts
```diff
@@ -37,6 +37,8 @@ export type PolicyAction =
   | "challenge_text_header"
   | "challenge_text_footer"
   | "getchallenges"
+  | "conditional_access_read"
+  | "conditional_access_write"
```
jelinaunger: replace these two everywhere with new names

privacyidea/static_new/src/app/services/conditional-access/conditional-access-policy.service.ts
```diff
@@ -0,0 +1,255 @@
...
+  "CHALLENGE_TRIGGERED",
+  "CHALLENGE_ANSWERED_OUT_OF_BAND",
+  "CHALLENGE_ANSWERED_FAIL",
+  "CHALLENGE_DECLINED",
+  "ENROLLMENT_TRIGGERED",
+  "ENROLLMENT_CANCELED_FAIL",
+  "UNKNOWN_FAIL_REASON"
+];
+
+// Mirrors privacyidea.lib.conditional_access.engine.LockoutAction
```
jelinaunger: maybe we should get all these constants from the api to avoid duplicated definition here and cover newly added once (already added in the auth log ui pr for the event types)

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/conditional-access-edit-page.component.html
```diff
@@ -0,0 +1,179 @@
...
+          <mat-error i18n>Must be at least 1.</mat-error>
+        }
+        <mat-hint i18n>Order across policies — highest first.</mat-hint>
+      </mat-form-field>
+
+      <p class="ca-field-note" i18n>
+        Priority decides how this policy ranks against the others: for an allow/deny decision the
+        highest-priority matching policy wins, while lock, block and email policies all run
+        regardless of priority.
+      </p>
```
jelinaunger: this hint is only related to the priority, hence it is a bit confusing that it spans the full width. Generally, I think we should add an i button/icon on fields/sections  where we want to explain more and put the explanation in an overlay only appearing on click / hover over the i icon. So users can get more information easily but it does not clutter the layout if a user already understood how it works.

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/stages-list/stage-item/conditional-access-stage-item.component.scss
```diff
@@ -0,0 +1,33 @@
+.ca-stage-item {
+  // Recessed panel matching the dashboard container: a subtle surface with the
+  // same triple inset shadow, so each stage reads as a "well" holding its card.
```
jelinaunger: to be honest I do not like this layout. I would prefer a simple box line with white background. Also actions might not need to be within an extra box.

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/stages-list/stage-item/actions-list/conditional-access-actions-list.component.html
```diff
@@ -0,0 +1,40 @@
...
+<div class="ca-actions-list">
+  <p class="ca-actions-list-label" i18n>Actions to run when this stage triggers</p>
+  @if (actions().length === 0) {
+    <p class="ca-actions-list-empty" i18n>No actions yet — add one below, e.g. lock the user or notify an admin.</p>
+  }
+  @for (action of actions(); track $index) {
+    <app-conditional-access-action-item
+      [action]="action"
+      (updateAction)="onUpdateAction($index, $event)"
+      (removeAction)="onRemoveAction($index)" />
```
jelinaunger: I think actions should be separated visually somehow, cause especially with the email action there are a lot of input fields and it is not clear where an action starts and a new one ends. adding a divider might be sufficient.

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/stages-list/stage-item/actions-list/action-item/conditional-access-action-item.component.html
```diff
@@ -0,0 +1,146 @@
...
+          class="input-width-l">
+          <mat-label i18n>Duration (seconds)</mat-label>
+          <input
+            matInput
+            type="number"
+            min="1"
+            [value]="durationValue()"
+            (input)="onDurationInput($any($event.target).value)" />
+          <mat-hint i18n>How long the lock or block lasts, in seconds.</mat-hint>
+        </mat-form-field>
```
jelinaunger, line 66: this could also be next to the action selection

privacyidea/static_new/src/app/components/conditional-access/conditional-access.component.html
```diff
@@ -0,0 +1,286 @@
...
+        mat-stroked-button>
+        <mat-icon>toggle_on</mat-icon>
+        <span i18n>(De)activate</span>
+      </button>
+      <button
+        (click)="toggleDryRunSelected()"
+        [disabled]="policySelection().length === 0"
+        class="action-button-secondary button-width-m"
+        mat-stroked-button>
+        <mat-icon>biotech</mat-icon>
```
jelinaunger, line 76: how is a microscope related to a dry run?

henrik-falk: experimental run, alternative should be a test tube (but there is none in the material icons list) or the science icon (fast so ein reagenzglas)

PR #5573, opened 2026-07-09
Don't fail fetching all machines if only some are corrupt/invalid

tizbac:
In a big active directory we have this problem occurred, attributes contains only {'dNSHostName': []}


nilsbehlen:
Thanks for this PR!
@jelinaunger please add this pr to the milestone if we merge

jelinaunger:
Hi, thanks for your contribution! Looking closer, though, there are a couple of other issues (#5584). Hence, we close this and handle it in a separate PR.

PR #5574, opened 2026-07-09
Add debug notice component and integrate with web UI settings

### LINKED ISSUE #5517: webui: add notice if privacyidea is running in debug
should be dismissable, but the admin should be notified

frankmer:
Introduce a debug notice component to the web UI that displays a message when the application runs in debug mode. Update related services and tests to handle the new debug state effectively. Ensure the debug notice is visible only to admin users.


privacyidea/api/lib/postpolicy.py
```diff
@@ -854,6 +854,8 @@ def get_webui_settings(request, response):
         content["result"]["value"]["container_wizard"] = container_wizard
 
         if role == ROLE.ADMIN:
+            content["result"]["value"]["is_debug"] = bool(current_app.debug)
```
henrik-falk: shouldnt it check the log_level in auth?

henrik-falk (changes requested):

privacyidea/api/lib/postpolicy.py
frankmer: current_app.debug (Flask) and PI_LOGLEVEL/log_level (logging verbosity) are two completely independent server settings. Neither controls nor implies the other. An admin can combine them independently as desired.
It just so happens that the built-in DevelopmentConfig sets both to “debug.” This is purely a convention of this one profile; there is no technical coupling (in ProductionConfig, for example, they are clearly separated).
The `is_debug` banner is intended to detect whether Flask is running in debug mode (tool debugger, RCE risk); `current_app.debug` is the correct and only reliable check for this.
A check on `log_level` (whether recalculated or the value sent upon login) would check something else (log verbosity, a separate real but different risk) and might miss the Flask debug case, so it’s not suitable as a replacement.

henrik-falk (approved):

PR #5603, opened 2026-07-16
adding comma-separated realm filtering logic

### LINKED ISSUE #4845: Feature Request: Support for Comma-Separated Realms in the /token/ Endpoint
## Description

The current `/token/` endpoint allows querying tokens for a specific realm using the `tokenrealm` parameter. When analyzing user accounts that may possess tokens in several independent realms (e.g., `staff`, `students`, `staff-rollout`, `students-rollout`), a separate API call must currently be made for each realm if filtering is desired.

However, when querying tokens for a **specific user** (`user=<username>`), it would be highly beneficial to retrieve the relevant tokens from multiple predefined realms in a single call. This is particularly relevant for scripts and integrations based on prioritized realm logic.

## Problem (Current State)

When querying tokens for a user who possesses tokens in, for example, `realm1` and `realm2`, the client must currently either:

1. Execute two separate API calls (`/token/?user=X&tokenrealm=realm1` and `/token/?user=X&tokenrealm=realm2`), or

2. Omit the `tokenrealm` filter and receive the user's entire token list (`/token/?user=X`) to perform filtering within the client application (e.g., PHP script).

The attempt to pass `tokenrealm=realm1,realm2` currently yields no or an incorrect result.

## Proposed Solution (Feature)

We request that the `tokenrealm` query parameter be extended to accept a comma-separated list of realm names (e.g., `tokenrealm=staff,staff-rollout,students`).

This would enable a single, more efficient API call per user and reduce the load on the PrivacyIDEA server by filtering out unnecessary tokens (those not belonging to the relevant realms) directly on the server side.

## Example of Desired Usage

```
GET /token/?user=username&tokenrealm=staff,students,staff-rollout,students-rollout
```
Should return all tokens belonging to the user `username` that are registered in any of the four specified realms.

jona-samuel:
closes #4845


privacyidea/api/token.py
```diff
@@ -660,7 +660,9 @@ def list_api():
         assigned to any user in this realm. Ignored for user-role
         callers.
     :query tokenrealm: filter to tokens that belong to this realm
-        (independent of the user's realm).
+        (independent of the user's realm). Substring match via ``*``;
+        comma-separated list of exact realm names also supported
+        (e.g. ``realm1,realm2,realm3``).
```
jelinaunger: to be consistent with the type parameter, we should add an own token_realm_list parameter

tests/test_lib_token.py
```diff
@@ -1292,6 +1292,58 @@ def test_41_get_tokens_paginate(self):
...
+        self.setUp_user_realm3()
+
+        tok_r1 = init_token({"type": "hotp", "genkey": True},
+                            tokenrealms=[self.realm1])
+        tok_r2 = init_token({"type": "hotp", "genkey": True},
+                            tokenrealms=[self.realm2])
+        tok_r3 = init_token({"type": "hotp", "genkey": True},
+                            tokenrealms=[self.realm3])
+
+        # Single realm still works as before
```
jelinaunger: already covered in the previous test

jelinaunger (commented):
Would be nice to have this already added to the new UI (multi select list with known realms) or create a follow up issue for it. In the conditional access PR we have already implemented a multi-select filter option, so we might cherry-pick this from there to re-use it

privacyidea/api/token.py
jona-samuel: This way it works the same way as serial, and I don't think it is problematic because you still can give him just one realm

privacyidea/lib/token/query.py
```diff
@@ -545,10 +568,14 @@ def get_tokens_paginate(tokentype: str | None = None, token_type_list: list[str]
     if serial and "*" not in serial and "," in serial:
         serial_list = serial.replace(" ", "").split(",")
         serial = None
+    realm_list = None
+    if realm and "," in realm:
+        realm_list = realm.replace(" ", "").split(",")
+        realm = None
```
jelinaunger: since they now behave exactly the same, there is no need to differentiate between single realm or realm list. can be always a list for simplicity.

jelinaunger (approved):

PR #5604, opened 2026-07-16
fix: update log_with decorator

### LINKED ISSUE #5524: log_with: two call sites use wrong hide_args_keywords form (over-hide via exception fallback)
### Summary

The `log_with` decorator's `hide_args_keywords` option is used with the wrong argument shape in two places. The redaction *appears* to work only because a broad `except Exception` in the decorator hides **all** arguments when the lookup fails. No secrets leak, but the feature isn't being used as intended and the affected DEBUG logs are needlessly empty.

### How `hide_args_keywords` actually works

In `privacyidea/lib/log.py`:

```python
for k, v in self.hide_args_keywords.items():
    for keyword in v:
        if keyword in args[k]:          # k is used as a positional INDEX into args
            log_args[k][keyword] = "HIDDEN"
```

- The dict **key** must be the **positional index** of the argument (e.g. `0`), not the parameter name.
- The dict **value** must be an **iterable of keywords** (a list), not a single string.

Correct, working usages already exist:

- `privacyidea/lib/resolver.py:69` — `hide_args_keywords={0: ["BINDPW", "Password"]}`
- `privacyidea/lib/user.py:779` — `hide_args_keywords={0: ["pass", "password"]}`

### The two incorrect usages

- `privacyidea/lib/token/lifecycle.py` (`init_token`) — `hide_args_keywords={'param': 'pin'}`
- `privacyidea/lib/tokens/radiustoken.py:134` — `hide_args_keywords={'param': 'pin'}`

With a string key, `args['param']` raises `TypeError: tuple indices must be integers`. That lands in the decorator's `except Exception`, which sets `log_args = ()` / `log_kwds = {}` — hiding everything. So nothing leaks, but:

- the redaction isn't selective as the API intends, and
- `init_token`'s param dict can carry `otpkey` (and `pin`), which is the secret we actually care about — currently protected only by the accidental catch-all.

Demonstration (real `log_with` on an `init_token`-shaped function called positionally):

| Config | Logged entry args |
|---|---|
| `{'param': 'pin'}` | `arguments () and keywords {}` |
| `{'param': ['pin', 'otpkey']}` | `arguments () and keywords {}` (identical — still the exception path) |
| `{0: ['pin', 'otpkey']}` | `[{'type': 'hotp', 'pin': 'HIDDEN', 'otpkey': 'HIDDEN'}]` |

### Proposed fix

- `init_token`: `hide_args_keywords={0: ['pin', 'otpkey']}`
- `radiustoken`: `hide_args_keywords={0: ['pin']}` (adjust to the actual secret keys in that param dict)

**Behavior-change caveat:** after the fix, DEBUG logs will show the non-secret fields of the param dict (`type`, `serial`, …) that are currently fully hidden by the exception fallback. The secrets stay hidden; this is an improvement for debuggability, but it is a visible change to log output.

### Optional hardening

The `except Exception: # hide everything` block silently masks this kind of misuse. Having `log_with` validate the `hide_args_keywords` config (or log a warning when the lookup fails) would have surfaced both bugs immediately. Worth considering as part of this fix or a follow-up.

### Notes

These are pre-existing usages; surfaced while reviewing CodeQL "clear-text logging" alerts. Note that CodeQL will likely keep flagging these log sites regardless, since it cannot reason about the dynamic redaction.

jona-samuel:
closes #5524


nilsbehlen (approved):
needs a follow up for 1. broader sweep and 2. way larger scope: redesign of this whole system that carries the information that a param is confidential by definion of same param and not via lists that have to be maintained.
but for now this is way better than the previous status quo.

PR #5609, opened 2026-07-16
CA: thread detection

### LINKED ISSUE #5434: Conditional Access: M2 - Threat Detection
* Password brute force
* MFA brute force
* password spraying
* blocklist

jelinaunger:
closes #5434


tests/conditional_access_lockout_base.py
```diff
@@ -0,0 +1,96 @@
...
+# You should have received a copy of the GNU Affero General Public
+# License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# SPDX-FileCopyrightText: 2026 NetKnights GmbH <https://netknights.it>
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""
+Shared test fixtures for the conditional-access lockout tests.
+
+:class:`LockoutTestCase` is the common base for the engine, template and
+snapshot suites: it resolves a real test user, wipes every lockout table before
```
henrik-falk: snapshot suite was deleted

privacyidea/lib/conditional_access/engine.py
```diff
@@ -385,41 +443,55 @@ def evaluate_access_decision(user: "User", source_ip: str | None = None,
...
+    the met stage carries only lockout-style actions, or the policy is in dry-run).
+    """
+    if policy.target == LockoutTarget.SOURCE_IP:
+        # IP-scoped: decide on the source IP regardless of user resolution. A
+        # never-block IP is never denied by an IP policy (mirrors the BLOCK_IP
+        # allowlist), so it contributes no decision.
+        if not source_ip or is_ip_never_block(source_ip):
+            return None
+        count = count_distinct_users_for_ip(source_ip, policy.counter_types_to_track,
+                                            policy.time_window_seconds, window_end=now)
```
henrik-falk, line 485: maybe restrict the SELECT to policies that have
at least one stage with an ALLOW or DENY action

privacyidea/lib/conditional_access/engine.py
```diff
@@ -385,41 +443,55 @@ def evaluate_access_decision(user: "User", source_ip: str | None = None,
...
-    :param user: the authenticating user; an unresolved user yields ``CONTINUE``
-    :param source_ip: the resolved client IP (reserved for IP-scoped decisions)
+    :param user: the authenticating user; only ``user``-target policies need it
+    :param source_ip: the resolved client IP; ``source_ip``-target policies decide on it
     :param now: the reference time; defaults to :func:`utc_now`
     :return: the :class:`AccessDecision` for this request
     """
-    if not _resolved(user):
-        return AccessDecision.CONTINUE
     now = _naive_utc(now) if now is not None else utc_now()
```
henrik-falk, line 458: should we allow admin self lockout?

privacyidea/lib/conditional_access/engine.py
```diff
@@ -591,7 +678,7 @@ def _evaluate_policy(policy: LockoutPolicy, user: "User", event_type: str,
         return []
 
     log.info(f"Policy {policy.name!r} triggered stage {triggered_stage.id} "
-             f"(threshold {triggered_stage.failure_threshold}) for {user!r}: "
+             f"(threshold {triggered_stage.failure_threshold}) for {subject_label}: "
```
henrik-falk, line 681: (line 843) maybe we should add another threshhold at which point the policy is not triggered anymore (to prevent infinite email spam)

privacyidea/lib/conditional_access/engine.py
```diff
@@ -385,41 +443,55 @@ def evaluate_access_decision(user: "User", source_ip: str | None = None,
...
+    :param user: the authenticating user; only ``user``-target policies need it
+    :param source_ip: the resolved client IP; ``source_ip``-target policies decide on it
     :param now: the reference time; defaults to :func:`utc_now`
     :return: the :class:`AccessDecision` for this request
     """
-    if not _resolved(user):
-        return AccessDecision.CONTINUE
     now = _naive_utc(now) if now is not None else utc_now()
     policies = db.session.scalars(
         select(LockoutPolicy)
```
henrik-falk, line 460: disable duplicate priority to prevent race condition

privacyidea/lib/conditional_access/engine.py
henrik-falk: so we count less on every auth

jelinaunger: it is not restricted to allow / deny. If target is set to IP, then there is the need to query the log, otherwise we do not reach there.

privacyidea/lib/conditional_access/engine.py
jelinaunger: Actually this is only for the IP path, so the admin itself would not be blocked, just the IP. (targe IP restricts the configurable actions and it excludes user specific ones)
However, I think we need conditions to make it configurable for who a policy should hold (users / external / internal admins, realms, ...)

privacyidea/lib/conditional_access/engine.py
jelinaunger: out of scope here -> #5643

PR #5610, opened 2026-07-16
Fix session timer and icon badges

henrik-falk:
+ username display in self service


jelinaunger (commented):
the small icon (settings/add) are slightly truncated at the right

Image

privacyidea/static_new/src/styles/icon.scss
```diff
@@ -0,0 +1,233 @@
...
+  vertical-align: middle;
+
+  // Small circular badge overlaid on the bottom-right corner of the base icon.
+  // The glyph is centered inside the circle; the solid background sits on top of
+  // the base icon (no cut-out). Offset defaults come from --icon-badge-right /
+  // --icon-badge-bottom, overridden by the icon-badge-right-* modifiers below.
+  .icon-badge-overlay {
+    position: absolute;
+    z-index: 1;
+    right: var(--icon-badge-right, -4px);
```
frankmer: Negative values will always lead to overflow. You could add 'overflow: visible;', but this would cause another problem: the badge could end up overlapping or sitting under the following components.

frankmer (approved):

PR #5612, opened 2026-07-17
5593/search policy and users without keyword

### LINKED ISSUE #5593: search without keyword in policy and user
All data for both policy and user objects is loaded at once, i.e. there is no pagination from the api.
Therefore, we should allow searching without keyword. For the users table, its especially clear that you would want to search a username, so adding "username:" to a query is redundant.

nilsbehlen:
* Search user table without keyword
* Search policy table without keyword
* Focus the filter input directly if you go to either of these pages
* Added highlighting to policy search (like events already had), and the highlighted entry are moved to the top of the cell so they are visible


frankmer:
When filtering the users there is no highlighting.

nilsbehlen:
> When filtering the users there is no highlighting.

highlighting is only in policies. user data is not very complex so i dont think its needed there

frankmer (approved):

PR #5614, opened 2026-07-17
4914/rate limiting

### LINKED ISSUE #5594: Conditional Access: Rate Limiting
### Goal
Let admins cap authentication volume, e.g. "max 10 auth attempts per user per minute".

### Core problem: events ≠ attempts
`authentication_log` stores one row per HTTP request. One logical attempt
(push / passkey / 4eyes) spans several requests, so counting *all* events
over-counts a single login as many. Each *completed* attempt, however, ends in
exactly one terminal event (a failure type or `LOGIN_SUCCESS`) — so counting
terminal events counts completed attempts without correlating requests. Only
abandoned / in-flight attempts have no terminal event.

### What's needed
event→attempt correlation via a request id
request id needs to be added as well, it could be stored in the challenge to avoid reading it from the previous auth log entry

### What might be useful
event type abstraction into attempt events (success, failed, abandoned, any)

jelinaunger:


henrik-falk:
- Templates should have count mode so that they can be saved without additional changes

- Count mode should have mat error since it is required

privacyidea/static_new/src/app/services/conditional-access/conditional-access-policy.service.ts
```diff
@@ -186,6 +225,64 @@ export class ConditionalAccessPolicyService implements ConditionalAccessPolicySe
...
+    () => this.templatesResource.value()?.result?.value ?? []
+  );
+
+  // Actions allowed for a target; falls back to the full list until /targets loads,
+  // so the select is never empty on first paint.
+  actionsForTarget(target: LockoutTarget): LockoutActionType[] {
+    return this.actionsByTarget()[target] ?? this.actionTypes();
+  }
+
+  // Count modes supported by a target (sorted; the UI treats the first as the default). Empty until /targets loads;
```
henrik-falk: the list is sorted alphabetically, so for user the first entry is PER_ATTEMPT while the actual default is PER_REQUEST

henrik-falk (approved):

PR #5616, opened 2026-07-20
Do not count WebUI challenge-trigger step towards auth_max_fail

### LINKED ISSUE #5628: WebUI challenge-trigger step counts towards auth_max_fail
In a two-step WebUI login (`login_mode=privacyIDEA` with a challenge-response token), the first `/auth` step submits the correct password to trigger the challenge. This step is recorded as a failed authentication and counted towards `auth_max_fail`, so a user performing a normal login can be locked out.

The challenge-trigger step should not count as a failed authentication.

Addressed by #5616.

QbDesu:
With login_mode=privacyIDEA and a challenge-response token, the first step of the two-step WebUI login (correct password, challenge is triggered) is logged with success=False at /auth and was counted as a failed authentication by check_max_auth_fail. A user who had used up all but one allowed failure could still trigger the OTP challenge, but was then locked out before being able to answer it.

The /validate/check count already excludes challenge-trigger entries via the audit "authentication" column; apply the same exclusion to the /auth count. For this, mark the /auth audit entries accordingly: CHALLENGE when a challenge is triggered, ACCEPT on success and REJECT on failed logins (the SQL negation would otherwise also exclude rows with an unset authentication column from the count).

Fixes #5628


nilsbehlen:
hi and thanks for you contribution, we will get to this soon.

nilsbehlen (approved):

PR #5621, opened 2026-07-20
Add notification delivery and resolver timing widgets

### LINKED ISSUE #5504: Dashboard metrics
Add metrics like ldap time, cert epiry, ... to the dashboard as done in #5266 but for the new UI.

Additionally, reload all resources on refresh (also for the existing ones, follow up of #5456)

frankmer:
Introduce Notification Delivery and Resolver Timing widgets to enhance dashboard metrics. Implemented functionality includes integration with SMS and SMTP services, performance metrics display, and data refresh capabilities. Updated the DashboardDataStore and SystemService to support these features, along with necessary tests and mock services for improved reliability.


privacyidea/static_new/src/app/components/dashboard/widgets/notification-delivery-widget/notification-delivery-widget.component.html
```diff
@@ -0,0 +1,98 @@
...
+              <td>{{ entry.failed }}</td>
+              <td><span [class]="badgeClass(entry)">{{ entry.error }}</span></td>
+            </tr>
+          }
+        </tbody>
+      </table>
+    }
+
+    @if (sections().push.length === 0 && sections().sms.length === 0 && sections().email.length === 0) {
+      <p i18n>No deliveries in this window.</p>
```
jelinaunger: If I have sms gateways etc. configured, but without any deliveries we do not reach here. Tables are still shown without any deliveries. Is that intentional?

privacyidea/static_new/src/app/components/dashboard/widgets/notification-delivery-widget/notification-delivery-widget.component.scss
```diff
@@ -0,0 +1,67 @@
...
+  border-collapse: collapse;
+
+  th,
+  td {
+    text-align: left;
+    padding: calc(var(--global-padding) / 2) var(--global-padding);
+    border-bottom: 1px solid color-mix(in srgb, currentColor 12%, transparent);
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
```
jelinaunger, line 57: names are truncated, while other columns could shrink further. Additionally, to make the full name accessible it could be shown on hover as type hint. (holds for all widgets)

privacyidea/static_new/src/app/components/dashboard/widgets/notification-delivery-widget/notification-delivery-widget.component.ts
```diff
@@ -0,0 +1,145 @@
...
+        return;
+      }
+      const value = ref.value();
+      if (value !== undefined) {
+        const ok =
+          value.delivery.result?.status === true &&
+          value.gateways.result?.status === true &&
+          value.smtp.result?.status === true;
+        this.state.set(ok ? "ready" : "error");
+      } else if (ref.error()) {
```
jelinaunger: error should be checked before reading the value

privacyidea/static_new/src/app/components/dashboard/widgets/resolver-timing-widget/resolver-timing-widget.component.html
```diff
@@ -0,0 +1,51 @@
...
+  <div class="table-container">
+    <table class="resolver-timing-table">
+      <thead>
+        <tr>
+          <th i18n>Resolver</th>
+          <th i18n>Operation</th>
+          <th i18n>Count</th>
+          <th i18n>Avg (ms)</th>
+          <th i18n>P95 (ms)</th>
+          <th i18n>Max (ms)</th>
```
jelinaunger: allow sorting by all columns might be useful (e.g. to better compare a specific operation over multiple resolvers or to see all operations of the same resolver)

privacyidea/static_new/src/app/components/dashboard/widgets/resolver-timing-widget/resolver-timing-widget.component.html
```diff
@@ -0,0 +1,51 @@
...
+<app-widget-state [state]="state()">
+  <div class="table-container">
+    <table class="resolver-timing-table">
+      <thead>
+        <tr>
+          <th i18n>Resolver</th>
+          <th i18n>Operation</th>
+          <th i18n>Count</th>
+          <th i18n>Avg (ms)</th>
+          <th i18n>P95 (ms)</th>
```
jelinaunger: info what p95 means would be nice

privacyidea/static_new/src/app/components/dashboard/widgets/resolver-timing-widget/resolver-timing-widget.component.html
```diff
@@ -0,0 +1,51 @@
...
+-->
+
+<app-widget-state [state]="state()">
+  <div class="table-container">
+    <table class="resolver-timing-table">
+      <thead>
+        <tr>
+          <th i18n>Resolver</th>
+          <th i18n>Operation</th>
+          <th i18n>Count</th>
```
jelinaunger: old UI names this Requests, might be more clear what this number stands for

privacyidea/static_new/src/app/services/dashboard/dashboard-data-store.service.ts
```diff
@@ -74,6 +78,10 @@ export class DashboardDataStore implements DashboardDataStoreInterface {
     return (this.entries.get(key) as CacheEntry<T> | undefined) ?? null;
   }
 
+  refreshAll(): void {
+    this.entries.forEach((entry) => this.startFetch(entry));
```
jelinaunger, line 66: token types only shows the result of a single token type and hangs up loading while the requests already finished.

privacyidea/static_new/src/app/services/dashboard/dashboard-data-store.service.ts
```diff
@@ -84,6 +92,29 @@ export class DashboardDataStore implements DashboardDataStoreInterface {
...
+          entry.revalidating.set(false);
+          entry.inFlight = false;
+          entry.subscription = undefined;
+        })
+      )
+      .subscribe({
+        next: (value) => entry.value.set(value),
+        error: () => entry.error.set(true)
+      });
+  }
```
jelinaunger, line 99: seems to be kind of duplicated of the load function

privacyidea/static_new/src/app/components/dashboard/widgets/resolver-timing-widget/resolver-timing-widget.component.html
```diff
@@ -0,0 +1,51 @@
...
+            <td>{{ row.resolver }}</td>
+            <td>{{ row.op !== null ? row.op : "—" }}</td>
+            <td>{{ row.count }}</td>
+            <td>{{ row.avgMs !== null ? row.avgMs : "—" }}</td>
+            <td><span [class]="badgeClass(row)">{{ row.p95Ms !== null ? row.p95Ms : "—" }}</span></td>
+            <td>{{ row.maxMs !== null ? row.maxMs : "—" }}</td>
+          </tr>
+        } @empty {
+          <tr>
+            <td colspan="6" i18n>No resolver activity in this window.</td>
```
jelinaunger: we should state what the time window is. Also if data is available and also for the other widgets.

privacyidea/static_new/src/app/components/dashboard/widgets/certificate-health-widget/certificate-health-widget.component.html
```diff
@@ -0,0 +1,47 @@
...
+          <th i18n>Status</th>
+        </tr>
+      </thead>
+      <tbody>
+        @for (entry of entries(); track entry.source + entry.name) {
+          <tr>
+            <td>{{ entry.source }}</td>
+            <td>{{ entry.name }}</td>
+            <td>{{ entry.days_remaining !== null ? entry.days_remaining : "—" }}</td>
+            <td><span [class]="badgeClass(entry.status)">{{ entry.status }}</span></td>
```
jelinaunger: status is untranslated

privacyidea/static_new/src/app/components/dashboard/widgets/certificate-health-widget/certificate-health-widget.component.html
```diff
@@ -0,0 +1,47 @@
...
+-->
+
+<app-widget-state [state]="state()">
+  <div class="table-container">
+    <table class="cert-health-table">
+      <thead>
+        <tr>
+          <th i18n>Source</th>
+          <th i18n>Name</th>
+          <th i18n>Days left</th>
```
jelinaunger: Exact expire date might also be useful to have

privacyidea/static_new/src/app/components/dashboard/widgets/certificate-health-widget/certificate-health-widget.component.html
```diff
@@ -0,0 +1,47 @@
...
+          <th i18n>Status</th>
+        </tr>
+      </thead>
+      <tbody>
+        @for (entry of entries(); track entry.source + entry.name) {
+          <tr>
+            <td>{{ entry.source }}</td>
+            <td>{{ entry.name }}</td>
```
jelinaunger: this should be a link to the resolver, if user has the right to see it. (also in the other tables)

privacyidea/static_new/src/app/services/dashboard/dashboard-data-store.service.ts
frankmer: entryFor is a helper function for load

privacyidea/static_new/src/app/components/dashboard/widgets/token-types-widget/token-types-widget.component.ts
```diff
@@ -144,6 +157,7 @@ export class TokenTypesWidgetComponent extends DashboardWidget implements OnInit
 
     const typeByKey = new Map(tokenTypes.map((type) => [type.key, type]));
     const initialCounts = cached ?? [];
+    const previousByKey = new Map((previousCounts ?? initialCounts).map((entry) => [entry.key, entry]));
```
nilsbehlen: **`refreshAll()` replays a scan seed that the first subscription already mutated → `undefined` rows and a change-detection `TypeError`.**

`initialAccumulator` (161-164) is built once per `load()` call, *outside* the factory closure that the store retains. The `scan` at 187 mutates `accumulated.indexByKey` in place (192) while `items` keeps pointing at the original array. `DashboardDataStore.refreshAll()` re-invokes the stored factory, so the second round starts from the same seed object: `items` is still `[]`, but `indexByKey` already holds every index from round one.

Repro: open the dashboard, then press the refresh button in the user-utils panel.
1. First response of round two hits the `index !== undefined` branch and does `nextItems[5] = entry` on an empty array.
2. The next emission spreads that sparse array, turning the holes into real `undefined` elements.
3. `typeCounts()` (line 78) evaluates `entry.count === null || entry.count > 0` on `undefined` → `TypeError: Cannot read properties of undefined (reading 'count')` during change detection. `allTypesFailed()`'s `.every()` throws the same way.

The widget then stops rendering until a full page reload. It only survives if the ~25 parallel `GET /token/?count` requests come back in exactly the same order both rounds, which is not guaranteed. I reproduced it by replaying this scan/seed logic standalone in node.

Fix would be to build the seed *inside* the factory (so each subscription gets a fresh `items`/`indexByKey` pair), or to treat the accumulator as immutable and never `set()` into the shared map.

privacyidea/static_new/src/app/components/layout/user-utils-panel/user-utils-panel.component.ts
```diff
@@ -217,6 +219,9 @@ export class UserUtilsPanelComponent {
     }
 
     switch (this.contentService.routeUrl()) {
+      case ROUTE_PATHS.DASHBOARD:
+        this.dashboardDataStore.refreshAll();
```
nilsbehlen, line 223: **Global refresh replays each entry's stored factory, so the token-types widget can never pick up a newly used token type.**

`refreshAll()` (dashboard-data-store.service.ts:65-67) re-invokes `entry.factory`, it does not call the widgets' own `reload()`. That matters for `TokenTypesWidgetComponent.load()`: when a cached value exists, `keysToLoad` is narrowed to types with `count > 0` (token-types-widget.component.ts:143-146) and *that reduced factory* is what gets stored.

Repro: visit the dashboard, navigate to Tokens and back (second `ngOnInit` → `load(false)` sees the cache and narrows the key set), enrol the very first WebAuthn token, press refresh. The stored factory never requests `type=webauthn`, so WebAuthn stays missing no matter how often refresh is pressed — only a hard page reload or the per-widget reload button (`reload()` → `load(true)`) surfaces it.

Same path also loses the new stale-value fallback: `reload()` passes `previous` counts into `previousByKey`, `refreshAll()` cannot, so a transient failure during a global refresh renders "unknown" instead of the last known counts.

Having the panel call each widget's `reload()` (or having the store re-ask the widget for a factory) would avoid both.

privacyidea/static_new/src/app/components/dashboard/widgets/resolver-timing-widget/resolver-timing-widget.component.ts
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  ngOnInit(): void {
+    this.loadData();
+  }
+
+  private loadData(): void {
+    this.dataRef.set(
+      this.store.load("dashboard:resolver-timing", () =>
+        forkJoin({
```
nilsbehlen, line 153: **`forkJoin` turns `resolverread` into a hard requirement for the whole widget.**

`listResolvers()` is `GET /resolver/`, which is guarded by `@prepolicy(check_base_action, request, PolicyAction.RESOLVERREAD)` (privacyidea/api/resolver.py:61). An admin with audit/system rights but no `resolverread` gets a 403, the `forkJoin` errors, and the widget shows the generic "Could not load data" — even though `/system/health/resolver_timing` (`@admin_required` only) returned complete timing data.

The component already computes `resolverLinkAllowed = authService.actionAllowed("resolverread")` for the row links; gating the request on the same flag (and dropping the idle-resolver rows when it is false) would degrade gracefully. Alternatively `catchError(() => of(null))` on just the `resolvers` stream.

Also, unlike `AdministrationWidgetComponent` (`static requiredAction = "auditlog"`), this widget declares no `requiredAction`, so a permission problem surfaces as "error" rather than "denied".

privacyidea/static_new/src/app/components/dashboard/widgets/resolver-timing-widget/resolver-timing-widget.component.html
```diff
@@ -0,0 +1,69 @@
...
+          <th matTooltip="95% of requests within the last hour completed within this time." i18n-matTooltip>
+            <app-table-sort-header key="p95Ms" [sortState]="sort" label="P95 (ms)" i18n-label [hasHintTooltip]="true" />
+          </th>
+          <th>
+            <app-table-sort-header key="maxMs" [sortState]="sort" label="Max (ms)" i18n-label />
+          </th>
+        </tr>
+      </thead>
+      <tbody>
+        @for (row of sortedRows(); track row.resolver + row.op) {
```
nilsbehlen: **`track row.resolver + row.op` is not a unique key.**

The backend groups timing rows by the full label set, which includes `resolver_type` (privacyidea/lib/metrics.py:257), and the label writer falls back to `resolver_name = "?"` when an instance has neither `name` nor `resolverId` (privacyidea/lib/metrics.py:319-321). So within one window you can get both `{resolver: "?", resolver_type: "unknown", op: "get_user_info"}` and `{resolver: "?", resolver_type: "passwdresolver", op: "get_user_info"}` — likewise if a resolver is deleted and recreated under the same name with a different type.

Both collapse to the track key `"?get_user_info"`: Angular raises NG0955 and the widget body fails to render in a dev build, while a production build reconciles rows against the wrong DOM nodes, showing one resolver's timings under another's label.

The row model already carries `resolverType` (component.ts:83-91), so including it in the track expression is enough.

nilsbehlen (commented):
Reviewed the widget/store changes. Five points below, roughly in order of impact. The first one reliably breaks the Token Types widget when the dashboard refresh button is pressed, so I'd like that one addressed before merge; the rest are smaller.

Scope-wise the PR looks well contained now that #5665 is in master — 54 files, all dashboard widgets plus the services and mocks they need, no backend changes.

privacyidea/static_new/src/app/components/dashboard/widgets/authentications-widget/authentications-widget.component.ts
```diff
@@ -87,11 +87,13 @@ export class AuthenticationsWidgetComponent extends DashboardWidget implements O
       if (!ref) {
         return;
       }
+      if (ref.error()) {
```
nilsbehlen: **Checking `ref.error()` before `ref.value()` defeats the store's stale-while-revalidate behaviour.** (Same guard was added to 10 widgets — authentications:90, tokens:112, administration:89, policies:80, events:80, subscriptions:68, certificate-health, notification-delivery, resolver-timing, token-types.)

`startFetch` deliberately keeps the previous value on a failed refresh — the new store spec asserts exactly that. But with the error branch first, a single momentary 500 during a global refresh sets `state = "error"`, and `app-widget-state` replaces `<ng-content>` with "Could not load data." Every widget that had a populated table goes blank, discarding data the store still holds.

Keeping the `value !== undefined` branch first, and surfacing the failed refresh separately (e.g. a `refreshFailed` flag rendered as a small badge or tooltip), would preserve the point of the retained value.

nilsbehlen:
image

there should not be a grey box around the -

privacyidea/static_new/src/app/components/dashboard/widgets/authentications-widget/authentications-widget.component.ts
frankmer: keeping a widget in ready when only the refresh fails. error now only for a failed first load.

privacyidea/static_new/src/app/components/dashboard/widgets/token-types-widget/token-types-widget.component.ts
frankmer: rebuilding the token types loading so a dashboard-wide refresh re-queries every type and out-of-order answers can no longer drop rows

privacyidea/static_new/src/app/components/layout/user-utils-panel/user-utils-panel.component.ts
frankmer: rebuilding the token types loading so a dashboard-wide refresh re-queries every type and out-of-order answers can no longer drop rows

privacyidea/static_new/src/app/components/dashboard/widgets/resolver-timing-widget/resolver-timing-widget.component.ts
frankmer: skipping the resolver list request when resolverread is missing so the timing table still renders

privacyidea/static_new/src/app/components/dashboard/widgets/resolver-timing-widget/resolver-timing-widget.component.html
frankmer: tracking resolver timing rows by resolver, type and operation so duplicate names no longer collide

henrik-falk (approved):

PR #5627, opened 2026-07-22
4914/user ip state management

### LINKED ISSUE #5496: Conditional Access: M3 - User State Management
* locked user and ip tables with action to unlock them
* Show the users state on the user details and allow to reset it
* Add link to show user in auth log (like for the audit log)

jelinaunger:


privacyidea/static_new/src/app/components/logs/locked-users/locked-users.component.html
```diff
@@ -0,0 +1,336 @@
...
+          *matHeaderCellDef
+          mat-header-cell>
+          <span class="header-cell">
+            <span i18n>Username</span>
+            <button
+              (click)="onKeywordClick('usernames')"
+              class="filter-button"
+              i18n-matTooltip
+              mat-icon-button
+              matTooltip="Filter by username"
```
henrik-falk: i guess we dont do tooltips here anymore

privacyidea/static_new/src/app/components/logs/locked-users/locked-users.component.html
```diff
@@ -0,0 +1,336 @@
...
+        <input
+          (input)="handleFilterInput($event)"
+          [value]="casService.lockedUsersFilter().value"
+          aria-label="Filter Input"
+          i18n-aria-label
+          i18n-placeholder
+          matInput
+          placeholder="Example query: 'usernames: alice realms: myrealm'" />
+      </app-clearable-input>
+      <mat-hint i18n>The search is case-insensitive. Use * as a wildcard. Wrap the filter value in quotation marks
```
henrik-falk: i would prefer an i icon here, so that there i more space for the table at 200% zoom

henrik-falk (commented):
pi-manage unlock-user silently does nothing without --resolver

privacyidea/static_new/src/app/components/logs/locked-users/locked-users.component.html
```diff
@@ -0,0 +1,336 @@
...
+
+      <!-- Username column -->
+      <ng-container matColumnDef="username">
+        <th
+          *matHeaderCellDef
+          mat-header-cell>
+          <span class="header-cell">
+            <span i18n>Username</span>
+            <button
+              (click)="onKeywordClick('usernames')"
```
henrik-falk, line 112: plural here but not in auth log

privacyidea/static_new/src/app/components/logs/locked-users/locked-users.component.html
jelinaunger: well, but this is the design we use in the other tables as well, so for consistency I would like to keep it as it is.

privacyidea/api/conditional_access.py
```diff
@@ -296,3 +314,200 @@ def delete_policy(policy_id):
...
+    get_required_one_of(request.all_data, ["user", "user_id"])
+    user_id = get_optional(request.all_data, "user_id")
+    username = get_optional(request.all_data, "user")
+    realm = get_required(request.all_data, "realm")
+    resolver = get_optional(request.all_data, "resolver")
+    visibility_scopes = get_policy_visibility_scopes(PolicyAction.USER_LOCKOUT_READ)
+
+    # User is already resolved in before request, but only for the login, realm, resolver triplet. If the uid is given
+    # instead we need to resolve the user here
+    user = getattr(request, "user", None)
```
henrik-falk: should it be "User" instead of "user"?

privacyidea/lib/conditional_access/engine.py
```diff
@@ -1166,8 +1223,7 @@ def _upsert_user_lockout_state(user: "User", *, is_locked: bool,
         db.session.rollback()
```
henrik-falk, line 1283: rollback should be nested. we could create a helper function for this like this:
try:
        with db.session.begin_nested():
            db.session.delete(state)
    except Exception as ex:
        log.warning(f"Failed to delete the expired {description}: {ex!r}")
        return
    try:
        db.session.commit()
    except Exception as ex:
        log.warning(f"Failed to commit the deletion of the expired {description}: {ex!r}")
        db.session.rollback()

privacyidea/static_new/src/app/components/logs/blocklist/blocklist.component.ts
```diff
@@ -0,0 +1,299 @@
...
+  readonly displayedColumns: string[] = ["select", "identifier", "state", "block_expires_at", "blocked_at"];
+
+  // Keep the previous rows while a reload is in flight to avoid flicker.
+  readonly dataSource = linkedSignal<PiResponse<BlocklistEntry[]> | undefined, MatTableDataSource<BlocklistEntry>>({
+    source: () =>
+      this.casService.blocklistResource.hasValue() ? this.casService.blocklistResource.value() : undefined,
+    computation: (resource, previous) => {
+      if (resource) {
+        const ds = new MatTableDataSource(this.sortRows(resource.result?.value ?? []));
+        ds.filterPredicate = this.blockFilterPredicate();
```
henrik-falk: filter is not restored here, so every reload drops the active filter while

henrik-falk (approved):

privacyidea/static_new/src/app/components/logs/locked-users/locked-users.component.html
jelinaunger: good catch we need to change the api for the auth log to use plural parameters

privacyidea/lib/conditional_access/engine.py
jelinaunger: I think we should use an own session for auth log, user lock and blocklist writes. Making nested transactions obsolete. But we could create a helper for this anyway.

PR #5632, opened 2026-07-23
5606/conditional access date time picker

### LINKED ISSUE #5606: Conditional Access: Date Time Picker
For the auth log, the combined date time slider should be splitted up into a date picker and a time slider.
This could be moved on top of the table to be always visible (histogram gives nice insights).

henrik-falk:


jelinaunger (approved):
Would be nice to have this as an own re-usable component

henrik-falk:
> Would be nice to have this as an own re-usable component

as soon as we use it a second time i will refactor it

PR #5648, opened 2026-07-24
enrollment data visibility in last step dialog

### LINKED ISSUE #5608: additional dialog after push enrollment
after push token is enrolled succesfully, i get this dialog

Image

probably not intended?

frankmer:
Added showEnrollDataInLastStep to EnrollTokenBase. So we can hide the enrollmentdata in die last step dialog.


nilsbehlen:
when i press cancel (unfinished enrollment of the push token), i get another dialog: 
image

but it should just cancel, i dont want another dialog
edit: it seems you need to have a failed attempt before you get this second dialog upon cancel? e.g. by using http instead of https.

nilsbehlen (changes requested):
see comment

nilsbehlen (approved):

PR #5653, opened 2026-07-25
Fix container registration finalize audit success

### LINKED ISSUE #5652: POST /container/register/finalize passes a dict as audit success and raises TypeError
### Top-level intent

Complete a normal smartphone container registration and record its result correctly in the privacyIDEA SQL audit log.

### Steps to reproduce

1. Start a smartphone container enrollment.
2. Scan the enrollment QR code with a compatible client.
3. Complete the key exchange and send a valid request to:
   `POST /container/register/finalize`
4. Verify that the container is registered.
5. Inspect the privacyIDEA application log and the corresponding SQL audit entry.

### Expected outcome

The endpoint should complete successfully and the audit event should contain a scalar success value, such as `true` or `1`.

No exception should be raised by the SQL audit module.

### Actual outcome

The API request succeeds and the smartphone container becomes registered, but SQL audit processing raises a `TypeError`.

The value passed to the audit field named `success` is a dictionary containing the actual success flag and client policies.

Consequently, the audit module attempts to execute the equivalent of `int(dict)` and cannot store the result correctly.

### Configuration

* privacyIDEA version: 3.13.3
* Installation method: Docker container; privacyIDEA installed as a Python package
* Python version: 3.12
* Operating system: Debian-based Linux container
* Webserver: Gunicorn / Flask
* Token database: MariaDB 10.11

### Log file

Sanitized log entry:

```text
[ERROR][privacyidea.lib.auditmodules.sqlaudit]
exception TypeError(
    "int() argument must be a string, a bytes-like object
     or a real number, not 'dict'"
)

DATA: {
    'success': {
        'success': True,
        'policies': {
            'container_client_rollover': False,
            'initially_add_tokens_to_container': False,
            'disable_client_token_deletion': False,
            'disable_client_container_unregister': False
        }
    },
    'action': 'POST /container/register/finalize'
}
```

The error is reproducible for each successful smartphone container finalization.

### Additional technical analysis

The endpoint obtains a dictionary from `finalize_registration()`, adds the client policies to it, and then passes the entire dictionary as the audit `success` field:

```python
res = finalize_registration(container_serial, params)
res["policies"] = request.all_data.get("client_policies", {})

g.audit_object.log({
    "container_serial": container_serial,
    "container_type": container.type,
    "success": res
})
```

Relevant source:

https://github.com/privacyidea/privacyidea/blob/v3.13.3/privacyidea/api/container.py#L874-L912

The SQL audit backend expects `success` to be a boolean or integer-compatible scalar.

The audit call should instead derive a boolean from the response, for example:

```python
"success": bool(res.get("success"))
```

The `policies` dictionary can remain in the API response or be stored in a separate audit detail field if required.

A regression test should verify that a successful smartphone registration finalization:

- returns the expected API response,
- stores audit success as `1`,
- does not emit an SQL audit exception,
- preserves the client policies in the API response.

frankii91:
## Summary

- log the scalar registration result in the SQL audit `success` field
- preserve the complete API response, including client policies
- assert that a successful smartphone registration is stored as audit success `1`

## Root cause

`finalize_registration()` returns a dictionary. `registration_finalize()` added client policies to that response and then passed the entire dictionary as the audit `success` value. During audit finalization, the SQL audit backend calls `int(success)`, which raises `TypeError` for a dictionary.

## Validation

- `python -m pytest -q tests/test_api_container.py`
- 58 passed

Fixes #5652


jelinaunger:
Hi, thanks for your contribution!

jelinaunger (approved):

PR #5654, opened 2026-07-25
Initialize validate request policy state

### LINKED ISSUE #5651: Smartphone container template enrollment via /validate/check crashes because g.policies is not initialized
### Top-level intent

Enroll a user without existing tokens into a smartphone container using the `enroll_via_multichallenge` authentication policy and a smartphone container template containing a Push token.

### Steps to reproduce

1. Configure a user resolver and realm.
2. Configure a working Push provider.
3. Create a smartphone container template containing a Push token assigned to the authenticating user.
4. Create an authentication policy with:
   - `enroll_via_multichallenge=smartphone`
   - `enroll_via_multichallenge_template=<template_name>`
   - userstore authentication as appropriate for the resolver
5. Use a test user that has no tokens or containers.
6. Send valid userstore credentials to `POST /validate/check`.
7. Inspect the HTTP response, privacyIDEA log, and the resulting container.

### Expected outcome

privacyIDEA should return the smartphone enrollment challenge and create both:

- the smartphone container,
- the Push token defined by the selected template.

If template token creation fails, the operation should not leave behind a partially initialized container.

### Actual outcome

`POST /validate/check` returns HTTP 500.

The container is created before the exception occurs. A subsequent retry can allow that container to be registered, but the template token is not created, leaving a registered smartphone container with zero tokens.

### Configuration

* privacyIDEA version: 3.13.3
* Installation method: Docker container; privacyIDEA installed as a Python package
* Python version: 3.12
* Operating system: Debian-based Linux container
* Webserver: Gunicorn / Flask
* Token database: MariaDB 10.11

### Log file

Sanitized traceback:

```text
Exception on /validate/check [POST]

  File "privacyidea/api/lib/postpolicy.py", line 1024,
    in multichallenge_enroll_via_validate
    content = container_create_via_multichallenge(
        request, content, enroll_type
    )

  File "privacyidea/api/lib/postpolicy.py", line 931,
    in container_create_via_multichallenge
    create_container_tokens_from_template(
        container.serial, template_tokens, request, ROLE.USER
    )

  File "privacyidea/lib/container.py", line 658,
    in create_container_tokens_from_template
    policies = copy.deepcopy(g.policies)

  File "flask/ctx.py", line 56, in __getattr__
    raise AttributeError(name)

AttributeError: policies
```

The problem is reproducible on every initial enrollment attempt using this flow.

### Additional technical analysis

The `/validate/check` blueprint initializes objects such as `g.policy_object`, but it does not initialize `g.policies`.

`create_container_tokens_from_template()` nevertheless assumes that `g.policies` already exists:

https://github.com/privacyidea/privacyidea/blob/v3.13.3/privacyidea/lib/container.py#L658

The validate blueprint setup can be seen here:

https://github.com/privacyidea/privacyidea/blob/v3.13.3/privacyidea/api/validate.py#L160

The same helper initializes or replaces `g.policies` later during token creation, but the initial `copy.deepcopy(g.policies)` happens before that code.

A possible fix would be to initialize this request context consistently, or safely preserve an empty dictionary when it is absent, for example:

```python
policies = copy.deepcopy(getattr(g, "policies", {}))
```

A regression test should exercise the complete `/validate/check` flow with:

- `enroll_via_multichallenge=smartphone`,
- a smartphone container template,
- at least one template token,
- a user without existing tokens.

frankii91:
## Summary

- initialize request-local policy data in the `/validate` blueprint
- make the full smartphone template enrollment regression test independent of state left by earlier requests

## Root cause

Unlike the shared API request setup, the `/validate` blueprint initialized `g.policy_object` but not `g.policies`. Smartphone template enrollment later calls `create_container_tokens_from_template()`, which preserves the current policy data before applying token enrollment policies. On a fresh request, that first read raised `AttributeError: policies` after the container had already been created.

The existing end-to-end test could pass as part of a larger run when a previous request happened to leave `g.policies` in the long-lived test app context. It failed reliably in isolation. The test now removes that inherited state before dispatching `/validate/check`.

## Design note

This draft initializes the state at the request boundary, consistent with the shared API setup, instead of adding a defensive default inside the container library helper. Maintainer feedback on that placement is welcome.

## Validation

- `python -m pytest -q tests/test_api_validate_multichallenge_enroll.py`
- 11 passed

Fixes #5651


jelinaunger:
Hi, thanks for mentioning this issue. I just did a broader refactor and added g clearing to more tests to avoid this fault in the future.

nilsbehlen (approved):

PR #5656, opened 2026-07-27
Challenge: data is always a dict

### LINKED ISSUE #5620: Challenge data should be a dict for all token tyes
Challenge data is sometimes a dict and sometimes a raw value. It should be a dict in all token types to add further info (e.g. attempt id for the auth log / conditional access)

jona-samuel:
closes #5620


privacyidea/lib/tokens/emailtoken.py
```diff
@@ -354,7 +354,9 @@ def check_otp(self, anOtpVal, counter=None, window=None, options=None):
         ret = HotpTokenClass.check_otp(self, anOtpVal, counter, window, options)
 
         if ret < 0 and is_true(get_from_config("email.concurrent_challenges")):
-            if options.get("data") is not None and safe_compare(options.get("data"), anOtpVal):
+            challenge_data = options.get("data")
+            saved_otp = challenge_data.get("otp") if challenge_data else None
```
jelinaunger, line 358: legacy entry would be value

privacyidea/lib/tokens/foureyestoken.py
```diff
@@ -373,9 +373,12 @@ def has_further_challenge(self, options=None):
...
                                               transaction_id=transaction_id)
         if len(challengeobject_list) == 1:
-            remaining_realms = self._get_remaining_realms(options.get("data", {}))
+            # options["data"] is set by check_challenge_response / _authenticate_remaining_realms
+            # with the current used_tokens dict
+            used_tokens_data = options.get("data", {})
+            remaining_realms = self._get_remaining_realms(used_tokens_data)
             if remaining_realms:
-                options["data"] = json.dumps(options.get("data", {}))
+                options["data"] = used_tokens_data
```
jelinaunger: it assigns exactly what you have read before from it

privacyidea/lib/tokens/foureyestoken.py
```diff
@@ -413,7 +416,8 @@ def check_challenge_response(self, user=None, passw=None, options=None):
             for challengeobject in challengeobject_list:
                 if challengeobject.is_valid():
                     # challenge is still valid
-                    used_tokens = json.loads(challengeobject_list[0].data or json.dumps({}))
+                    challenge_data = challengeobject.get_data()
+                    used_tokens = challenge_data.get("used_tokens", {})
```
jelinaunger, line 419: add pure challenge_data here as fallback if used_tokens does not exist

privacyidea/lib/tokens/indexedsecrettoken.py
```diff
@@ -278,7 +278,9 @@ def check_challenge_response(self, user=None, passw=None, options=None):
                     # challenge is still valid
                     # Add the challenge to the options for check_otp
                     options["challenge"] = challengeobject.challenge
-                    options["data"] = [int(c) for c in challengeobject.data.split(",")]
+                    challenge_data = challengeobject.get_data()
+                    position_str = challenge_data["positions"]
```
jelinaunger: same here and probably also for the other token types, add the legacy key "value" as fallback

privacyidea/lib/cache/redis.py
```diff
@@ -392,15 +407,14 @@ def set_otp_status(self, valid: bool = False):
...
-            self.data = json.dumps(data)
+        """
+        Set the challenge data. Must be a dict (or None/empty).
+        Mirrors Challenge.set_data (models/challenge.py).
+        """
+        if not data:
+            self.data = ''
         else:
-            self.data = convert_column_to_unicode(data)
+            self.data = json.dumps(data)
```
jelinaunger, line 415: Unrelated to this PR, but Redis is missing the password encryption here the db function uses

tests/test_db_model.py
```diff
@@ -410,11 +410,9 @@ def test_12_challenge(self):
         self.assertEqual(20, len(c.transaction_id), c.transaction_id)
         self.assertEqual(20, len(c.get_transaction_id()), c.transaction_id)
 
-        c.set_data("some data")
-        self.assertEqual("some data", c.data)
-        self.assertEqual("some data", c.get_data(), c.data)
         c.set_data({"some": "data"})
-        self.assertIn("some", c.data, c.data)
+        self.assertEqual({"some": "data"}, c.get_data())
```
jelinaunger: use assertDictEqual

jelinaunger (commented):
Also add tests for the legacy authentication cases. Even if they should occure rarely, we should ensure no unexpected errors are raised

privacyidea/lib/tokens/foureyestoken.py
```diff
@@ -413,7 +415,8 @@ def check_challenge_response(self, user=None, passw=None, options=None):
             for challengeobject in challengeobject_list:
                 if challengeobject.is_valid():
                     # challenge is still valid
-                    used_tokens = json.loads(challengeobject_list[0].data or json.dumps({}))
+                    challenge_data = challengeobject.get_data()
+                    used_tokens = challenge_data.get("used_tokens", challenge_data)
```
jelinaunger: no fallback needed anymore as we dropped the legacy cases (my original comment was wrong anyway, wrote this before I saw the default dict with value shape and missed to remove it)

privacyidea/migrations/versions/c3d4e5f6a7b8_clear_challenge_table.py
```diff
@@ -0,0 +1,29 @@
+"""Clear challenge table for dict-only data format
```
jelinaunger: This is in the wrong directory, should be in privacyidea/privacyidea/migrations/versions

privacyidea/lib/cache/redis.py
```diff
@@ -355,12 +356,15 @@ def get_transaction_id(self) -> str:
...
+        :return: The challenge data as a dict. Returns ``{}`` if no data is stored.
+        :rtype: dict
+        """
         if not self.data:
             return {}
-        try:
-            return json.loads(self.data)
-        except (json.JSONDecodeError, ValueError):
-            return self.data
+        return json.loads(self.data)
```
jelinaunger: We should still catch the decode error here as the data could still be unparsable theoretically. If not parsable to a dict I would return an empty dict to ensure the dict data type from here on. And we can not handle the data correctly anyway if not a dict.

privacyidea/lib/tokens/emailtoken.py
```diff
@@ -354,7 +354,9 @@ def check_otp(self, anOtpVal, counter=None, window=None, options=None):
         ret = HotpTokenClass.check_otp(self, anOtpVal, counter, window, options)
 
         if ret < 0 and is_true(get_from_config("email.concurrent_challenges")):
-            if options.get("data") is not None and safe_compare(options.get("data"), anOtpVal):
+            challenge_data = options.get("data")
```
jelinaunger: add empty dict fallback

privacyidea/lib/tokens/indexedsecrettoken.py
jelinaunger: use the get function to avoid KeyError

privacyidea/lib/cache/redis.py
jelinaunger: And from this point on the return type should always be a dict. json.loads can also return string, list, .. depending on the data. So a type check should be here as well.

jelinaunger: Same on db level

privacyidea/lib/tokens/smstoken.py
```diff
@@ -387,7 +387,9 @@ def check_otp(self, anOtpVal, counter=None, window=None, options=None):
 
         ret = HotpTokenClass.check_otp(self, anOtpVal, counter, window, options)
         if ret < 0 and is_true(get_from_config("sms.concurrent_challenges")):
-            if options.get("data") is not None and safe_compare(options.get("data"), anOtpVal):
+            challenge_data = options.get("data")
```
jelinaunger: empty dict fallback

privacyidea/models/challenge.py
```diff
@@ -60,15 +60,11 @@ class Challenge(MethodsMixin, db.Model):
...
+        """Return the decrypted challenge data as a JSON string, or empty string."""
         raw = self._data
         if not raw:
             return raw
-        decrypted = decryptPassword(raw)
-        if decrypted and not decrypted.startswith("FAILED TO DECRYPT"):
-            return decrypted
-        # Legacy unencrypted data or decryption failure - return as-is
-        return raw
+        return decryptPassword(raw)
```
jelinaunger: decryption can still fail, keep the check

privacyidea/models/challenge.py
```diff
@@ -123,25 +119,24 @@ def set_data(self, data):
         The data is encrypted before being stored in the database since it
         may contain OTP values.
 
-        :param data: The challenge data (string, dict, or other serializable value)
+        :param data: The challenge data. Must be a dict (or None/empty).
+        :type data: dict or None
```
jelinaunger: type hint should go to the function signature

tests/test_migration_c3d4e5f6a7b8.py
```diff
@@ -0,0 +1,59 @@
...
+        c3.save()
+
+        # Verify rows exist
+        count = db.session.execute(
+            text("SELECT COUNT(*) FROM challenge")
+        ).scalar()
+        self.assertGreaterEqual(count, 3)
+
+        # Simulate the migration
+        db.session.execute(text("DELETE FROM challenge"))
```
jelinaunger: This tests nothing except that the sql statement exist. Check existing migration tests and the README file how to do this tests correctly. It should execute the real migration script and not only mirror its content

jelinaunger (changes requested):

privacyidea/models/challenge.py
```diff
@@ -123,25 +119,24 @@ def set_data(self, data):
...
+        :rtype: dict
+        """
         if not self.data:
             return {}
-        try:
-            data = json.loads(self.data)
-        except (json.JSONDecodeError, UnicodeDecodeError, ValueError):
-            data = self.data
-        return data
+        return json.loads(self.data)
```
jelinaunger: assign self.data to a variable to avoid doing the decryption multiple times

jelinaunger (commented):
api/validate.py:971-976: can be replaced with challenge.get_data()

lib/tokens/pushtoken.py: 
* dead isinstance(…, str) and redundant ininstance(..., dict) branches at :880, :1007, :1117, :1511
* _get_existing_challenge_data update doctring

privacyidea/models/challenge.py
```diff
@@ -134,9 +137,13 @@ def get_data(self):
         :return: The challenge data as a dict. Returns ``{}`` if no data is stored.
         :rtype: dict
         """
-        if not self.data:
+        data = self.data
+        if not data:
+            return {}
+        try:
+            return json.loads(data)
```
jelinaunger: the type check is still missing. json.loads does not alays return dicts. It can be a list or string as well, which should be avoided

jelinaunger (approved):

PR #5660, opened 2026-07-27
4914/conditional access fix threshold

### LINKED ISSUE #5644: Conditional Access: actions are only triggered on exact threshold count
actions should only be triggered once if event count == threshold (not >=)
This makes the last_stage_triggered column obsolete for the locked user and blocklist tables.

henrik-falk:


privacyidea/models/lockout_policy.py
```diff
@@ -121,9 +121,13 @@ class LockoutPolicyStage(MethodsMixin, db.Model):
...
     N stages (e.g. 5, 10 and 15 failures).
 
-    Within a policy the stages are evaluated highest ``priority`` first,
-    so the most severe matching stage wins (e.g. evaluate the 15-fail
-    stage before the 5-fail stage).
+    Post-response, each of a stage's actions fires when the failure count reaches
+    the stage's ``failure_threshold`` (see
+    :attr:`LockoutStageAction.retrigger_above_threshold` for the per-action
+    fire-once vs re-trigger choice); escalation is expressed as separate stages
+    per threshold. ``priority`` orders the stages for the pre-auth ALLOW/DENY
```
jelinaunger: mentioning priority is a bit confusing as it is a per policy priority and not per stage priority

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/stages-list/stage-item/actions-list/action-item/conditional-access-action-item.component.html
```diff
@@ -65,21 +66,63 @@
...
+          <mat-label i18n>Duration</mat-label>
           <input
             matInput
             type="number"
             min="1"
             [value]="durationValue()"
             (input)="onDurationInput($any($event.target).value)" />
-          <mat-hint i18n>How long the lock or block lasts, in seconds.</mat-hint>
+          @if (durationUnit() !== "seconds") {
+            <mat-hint i18n>= {{ durationSecondsHint() }} seconds</mat-hint>
```
jelinaunger: not sure if we need this hint on how many seconds this is (we need it for the backend right, but the user do not care about it)

tests/test_api_conditional_access.py
```diff
@@ -229,10 +230,17 @@ def test_user_locked_again_after_lock_expires(self):
...
+        # it does not re-fire: the user stays unlocked.
         body = self._check({"user": "cornelius", "pass": "pin000000"})
         self.assertFalse(body["result"]["value"], body)
+        self.assertFalse(is_user_locked(self.user))
+
+        # A successful login resets the counter; climbing back to exactly 3 re-locks.
+        body = self._check({"user": "cornelius", "pass": "pin755224"})
+        self.assertTrue(body["result"]["value"], body)
+        for _ in range(3):
+            self._check({"user": "cornelius", "pass": "pin000000"})
```
jelinaunger, line 246: to make this more explicit, check that the user is not yet locked within the loop

jelinaunger (approved):

privacyidea/models/lockout_policy.py
jelinaunger: oh I just realized that actually stages can have a priority (db model). But I think this is not required and can be dropped as the threshold now serves as kind of priority (higher threshold takes precedence)

Issue #5617, opened 2026-07-20
Multi select filter for realm and token-type


jona-samuel:
After merging 4914/conditional_access, use the multi-select component also for the realm and token type filter in the token list.

jelinaunger:
multi select component is in the master now

PR #5665, opened 2026-07-27
unrelated ui changes from conditional access

henrik-falk:
These changes are independent of the conditional-access feature and are split onto this master-based branch so they no longer appear in the conditional-access pull request once merged into master.

Included:
- Detail-card & global style refactor: consolidate the .h3-color class into the global stylesheet (serial/username-header -> h3-color), drop redundant font-size/color rules, add a --copy-button-opacity visibility hook, minor spacing/width utility-class tweaks across token/container/user/machine details, details-header, policies-table; replace em dashes with hyphens in detail text.
- External-services bulk row-selection: checkbox column, select-all and bulk-delete for ca-connectors, privacyidea-servers, radius-servers, service-ids, sms-gateways, smtp-servers and tokengroups.
- Shared filter components: app-multi-select-filter (+ menu) and app-filter-value-button, with the supporting table-utils sort fallback and the filter_value colon-quoting fix, plus the .filter-select-menu panel style.


jelinaunger (approved):

nilsbehlen:
These surfaced while reviewing #5621, which had this branch merged in. Posting here rather than opening an issue so the context stays with the change — happy to split them into issues if you prefer.

Line numbers are as of current master.

### 1. Select-all ignores the active filter, so bulk delete destroys rows the admin cannot see

`external-services/smtp-servers/smtp-servers.component.ts:104` — and the same code in all eight converted tables (`sms-gateways:109`, `tokengroups:104`, `ca-connectors`, `privacyidea-servers`, `radius-servers`, `service-ids`).

`toggleAllRows()` and `isAllSelected()` read `dataSource().data`, but the table renders `filteredData` — `MatTableDataSource.filter` only affects the latter.

Repro: with 10 SMTP servers configured, type a filter that leaves exactly one visible row, tick the header select-all checkbox intending to select that row, click Delete, confirm. All 10 servers are deleted, including nine that were not on screen. The per-row delete button this replaced could only ever delete its own row, so this is a new way to lose configuration.

Fix: select from `dataSource().filteredData` (and compare against its length) in both methods.

### 2. Bulk delete fires all requests in parallel and races `resource.reload()`

`external-services/sms-gateways/sms-gateways.component.ts:144` — same in the other seven tables (`tokengroups:143`, `smtp-servers:143`, `radius-servers:148`, `privacyidea-servers:148`, `ca-connectors:148`, `service-ids:141`).

```ts
selected.forEach((gateway) => this.smsGatewayService.deleteSmsGateway(gateway.name));
this.selection.set([]);
```

Two problems:

* Each `deleteX()` ends in `resource.reload()`, but Angular's `reload()` returns `false` and does nothing while a reload is already in `loading` state (`_resource-chunk.mjs:255`). With A, B, C selected, A's reload starts a `GET /smsgateway/`; B's and C's reloads land while it is in flight and are dropped. The table repaints from that response, which still lists B and C. Clicking one then navigates to a details page for a gateway that no longer exists, and only a manual page refresh corrects the list.
* Nothing attaches a `.catch()`, while each service method rejects with `throw new Error("delete-failed")`. A partial failure therefore produces N unhandled promise rejections, and the selection is cleared regardless of which deletes actually succeeded.

Fix: `await Promise.allSettled(...)` the deletes, then reload once, and report failures.

### 3. Container token table paginator is never wired up

`container/container-details/container-details-token-table/container-details-token-table.component.html:56`

The `<mat-paginator>` moved inside the `@if (dataSource().data.length > 1)` block, but it is bound to the data sources exactly once, in `ngAfterViewInit`. The child renders before the container-details `httpResource` resolves, so at that moment `data.length === 0`, the paginator is not in the DOM, `@ViewChild(MatPaginator)` is `undefined`, and `ngAfterViewInit` assigns `dataSource.paginator = undefined` / `externalDataSource.paginator = undefined`.

Repro: open a container holding 30 tokens. The paginator reappears once the tokens arrive but is attached to nothing: all 30 rows are listed instead of `pageSize = 5`, the range label reads "0 of 0", and next/previous do nothing.

Fix: use a `viewChild` signal plus an `effect` to (re)bind whenever the paginator appears, or keep it rendered and hide it with CSS.

### 4. "More Filter" on the container table lost the boolean handling for `assigned`

`container/container-table/container-table.component.html:313`

The menu moved out of `container-table-actions` and was rebound to the pre-existing `onKeywordClick()`. Previously `onAdvancedFilterClick()` routed `assigned` through `toggleBooleanInFilter`, so one click gave `assigned: true`, a second `assigned: false`, a third cleared it.

Now `onKeywordClick()` calls `toggleKeywordInFilter`, which inserts a bare `assigned:` with no value; `toWildcardParam` drops empty values, so no `assigned` query parameter is sent and the container list is not filtered at all until the admin types `true`/`false` by hand.

`container-table-actions.component.ts` also still carries the now-orphaned `onAdvancedFilterClick()` and `advancedFilterTrigger` ViewChild, whose template anchor was deleted.

### 5. The new shared filter components have no consumers

`shared/multi-select-filter/` (+ `multi-select-menu/`), `shared/filter-value-button/`, and the `.filter-select-menu.mat-mdc-menu-panel` override in `styles/styles.scss:848`.

Roughly 250 lines of component code, three spec files, a global CSS override and the `--copy-button-opacity` / `--filter-button-opacity` hooks landed with no caller — nothing outside those folders references `app-multi-select-filter`, `app-multi-select-menu` or `app-filter-value-button`. They also overlap with the existing `shared/multi-select-only`. Since the conditional-access work they were extracted from is still to come, they are unexercised in the real UI in the meantime.

Two small extras in the same area:

* `styles/styles.scss:848` still contains a leftover `// REVIEW: ...` comment suggesting `min(90vw, 480px)` instead of `max-width: none`.
* `container-details.component.html:61` and `token-details-description.component.html:60` changed the em dash `—` to a plain hyphen `-`; the em dash was the deliberate empty-value placeholder.

jelinaunger:
1) can be done here #5668

3) I think the paginator can completly go. Usually a container will not contain more than 5 tokens.

5) is used in conditional access branch and can now be used for the realm filter on token / container lists (#5617)

PR #5675, opened 2026-07-28
container audit tests

### LINKED ISSUE #5664: container set_realms sets wrong audit success value
even on success, request is logged as failed

jelinaunger:
closes #5664


nilsbehlen (commented):
Reviewed only the changes on this branch (`git diff origin/master..HEAD`), leaving out what came in with the master merge.

The audit-coverage direction is good, and the shared helpers in `api_container_common.py` are a real improvement over the per-file copies. Two themes in the inline comments:

**1. The `set_realms` success expression.** The change removes a genuine false negative, but the new comment claims more than the code delivers: refused realm removals still audit as successes, and in fact can't be detected from the returned dict at all. Details inline.

**2. The new assertions are weaker than their volume suggests.** Roughly 200 of the `success=0` assertions are satisfied by the `before_request` default and stay green even if the endpoint's audit call is deleted outright; all twelve `assert_no_audit_entry` calls are unfalsifiable by construction; and `try_unspecific=True` means the dispatch that gets inspected is never the one production uses by default. Since audit coverage is the whole point of the branch, these look worth tightening before merge.

Two notes that don't fit on a diff line:

- **Unused imports.** The cleanup covered `test_api_container.py`, `_api.py` and `_helpdesk.py` (zero findings each), but `ruff check --select F401 tests/test_api_container*.py` still reports ~190 in the rest: `_admin.py` 52, `_user.py` 47, `_template.py` 46, `_policy_conditions.py` 44. Presumably leftovers from the original file split.
- **Pre-existing, not from this PR.** In `delete_container_info_entry` the audit call passes `"key": key`, but `key` is not an audit column, so `finalize_log` drops it and the row never records *which* info key was deleted (`action_detail` stays empty). That is already on master, so it is out of scope here, but you are touching that call and `action_detail` is right there if you want it.

privacyidea/api/container.py
```diff
@@ -727,7 +727,8 @@ def set_realms(container_serial):
 
     # Set realms
     result = set_container_realms(container_serial, realm_list, allowed_realms)
-    success = False not in result.values()
+    # "deleted" is a status flag, not a per-realm result: exclude it so it does not poison the success value.
+    success = all(v for k, v in result.items() if k != "deleted")
```
nilsbehlen: The direction here is right: `False not in result.values()` audited a *successful* no-op replacement as `success=0`, because `containerclass.set_realms` seeds `result = {"deleted": False}` and leaves it `False` whenever nothing needed removing. Worth being precise about what the new expression does and does not buy, though.

`deleted` means "at least one realm was actually removed", so it cannot distinguish success from refusal in either direction:

| Container in | Admin allows | Request | `result` | old | new |
|---|---|---|---|---|---|
| r1, r2 | r1 | `realms=realm1` | `{deleted: F, r1: T, r2: T}` | 0 | 1 |
| r1, r2, r3 | r1, r3 | `realms=realm1` | `{deleted: T, r1: T, r2: T}` | 1 | 1 |

Both rows are the same refusal: `set_container_realms` unions `not_allowed_realms = set(old_realms) - set(allowed_realms)` back into `matching_realms`, so the removal is silently dropped and the container keeps a realm the admin asked to have removed. Row 1 is the case `test_api_container_helpdesk.py:646` now asserts as `success=1`, while also asserting both realms are still attached. Row 2 shows the old expression logged `success=1` for the very same refusal, so this is not a regression you introduced — but the new comment ("so it does not poison the success value") reads as though the refusal case is handled, and it is not.

The underlying reason is that refused removals never leave the lib: admin-scope refusals are folded back into `matching_realms` and never reach `res_failed` (which only covers refused *additions*), and owner-realm refusals are only `log.info`'d inside `containerclass.set_realms`. So no expression over this dict can detect them. If they should audit as failures, the place to fix it is the lib — report those realms as `{realm: False}`, both sets are already computed there — after which `all(...)` here becomes correct with no special-casing.

Minor, same line: `k != "deleted"` is a name comparison, so a realm literally named `deleted` is excluded too. If it is the only requested realm and it fails, the generator is empty, `all([])` is `True`, and you get `success=1` with no `info` detail. Reporting refusals from the lib and dropping the name filter would fix that as a side effect.

privacyidea/api/container.py
```diff
@@ -793,16 +794,25 @@ def delete_container_info_entry(container_serial, key):
...
-                        "success": res[key]})
-    return send_result(res[key])
+                        "success": success})
+
+    # delete_container_info returns False both for a reserved internal key and for a key that does
+    # not exist. Only the reserved-key case is an error; a missing key is a no-op returning False.
+    if not success:
+        container = find_container_by_serial(container_serial)
+        if key in container.get_internal_info_keys():
+            raise ParameterError("The key is reserved for internal use and cannot be deleted.")
```
nilsbehlen: This makes the reserved-key case 400 / `ERR905` (`ParameterError` carries `Error.PARAMETER = 905`), while the sibling "set info" path reports the same condition as 403 / `ERR303`, because `lib.container.add_container_info` raises `PolicyError`. The docstring above moves from `:status 403:` to `:status 400:`, so the two halves of the same feature now disagree on the wire, and a client that already special-cases the reserved-key message on 403/303 will fall through to generic bad-parameter handling here.

Probably worth picking one and using it on both endpoints. `PolicyError` / 403 reads as the more accurate of the two: the key is not a malformed parameter, it is an operation that is not permitted.

Separately, the message is a bare literal, so it is never translated. Other user-facing errors in this module wrap it, e.g. `raise ParameterError(_("Missing parameter 'realm', 'resolver', and/or 'user_id'"))`.

privacyidea/api/container.py
```diff
@@ -793,16 +794,25 @@ def delete_container_info_entry(container_serial, key):
...
     g.audit_object.log({"container_serial": container_serial,
                         "key": key,
-                        "success": res[key]})
-    return send_result(res[key])
+                        "success": success})
+
+    # delete_container_info returns False both for a reserved internal key and for a key that does
+    # not exist. Only the reserved-key case is an error; a missing key is a no-op returning False.
+    if not success:
+        container = find_container_by_serial(container_serial)
```
nilsbehlen: This re-loads the container to work out *why* the lib returned `False`, but the lib already knew: `containerclass.delete_container_info` is the code that skipped the `PI_INTERNAL` entry, and it discards that by returning an indistinguishable `False`.

Because the rule lives in the API layer, the other callers keep the old silent behavior — `lib/eventhandler/containerhandler.py` (delete all info) and `cli/pitokenjanitor/utils/findcontainer.py` still drop reserved keys without a word. So the same operation now errors over REST but quietly succeeds via the event handler.

Moving the distinction into the lib (raise there, or return the skipped keys separately) would cover all three entry points and let this drop the extra `find_container_by_serial` round trip.

tests/api_container_common.py
```diff
@@ -89,9 +95,39 @@ def request_assert_404_no_result(self, url, data: dict, auth_token, method='POST
...
+    def assert_audit_entry(self, action, **expected):
+        """Assert the most recent audit entry for the given action holds the expected column values.
+
+        Clears the audit log afterwards so the next assertion can only match a fresh entry: a request
+        that logs nothing (e.g. rejected before the audit is written) then finds an empty log and fails
+        loudly instead of silently matching a stale entry from an earlier request.
+        """
+        entries = self._audit_entries(action)
+        self.assertTrue(entries, f"No audit entry found for action {action!r}")
+        entry = entries[0]
```
nilsbehlen, line 150: `entries[0]` combined with the `clear()` on line 118 means only the newest entry is ever inspected and the rest are destroyed. That interacts badly with `request_assert_error(..., try_unspecific=True)`, which dispatches the same request twice: once with specific messages, then again under `hide_specific_error_message`. The following `assert_audit_entry(...)` therefore only ever checks the *hidden-error* dispatch, and wipes the specific-error entry unexamined.

So at the eight `try_unspecific=True` call sites in `test_api_container.py` (260, 286, 318, 388, 397, 563, 580, 612), the path production uses by default is never actually asserted. If it regressed to writing no entry at all, or to `success=1`, those tests would still pass.

Clearing the log between the two dispatches, or asserting both entries, would close the gap.

tests/api_container_common.py
```diff
@@ -89,9 +95,39 @@ def request_assert_404_no_result(self, url, data: dict, auth_token, method='POST
             self.assertEqual(404, res.status_code, res.json)
         self.reset_flask_g()
 
+    def _audit_entries(self, action):
+        """Most-recent-first audit entries for an action, read directly via the audit lib.
+
+        Querying the audit backend directly (instead of the ``/audit`` API) avoids depending on the
+        ``auditlog`` policy, which tests may restrict, and skips the full request dispatch.
+        """
+        return getAudit(self.app.config).search({"action": action}, sortorder="desc", page_size=1).auditdata
```
nilsbehlen: Each call here builds a brand-new audit object, and `assert_audit_entry` calls this once and then `getAudit(...).clear()` again — two per assertion. `TestingConfig` leaves `PI_AUDIT_NO_SIGN` and `PI_AUDIT_NO_PRIVATE_KEY_CHECK` unset, so every one of those re-reads and validates the RSA audit keys from disk and opens a new session.

With ~470 `assert_audit_entry` calls, that is on the order of a thousand RSA key loads spent purely on assertions, in what is already one of the slower test packages. Building one object in `setUp` (`self.audit_object = getAudit(self.app.config)`) and reusing it would remove all of it.

tests/test_api_container_api.py
```diff
@@ -879,9 +925,16 @@ def test_24_delete_container_info_fail(self):
                                                                 info_type=PI_INTERNAL)])
 
         # Try to delete internal key
-        result = self.request_assert_success(f"/container/{container_serial}/info/delete/internal_test",
+        result = self.request_assert_error(400, f"/container/{container_serial}/info/delete/internal_test",
+                                           {}, self.at, "DELETE", 905,
+                                           "ERR905: The key is reserved for internal use and cannot be deleted.")
+        self.assert_audit_entry('DELETE /container/<string:container_serial>/info/delete/<key>', success=0)
```
nilsbehlen: This assertion passes without the endpoint doing anything: `before_container_request` already seeds the entry with `success: False` and `container_serial` for every request (`api/before_after.py`; `container_serial` arrives through `view_args` via `get_all_params`).

I checked it by deleting the entire `g.audit_object.log({...})` block from `delete_container_info_entry` — every new `success=0` assertion on this branch stays green, including this one. That is roughly 200 of the new assertions, so the failure paths are not really covered: a change that stopped auditing them would not be caught.

The `success=1` assertions are meaningful, since those require the endpoint to actively overwrite the default. For the failure paths, asserting a column the endpoint itself sets — `action_detail`, `info`, or `serial` — alongside `success` would make them bite.

tests/test_api_container_api.py
```diff
@@ -419,15 +423,18 @@ def test_09_set_realms_fail(self):
...
         # Set non-existing realm
         payload = {"realms": "nonexistingrealm"}
         result = self.request_assert_success(f'/container/{container_serial}/realms', payload, self.at, 'POST')
+        self.assert_audit_entry('POST /container/<string:container_serial>/realms', success=0)
         result = result.get("result")
         self.assertFalse(result["value"]["nonexistingrealm"])
 
         # Missing container serial
         self.request_assert_405('/container/realms', {"realms": [self.realm1]}, self.at, 'POST')
+        self.assert_no_audit_entry('POST /container/<string:container_serial>')
```
nilsbehlen: This one cannot fail, for two independent reasons:

1. `POST /container/<string:container_serial>` is not a registered rule — that path only has a `DELETE` route — so the string can never appear in the audit `action` column, which is built as `f"{request.method} {request.url_rule}"`.
2. For the 405 on the line above, no rule matched at all, so no blueprint `before_request` runs, no `g.audit_object` is created, and no entry is written. (I confirmed the empty-log half with a throwaway probe test.)

So this passes for any string, `''` included. The same pattern appears at lines 71, 482, 516, 552, 557, 636, 656 and in `test_api_container_template.py:77,84`. The ones at 557 interpolate a concrete serial (`f'POST /container/{container_serial}/info/'`), which likewise never matches, because the action column stores the rule pattern rather than the request path.

If the intent is "a mis-routed request writes no audit entry at all", asserting that the log is empty (without an action filter) would express that and would actually fail if it ever stopped holding.

privacyidea/api/container.py
jelinaunger: Kept the {realm: bool} response shape and fixed the correctness behind it: set_container_realms now returns 
  a ContainerRealmsResult dataclass (attached / not_added / removed / not_removed) with success = not         
  not_added and not not_removed, so refused removals and unremovable owner realms now audit as success=0      
  instead of silently passing; the response stays a consistent "is this realm attached" map (attached → True, 
  requested-but-unattachable → False, plus deleted), and the full breakdown goes into the audit info. Deliberately didn't switch to an entirely new return structure (state strings / outcome lists / a public dataclass), which would be a breaking API change.

nilsbehlen (approved):
great

PR #5680, opened 2026-07-28
Send the assigned user's realm as a container list filter

nilsbehlen:
handleFilterInput adds a realm entry when filtering by user, but realm was not part of apiFilter, so filterParams dropped it and it never reached the backend. Add realm as an exact-match filter key, distinct from container_realm, and map the user_realm column to it.

### DIFF
--- a/privacyidea/static_new/src/app/services/container/container.service.spec.ts
+++ b/privacyidea/static_new/src/app/services/container/container.service.spec.ts
@@ -461,10 +461,16 @@ describe("ContainerService", () => {
       expect(containerService.containerFilter().filterMap.get("realm")).toBe("otherrealm");
     });
 
-    it("does not add a realm without a user filter", () => {
+    it("does not default the realm when no user filter is set", () => {
       containerService.handleFilterInput(filterInputEvent("type: generic"));
 
       expect(containerService.containerFilter().hasKey("realm")).toBe(false);
+    });
+
+    it("keeps a realm that is filtered without a user", () => {
+      containerService.handleFilterInput(filterInputEvent("realm: otherrealm"));
+
+      expect(containerService.containerFilter().filterMap.get("realm")).toBe("otherrealm");
     });
   });
 
--- a/privacyidea/static_new/src/app/services/container/container.service.spec.ts
+++ b/privacyidea/static_new/src/app/services/container/container.service.spec.ts
@@ -446,6 +446,40 @@ describe("ContainerService", () => {
     expect(containerService.filterParams()).toEqual({ type_list: "smartphone,yubikey" });
   });
 
+  describe("handleFilterInput realm handling", () => {
+    const filterInputEvent = (value: string): Event => ({ target: { value } as HTMLInputElement }) as unknown as Event;
+
+    it("adds the default realm when filtering by user without a realm", () => {
+      containerService.handleFilterInput(filterInputEvent("user: alice"));
+
+      expect(containerService.containerFilter().filterMap.get("realm")).toBe("realm1");
+    });
+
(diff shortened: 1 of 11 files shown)


privacyidea/static_new/src/app/services/container/container.service.spec.ts
```diff
@@ -446,6 +446,40 @@ describe("ContainerService", () => {
...
+    });
+
+    it("keeps an explicitly given realm", () => {
+      containerService.handleFilterInput(filterInputEvent("user: alice realm: otherrealm"));
+
+      expect(containerService.containerFilter().filterMap.get("realm")).toBe("otherrealm");
+    });
+
+    it("does not add a realm without a user filter", () => {
+      containerService.handleFilterInput(filterInputEvent("type: generic"));
```
jelinaunger, line 465: from the test description I would expect the realm filter needs to be set here, such that realm filter can only be used in combination with the username which is the current api behaviour. Actually, I can set only the realm filter returning no results.

jelinaunger: Or we change the api to allow user realm only filtering

nilsbehlen: thats a good idea, i will do that

nilsbehlen:
Addresses the second option from @jelinaunger's review — the API now supports filtering by the realm of the assigned user alone, instead of requiring it to be combined with a username. Tracked in #5684.

`GET /container/` folded `realm` into `request.User` and then always filtered on `TokenContainerOwner.user_id == user.uid`, which is `NULL` when no username was given — hence the empty result. `_create_container_query()` now applies realm, resolver and user id independently, the way `lib/token/query.py` already does.

Two cases that were silently answered with an empty list now return an error, matching `GET /token/`: an unknown `realm` gives a 404, and a `user` that cannot be resolved gives a 400. Without the latter the query would fall back to filtering by realm alone and return every container of that realm. Noted in `READ_BEFORE_UPDATE.md`.

The spec case you commented on keeps its behaviour — no realm is defaulted when there is no user filter — but is renamed to `does not default the realm when no user filter is set`, since the old name read as if realm required a user.

nilsbehlen:
Follow-up in db74307ab after reviewing the branch. **Correction to my earlier comment: an unknown `realm` no longer returns a 404.** It matches nothing and the endpoint answers 200 with an empty list, consistent with every other filter here (`type`, `container_realm`, `token_serial` all return an empty list for a value that does not exist). `READ_BEFORE_UPDATE.md` is updated accordingly and no longer announces a 404. As a side effect, a partially typed realm in the WebUI filter box now yields an empty table instead of an error response.

The realm is evaluated before the username, so an unknown realm behaves the same whether or not a `user` was given.

More importantly, the first version of the guard was keyed on the resolver instead of the user id. A user object can carry a resolver and still have no user id — namely when the login does not exist in that resolver — in which case the query fell back to filtering by realm and resolver alone. Since the user filter is the only scoping of the container list for a caller with role `user`, that returned other users' containers. The guard now requires the user id, and `test_17b_user_container_list_only_returns_own_containers` covers it.

Also in this commit: the resolver of the user object is matched with wildcards, so `realm` and `resolver=res*` can be combined (it silently matched nothing before); the user details page only sends `realm` together with the username; and the docstrings no longer claim that a resolver-only user object filters by resolver (such an object is empty and does not filter at all).

@jelinaunger the spec case you commented on keeps its behaviour and is renamed to `does not default the realm when no user filter is set`, plus a case asserting that a realm filtered without a username survives. Does that settle your point, given we went with your second option of changing the API?

jelinaunger (approved):

Issue #5638, opened 2026-07-23
Policy condition on the API client


nilsbehlen:
## Summary
Add a `client` policy **condition section** so policies can match on attributes
of the API client that made the request (resolved from `X-API-Key`).

## Motivation
Once requests carry a client identity, admins should be able to scope policies
to specific clients or client types — e.g. "enable feature X only for
`client_type == windows_cp`" — using the existing extended-conditions system
rather than a bespoke mechanism.

## Scope
- [x] New `ConditionSection.CLIENT` exposing non-secret client attributes:
      `id`, `display_name`, `client_type`, `status`, `key_id`.
- [x] Evaluate against `g.client_id`; thread `client_id` through the policy
      matching path (`Match` → `match_policies` → `filter_policies_by_conditions`
      → `condition.match`).
- [x] Section description in the policy-condition metadata (WebUI editor).
- [x] Tests, including missing-client behaviour.

## Acceptance criteria
- A policy with a `client` condition matches only when the request's client
  satisfies it.
- **Missing-client handling is documented**: on a request without an API key,
  the whole object is absent, so the condition must use
  `handle_missing_data = condition_is_false` to avoid raising and breaking
  policy matching for non-API-key requests.

---
**Part of** #5636
**Depends on** #5637

nilsbehlen:
not done in 3.14

nilsbehlen:
The `client` policy condition was implemented on branch `clients-api-keys` and then **removed** before merge. Reasons: (1) a free-text extended-condition with no client picker was not an acceptable way to select a client in the WebUI, and (2) client identity belongs as a first-class, secure policy dimension rather than a generic condition — the eventual replacement for the spoofable `user_agents` app identification. Superseded by #5711 (client-as-principal) and #5705 (integration vocabulary). Closing as not planned; PR #5694 no longer delivers this.

PR #5694, opened 2026-07-30
API clients with API keys and rotating remember-device cookie

### LINKED ISSUE #5636: API clients and persistent remember-device sessions (tracking)
## Overview
Introduce first-class **API clients** that authenticate with an API key
(`X-API-Key` header), and build a persistent "remember this device" mechanism
on top of them.

#5637, #5639 and #5640 are implemented on branch `clients-api-keys` (PR #5694),
documentation included. #5638 (a `client` policy condition) was implemented and
then **removed** — see below.

## Design notes
- API key format is `pi_<key_id>_<secret>`: `key_id` is a public, indexed
  lookup handle; only the secret half is hashed (HMAC-SHA256 keyed with
  `PI_PEPPER`). This keeps the lookup O(1) and never exposes the secret.
- Issuance is on `/validate/check`; **recognition** is a dedicated,
  non-authenticating endpoint `POST /validate/remember_device`.
- The persistent cookie holds only a rotating `series_id:counter` token — never
  the API key.
- Everything is gated by policy and defaults to **off**.

## Sub-issues
- #5637 — API clients authenticating with API keys (done)
- #5638 — Policy condition on the API client (**removed / superseded**: the
  free-text condition UX was unacceptable; client identity should be a
  first-class, secure policy dimension — see #5711 and #5705)
- #5639 — Rotating "remember device" session cookie / issuance (done)
- #5640 — Remember-device recognition, hardening & docs (done)
- #5696 — Admin: view and revoke remember-device sessions per client (done)

## Related
- #5695 — Remember this browser (separate WebUI/user-bound feature)
- #5705 — Unify the ecosystem-integration vocabulary (client types / user_agents / subscriptions)
- #5707 — Consolidate stale-row cleanup behind one scheduled mechanism
- #5711 — API client as a first-class principal (phased key enforcement + key-scoped rights)

### LINKED ISSUE #5637: API clients authenticating with API keys
## Summary
Add API clients that authenticate to privacyIDEA with an API key sent in the
`X-API-Key` header, plus admin management of those clients.

## Motivation
External integrations (e.g. credential providers, IdP plugins) need a stable,
revocable machine identity that is independent of a user session, so that
per-client behaviour can be configured and audited.

## Scope
- [x] `Client` model / `clients` table: UUID id, `key_id`, hashed secret,
      `status` (active/suspended), `config` (JSON), timestamps.
- [x] Key generation: `pi_<key_id>_<secret>`; store `key_id` (public, indexed)
      plus HMAC-SHA256 of the secret only. Plaintext shown to the admin **once**.
- [x] Request middleware: resolve `X-API-Key` into `g.client_id`.
      - Missing header: `g.client_id = None` (legacy callers unaffected).
      - Present but unknown/inactive: `g.client_id = None` (the request is unidentified, not rejected; endpoints that require a client enforce it).
- [x] CRUD API (`/clients`): create, list, update, rotate key, delete.
      Create/rotate return the plaintext key exactly once.
- [x] Admin policy actions: `clients_list`, `clients_add`, `clients_delete`,
      `clients_rotate`.
- [x] Old-WebUI management page (list / create / inline rename / inline status /
      rotate / delete).
- [x] Alembic migration and unit/API tests.

## Acceptance criteria
- Plaintext key is never stored and never retrievable after creation/rotation.
- An invalid/inactive/missing key leaves the request unidentified; endpoints that require a client (e.g. `/validate/capabilities`, `/validate/remember_device`) reject it, others are unaffected.
- Rotating a key invalidates the previous one immediately.

## Notes
- Client status is `active` / `suspended` (only `active` authenticates).
  Permanent removal is delete; a compromised key is handled by rotate — so there
  is no separate `revoked` status.

---
**Part of** #5636

### LINKED ISSUE #5639: Rotating "remember device" session cookie
## Summary
Issue a persistent "remember this device" cookie to an API client on successful
authentication at `/validate/check`, using a rotating-token scheme with theft
detection. Gated by policy, off by default.

## Motivation
Clients that repeatedly authenticate the same user on the same device should be
able to skip a step on subsequent auths, without weakening security: a stolen
cookie must be detectable and revocable.

## Scope
- [x] `AuthSession` model / `auth_sessions` table: `series_id` (PK), `counter`,
      `user_id`, `client_id` (FK to `clients.id`), ip/user-agent,
      created/last-used, `expires_at` (configurable via the
      `remember_device_validity` policy, default 30 days).
- [x] Cookie = `series_id:counter`, `HttpOnly` + `Secure` + `SameSite=Strict`;
      **never** contains the API key.
- [x] Rotating-token logic: create, parse, `consume_remember_device_cookie`
      (validate + increment counter; on counter reuse: delete the whole series —
      theft detection; on expiry: delete).
- [x] `remember_device` policy action (SCOPE.AUTH), **fail-closed** (no policy
      means no cookie).
- [x] Issue the cookie on `/validate/check` success when: the client opts in
      (`request_persistent_cookie`), `g.client_id` is set, the user is resolvable,
      and the `remember_device` policy allows it.
- [x] `GET /validate/capabilities` reports `remember_device` availability per client.
- [x] Migration and lib/API tests.

## Acceptance criteria
- Cookie only issued when opted-in, client-identified, user-resolvable, and policy-allowed.
- Presenting a stale counter invalidates the series (theft) - except the immediately-previous counter within a short, IP-bounded grace window (see #5640).
- Cookie carries no secret or API key.

## Note
Consumption is **recognition** on the dedicated `POST /validate/remember_device`
endpoint (#5640), not on `/validate/check`. Theft detection includes a bounded
grace window for concurrent requests.

---
**Part of** #5636
**Depends on** #5637

### LINKED ISSUE #5640: Persistent remember-device cookie: recognition, hardening & docs
## Summary
Let a remembered device be **recognised** so the calling client can skip the
second factor, with hardening and documentation. Finishing issue of the arc.

## Recognition is a dedicated endpoint (not consumption on /validate/check)
Recognition lives on **`POST /validate/remember_device`**, separate from
`/validate/check`:
- It answers one question — *is this device remembered for this user and
  client?* — and rotates the cookie on a hit.
- It is **not an authentication**: no credential is verified, no challenge is
  triggered, and it writes its own audit action rather than an auth ACCEPT. The
  client decides whether to skip the second factor from the boolean.
- `/validate/check` only **issues** (#5639); it no longer consumes a cookie.

Rationale: recognition ≠ authentication, so it must not pollute the auth endpoint
(audit noise, challenge side effects, CA accounting). Discovery is
`GET /validate/capabilities` (client-level).

## Grant vs. Conditional Access (decoupled)
The grant — "may this recognised device skip the second factor?" — is the
`remember_device` policy (scope auth) plus its realm/user scope. No dependency on
a Conditional Access feature. A future CA/risk layer is orthogonal: it would veto
auth globally and could consume the theft signal, but neither enables nor is
required by this flow.

## Scope
- [x] `POST /validate/remember_device` recognition endpoint: consume + rotate in
      one place. Hit → rotate + `Set-Cookie` + `value:true`; grace hit → `true`,
      no rotate; miss/theft → `false` + clear cookie. 401 without an identified client.
- [x] Bound to the authenticating user: matched on `(series_id, client_id,
      user_id)` **and** `user.exist()`, so deleting/removing the user revokes
      their remembered devices.
- [x] Theft/replay detection + baseline response: a stale counter invalidates the
      whole series and is audited (`action_detail`).
- [x] Concurrency grace window: `PI_REMEMBER_DEVICE_GRACE_SECONDS` (pi.cfg,
      default 10, `0` = strict). Single-step, time- and IP-bounded, non-forking.
- [x] Configurable cookie lifetime: `remember_device_validity` policy (days,
      default 30), scopable per realm/user.
- [x] Expired-session cleanup: `pi-manage config authsession cleanup` (chunked,
      deadlock-safe), shipped in the packaged crontab (NetKnights-GmbH/ubuntu#74).
- [x] Admin session management: `GET/DELETE /clients/<id>/sessions[/<series_id>]`
      + old-WebUI view (this is #5696).
- [x] Documentation: the recognition + capabilities endpoints, `remember_device`,
      `remember_device_validity`, `PI_REMEMBER_DEVICE_GRACE_SECONDS`, and the
      documented limitations below.

## Hardening (from the code review)
- Serial-only (userless) auth issues no cookie (it could never be recognised);
  over-long `user_id`/`user_agent` are truncated so a passed auth never 500s.
- Recognition evaluates the policy only when a cookie **and** a bindable user are
  present (no wasted eval/audit on empty polls).
- `touch_client` throttled; an unknown/inactive key leaves the request
  unidentified rather than 401-ing app-wide.

## Documented limitations (accepted, by design)
- `/validate/capabilities` is **client-level**, not per-user: `remember_device:
  true` means "available to this client / worth attempting"; the per-user answer
  comes from issuance/recognition.
- A lost-rotation retry beyond the grace window is treated as theft (series
  destroyed, re-registration required) — accepted cost of rotating-token theft
  detection; the user is never wrongly authenticated.

## Theft-response escalation (future seam)
Baseline theft response (invalidate + audit) ships now. The `except AuthError`
path in the recognition endpoint is the seam for escalations (notify, block, feed
a CA risk score). Not required for the initial feature.

---
**Part of** #5636
**Depends on** #5637, #5639

### LINKED ISSUE #5696: Admin: view and revoke persistent remember-device sessions per client
**Status:** implemented in PR #5694 (lib helpers + `/clients/<id>/sessions` API + old-WebUI "Devices" view).

## Summary
Give admins visibility into the persistent "remember device" sessions
(`auth_sessions`) issued to an API client, and the ability to revoke a single
session — without deleting the whole client or waiting for expiry.

## Motivation
The rotating cookie (#5639) creates `auth_sessions` rows, but today they are
never surfaced: an admin can only clear them via theft detection (automatic),
expiry, or deleting the client (cascade). There is no way to see active
remembered devices or revoke one. This is the companion to theft detection and
the standard "active devices / sign out this device" capability — it becomes
important for lost/compromised-device response once #5640's Conditional Access
lets a remembered device actually relax authentication.

## Scope
- [x] lib: `get_client_sessions(client_id)`, `revoke_client_session(client_id,
      series_id)` (scoped to the client so one client's id can't target
      another's session), `session_to_dict` (never exposes the token/counter as
      a usable secret).
- [x] API: `GET /clients/<client_id>/sessions` (list) and
      `DELETE /clients/<client_id>/sessions/<series_id>` (revoke one). Listing
      gated by `clients_list`, revoke by `clients_delete`.
- [x] Old WebUI: a per-client "Sessions" view (reached from the client list)
      showing user, IP, user agent, created / last used / expires, with a
      Revoke button per row.
- [x] Tests: list, revoke, revoke scoped to the owning client (404 for a
      series that belongs to another client).

## Notes
- Admin-side only. End-user "manage my trusted devices" is the separate
  remember-this-browser feature (#5695), where sessions are user-bound.
- The response must never return anything that lets the cookie be reconstructed
  as a usable credential.

## Relationship
Part of #5636. Complements #5639 (issuance) and #5640 (consumption/CA).

nilsbehlen:
## Summary
Introduces first-class **API clients** that authenticate with an API key
(`X-API-Key` header) and a persistent **"remember this device"** mechanism built
on top of them. Everything is policy-gated and off by default.

### API clients
- `Client` model / `clients` table. API key format `pi_<key_id>_<secret>`:
  `key_id` is a public, indexed lookup handle; only the secret half is hashed
  (HMAC-SHA256 keyed with `PI_PEPPER`), compared with `hmac.compare_digest`.
- Admin CRUD blueprint (`/clients`); create/rotate return the plaintext key
  exactly once. Gated by new `api_client_*` policy actions
  (`api_client_list/add/edit/rotate/delete`).
- `X-API-Key` middleware resolves the key into `g.client_id`. A missing, unknown
  or inactive key leaves the request **unidentified** (`g.client_id = None`);
  legacy callers are unaffected. Endpoints that require a client enforce it
  themselves (401) — an invalid key is not an app-wide 401.
- Client status is `active` / `suspended` (only `active` authenticates);
  permanent removal is delete, a compromised key is handled by rotate.
- Old-WebUI page: list, create, inline rename, inline status change, rotate
  (one-time key panel with copy), delete (inline confirm), and a per-client
  remembered-devices view.

### Remember-device (rotating cookie)
- `RememberedDevice` model: rotating `series_id:counter` token bound to the
  user's resolver-stable identity `(client_id, resolver, user_id, realm_id)` —
  so a remembered device survives a login rename and is never recognised for a
  different account that later reuses a freed login. Theft detection (a stale
  counter invalidates the whole series), a bounded grace window for concurrent
  requests (`PI_REMEMBER_DEVICE_GRACE_SECONDS`, default 10), and a configurable
  lifetime (`remember_device_validity` policy, default 30 days). An optional
  `remember_device_max_devices` policy caps the number of live devices per user.
  Cookie is `HttpOnly` + `Secure` + `SameSite=Strict` and never carries the API
  key.
- `remember_device` policy action (SCOPE.AUTH), fail-closed; requires an
  identified API client.
- **Issuance:** `POST /validate/check` with `request_persistent_cookie=1` issues
  the cookie on success (opt-in + policy + resolvable user). It does **not**
  consume cookies.
- **Recognition:** dedicated `POST /validate/remember_device` — answers "is this
  device remembered for this user+client?" and rotates the cookie. It is **not**
  an authentication (no credential, no challenge, its own audit action, no
  `result.authentication`); the calling client decides whether to skip the
  second factor. Recognition confirms the bound user still resolves, so deleting
  or removing a user revokes their remembered devices.
- **Discovery:** `GET /validate/capabilities` reports `remember_device`
  availability at the client level.
- Admin management of remembered devices, gated by `remembered_device_list` /
  `remembered_device_revoke`: `GET /clients/<id>/remembered_devices`,
  `DELETE /clients/<id>/remembered_devices[/<series_id>]` (per client, optionally
  narrowed to a `realm` or a `user`), and a cross-client
  `DELETE /clients/remembered_devices?realm=&user=` (single atomic delete) for
  realm-wide incident response.
- Expired-device GC: `pi-manage config remembered_device cleanup` (chunked,
  deadlock-safe), shipped in the packaged crontab (packaging repo PR:
  NetKnights-GmbH/ubuntu#74).

## Issues
- Closes #5637 — API clients authenticating with API keys
- Closes #5639 — rotating remember-device cookie (issuance)
- Closes #5640 — remember-device recognition, hardening & docs
- Closes #5696 — admin: view/revoke remembered devices per client
- Part of #5636 (tracking)

The `client` policy **condition** (#5638) was implemented and then **removed**
from this branch: a free-text extended-condition was not an acceptable way to
select a client, and client identity belongs as a first-class, secure policy
dimension (the eventual replacement for the spoofable `user_agents`). Tracked in
#5711 (client-as-principal) and #5705 (integration vocabulary).

## Notes
- Migrations included for both new tables.
- Documented limitations (in code + spec): `/validate/capabilities` is
  client-level, not per-user; a lost-rotation retry beyond the grace window is
  treated as theft and forces re-registration (accepted rotating-token trade-off;
  the user is never wrongly authenticated).


doc/configuration/api_clients.rst
```diff
@@ -0,0 +1,155 @@
...
+An absent, unknown or inactive ``X-API-Key`` simply leaves the request
+*unidentified* — existing integrations that send no key are unaffected. Endpoints
+that require an identified client enforce it themselves.
+
+
+"Remember this device"
+----------------------
+
+The remember-device feature lets a client skip the second factor for a device it
+has seen before, without weakening security: a stolen cookie is detectable and
```
jelinaunger: without weakening security might be misleading. People might think they can use this feature without any loss of security, but always using MFA is still the more secure way.

doc/installation/system/inifile.rst
```diff
@@ -632,3 +632,21 @@ accepted when set via an environment variable).
    the document structure and a size limit are enforced) so the Web UI can evolve
    its settings freely. ``PI_USER_SETTINGS_ALLOWED_KEYS`` will take effect once
    key enforcement is enabled.
+
+.. _ini_remember_device_grace:
+
+Remember-device grace window
```
jelinaunger, line 638: Out of curiosity, why is this a config option and not a policy? As it might be global and not scoped?

privacyidea/api/clients.py
```diff
@@ -0,0 +1,224 @@
...
+    if client_id:
+        clients = [get_client(client_id)]
+    else:
+        clients = get_clients()
+
+    g.audit_object.log({"success": True})
+    return send_result([client_to_dict(client) for client in clients])
+
+
+@clients_blueprint.route('/<client_id>', methods=['POST'])
```
jelinaunger: these are partial updates, right? Then it should be PATCH.

privacyidea/api/clients.py
```diff
@@ -0,0 +1,224 @@
...
+    :status 200: the updated client in ``result.value``.
+    """
+    param = request.all_data
+    display_name = get_optional(param, "display_name")
+    status = get_optional(param, "status")
+    config = get_optional(param, "config")
+
+    client = update_client(client_id, display_name=display_name, status=status, config=config)
+
+    g.audit_object.log({"success": True, "info": client_id})
```
jelinaunger: info could be f"Client ID: {client_id}" so it is no random number appearing there

jelinaunger: By the way, I think function arguments are commonly written to action details column and info mostly holds outcomes. But not sure, if this is a consistent pattern.

privacyidea/api/validate.py
```diff
@@ -735,6 +753,203 @@ def _finalize_auth_response(context):
     return ret
 
 
+def _resolve_persistent_cookie(user, success):
```
jelinaunger: add type hints

privacyidea/lib/authsession.py
```diff
@@ -0,0 +1,403 @@
...
+    # 500. user_id/user_agent are Unicode(255), ip_address is Unicode(64).
+    session = AuthSession(series_id=series_id, user_id=(user_id or "")[:255], client_id=client_id,
+                          ip_address=(ip_address or None) and ip_address[:64],
+                          user_agent=(user_agent or None) and user_agent[:255], counter=1,
+                          expires_at=expires_at)
+    session.save()
+    return session, build_cookie_value(series_id, session.counter)
+
+
+def session_user_id(user) -> str | None:
```
jelinaunger: type hint for the user

privacyidea/lib/authsession.py
```diff
@@ -0,0 +1,403 @@
...
+    only cost is that the device must re-register; the user is never wrongly
+    authenticated. Widening ``PI_REMEMBER_DEVICE_GRACE_SECONDS`` trades
+    theft-detection tightness for fewer such re-registrations.
+
+    :param cookie_value: the raw cookie value from the request
+    :param client_id: the id of the API client making the request (g.client_id)
+    :param user_id: the identifier of the authenticating user (see
+        :func:`session_user_id`)
+    :param client_ip: the source IP of the request; used to bound the grace
+        window (a grace hit must come from the IP the session was last used from)
```
jelinaunger: Can you add a short explanation of the return values here, especially the cool value, so it is clear what it means without reading the entire docstring

privacyidea/lib/clients.py
```diff
@@ -0,0 +1,283 @@
...
+        the secret does not match. ``client`` is the matched client (even when
+        disabled), or ``None`` for an unknown/invalid key.
+    """
+    key_id, secret = parse_api_key(plaintext)
+    if not key_id or not secret:
+        return None, "unknown"
+    client = Client.query.filter_by(key_id=key_id).first()
+    if not client or not hmac.compare_digest(client.key_hash, _hash_secret(secret)):
+        return None, "unknown"
+    return client, client.status
```
jelinaunger: No need for a tuple return type, status info is already within the returned client object. In general, you are using tuple return types quite often. Some could be replaced by dataclasses, e.g. (cookie_value, expires_at) occurs multiple times.

privacyidea/models/client.py
```diff
@@ -0,0 +1,65 @@
...
+    client_type: Mapped[str] = mapped_column(Unicode(64), default='', nullable=False)
+    key_id: Mapped[str] = mapped_column(Unicode(32), unique=True, index=True, nullable=False)
+    key_hash: Mapped[str] = mapped_column(Unicode(64), nullable=False)
+    status: Mapped[str] = mapped_column(Unicode(16), default='active', nullable=False)
+    created_at: Mapped[datetime] = mapped_column(DateTime, default=utc_now, nullable=False)
+    last_used_at: Mapped[datetime | None] = mapped_column(DateTime, nullable=True)
+    config: Mapped[dict[str, Any]] = mapped_column(JSON, default=dict, nullable=False)
+
+    @log_with(log)
+    def __init__(self, display_name, client_type, key_id, key_hash,
```
jelinaunger: **`__init__`s are removable.** They only copy fields; SQLAlchemy already generates a keyword `__init__`, and both models are constructed at one keyword-only call site each. `AuthSession.counter` is asserted three times (column `default=1`, `__init__` default, and the call site), and `expires_at` is `nullable=False` with no column default, so the NOT NULL guarantee lives only in Python. `Client.config`/`status` already have column defaults, so that `__init__` is duplication too.                                                                           
                                                                                                                
  Caveat: a column `default=` fires only when the attribute is *unset* — explicit `None` suppresses it. Both callers pass `None` (`expires_at=None` when no `validity_days`, `config=config`), so the callers must omit the kwarg rather than pass `None`, or the default path inserts NULL. Note also that column defaults apply at flush, not construction.

tests/test_api_authsession.py
```diff
@@ -0,0 +1,353 @@
...
+    def test_02_invalid_api_key_rejected(self):
+        self.assertEqual(self._get("pi_deadbeef_nope").status_code, 401)
+
+    def test_03_remember_device_reflects_policy(self):
+        _client, api_key = create_client("caps client", "keycloak")
+
+        # No remember_device policy -> capability is False (default off).
+        res = self._get(api_key)
+        self.assertEqual(res.status_code, 200, res)
+        self.assertEqual(res.json['result']['value'], {"capabilities": {"remember_device": False}})
```
jelinaunger: use assertDictEqual and the argument order is wrong (not only here)

tests/test_api_clients.py
```diff
@@ -0,0 +1,308 @@
...
+    def test_07_unknown_key_is_not_audited(self):
+        # An unknown/garbage key must NOT create an audit note (avoid flooding).
+        self.setUp_user_realms()
+        with self.app.test_request_context('/validate/check', method='POST',
+                                           data={"user": "cornelius", "realm": self.realm1, "pass": "x"},
+                                           headers={'X-API-Key': 'pi_deadbeef00000000_nope'}):
+            res = self.app.full_dispatch_request()
+            self.assertEqual(res.status_code, 200, res)
+        entry = self.find_most_recent_audit_entry()
+        self.assertNotIn("API key presented", entry.get("action_detail", "") or "")
```
jelinaunger, line 309: would be nice to have these audit assertions for all api tests

privacyidea/lib/authsession.py
```diff
@@ -0,0 +1,403 @@
...
+        :func:`session_user_id`)
+    :param client_ip: the source IP of the request (bounds the grace window)
+    :return: a :class:`ConsumeResult`. Theft is reported as ``status ==
+        "theft"`` (the series has already been invalidated) rather than raised,
+        so the caller can answer "not recognised" uniformly.
+    """
+    try:
+        result = get_valid_session(cookie_value, client_id, user_id, client_ip)
+    except AuthError:
+        return ConsumeResult("theft", None, None)
```
jelinaunger: outcomes could be an enum

privacyidea/lib/clients.py
```diff
@@ -0,0 +1,283 @@
...
+KEY_ID_BYTES = 8
+# Bytes of entropy in the secret half. 32 random bytes (~256 bits) make an
+# offline brute force against the keyed hash infeasible regardless of the hash
+# speed, which is why a fast keyed hash (rather than a slow KDF) is used and why
+# exposing the key id publicly leaks nothing useful about the secret.
+KEY_SECRET_BYTES = 32
+# The states a client may be in. Only "active" clients can authenticate;
+# "suspended" is a reversible off-switch. Permanent removal is a delete, and a
+# compromised key is handled by rotation - so there is no separate "revoked".
+CLIENT_STATUS = ("active", "suspended")
```
jelinaunger: Could be an enum

tests/test_lib_authsession.py
```diff
@@ -0,0 +1,186 @@
...
+        expired.save()
+        live, _ = create_auth_session("john", client_id)
+        # Capture ids before the delete+commit expires the ORM instances.
+        expired_series = expired.series_id
+        live_series = live.series_id
+
+        deleted = cleanup_expired_auth_sessions()
+        self.assertGreaterEqual(deleted, 1)
+        self.assertIsNone(AuthSession.query.filter_by(series_id=expired_series).first())
+        self.assertIsNotNone(AuthSession.query.filter_by(series_id=live_series).first())
```
jelinaunger: Another test worth adding: An attacker has the current cookie and tries to create a fresh cookie by incrementing the counter

.gitignore
```diff
@@ -39,3 +39,8 @@ dist/
 
 # Local development environment variables (use .env.example as template)
 .env
+
+# Local Docker deployment secrets and backups (never commit these)
```
jelinaunger: hmm that didn't work :D

nilsbehlen: why? they are not in the diff

doc/installation/system/inifile.rst
nilsbehlen: this is a protocol/networking-type setting for handling duplicate requests. its not a configuration in the sense of the others, like "is the feature enabled", "how many days".

.gitignore
nilsbehlen: understood, removed the entries

jelinaunger (approved):

PR #5699, opened 2026-07-30
Fix list_policies

### LINKED ISSUE #5683: Policies with specified User Agent are missing from pi-manage config export
### Top-level intent
`pi-manage config export` drops/omits policies if they have a specific `User-Agent` condition defined (e.g., `privacyIDEA-WebUI`).

### Steps to reproduce

1. Navigate to the WebUI and create a new policy named `PolicyWithUA`.
2. Set a condition for this policy: **User Agent** = `privacyIDEA-WebUI`.
3. Run the export command via CLI: 
   ```bash
   pi-manage config export -t policy
   ```
4. Check the generated output file or stdout. The `PolicyWithUA` will be completely missing from the export list.

**Relevant DEBUG log snippet from the execution:**
```text
...
[2026-07-28 18:33:46,317][26280][140044797501504][DEBUG][privacyidea.lib.policy:383] Entering list_policies with arguments (<privacyidea.lib.policy.PolicyClass object at 0x7f5eb21bf310>,) and keywords {'name': None} (called from policy.py:3564)
...
[2026-07-28 18:33:46,490][26280][140044797501504][DEBUG][privacyidea.lib.crypto:343] Exiting decryptPassword with result HIDDEN (called from config.py:140)
[2026-07-28 18:33:46,534][26280][140044797501504][DEBUG][privacyidea.lib.policy:564] Policies after matching the user_agent=None: ['Policy1_UAnone', 'Policy2_UAnone', 'Policy3_UAnone']
[2026-07-28 18:33:46,542][26280][140044797501504][DEBUG][privacyidea.lib.policy:383] Exiting list_policies with result HIDDEN (called from policy.py:3564)
```

### Expected outcome

The `pi-manage config export` command should perform a complete dump of the specified configuration tables from the database regardless of the runtime execution context. It should bypass the active matching logic (like checking the environment's `user_agent`, which is inherently `None` when executed from the CLI).

### Actual outcome

The export utility utilizes the internal `list_policies()` method, which applies active contextual filters. Since the CLI context initializes `user_agent` as `None`, any policy that explicitly enforces a non-empty `User-Agent` condition gets dropped during the configuration export phase.

### Configuration

* **privacyIDEA version**: 3.13.3
* **Installation method**: Virtual Environment (venv / pip)
* **Python version**: 3.11.2
* **Operating system**: Debian GNU/Linux 12 (bookworm)
* **Webserver**: Apache/2.4.67 (Debian)
* **Token database**: MariaDB Ver 15.1

jona-samuel:
closes #5683


privacyidea/lib/policy.py
```diff
@@ -560,24 +560,26 @@ def list_policies(self, name: str | None = None, scope: str | None = None, realm
...
-        for policy in reduced_policies:
-            policy_matches = False
-            # The policy either matches if it has no user agents defined or if the user agent is contained in the list
-            if not policy.get("user_agents"):
-                # If no user agent is defined, we match this policy
-                policy_matches = True
-            elif user_agent:
-                policy_agents = [agent.lower() for agent in policy.get("user_agents") if agent]
-                if user_agent.lower() in policy_agents:
+        if user_agent is not None:
```
jelinaunger, line 563: this gard is right for listing policies and consistent with the other filter parameters. However, this function as a structural problem: It is used for three different purposes:
* listing: None = no filtering is fine
* Are there any policies in this scope?: None = no filtering fine
* policy matching: The user agent "filter" input comes from the request. If the caller does not define a user agent, this should only match generic policies without a user agent defined as well. However, it would now also match policies with a user agent condition.

To get rid of this problem without changing the entire code structure, the filter behaviour should be 
* None -> no filter applied at all, returns all policies
* "" -> only return policies without a user agent condition
Which should work as it is now, just add a test for the empty filter string behaviour.

The only thing we need to ensure to make this work is that the user agent read from the request always defaults to an empty string and never to None. Could be done e.g. in the Match init function.
Even if it shouldn't be a problem now, we should ensure no empty user agent strings can be set in a policy.

jelinaunger (approved):

PR #5703, opened 2026-07-30
Usage of app-multi-select-filter

### LINKED ISSUE #5617: Multi select filter for realm and token-type

frankmer:
Used app-multi-select-filter for realm and token type filter in the token list.


privacyidea/static_new/src/app/services/token/token.service.spec.ts
```diff
@@ -446,6 +446,49 @@ describe("TokenService", () => {
...
       req.flush(MockPiResponse.fromValue({ count: 0, current: 1, tokens: [] }));
     });
+
+    it("sends a single token type as a wildcarded type param", () => {
+      contentServiceMock.onTokens = signal(true);
+      tokenService.activeFilter.set(new FilterValue({ value: "type: hotp" }));
+      TestBed.tick();
+
+      const req = mockBackend.expectOne((r) => r.url === "/token/");
+      expect(req.request.params.get("type")).toBe("*hotp*");
```
jelinaunger, line 456: no need to make it a wildcard match if we know the exact type

jelinaunger (approved):

privacyidea/static_new/src/app/services/token/token.service.spec.ts
jelinaunger: Is fine for now, otherwise we are running into inconsistencies and it does not have huge performance issues anyway. But in general we might think about showing added wildcards in the filter input and not only adding them silently. Which would allow to explicitly not add them on selections, but not dropping them silently. Out of scope here anyway.

PR #5709, opened 2026-08-03
Introduced News /RSS Feed components

### LINKED ISSUE #5384: News / RSS Feed in the privacyIDEA Dashboard
The privacyIDEA Dashboard should contain a short sections with important information / news.
These could be sent via RSS Feeds.

# Why

It is important to note, that an admin running privacyIDEA does not only run a single system, but should be part of a privacyIDEA community. So the goal is, to help the admin get in touch.
Get in touch with community, get in touch with events, when privacyIDEA is joining some open source events or when technical webinars are held.

privacyIDEA admins are sometimes difficult to get in touch with. 
So this is a way to provide information to privacyIDEA admins, where they actually are.

We are all humans. So the privacyIDEA admin should also get the possibility to become part of a privacyIDEA community. Join webinars. Go to events and learn to know other humans, who are using privacyIDEA. 
So it is about building a community and getting in touch and not only running my own single privacyIDEA system in the dark.

# What

To do so, there should be a fixed info field on the dashboard. It should only consume maybe 10% of the space, but it should be visible.
The RSS should be used to

* display the latest news about the latest privacyIDEA version. This way, the admin always knows, if there is something to do (update)
* maybe displays latest news about the latest versions of the Plugins, he is using (see #5383 )
* should display news about the next webinar. So that the admin can increase his knowledge about privacyIDEA.
* contain information about the next fair /trade show, so that the admin can get in touch.

In addition to the RSS Feed we can also display:

* Which services are contained in your support level (attended update)

We might need to define different RSS Types, so that the privacyIDEA server can easily display the correct type of information at the right place.

These types could be:

* update information
* event hint
* security issues (thus the admin knows, if there is something to worry about)
* public holidays (Thus, admins with support know, when support will be available and when not)
* your servivces

frankmer:
Added an widget component to show the titles of the news. Also added a News Page to show all the News with more detail.


privacyidea/static_new/src/app/components/layout/navigation-self-service/navigation-self-service.component.html
```diff
@@ -92,5 +92,12 @@
         matIconName="receipt_long"
         title="Audit Log" />
     }
+    @if (authService.rssAge() > 0) {
+      <app-navigation-self-service-button
+        i18n-title
+        key="{{ this.ROUTE_PATHS.NEWS }}"
+        matIconName="campaign"
+        title="News" />
+    }
```
jelinaunger, line 101: should appear in the navigation for admins as well to make it accessible even without the widget and to show on which page we are currently. We might then need a sub menu entry for the dashboard itself as well. edit dashboard button should disappear on the news page (or move to an entire different location).

privacyidea/static_new/src/app/components/news/news.component.html
```diff
@@ -0,0 +1,51 @@
...
+        i18n>
+        The news feed is disabled by the
+        <a [routerLink]="rssAgePolicyLink">rss_age</a>
+        policy.
+      </p>
+    } @else {
+      <p
+        class="news-hint margin-8"
+        i18n>
+        The news feed is disabled.
```
jelinaunger, line 38: I would hide it entirely if disabled (consistent with other pages hidden if users do not have the right)

privacyidea/static_new/src/app/components/shared/news-list/news-list.component.html
```diff
@@ -0,0 +1,52 @@
...
+        <a
+          class="news-title"
+          [href]="item.link"
+          rel="noopener"
+          target="_blank">
+          {{ item.title }}
+        </a>
+        @if (item.date; as date) {
+          <span class="news-date">
+            {{ date | date: "shortDate" }}
```
jelinaunger: Should be the same format as other dates (just dropping the time). I would prefer the format used for logs YYYY-MM-DD (better readability if multiple dates occur below each other than the Aug 11, 2026 format)

privacyidea/static_new/src/app/components/shared/news-list/news-list.component.html
```diff
@@ -0,0 +1,52 @@
...
+            class="news-summary"
+            [innerHTML]="item.summary"></div>
+        }
+      </li>
+    }
+  </ul>
+} @else {
+  <div class="news-empty">
+    <mat-icon class="news-empty-icon">newspaper</mat-icon>
+    <span i18n>No news available.</span>
```
jelinaunger: add the configured time from the policy here, e.g. "No news within the last 90 days"

privacyidea/static_new/src/app/components/dashboard/widgets/news-widget/news-widget.component.html
```diff
@@ -0,0 +1,40 @@
...
+-->
+
+<app-widget-state [state]="state()">
+  @if (newsDisabled()) {
+    @if (canReadPolicies()) {
+      <p
+        class="news-hint"
+        i18n>
+        The news feed is disabled by the
+        <a [routerLink]="rssAgePolicyLink">rss_age</a>
```
jelinaunger, line 27: A link directly to the edit page might make more sense.

privacyidea/static_new/src/app/services/info/info.service.ts
```diff
@@ -0,0 +1,116 @@
...
+
+export interface NewsListItem {
+  title: string;
+  link: string;
+  channel: string;
+  summary: string;
+  date: Date | null;
+}
+
+export const NEWS_ITEM_LIMIT = 20;
```
jelinaunger: And what about the rest? They are silently hidden.

privacyidea/static_new/src/app/components/news/news.component.scss
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ **/
+
+@use "styles";
+
+.news-page {
+  @include styles.main-component-style;
+  box-sizing: border-box;
+  height: var(--content-max-height);
```
jelinaunger: height does not fit available space on the news page, content is truncated

privacyidea/static_new/src/app/components/shared/news-list/news-list.component.html
```diff
@@ -0,0 +1,52 @@
...
+
+ SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+@if (items().length) {
+  <ul class="news-list">
+    @for (item of items(); track $index) {
+      <li class="news-item">
+        <span class="news-channel">
+          {{ item.channel }}
```
jelinaunger, line 25: font size of the head row is too small on the news page. It is hardly distinguishable from the content

privacyidea/static_new/src/app/components/news/news.component.html
frankmer: It's just a fallback. A user should not see the navgation button when the news is disabled.

nilsbehlen (approved):

PR #5720, opened 2026-08-04
5719 new webui UI settings

### LINKED ISSUE #5719: New WebUI: UI Settings
should be continued..

- [x] create elevation and border-radius selector
- [x] group requests and add toggle to settings

henrik-falk:


privacyidea/static_new/src/app/components/dashboard/widgets/appearance-widget/appearance-widget.component.scss
```diff
@@ -0,0 +1,49 @@
...
+  @include widget.host;
+  width: 100%;
+}
+
+.appearance-widget {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  gap: var(--double-global-gap);
```
frankmer: --double-global-gap does not exist. However, you have refactored a lot in styles.scss, and nowadays we often use size suffixes like -xs, which we could use here too, e.g. --gap-m, --gap-l.

privacyidea/static_new/src/styles/styles.scss
```diff
@@ -81,28 +69,102 @@ body {
...
+    inset calc(-1 * var(--depth-x-3)) calc(-1 * var(--depth-y-3)) var(--depth-blur-3) var(--mat-sys-light);
+  // --elevation-5 with the vertical offset dropped, for surfaces whose bottom edge meets a bar
+  // (the active toolbar tab, the util groups). With the light straight up or down there is no
+  // horizontal component left and the glow disappears.
+  --side-glow:
+    var(--depth-x-5) 0 var(--depth-blur-5) var(--mat-sys-shadow),
+    calc(-1 * var(--depth-x-5)) 0 var(--depth-blur-5) var(--mat-sys-light);
+  // Surface shading: a gradient running light to dark when raised and dark to light when
+  // recessed. 90deg behind the dial angle puts its light end on the corner the light comes from.
+  --depth-gradient-angle: calc(var(--dial-angle, 340deg) - 90deg);
```
frankmer: this has default 340deg in appearance.service.ts line 60 the default is 320 and in light-source-sial.component.scss the default is 225deg

privacyidea/static_new/src/app/components/configuration/ui-settings/ui-settings.component.html
```diff
@@ -0,0 +1,137 @@
...
+  class="settings-container"
+  i18n-aria-label>
+  <div class="settings-header">
+    <h3
+      class="margin-8"
+      i18n>
+      UI Settings
+    </h3>
+    <button
+      (click)="resetSettings()"
```
frankmer, line 31: missing aria-label

privacyidea/static_new/src/app/components/dashboard/widgets/appearance-widget/appearance-widget.component.ts
```diff
@@ -0,0 +1,159 @@
...
+
+const DEPTH_LABELS: Record<DepthLevel, string> = {
+  flat: $localize`flat`,
+  subtle: $localize`subtle`,
+  default: $localize`default`,
+  strong: $localize`strong`,
+  "very-strong": $localize`very strong`
+};
+
+const CORNER_LABELS: Record<CornerLevel, string> = {
```
frankmer, line 81: corner labels also defined in ui-settings.components.ts but under an another label

privacyidea/static_new/src/app/components/configuration/ui-settings/ui-settings.component.scss
```diff
@@ -0,0 +1,75 @@
...
+  --mat-button-toggle-selected-state-text-color: var(--mat-sys-on-primary);
+}
+
+.settings-row {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  align-items: center;
+  justify-content: center;
+  margin: 8px;
```
frankmer, line 105: margin size is not a variable

privacyidea/static_new/src/app/services/appearance/appearance.service.ts
```diff
@@ -0,0 +1,196 @@
...
+  public applyStoredLightSource(level: unknown): void {
+    this.apply(this.lightSourceGroup, level);
+  }
+
+  public applyStoredCorners(level: unknown): void {
+    this.apply(this.cornerGroup, level);
+  }
+
+  public resetToDefaults(): void {
+    this.setDepth(this.depthGroup.fallback);
```
frankmer: Writes the cookie and it gets instantly overwritten two times.

privacyidea/static_new/src/app/components/shared/theme-toggle/theme-toggle.component.html
```diff
@@ -0,0 +1,32 @@
...
+  [matTooltip]="label()"
+  class="theme-toggle">
+  <input
+    (change)="toggle()"
+    [attr.aria-label]="label()"
+    [checked]="theme() === 'dark'"
+    type="checkbox" />
+  <span class="theme-toggle__button"></span>
+  <span class="theme-toggle__label">
+    <mat-icon>{{ theme() === "dark" ? "dark_mode" : "light_mode" }}</mat-icon>
```
jelinaunger, line 30: I think the button should show the target state. If I'm in light mode there should be the moon so it indicates switching to dark mode on click.

privacyidea/static_new/src/app/components/configuration/ui-settings/ui-settings.component.html
```diff
@@ -0,0 +1,137 @@
...
+
+  <div class="details-cards-grid">
+    <app-details-card
+      class="settings-card-xs"
+      i18n-title
+      title="Theme">
+      <div class="settings-row">
+        <app-theme-toggle />
+      </div>
+    </app-details-card>
```
jelinaunger, line 48: the and language have a slightly different height. Either make the theme button smaller (same height as input field) or increase the height of the language card.

privacyidea/static_new/src/app/components/configuration/ui-settings/ui-settings.component.html
```diff
@@ -0,0 +1,137 @@
...
+        <mat-button-toggle-group
+          (change)="selectDepth($event.value)"
+          [hideSingleSelectionIndicator]="true"
+          [value]="depth()"
+          aria-label="Depth"
+          i18n-aria-label>
+          @for (level of depthLevels; track level.value) {
+            <mat-button-toggle [value]="level.value">{{ level.label }}</mat-button-toggle>
+          }
+        </mat-button-toggle-group>
```
jelinaunger: selecting flat here does not remove the shadow and it is also not possible to remove it entirely using light source setting.

privacyidea/static_new/src/app/components/configuration/ui-settings/ui-settings.component.html
```diff
@@ -0,0 +1,137 @@
...
+      <div class="settings-row">
+        <app-light-source-dial
+          [items]="lightSourceDialItems"
+          [selected]="lightSource()"
+          groupName="light-source"
+          legend="Light source"
+          i18n-legend
+          (pick)="selectLightSource($event)" />
+      </div>
+    </app-details-card>
```
jelinaunger: A short explanation on these UI styling settings might be helpful. Not always visible, but e.g. using i icon

privacyidea/static_new/src/app/components/configuration/ui-settings/ui-settings.component.html
```diff
@@ -0,0 +1,137 @@
...
+            <mat-button-toggle [value]="level.value">{{ level.label }}</mat-button-toggle>
+          }
+        </mat-button-toggle-group>
+      </div>
+    </app-details-card>
+
+    <app-details-card
+      class="settings-card-m"
+      i18n-title
+      title="Light source">
```
jelinaunger: titlecase

privacyidea/static_new/src/app/components/configuration/ui-settings/ui-settings.component.html
```diff
@@ -0,0 +1,137 @@
...
+            }
+          </mat-select>
+        </mat-form-field>
+      </div>
+    </app-details-card>
+
+    <app-details-card
+      class="settings-card-m"
+      i18n-title
+      title="Pending requests">
```
jelinaunger: titlecase

privacyidea/static_new/src/app/components/dashboard/widgets/appearance-widget/appearance-widget.component.html
```diff
@@ -0,0 +1,41 @@
...
+  <div class="appearance-widget__row">
+    <app-theme-toggle />
+
+    <app-light-source-dial
+      [items]="items"
+      [selected]="selected()"
+      groupName="appearance-preset"
+      legend="Appearance preset"
+      i18n-legend
+      (pick)="selectPreset($event)" />
```
jelinaunger, line 30: the tooltip is broken here. Multiple different ones appear here while hovering over it.

nilsbehlen (commented):
put the "design" settings in an expansion panel (collapsed by default) please

privacyidea/static_new/src/app/components/dashboard/widgets/appearance-widget/appearance-widget.component.scss
henrik-falk: I will fix this soon. global gap should not exist in the first place, all values that are simply multiples of 8 dont need a separate variable.

privacyidea/static_new/src/app/components/shared/theme-toggle/theme-toggle.component.html
henrik-falk: The icon shows current state because this is a switch, not a button. i kind of like it like this better. if its a big problem i can change it, but its not very confusing either.

privacyidea/static_new/src/app/components/configuration/ui-settings/ui-settings.component.html
henrik-falk: i prefer not to have fixed height and rather let it have as much height as the content need. the width is fixed so both fit in the same row. a little bit of a gap but i think its better  without hard coded margins

privacyidea/static_new/src/app/components/configuration/ui-settings/ui-settings.component.html
henrik-falk: intentional. flat = no blur and 2px border all around. the light source only changes the color slightly

privacyidea/static_new/src/app/components/dashboard/widgets/appearance-widget/appearance-widget.component.html
henrik-falk: unable to recreate. i only get one tool tip depending on where i hover. each selection gets its own tooltip

privacyidea/static_new/src/app/components/configuration/ui-settings/ui-settings.component.html
henrik-falk: missed this one. i will add an explanation to the expansion panel

henrik-falk: done

privacyidea/static_new/src/app/components/configuration/ui-settings/ui-settings.component.ts
```diff
@@ -0,0 +1,123 @@
...
+  ];
+  protected readonly resetTooltip = $localize`Reset UI settings to defaults`;
+  protected readonly appearanceHint = $localize`Changes how the interface looks: corner rounding, shadow depth and the direction the light comes from. Saved for your account.`;
+
+  protected resetSettings(): void {
+    this.appearanceService.resetToDefaults();
+    this.themeService.setTheme("light");
+    this.uiPreferencesService.setShowLoadingUrls(false);
+    // Last: switching locale is a full-page navigation, which would cancel the requests above.
+    this.uiPreferencesService.switchLocale("en");
```
nilsbehlen: Being last does not protect the writes above. `switchLocale` awaits only its own locale POST and then calls `window.location.assign` (`ui-preferences.service.ts`), and that navigation aborts every other request still in flight.

So an admin on the German UI with depth `flat` who clicks reset gets an English page and a locally-default appearance (from the cookie cache), but the `depth` / `light_source` / `corner_radius` / `theme` / `show_loading_urls` POSTs are cancelled — the stored settings keep the old values, and the old appearance comes back on the next login in another browser.

Awaiting the writes (e.g. `forkJoin`) before switching, or skipping the locale step when it is already `en`, would make the reset stick.

privacyidea/static_new/src/app/components/configuration/ui-settings/ui-settings.component.ts
```diff
@@ -0,0 +1,123 @@
...
+    { value: "default", label: $localize`Default` },
+    { value: "round", label: $localize`Round` },
+    { value: "extra-round", label: $localize`Extra round` }
+  ];
+  protected readonly resetTooltip = $localize`Reset UI settings to defaults`;
+  protected readonly appearanceHint = $localize`Changes how the interface looks: corner rounding, shadow depth and the direction the light comes from. Saved for your account.`;
+
+  protected resetSettings(): void {
+    this.appearanceService.resetToDefaults();
+    this.themeService.setTheme("light");
```
nilsbehlen: The theme default is `"system"`, not `"light"` — `ThemeService` initialises `currentTheme` to `"system"` and falls back to it for absent or unknown stored values. Same line in `appearance-widget.component.ts` (`resetAppearance`).

An admin on a dark-mode OS who never picked a theme (i.e. following the system) presses reset and is pinned to light mode, stored as `"light"`. Since both new controls only toggle light/dark, there is then no way back to `"system"` from the new UI. `setTheme("system")` is the actual default.

nilsbehlen (commented):
Review of the branch-only changes (merge base `475fb7ee0` → `7b21605ae`), skipping what came in with the master merges. Jest (328 suites / 4015 tests), `ng lint` and `ng build` are green, and the deleted globals (`.action-card*`, `.action-docu`, `--global-*`, `table-container-styles`, `dialog-container-styles`, `MasonryDirective`) have no remaining users.

Four comments below. A fifth finding — `proxy.conf.json` pointing at a personal host — is already fixed by `bb5b266ee`.

privacyidea/static_new/src/app/components/dashboard/widgets/appearance-widget/appearance-widget.component.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  protected selectPreset(lightSource: string): void {
+    const preset = APPEARANCE_PRESETS.find((candidate) => candidate.lightSource === lightSource);
+    if (!preset) {
+      return;
+    }
+    this.appearanceService.setDepth(preset.depth);
+    this.appearanceService.setCorners(preset.corner);
+    this.appearanceService.setLightSource(preset.lightSource);
```
nilsbehlen: Each of these three setters persists on its own — `AppearanceService.set` → `store` → one `setSetting` POST — so a single preset click is three writes to `/user/settings`, with no debounce.

The dial is a native radio group, so arrow keys move the selection: focusing it and holding an arrow walks all stops and fires ~3 audited API calls per stop. The three writes also race, so the cached settings document can end up being the response that predates one of the keys just written (last response wins).

A single batched write per interaction, or a debounce in `store`, would fix both.

privacyidea/static_new/src/app/components/layout/navigation/navigation.component.scss
```diff
@@ -1,9 +1,35 @@
...
+// hex step, so the bar stands off the page by the same amount in both themes: light runs 5 to 8
+// points above a 90% page, dark the same above a 20% one, starting on surface-container-high.
+$nav-active-first: #f3f3f3;
+$nav-active-start: #f4f4f4;
+$nav-active-end: #f7f7f7;
+$nav-active-first-dark: #404040;
+$nav-active-start-dark: #414141;
+$nav-active-end-dark: #444444;
+// One per section, plus the overflow "More" tab.
+$nav-item-count: 10;
```
nilsbehlen, line 24: The tone ramp is sliced with `nth-child` against this fixed count, but the number of `.nav-item` children is dynamic: `getFilteredNavItems()` drops sections the admin has no rights for, and `visibleNavCount` shrinks further on a narrow window (overflow "More").

For a restricted admin, or a 4-tabs-plus-More window, the active tones then come from only the first fraction of the ramp while `.utils-group` keeps the fixed far-end tone (`#f8f8f8` / `#454545`), so the tabs no longer line up with the bar gradient they are meant to be a slice of. Deriving the weight from the actual count (or using `nth-last-child`) keeps it in step.

nilsbehlen (approved):

PR #5722, opened 2026-08-05
Added state gates to all tables

### LINKED ISSUE #5663: zero state of tables
Image
this is the "zero state", exemplary here for the container table.

Firstly, "No data matching the filter" is irritating, since no filter was applied, so it looks like a broken state (only in case there are actually no objects at all, if a filter was applied and existing objects were filtered, everything should stay as it is now, its just about the zero state)

A nicer approach would be to show a text like "No containers exist yet, create your first" and then the create button, prominently in the middle. all other elements have no use if there is no object (container in this case), so no filter, pagination or action buttons, just the text and the create button.

frankmer:
closes #5663


henrik-falk (changes requested):
Image

henrik-falk (changes requested):
font size and icon size should be consistent across the whole webui.

nilsbehlen:
> font size and icon size should be consistent across the whole webui.

i dont think that is true, depending on what the context is. in the screenshot you showed, the zero state is surrounded by other elements. in the token table and other overviews, there is nothing. a small text and small icon - which would fit the previous context - would look lost there.

But also the message in that screenshot does not fit, i didnt check the user page when i looked at this.

henrik-falk (dismissed):
configuration/machines and /events is broken for some reason

henrik-falk:
> > font size and icon size should be consistent across the whole webui.
> 
> i dont think that is true, depending on what the context is. in the screenshot you showed, the zero state is surrounded by other elements. in the token table and other overviews, there is nothing. a small text and small icon - which would fit the previous context - would look lost there.
> 
> But also the message in that screenshot does not fit, i didnt check the user page when i looked at this.

At least keep the h2 font size (better h3 or h4) and not this hard coded font var(--mat-sys-title-large). 
For the icon why 5.5rem? it just doesnt fit with any other style.

nilsbehlen:
image
looks like that now.
machines list also fixed

nilsbehlen:
## Adjacent fixes and improvements

Gating the tables on their state turned up a number of problems next to the zero state itself. Listing them here so they are not a surprise in the diff — each is its own commit.

### Bugs found and fixed

**`configuration/machines` and `/events` were broken** — 28b7039a1

Both pages threw on first paint and never rendered their rows. The loading placeholder rows filled every column with `""`, and those two tables call `.join()` on their `hostname` / `event` value, so `"".join` threw once per placeholder row before any data arrived.

**Assigning a token to a user showed an empty table** — d628c0bb7

Four places assigned a signal to `MatTableDataSource`'s private `_sort`. Material reads `_sort.sortChange` / `_sort.initialized` whenever the table reconnects, and a signal has neither, so that became `merge(undefined, undefined)` and threw. Gating a table on its state is what introduced the teardown and rebuild that reaches it. The assignment was also doing nothing: `_orderData` passes the value to `sortData`, whose first statement reads `sort.active` — `undefined` on a signal — so the data came back untouched. Ordering is done by each component, not the data source.

Three specs asserted that assignment, so they passed while the feature was broken and went red when it was removed. They now assert the opposite.

**The token assignment control ignored its right** — 13556d9da

"Assign Token to User" on the user details page rendered with no rights check at all, unlike every other action there. Now gated on `assign`.

**Cancelling a confirmation still revoked or deleted a self-service token** — 963dc48b9

`revokeToken` and `deleteToken` on the self-service token table called the service as soon as the dialog closed, whatever the user had clicked, and used the answer only to decide whether to reload the table. So cancelling still performed the action, and skipping the reload hid it — the row stayed on screen until the next refresh. The sibling container table already guarded the call itself, which is what both now do.

**Wrong hint on the container zero state** — 13556d9da

It told admins to "assign a container to this user" on a page that only offers creating one.

**The session countdown made the user panel jump** — 1a5527ebd

The countdown changes format at whole hours and at ten minutes, and a session length set to exactly one of those — 3600 seconds — resets it onto the boundary, so the first tick after every reset flipped "1 h 00 min" to "59 min" and the next activity flipped it back. The format is now chosen from the time rounded up to the minute, and the countdown holds one width with equally-spaced digits.

### From the code review

**Self-service token list was empty for every user** — 52d27cd1d

The token table returns an empty data source when `tokenlist` is not allowed, and the self-service table inherits it. `tokenlist` exists in the admin policy scope only, so self-service users got an empty table under a paginator reporting the real count.

**`/users/realms` threw during its first load** — ff75addce

The resolvers cell reads `resolverGroups`, which is not one of the column keys, so the placeholder rows reached it as `undefined`. The same commit drops a `variant` attribute from that table's state panel — the component has no such input, so it never did anything.

**A container could hold on to its empty state while rows were present** — 0e0b831bf

The data source keeps one `MatTableDataSource` and swaps its rows, handing back the same reference, so the default reference equality reported "unchanged" and the table state kept its previous count.

**A failed request with an active filter was a dead end** — 8825f04bf

The state panel takes the place of the filter row, so a failure left the retry as the only control. Where the request carries the filter — audit, token, container, challenges, user — retrying repeats exactly what failed with no way to reach the filter. The error panel now offers to clear the filter alongside the retry, where a filter is actually narrowing the list.

### What a table shows while it loads — f3a14db55

Opening a table drew the filter row and a table of placeholder rows, then threw all of it away and put the empty panel in its place. The placeholders are shaped like data, so the page appeared to load rows and then lose them.

Withholding everything for a moment was tried first (61f3856c2), on the assumption that most responses beat it. Measuring this dev setup showed they do not: the five requests the token page fires take 320–480ms each in the browser. Serially each endpoint is 16–114ms; they are issued together and contend inside the single Flask process, and four of them are configuration that barely changes and need not be re-fetched per navigation — worth a look on its own.

**The panel now speaks for the load as well**: a spinner in the footprint the outcome will occupy, so an empty result fills in around it and rows replace it. One transition instead of two, in both directions, and it does not depend on the response being fast.

**That left the placeholder rows unreachable, so they were removed** in the same commit: the load guards in 25 components, the row builders behind them, the class the templates put on those rows, its styling, and the two helpers that existed to make them. Net −626/+312 lines across 78 files.

Both placeholder crashes above were in that machinery and cannot recur. The helpers pre-date this branch, but had two callers before it and twenty-two after, so nearly all of what went was added here — worth saying plainly rather than filing as inherited debt.

Two behaviours that had been sharing that code are kept and still covered: the token list keeps its count while a reload is in flight, and the user list keeps the rows already on screen.

### Improvements

**The zero state has two sizes** — 8d4430f10

Both steps come from the shared Material typescale rather than a hard-coded value, so a panel that owns the page and one sitting among a details page's sections are the same component at two sizes. This addresses the review comment about font and icon size consistency.

**Zero state on the subscription page** — 65f89125b

It renders cards rather than a table, so the state gating had nothing to attach to and the page showed an empty panel with just a title and an upload button.

**The user detail lists offer the ways out of the empty state** — 13556d9da

The actions are passed as templates instead of projected content, so the table places them by its own state: heading the bulk actions when there are rows, inside the state panel when there are none. A single `ng-content` slot renders in one place only, which is why the button could not follow the panel before. The token hint names only the routes this admin actually has, since enrolling and assigning are separately rights-gated.

### Tests

Patch coverage is **92.85%**, up from 66.14% when the gate first failed.

The original gap was not the new shared component but the ~29 call sites: the options handed to a `TableState` are callbacks, so they only run when something asks the state for its status, and nothing did. That left the read right, the row count and the filter predicate of every gated table unexecuted — including the name of the right, which is a plain string that nothing else checks, and where a typo silently leaves a table permanently denied or permanently visible.

A shared helper (`src/testing/table-state-gating.ts`) drives the state from denied to allowed and calls the two paths the state panel offers back to the page, so it is one line per spec across 27 specs. The shared panel itself, and the two self-service tables, also gained specs — one of those tables had only a `should create` case, the other lived as a second `describe` inside the admin table's spec.

One thing worth knowing when writing tests around this: do not stub `actionAllowed` with a constant. It reads no signal, so any `computed` over it caches its first verdict and never sees a rights change. Drive `authData` (or `jwtData`) instead.

### Deliberately left out

Three findings were kept out of scope, each its own change:

- The white content surface changes width per route, because nothing sets a width contract for it.
- The dark theme's container ramp straddles its own background, so no token reads as a raised surface.
- The first-level menu folds and unfolds "Subscription" at one specific window width, because the fold decision uses a different width budget depending on whether the "More" button is currently rendered — filed as #5781.

String extraction for the new zero-state wording is on its own schedule.

nilsbehlen (approved):

PR #5723, opened 2026-08-06
Generalize table row selection

### LINKED ISSUE #5668: Generalize table row selection
Row selection in tables is duplicated in many different components (toggleRow/toggleAllRows/isAllSelected functions). This should be generalized to remove redundant code and ensure same behaviour for all components.

frankmer:
Introduced a new RowSelector class that will be used by every Table that has selectable rows. 

I didn't use a service here because it would have been too complicated to handle with a service. Each table has different data and a different way to get the column key, so we would have to create an initialiser in the component constructor that populates the signal with data and a key getter, as well as an effect that keeps the data in the RowSelectorService up to date. I can think of many more ways in which this could be a service, but they would be even messier.


privacyidea/static_new/src/app/components/external-services/privacyidea-servers/privacyidea-servers.component.ts
```diff
@@ -92,6 +93,18 @@ export class PrivacyideaServersComponent {
     return dataSource;
   });
 
+  private readonly renderedRows = toSignal(
+    toObservable(this.privacyideaDataSource).pipe(
+      switchMap((dataSource) => dataSource.connect().pipe(finalize(() => dataSource.disconnect())))
+    ),
+    { initialValue: [] as PrivacyideaServer[] }
+  );
```
jelinaunger: might make sense to move this also into the selector as it is duplicated in multiple components.

jelinaunger (commented):
Tables have a slightly different behaviour, would be nice to unify this:
e.g. token table clears selection after filtering even if selected rows are still there
while e.g. in policies the selection is preserved

frankmer:
> Tables have a slightly different behaviour, would be nice to unify this: e.g. token table clears selection after filtering even if selected rows are still there while e.g. in policies the selection is preserved

Fixed it

jelinaunger (approved):

PR #5728, opened 2026-08-06
Keep translations attached when a template is reformatted

### LINKED ISSUE #5704: Reformatting a template silently orphans its translations
## Problem                                                                                                    
                                                                                                                
  Moving an `i18n` element from prettier's dangling-bracket form to the readable multi-line                     
  form adds a leading and trailing space to the extracted message. That changes the generated                   
  message id, so the translation no longer matches and the string falls back to English in all                  
  localised builds. Nothing fails — no build error, no lint warning, no test.                                   
                                                                                                                
  Angular collapses whitespace but deliberately does not trim it (a trailing space is sometimes                 
  content), and there is no compiler option to change that.                                                     
                                                                                                                
  ## Example                                                                                                    
                                                                                                                
  Same words, same `ng extract-i18n` run — only the formatting differs:

```html                                                                                                       
  <mat-hint align="end" i18n                                                                                    
    >Press enter to apply the filter</mat-hint                                                                  
  >                                                                                                             
  <!-- id 123340015457578972 — translated in all 12 locales -->                                                 
                                                                                                                
  <mat-hint align="end" i18n>                                                                                   
    Press enter to apply the filter                                                                             
  </mat-hint>                                                                                                   
  <!-- id 8324922110642776350 — source " Press enter … filter ", no translation anywhere -->                    
  ```                                                                                                           
                                                                                                                
  Roughly 13 strings on `master` are already affected this way.                                                 
                                                                                                                
  ## Options

  **A — Explicit message ids** (`i18n="@@tokenTable.filterEnterHint"`)                                          
  Matching happens on the id, so text and formatting stop mattering entirely. Also survives                     
  copy edits. Kills the whole class of bug. Costs a naming convention, and retrofitting                         
  everything would be a large mechanical diff across the `.xlf` files.                                          
                                                                                                                
  **B — Wrapper `<span i18n>text</span>`**                                                                      
  A single-attribute element stays on one line, so no whitespace gets injected. Cheap and                       
  readable, but adds a DOM node, and breaks again as soon as someone adds a second attribute                    
  to the span.                                                                                                  
                                                                                                                
  **C — `$localize` in the component, interpolated into the template**                                          
  Immune to template formatting. Good for long strings; poor as a blanket policy, since it                      
  separates the text from the markup.                                                                           
                                                                                                                
  **D — Keep prettier's dangling-bracket form**                                                                 
  Free and stable, but it is the form we reformatted away because it reads badly, so it will                    
  happen again.

### PR DESCRIPTION
**Pull request by frankmer**
Use ng-extract-i18n-merge to update the i18n hashes of the target
localizations when the source text is unchanged.

When the source text changes slightly, the target translation is kept
but marked as needing review. When the source text changes too much,
the target translation is dropped.

Also regenerate the xlf files.

Add a workflow that ensures the
localizations do not drift.
When the workflow fails run "npm run i18n:extract" in the project folder

### CODE REVIEW

File: .github/workflows/webui-i18n-drift.yml [hunk 1]
```diff
@@ -0,0 +1,120 @@
...
+# retires an id and mints a new one, and the localized build only warns about
+# the now-missing translation (i18nMissingTranslation defaults to "warning"),
+# so the string silently falls back to English in all 12 locales. This job
+# re-runs the extraction and reports whether src/locale is still in sync.
+#
+# Note for maintainers: src/locale is co-owned by Weblate (component
+# privacyidea-webui-next). A Weblate pull request that re-serializes a file in
+# Weblate's own XLIFF dialect will fail this check until the file is
+# normalized by running the extraction locally.
+name: WebUI i18n drift
```
jelinaunger, line 17: I know I said something different before, but taking a closer look I don't think we should add this workflow. It reports drift from reformatting, but equally from any new translatable string — and since we usually extract strings in their own PRs, it would be triggered constantly without telling us what we actually want to know.

privacyidea/static_new/angular.json
```diff
@@ -206,7 +206,33 @@
...
         "extract-i18n": {
-          "builder": "@angular-devkit/build-angular:extract-i18n"
+          "builder": "ng-extract-i18n-merge:ng-extract-i18n-merge",
+          "options": {
+            "buildTarget": "privacyidea-webui:build",
+            "builderI18n": "@angular-devkit/build-angular:extract-i18n",
+            "format": "xlf",
+            "outputPath": "src/locale",
+            "sourceFile": "messages.xlf",
+            "targetFiles": [
```
jelinaunger, line 216: Missing target files for da, lo, and pl. Don't forget to normalize them before adding here.

jelinaunger, line 216: To avoid forgetting to add new files here (or if new languages are added automatically by weblate), we can add a test checking each language file is added at all  6 places.

privacyidea/static_new/angular.json
```diff
@@ -206,7 +206,33 @@
...
+              "messages.it.xlf",
+              "messages.ta.xlf",
+              "messages.pt.xlf",
+              "messages.ru.xlf",
+              "messages.uk.xlf"
+            ],
+            "includeContext": true,
+            "includeContextLineNumber": false,
+            "fuzzyMatch": true,
+            "newTranslationTargetsBlank": false,
```
jelinaunger, line 233: Should be "omit" instead of false. Otherwise, the english text will be added as the translation and marked as translated in weblate.

jelinaunger (commented):
While it fixes most cases of minor changes requiring no changes in the translations (white spaces due to reformatting, typos, ...) it can also falsly match translations, e.g. "At least one event must be selected." now matches "Mindestens ein Status muss ausgewählt werden." wehre just the word "event" changed but the translations is now wrong.
It shouldn't be a big issue, as these translations should be marked in weblate to be reviewed. We just need to double-check this works well and fix them. 

Just as a note to be aware of the costs this introduces.

nilsbehlen:
we go with a different approach:
- daily extraction of strings via workflow. should then open a PR
- ideally these PRs would stack/overwrite so there only one open
- There is no "guessing" if strings are equal, they must be identical to have the translation preserved

PR #5548, opened 2026-07-02
update docker, add ci

nilsbehlen:
## Standalone single-node Docker deployment

Adds a self-contained, single-node Docker deployment for privacyIDEA under `deploy/docker/`, plus CI that builds it and boots the full stack. privacyIDEA previously had only a minimal/WIP Dockerfile; this gives a tested, easy-to-operate containerized deployment and a CI gate that proves it builds and runs.

### What's included
- **One image, three roles** (`compose.yaml`): a MariaDB plus the same privacyIDEA image run as `pi-init` (one-shot — create tables / migrate / bootstrap admin / install the enckey canary), `pi` (gunicorn web), and `pi-cron` (maintenance scheduler). `pi`/`pi-cron` wait for `pi-init` to complete, so migrations never race the workers.
- **Config is environment + secret files only** (no `pi.cfg`). enckey, pepper, secret_key, DB passwords and the audit-signing keypair are Docker file-secrets; `make init` generates them.
- **Maintenance** (`cron-runner.py`): challenge cleanup, audit rotation, usercache cleanup, and UI-configured periodic tasks — each toggleable/tunable via `PI_CRON_*`.
- **enckey canary**: refuses to start when the on-disk enckey does not match the database.
- **Backup / restore** scripts: logical DB dump bundled with the crypto keys, optional `age` encryption; restore also runs `db upgrade` so cross-version restores are safe.
- **Optional TLS** via a Caddy compose profile; otherwise the app binds localhost only.
- **Kerberos**: the image bundles `gssapi`, so SASL-Kerberos LDAP binds work (mount your `krb5.conf`).
- **`Makefile`** (`init`/`up`/`down`/`backup`/`restore`/`upgrade`/…) and a README covering configuration, env vars, TLS, backup/restore/DR, upgrades and troubleshooting.
- **CI** (`docker-build.yml`): static checks (shellcheck, hadolint, compose config), then build + `compose up` + smoke test (readiness, admin auth, cron) + a restart-idempotency check.

### Also
- `DockerConfig` accepts `PI_SECRET_KEY`(`_FILE`) as an alias for `SECRET_KEY`(`_FILE`) so all Docker secret variables share the `PI_` prefix (backwards compatible, unit-tested).
- The pip install docs now document the optional `[postgres]` / `[kerberos]` / `[hsm]` extras and their OS prerequisites.

PR #5733, opened 2026-08-07
build: add Docker publish workflow for GHCR

rophy:
## Docker publish workflow

Builds on the Docker deployment added in #5548 — that PR included CI (`docker-build.yml`) and a throwaway dev image workflow (`docker-dev-image.yml`), but neither publishes a production image. This adds the missing piece.

### What it does

- Triggers on `v*` tag pushes and via `workflow_dispatch` (manual, with a tag input)
- Filters out dev/pre-release tags (`dev`, `alpha`, `beta`, `rc`) — only clean releases publish
- Tags the image with the bare version number (e.g. `v3.13.3` → `ghcr.io/privacyidea/privacyidea:3.13.3`)
- Uses `fetch-depth: 0` so setuptools-scm derives the version from the git tag natively
- Follows the same conventions as the existing Docker workflows (SHA-pinned actions, `persist-credentials: false`, SPDX header, concurrency group, 40-minute timeout)

### What it doesn't do

- No `latest` tag — avoids ambiguity about which release is "latest"
- No multi-arch build — can be added later if needed
- No smoke test — that's already covered by `docker-build.yml` on PRs

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,0 +1,83 @@
+# SPDX-FileCopyrightText: (C) 2026 NetKnights GmbH <https://netknights.it>
+#
+# SPDX-License-Identifier: CC0-1.0
+#
+# Builds the single-node Docker image from a release tag and pushes it to GHCR.
+# Only fires on semver release tags (v1.2.3); dev/pre-release tags are ignored.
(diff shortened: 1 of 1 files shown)


nilsbehlen:
already exists https://github.com/privacyidea/privacyidea/blob/master/.github/workflows/docker-build.yml

how and if an image will be released is currently under discussion.

PR #5734, opened 2026-08-07
audit component: update download button styling and position

### LINKED ISSUE #5662: audit download button position and style
currently:

Image

move the download button to the top row. also make it either an icon only or make it not filled - we dont need to encourage to use this broken operation.

frankmer:
closes #5662


privacyidea/static_new/src/app/components/audit/audit.component.html
```diff
@@ -37,25 +37,26 @@
...
-        mat-stroked-button>
+        aria-label="Download Audit Log"
+        class="margin-left-8"
+        i18n-aria-label
+        i18n-matTooltip
+        mat-icon-button
+        matTooltip="Download Audit Log">
         <mat-icon>download</mat-icon>
-        <span i18n>Download</span>
       </button>
```
jelinaunger, line 54: the button should be vertically centered with the input only (excluding the hint)

jelinaunger (approved):

PR #5740, opened 2026-08-07
Enforce only known user setting keys

### LINKED ISSUE #5478: User Settings: Enforce only known keys
Follow up for #4220.
After implementing usages in the frontend and defining the keys used, we need to enforce only known keys.

frankmer:
closes #5478


privacyidea/lib/usersetting.py
```diff
@@ -48,33 +49,28 @@
...
+# Hard cap on the serialized document, so the column cannot be abused as
+# arbitrary per-principal storage.
 MAX_SETTINGS_BYTES = 16384
 
-# pi.cfg / environment option letting admins extend the set of accepted setting
-# keys without a code change, e.g. PI_USER_SETTINGS_ALLOWED_KEYS = ["foo", "bar"].
-# Accepts a list (pi.cfg) or a comma-separated string (environment variable).
-# Placeholder for now: get_allowed_keys() is wired up but not yet enforced (see
-# the TODO in validate_user_settings).
 USER_SETTINGS_ALLOWED_KEYS_CONFIG = "PI_USER_SETTINGS_ALLOWED_KEYS"
```
jelinaunger, line 57: add a short comment of the purpose of this config

privacyidea/lib/usersetting.py
```diff
@@ -147,14 +143,48 @@ def from_logged_in_user(cls, logged_in_user: dict, resolved_user: "User | None"
                    user_id=user.uid or "", resolver=user.resolver or "", realm_id=user.realm_id)
 
 
-def validate_user_settings(settings: dict) -> None:
+def _describe_unknown_keys(unknown: list) -> str:
```
jelinaunger, line 150: I think this can be a lot smaller. The worry is real, but I think both halves belong in another layer.

  **Escaping**: all three consumers already handle it — SecureFormatter for the log (lib/log.py:107-110), jsonify for the error response (api/lib/utils.py:152), and in the DB it only matters for the unquoted CSV
  export (sqlaudit.py:531). That export is genuinely broken, but that belongs in csv_generator (different issue).


  **Truncation**: that's the audit module's job and it's operator-configurable — PI_AUDIT_SQL_TRUNCATE and PI_AUDIT_SQL_COLUMN_LENGTH (doc/audit/index.rst:170-195), honoured in _truncate_data. A hardcoded 200   
  overrides the admin's setting from a lib function that can't see it, and under loggeraudit there's no column limit at all.


  **Suggestion**: report the count plus the first ~3 keys as examples, log the full list on debug, drop the escaping and MAX_REPORTED_KEYS_LENGTH.

privacyidea/lib/usersetting.py
```diff
@@ -48,33 +49,28 @@
...
-# Registry of the top-level setting keys the WebUI may store. Only used to seed
-# the (not-yet-enforced) allow-list; the backend stores the *values* verbatim
-# and never validates their structure. Admins can add further accepted keys via
-# PI_USER_SETTINGS_ALLOWED_KEYS without touching this registry.
-#
-# NOTE: these names are PLACEHOLDERS reflecting the intended settings. The
-# frontend team defines the real set; do not treat them as a committed API yet.
+# These are the keys the WebUI writes, as declared by the ``UserSettingKey`` type in
+# static_new/src/app/services/user-settings/user-settings.service.ts. A setting that
+# is added there has to be added here as well.
```
jelinaunger, line 61: an endpoint could return them instead of defining the list twice and keeping them in sync

tests/test_api_user_settings.py
```diff
@@ -39,15 +39,23 @@ def test_02_post_roundtrip_and_get_match(self):
         # GET returns exactly the same shape
         self.assertEqual(post_value, self._get().json["result"]["value"])
 
-    def test_03_open_mode_accepts_unknown_key(self):
-        # Key enforcement is not active yet (see the TODO in validate_user_settings),
-        # so the frontend may store any key.
+    def test_03_post_rejects_unknown_key(self):
         res = self._post({"settings": {"frontend_key": "v"}})
+        self.assertEqual(400, res.status_code, res)
```
jelinaunger, line 45: also check the parameter error code here, so it is clear if the test fails another error was raised before / instead

henrik-falk (approved):

PR #5756, opened 2026-08-12
4914/conditional access smtp identifier  selector

### LINKED ISSUE #5729: Conditional Access: SMTP identifier selector
- [x] get smtp server from backend
- [x] make it a selector
- [x] remove email actions if smtp server identifer are not readable

henrik-falk:


jelinaunger (commented):
Not sure if we should really remove the mail actions at all. Admins now do never know these actions exist and it might only be mistake they do not have the right as it is not that straight forward that CA policy configuration might require smtp server read rights. 
I would keep the actions and either fallback on a free text input field with a hint flagging the missing right or showing the empty selection with a hint / error  avoiding users create broken actions.

jelinaunger (approved):

PR #5765, opened 2026-08-13
Apply the resolver and userid filters of GET /token/

### LINKED ISSUE #5624: Token list: "userid" and "resolver" filters have no effect
The "userid" and "resolver" query parameters of the token list are accepted but never applied.

### LINKED ISSUE #5689: GET /token/ ignores the resolver filter
### Top-level intent

List the tokens of the users of one resolver, i.e. `GET /token/?resolver=<name>`, as an
administrator. The parameter is documented both in the endpoint and in
`get_tokens_paginate()`/`_create_token_query()`:

> `:param resolver: filter for the given resolver name`

### Steps to reproduce

1. A realm `realm1` with two resolvers, `resolver1` and `reso3`, holding different users.
2. A token assigned to a user of `reso3`, another token assigned to a user of `resolver1`, and a
   token with no owner at all.
3. As an administrator:
   * `GET /token/?resolver=reso3`
   * `GET /token/?realm=realm1&resolver=reso3`

### Expected outcome

Both requests return only the tokens whose owner is in `reso3`.

### Actual outcome

* `GET /token/?resolver=reso3` returns **all** tokens, including the tokens of `resolver1` users and
  the token that has no owner. The filter is not applied at all.
* `GET /token/?realm=realm1&resolver=reso3` returns **all tokens of the realm**, regardless of the
  resolver of their owner.

### Cause

The `resolver` parameter of `_create_token_query()` is never turned into a `where()` clause. It only
takes part in the decision whether to join `TokenOwner`:

```python
should_join_token_owner = (bool(userid and userid.strip("*")) or
                           bool(resolver and resolver.strip("*")) or
                           ...
```

The only resolver condition in the function is derived from the *user object*
(`if user.resolver: sql_query = sql_query.where(TokenOwner.resolver == user.resolver)`). That is why
the outcome depends on how the user object is built in the endpoint:

* With `resolver` alone, the user object is `User(login="", realm="", resolver="reso3")`.
  `User.is_empty()` deliberately ignores the resolver, so the object is empty, the whole
  `if user and not user.is_empty():` block is skipped and no filter is applied.
* With `realm` and `resolver`, `api/token.py` takes the branch
  `elif is_admin and realm_param: user = User(login="", realm=realm_param)`, which rebuilds the user
  object from the realm only and thereby **drops the resolver**.

### Suggested fix

Apply `resolver` as a filter of its own, independent of the user object, the way the container list
already does it in `_create_container_query()` (case-insensitive, `*` as wildcard, matched against
`TokenOwner.resolver`). The resolver then no longer needs to travel inside the user object to act as
a filter.

While doing that, the documented wildcard support should be verified: the container list accepts
`resolver=res*`, the token list currently documents `filter for the given resolver name` without
saying whether wildcards are honored.

nilsbehlen:
Fixes the `resolver` filter of `GET /token/` reported in #5689, and the `userid` filter which was broken the same way.

### What was wrong

The `resolver` and `userid` query parameters were accepted and documented, but never turned into a `where()` clause — they only took part in the decision whether to join `TokenOwner`. As a result:

* `GET /token/?resolver=<name>` returned **all** tokens, not the tokens of that resolver.
* `GET /token/?userid=<id>` likewise returned all tokens.
* `GET /token/?realm=<unknown>` raised a `404 ResourceNotFoundError`, while every other filter of the endpoint (including `tokenrealm`) returns an empty list for a value that does not exist.

### What changed

* `resolver` and `userid` are applied as independent filters of the token owner, mirroring `_create_container_query()`. `resolver` is matched case-insensitively and both honor `*` as a wildcard. Only tokens that are assigned to a user can match either filter.
* The user-object filter applies realm, resolver and user id independently, so a user object carrying only a realm lists the tokens of all users of that realm. A login that can not be resolved to a user id is rejected instead of falling back to a realm/resolver filter.
* An unknown realm now returns an empty list instead of a 404, consistent with `tokenrealm` and `GET /container/`.
* For user-role callers the request resolver is dropped before `request.User` is built, so the user object always reflects the caller's own identity.
* WebUI: the token table no longer greys out `resolver`/`userid` as "unsupported" filter keys, since they work again; the temporary "unsupported filter key" mechanism is removed.
* `READ_BEFORE_UPDATE.md` documents the two API behaviour changes (filters now applied; unknown realm returns an empty list).

### Tests

New coverage in `tests/test_lib_token.py` and `tests/test_api_token.py` for the resolver/userid filters (exact, case-insensitive, wildcard, unknown value, ownerless tokens), the unknown-realm empty list, and user-role scoping. Frontend specs updated for the removed mechanism.

Suites run individually: `test_lib_token.py` 77, `test_api_token.py` 86, `test_api_validate.py` 52, `test_api_container_user.py` 35, `test_api_roles.py` 42, `test_api_lib_policy.py` 35; jest 220 for the touched specs.

Closes #5689


jona-samuel (dismissed):
1. Add the userid filter to get_tokens so that get_tokens and get_tokens_paginate have the same parameter.


2. You automatically fixed the get_tokens_in_resolver function. Fix this test so that it actually shows if the function is broken, or remove it. https://github.com/privacyidea/privacyidea/blob/a2239968d4ed6ec8fd59fd384b40e01ee791bafa/tests/test_lib_token.py#L286-L288

nilsbehlen:
@jona-samuel good catch, added that.

jelinaunger (approved):

PR #5768, opened 2026-08-13
Fix replaceholder handling in WebhookHandler

### LINKED ISSUE #5748: WebHook handler: tag replacement drops all but the first key of a JSON payload
## Description

The WebHook event handler drops most of the payload when `replace` is enabled and the content type is `application/json`.

`replace_recursive()` in `privacyidea/lib/eventhandler/webhookeventhandler.py:169` returns from inside its `for` loop, so it processes only the **first** key of each dict and silently discards every other key.

```python
def replace_recursive(val):
    for k, v in val.items():
        k = k.format(**attributes)
        if isinstance(v, dict):
            return {k: replace_recursive(v)}   # <-- returns on the first key
        else:
            return {k: v.format(**attributes)} # <-- same here
```

## Reproduction

Configure a WebHook event handler with `content_type = application/json`, `replace = True` and a `data` payload with more than one key:

| `data` option | posted body |
| --- | --- |
| `{"user": "{logged_in_user}", "serial": "{token_serial}", "event": "login"}` | `{"user": "john"}` |
| `{"outer": {"a": "{realm}", "b": "keep"}, "top": "{token_serial}"}` | `{"outer": {"a": "realm1"}}` |

The keys are not merely left un-replaced — they are removed from the request body entirely, so the receiving system gets a truncated payload with no indication that anything was lost.

## Second problem: non-string values raise

`v.format(...)` is called on every non-dict value, so any JSON value that is not a string raises `AttributeError`:

- `{"count": 5}` → `AttributeError: 'int' object has no attribute 'format'`
- `{"tags": ["a"]}` → `AttributeError: 'list' object has no attribute 'format'`

The surrounding `try` only catches `KeyError` and `ValueError` (`webhookeventhandler.py:182-185`), and `lib/event.py:87` does not guard the `event_handler.do()` call. For a handler in the **pre** position the exception therefore propagates before the wrapped API function runs — so a `/validate/check` with such a webhook configured fails the request, rather than just failing to send the webhook.

JSON lists are not traversed at all, so placeholders inside arrays are never replaced even in the single-key case.

## Expected behaviour

- Every key of the payload is preserved, and placeholders are replaced in all of them, at any nesting depth.
- Values inside JSON arrays are traversed.
- Non-string values (numbers, booleans, `null`) are passed through unchanged instead of raising.
- A failure while rendering the payload never propagates out of the event handler.

## Notes

`replace_recursive()` only supports six placeholders, a subset of the tags the notification and logging handlers offer through `create_tag_dict()` (`privacyidea/lib/utils/__init__.py:1276`); `token_owner` is also filled with the given name rather than a full owner name. Replacing this function with `create_tag_dict()` would fix the bug and align the tag set at the same time, but that is a larger change than this bug needs — noting it here as the possible direction.

jona-samuel:
closes #5748


nilsbehlen:
Code review (medium effort) of this branch's changes against #5748's expected behaviour:

**1. `webhookeventhandler.py:162` — token lookups still run outside the try/except, so a failure there can propagate (CONFIRMED)**
The `get_tokens(serial=token_serial)` (line 163) and `get_tokens(user=tokenowner)` (line 168) calls added to resolve `tokentype`/`tokendescription`/`token_serial` execute before the try/except block that this branch added to swallow rendering failures. If either call raises — e.g. a resolver/DB error while looking up the token owner — the exception isn't caught by the `except (KeyError, AttributeError, IndexError)` / `except (ValueError, TypeError)` clauses further down, so it still reaches `lib/event.py`'s handler wrapper. Issue #5748's stated expected behaviour is "A failure while rendering the payload never propagates out of the event handler" — with `abort_on_error` set (or a `HandlerAbortError` subtype), this reintroduces exactly that class of problem, just from the data-gathering step instead of formatting. No test in this branch's diff exercises `get_tokens` raising here.

Suggest moving these two lookups inside the existing try block (or otherwise covering them) so a lookup failure is handled the same way a rendering failure is.

**2. `webhookeventhandler.py:160-169` — token lookup logic duplicates `usernotification.py:443-451` almost verbatim**
The two copies have already diverged slightly (this one guards `if tokenowner else []`, `usernotification.py` doesn't), so a future fix to this lookup logic is easy to apply in one handler and miss in the other.

**3. `webhookeventhandler.py:198` — dead branch in `replace_recursive`**
`k.format(**tags) if isinstance(k, str) else k` — `json.loads` always produces `str` keys for JSON objects, so the `else k` path is unreachable. Not a bug, just a branch that reads as reachable but isn't.

Overall: the recursive-replace rewrite fixes the reported truncation bug, traverses lists, and passes non-strings through correctly. Finding 1 is the one that matters for fully closing #5748's "never propagates" requirement.

nilsbehlen (changes requested):
see comment

nilsbehlen (approved):

PR #5770, opened 2026-08-13
Add biometric protection policies for Push enrollment

frankii91:
## Summary

Add two optional Push enrollment policy actions:

- `push_app_biometric_level=any|strong`
- `push_app_invalidate_on_biometric_change=true|false`

The selected values are added to the Push enrollment URL as
`app_biometric_level` and `app_invalidate_on_biometric_change`, allowing a
supporting authenticator app to select the required biometric class and bind
the Push key to the current biometric enrollment.

## Motivation

`push_app_force_unlock=biometric` specifies that the app should request
biometric authentication, but it does not describe the required biometric
strength or whether a change to the enrolled biometrics should invalidate the
existing Push key. These are separate security decisions and need separate
policy controls.

## Behavior

- `strong` requests strong biometric authentication from supporting clients.
- `any` permits the client to use any biometric accepted by the platform.
- `true` requests binding the Push key to the current biometric enrollment.
- `false` explicitly opts out of enrollment-change invalidation.
- If either policy is not configured, its parameter is omitted from the
  enrollment URL. Existing deployments therefore retain their current server
  behavior, and supporting clients may apply their own default.

The invalidation action is represented as a string enum rather than a boolean
policy action so that an unset policy remains distinguishable from an explicit
`false` value.

## Validation

- policy definition and policy helper tests
- REST policy create, update, read, and delete coverage
- Push enrollment API coverage for generated QR parameters
- Push token enrollment coverage for `strong/true` and `any/false`

Six focused regression tests pass. The changed Python modules also pass syntax
compilation after rebasing onto the current `master`.

### DIFF
--- a/doc/policies/enrollment.rst
+++ b/doc/policies/enrollment.rst
@@ -951,3 +951,26 @@ This policy can be used to force the privacyIDEA Authenticator App to secure the
 biometric. If you select any, the token can be unlocked with both.
 
 .. note:: This needs the privacyIDEA Authenticator app 4.6.1 or higher.
+
+push_app_biometric_level
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+type: ``string``
+
+For Push tokens, this policy requests either any system biometric (``any``) or
+a strong biometric (``strong``) from a supporting Authenticator App. It only
+refines ``push_app_force_unlock=biometric``; older applications ignore the
+additional enrollment parameter.
+
+push_app_invalidate_on_biometric_change
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+type: ``string``
+
+For Push tokens, this policy asks a supporting Authenticator App to protect the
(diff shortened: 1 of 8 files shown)


nilsbehlen:
hello, 
while we appriciate the engagement, you cannot just open pullrequests without an issue (bar one line fixes maybe), where we can discuss if the suggested enhancement is needed, where a bug comes from etc. moreover, the changelog is generated from issues in a milestone, so a minimum level of organized work is required.
there is already a setting to set if biometrics should be used, as your description correctly states. but this explanation:
> but it does not describe the required biometric strength or whether a change to the enrolled biometrics should invalidate the existing Push key. These are separate security decisions and need separate policy controls.

of how this differentiates from or extends the existing setting is insufficient and also does not convey why there should be another setting (`push_app_biometric_level=any|strong`) for the same thing. Also there is no link to the PR you opened for the authenticator app, which would be critical to further explain the context, e.g. what does the app do with `push_app_invalidate_on_biometric_change=true|false` , what behavior (change) can the server expect from the app etc.

PR #5784, opened 2026-08-18
CA - error messages

### LINKED ISSUE #5746: Conditional Access: Error messages
Currently, CA specific error message are only surfaced at /auth and can be suppressed with the policy `hide_specific_error_message`. All other auth endpoints always return a generic error message. This should be configurable at all. It should get it's own policy action with values none (do not show any ca error at all, default), webui (only surface error at /auth), all (show ca errors for all authentication endpoints.
We could also move this configuration into the LockoutPolicy itself, but I think it is not that policy specific and more a global setting which should be once defined for all ca policies.

Additionally, it would be nice to allow custom error messages. They can be defined per stage. This allows admins to define which information should be surfaced to the user (e.g. show the lock duration, but not that the user/admin got an email) and allow to give custom instructions.

jelinaunger:
Adds the one thing a conditional-access rejection could not do before: tell the user why. Opt-in per stage, silent by
default, and answered identically on every gated endpoint.

## What an admin configures

* **One free-text `error_message` per stage** (`Unicode(500)`, validated on save). A stage can lock, block and email
  at once; the admin writes one sentence and thereby resolves the arbitration themselves. Empty is the default and
  means *nothing is said*.
* **`{duration}` is the only substituted tag.** Every other brace expression is left exactly as written, so braces in
  prose need no escaping. Plain `str.replace`, not `format_map` — a bare `{}` would make that raise and silently
  return the template with the duration unsubstituted.
* **`show_default_ca_error_message`** (new, `conditional_access` scope, boolean): where a stage carries no wording of
  its own, fall back to the default sentence for the action that turned the request away. A stage's own message
  always wins. Without the policy, a stage with no message stays silent whatever its actions.
* **`GET /conditionalaccess/defaulterrormessages`** returns those suggested sentences as
  `[{"action_type", "message"}]`, most severe first — an authoring aid for the editor, and the same table the
  runtime fallback uses, so editing a suggestion edits what you would otherwise have got.

## What a rejection actually says

Silent by default: with nothing configured, a rejection says what every other failed authentication on that endpoint
says, and no more. It never says "you are locked" unless an admin wrote that.

| endpoint | ordinary failure | CA rejection, silent | CA rejection, worded |
|---|---|---|---|
| `POST /auth` | 401, `error.code=4031`, `detail` present | 401, `error.code=403`, same message, `details={}` | 401, `403`, the wording, `details={}` |
| `/validate/check` | 200, `value=false`, `REJECT`, `detail.message="wrong otp pin"` | same, `detail.message="Authentication failed."` | same, the wording |
| `/validate/radiuscheck` | empty body, 400 | empty body, 400 | empty body, 400 (no channel for wording) |
| `/validate/triggerchallenge` | 200, `value=<count>`, `CHALLENGE` | 200, `value=0`, `REJECT`, generic message | same, the wording |
| `/validate/initialize` | 200, `value=false`, `CHALLENGE`, `detail.passkey` | 200, `value=false`, `REJECT`, generic message | same, the wording |
| `/ttype/push` (signed answer) | 200, `value=false`, **no `detail`** | 200, `value=false`, **no `detail`** | 200, `value=false`, `detail.message` |
| `/validate/polltransaction` | 200, `detail.challenge_status` | *not gated* | *not gated* |

Two endpoints resolve "look like an ordinary failure" in opposite directions, deliberately. On `/validate/*` every
failure carries a `detail`, so a silent rejection carries the generic message to have one too — a response with no
detail could only have come from conditional access. On `/ttype/push` no failed answer carries a detail, so a silent
rejection carries none either; the generic message there would be exactly the tell it avoids elsewhere.

## Open to discuss

Three decisions I would rather have argued than assumed. All three are load-bearing enough that changing them later
is a behaviour change for integrators, so now is the moment.

### 1. Should the error code be consistent *across* endpoints, or *within* each endpoint?

**Current: within each endpoint.** A rejection mirrors whatever a failed authentication looks like where it happens —
401 with `error.code` 403 at `/auth`, an ordinary 200 with `value=false` and no error object at `/validate/*`, an
empty 400 at `/radiuscheck`, a 200 with `value=false` at `/ttype/push`.

*For keeping it:* indistinguishability is the whole point of silent-by-default, and the status code is the loudest
channel there is. A CA rejection returning 401 on `/validate/check`, where every other failure returns 200, would
announce "conditional access refused you" on every silent lock — the default would leak by construction. It would
also break callers: RADIUS plugins, SAML adapters and PAM modules read a non-200 there as a transport or server
fault, not as a failed authentication, so a lock would surface to users as an outage.

*For changing it:* one rule for an integrator writing against several endpoints, and one greppable signal.
`hide_specific_error_message` already carries a `TODO` admitting privacyIDEA does not consistently return 401 for
failed authentications, and the `hide_auth_error_status` hardening policy exists precisely to collapse distinct 4xx
codes to 401 — so there is appetite for uniformity somewhere.

### 2. On `/validate/triggerchallenge`, is it valid to suppress the challenge data?

**Current:** a request that successfully triggers a challenge *and* trips a restriction in the same breath is answered
as the rejection it now is — `value` becomes `0`, `authentication` `REJECT`, and the detail is replaced by the wording
(the `transaction_id`, `multi_challenge` and messages go with it). **The challenge row is left in the database** to
expire unanswered.

*For it:* the request was refused, and handing back a `transaction_id` that the very next request would refuse is
incoherent — the client would prompt or poll for something already dead. It also preserves the invariant that the
request which writes a lock is answered exactly like the requests the lock then refuses; leave the challenge in and
that one response becomes the distinguishable one.

*Against it:* `triggerchallenge` is admin-only and is not an authentication of the target user — it is an admin
action on their behalf, and the admin is not the one being restricted. The challenge genuinely was created, so the
response now understates server state, and an automation that records transaction_ids loses one silently.

*The uncomfortable middle:* we withhold the transaction_id but do not invalidate the challenge, so the response and
the database disagree. If we keep suppressing, cancelling the challenge as well would at least make them agree. The
three options are (a) suppress and cancel, (b) suppress and leave (current), (c) answer the admin honestly and let
the lock bite when the user answers.

### 3. On a notify-only stage, is it valid to *append* the message to the existing one?

**Current:** a stage that only notified refused nothing, so the credential failure keeps its own id, message and
detail, and the notification is appended into the same field:
`"wrong otp pin. Your administrator has been notified by email."` — with masking on it composes against the generic
failure instead: `"Authentication failed. Your administrator has been notified by email."`

*For it:* the message is failure-only and the credential failure really is still the reason; the notification is
additional information, not a replacement. A separate detail key was tried (`login_notice`) and dropped, because a
field of its own escapes both masking policies and re-opens the disclosure question the whole design closes.

*Against it:* it concatenates two sentences from different authors — the token layer and an admin's free text — into
one field that clients display verbatim, and it joins them by stripping a trailing `.` from the first
(`compose_failure_message`), which is crude for a reason ending in `?` or `:` or for non-Latin punctuation. It also
means a stage that both locks *and* emails never reaches this path at all (its message is treated as describing the
restriction), so the append only ever carries notify-only wording — a subtlety an admin authoring one message per
stage will not expect.

*And a question underneath it:* should the user be told at all? `EMAIL_USER`'s default —
*"A notification email has been sent to your email address."* — is plainly useful to them. `EMAIL_ADMIN`'s —
*"Your administrator has been notified by email."* — reassures a legitimate user and tells an attacker their probing
was noticed. It is opt-in either way, so this is about whether the default wording we ship is the right default.


privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/stages-list/stage-item/conditional-access-stage-item.component.ts
```diff
@@ -50,6 +77,72 @@ export class ConditionalAccessStageItemComponent {
...
+  readonly maxErrorMessageLength = MAX_ERROR_MESSAGE_LENGTH;
+
+  // One string for the reset button's tooltip and its accessible name: a sighted user hovering and a
+  // screen-reader user tabbing must be told the same thing, and two literals would drift apart.
+  readonly resetErrorMessageLabel = $localize`Replace with the suggested wording for this stage's actions`;
+
+  readonly durationTagUnusableHint = $localize`{duration} needs a temporary lock or block to count down. This stage \
+has none, so it would be shown to the user as written - remove the tag, or add a temporary action.`;
+
+  readonly errorMessageHint = $localize`When enabled, the text you enter is shown to the user whenever authentication \
```
henrik-falk: I thinik this is a little bit verbose

privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/stages-list/stage-item/conditional-access-stage-item.component.html
```diff
@@ -88,6 +88,61 @@
...
+                </span>
+              }
+              @if (durationTagUnusable()) {
+                <span class="ca-stage-message-warning">{{ durationTagUnusableHint }}</span>
+              }
+            </mat-hint>
+            <mat-hint align="end">{{ errorMessageLength() }}/{{ maxErrorMessageLength }}</mat-hint>
+          </mat-form-field>
+          <button
+            class="padding-top-8"
```
henrik-falk: should be centered to the input field (without hint)

henrik-falk (dismissed):

privacyidea/api/validate.py
```diff
@@ -389,6 +378,8 @@ def _poll_transaction_identity() -> User:
 @postpolicy(check_serial, request=request)
 @postpolicy(autoassign, request=request)
 @add_serial_from_response_to_g
+# First decorator to act on the request; see conditional_access_gate for why it sits exactly here.
+@conditional_access_gate(_conditional_access_identity)
 @prepolicy(check_application_tokentype, request=request)
 @prepolicy(pushtoken_validate, request=request)
 @prepolicy(set_realm, request=request)
```
henrik-falk, line 382: A second gate here, below set_realm and mangle, in addition to the one above?                                                                 
Both pre-policies replace request.User (prepolicy.py:1022 and :1220), and a lock is stored under — and written for — the identity the authentication log records, which is the rewritten one. The gate above therefore looks the lock up under a name the engine never uses: a lock written for cornelius@realm2 is never found again on a request that arrives as cornelius@defrealm, and it refuses nothing. /auth resolves the realm in before_request, so the two endpoints also disagree about who is locked.                                                              
The gate above has to stay where it is — nothing may run for a locked user — so this would be a second decorator rather than a move:          
 @conditional_access_gate(_conditional_access_identity, only_if_rewritten=True)                                                                                                                                                                                                              
only_if_rewritten does not exist yet: it would have the gate record the identity key it checked on the ConditionalAccessContext and skip the  query when it is unchanged, so an untouched request pays nothing for this.

privacyidea/api/lib/conditional_access.py
```diff
@@ -19,302 +19,590 @@
...
+        detail = content.get("detail") or {}
+        # An error-shaped body means the view raised and an error handler built this response.
+        errored = "error" in result
+        if evaluation.restricted:
+            # An error body never met the post-policies, so a detail missing *there* says nothing about them. On
+            # any other response one that is gone was stripped by no_detail_on_fail, and a silent restriction then
+            # has nothing to say - the pre-check answers such a request with no detail either.
+            message = rejection_message(context.rejection_shape, evaluation.messages,
+                                        detail_stripped=not errored and "detail" not in content)
+            if errored:
```
henrik-falk: if errored or context.rejection_shape.as_error:?                                                                                              
/auth returns its challenge from send_result(False, rid=2, details=details) before it evaluates, so a restriction written on that very request
arrives here with a 200 in hand. errored is False, the body is edited in place, and the login is refused as a 200 with value false and REJECT - a shape no failed login on that endpoint ever has, and therefore the one thing that identifies a rejection.                                                                                                                                                                      
RejectionShape(as_error=True) already states how this endpoint answers a refusal; the condition just never asks it.

privacyidea/lib/conditional_access/engine.py
```diff
@@ -1495,33 +1768,42 @@ def _upsert_user_lockout_state(user: "User", *, lock_expires_at: datetime | None
...
+            weakening_declined = True
+        elif lock_expires_at is not None and lock_expires_at < state.lock_expires_at:
+            log.info(f"Not shortening the existing lock for {user!r}: it already runs until "
+                     f"{state.lock_expires_at}.")
+            weakening_declined = True
+        if not weakening_declined:
             state.username = user.login
             state.lock_expires_at = lock_expires_at
-    return write.succeeded and not downgrade_declined
+            # Written together with the expiry, so the error message always describes the lock now in force. A write
```
henrik-falk: Should the error message be kept when this write only restates the lock already in force?                                                     
Weakening is declined above, but an equal write is not weakening: a silent PERMANENT_LOCK_USER over a permanent lock - or a timed one with the same expiry, since the check above is a strict <- leaves the lock exactly as it was and replaces its wording with None. The user is then told nothing about a restriction that did not change. Message and lock are coupled only in the declining direction.                           
A stage that restates the lock and carries wording of its own should still win, so the rule would be "keep the existing message only where  this write changes nothing", not "never overwrite".                                                                                                                                                                                                                              _upsert_ip_block has the same line and the same question.

henrik-falk (approved):

privacyidea/static_new/src/app/components/shared/expandable-message/expandable-message.component.ts
```diff
@@ -0,0 +1,66 @@
...
+let nextId = 0;
+
+/**
+ * A one-line piece of text in a dense table, expandable in place.
+ *
+ * Long text would otherwise stretch its column past every other one, and a tooltip is the wrong answer: it is
+ * mouse-only, cannot be selected or copied, and never reaches a keyboard or touch user. This clamps to one line
+ * and offers a real button to reveal the rest, so the full text stays reachable and selectable by anyone.
+ *
+ * Renders an em dash when there is no text, so an empty value reads as deliberate rather than missing.
```
henrik-falk: renders hyphen not em dash ;)

privacyidea/lib/conditional_access/engine.py
jelinaunger: The higher priority policy now wins regarding the message, which is always the first policy.

Issue #5787, opened 2026-08-18
Make the triggering client available to the push notification text (also when polling)


nilsbehlen:
## Summary

`push_text_on_mobile` supports the tags `{client_ip}`, `{ua_browser}`, `{ua_string}` and
`{action}`, which describe the client that triggered the challenge. A text like

    push_text_on_mobile=Confirm the login to {ua_browser} from {client_ip}

only works for the notification that is sent via Firebase. If the smartphone fetches its open
challenges by polling (`GET /ttype/push`), the same text is rendered again for the request of
the smartphone, which knows nothing about the client that triggered the challenge. The four
tags are empty in that case, so the user sees a notification without the application they are
logging in to - which is exactly the information that makes the notification worth reading.

## Cause

`_build_smartphone_data()` in `privacyidea/lib/tokens/pushtoken.py` reads the triggering
request from `g.request_headers`. The `validate` blueprint sets this in its `before_request`
(via `get_before_request_config()`), the `ttype` blueprint does not, and it also cannot: the
polling request *is* the request of the smartphone. Nothing about the triggering client is
persisted with the challenge, so it cannot be recovered later.

The notification text is not stored either. It is rendered from the policy on every poll,
which is why the two ways of delivering the same challenge can produce different texts.

## Proposal

Store the information about the triggering client in the challenge data when the challenge is
created, and use it when the notification is built:

* Add the name of the client application, the user agent string and the client IP to the
  challenge data in `PushTokenClass.create_challenge()`.
* In `_build_smartphone_data()`, prefer these stored values over the current request, so that
  both the Firebase notification and the polled challenge describe the same client.

Points to decide:

* The challenge data is user-controlled input (the user agent is a request header). It is sent
  to the smartphone and displayed there, so the length has to be limited.
* Whether the rendered text itself should be stored instead of the single values. This would
  make the notification stable, but a policy change would no longer affect challenges that
  are already open.
* Whether a dedicated tag for the client application (for example `{client_application}`)
  is clearer than reusing `{ua_browser}` for the name of a plugin.

## Related

The name of the application is already determined for the policy condition `user_agents` and
for the audit log (`get_plugin_info_from_useragent()`). The change that documents the tags of
`push_text_on_mobile` adds `get_useragent_name()`, which fills `{ua_browser}` with the name of
a client application that is not a browser - for push, email, SMS and the user notification
handler.

The token event handler has the same problem in its own tag substitution, see #5786.

Issue #5786, opened 2026-08-18
Token event handler: {ua_browser} is always written as "None"


nilsbehlen:
## Summary

The `set_description` and `set_tokeninfo` actions of the token event handler support the tags
`{ua_browser}` and `{ua_string}`. `{ua_browser}` is taken directly from
`request.user_agent.browser`, which is **always** `None`: werkzeug does not parse the user
agent anymore (the `UserAgent` class sets `browser`, `platform` and `version` to `None` unless
a parser is registered as `Request.user_agent_class`, which privacyIDEA does not do).

So a handler with the description `Enrolled via {ua_browser}` writes the literal string
`Enrolled via None` into the token description - for browsers, for the plugins, for the app,
for every client.

## Cause

`privacyidea/lib/eventhandler/tokenhandler.py:516` and `:540` build their own tag values
instead of using `create_tag_dict()`:

    ua_browser=request.user_agent.browser,
    ua_string=request.user_agent.string))

## Proposal

Use `get_useragent_name(request)` from `privacyidea/lib/utils` for both actions. It returns the
name the client sends in its user agent string - the same value that the policy condition
`user_agents` and the audit log use - and an empty string if there is no user agent. Browsers
report `Mozilla` there, the complete user agent stays available as `{ua_string}`.

`get_useragent_name()` is added by the change that documents the tags of
`push_text_on_mobile` (#5789), so this needs that branch first.

## Note

These two actions are the only places left that assemble notification tags by hand. Letting
them use `create_tag_dict()` would fix this and keep the tags of all handlers in sync, but it
changes which tags the two actions accept, so it is a bigger change than the fix above.

nilsbehlen:
The notification tags of the push token have the same gap when the smartphone polls, see #5787.

PR #5789, opened 2026-08-18
Document the tags of push_text_on_mobile and make {ua_browser} useful

### LINKED ISSUE #5788: A push_text_on_mobile that cannot be formatted breaks the authentication
## Summary

If the value of the policy `push_text_on_mobile` cannot be formatted, the authentication fails
instead of falling back to the default notification text.

`_build_smartphone_data()` in `privacyidea/lib/tokens/pushtoken.py` renders the text with
`str.format()` and catches only `KeyError`:

    try:
        message_on_mobile = message_on_mobile.format(**tags)
    except KeyError as e:
        log.warning(...)
        message_on_mobile = default_message

`str.format()` raises two more exceptions that a typo in the policy value can trigger:

* `ValueError` for an unbalanced brace, for example `Confirm the login with a 50{ discount`
* `IndexError` for a positional field, for example `Confirm the login of {0}`

Both propagate out of the challenge creation, so the user cannot authenticate with their push
token at all until the policy is corrected. An unknown tag (`{unknown}`) is handled and falls
back to the default text as intended.

## Steps to reproduce

1. Enroll a push token for a user.
2. Set an authentication policy with `push_text_on_mobile=Confirm the login of {0}`.
3. Trigger a challenge for that user.

## Expected

The default text *Do you want to confirm the login?* is used and a warning is logged, the same
way it already happens for an unknown tag.

## Actual

The request fails with `IndexError: Replacement index 0 out of range for positional args
tuple` (or `ValueError: expected '}' before end of string` for an unbalanced brace).

## Fix

Catch `IndexError` and `ValueError` in addition to `KeyError`.

nilsbehlen:
Fixes #5788

## What

`push_text_on_mobile` only said "You can also use certain tags in the text, just like in
:ref:`emailtext`". That points at a tag list containing `{otp}` and, in a separate paragraph,
`{challenge}` - neither of which can ever be filled for a push token:

* a push notification carries no OTP value, and `{otp}` is not part of the tag dictionary of
  the push token, so using it discards the whole text (see below),
* `{challenge}` is the `challenge` **parameter of the authentication request**, which only
  transaction signing tokens send (DisplayTAN, see the OCRA token documentation). It is empty
  in every normal push flow. It is not the push nonce and not the transaction id.

This documents the tags that actually work, plus two limitations that were not written down
anywhere:

* `{client_ip}`, `{ua_browser}`, `{ua_string}` and `{action}` describe the client that
  triggered the challenge, so they can only be filled for the notification that is sent via
  Firebase. When the smartphone fetches its open challenges by polling, the text is rendered
  again for the request of the smartphone, which knows nothing about the triggering client.
  The tags of the token owner are filled on both ways.
* `push_title_on_mobile` is never run through the tag replacement, so tags do not work in the
  title.

The `{challenge}` paragraphs of `smstext` and `emailtext` now name the DisplayTAN token and
link to the OCRA documentation, because "banking transaction data" alone is hard to place.

## Fixes

**A text that cannot be formatted no longer breaks the authentication** (#5788). Only
`KeyError` was caught, so a positional field (`{0}`) or an unbalanced brace (`50{ discount`) in
the policy value raised out of the challenge creation instead of falling back to the default
message.

**`{ua_browser}` now names clients that are not a browser.** The value came straight from
werkzeug, which only recognizes well-known browsers and left the tag empty for everything else
- so exactly for the plugins, where the name of the application is the interesting part. The
name is now taken from the user agent string in that case, using the same parse that the
`user_agents` policy condition and the audit log already use. A notification can therefore say
`Confirm the login to privacyidea-keycloak`. Since this lives in `create_tag_dict()`, it
applies to the email, SMS and user notification texts as well.

## Tests

* `test_api_validate_pushchallenge.py`: the enrollment and the trigger/poll sequence are
  extracted into helpers, and a new test asserts that an unknown tag, a positional field and
  an unbalanced brace all fall back to the default message.
* `test_lib_utils.py`: `create_tag_dict()` fills `ua_browser` with the plugin name for a
  non-browser user agent, keeps the browser name for browsers, and stays empty without a
  request.

## Follow-ups

* #5787 - store the triggering client in the challenge data, so the push text can name the
  application even when the smartphone polls.
* #5786 - `set_description` and `set_tokeninfo` of the token event handler build their own tag
  values and still write the literal `None` for `{ua_browser}`. The fix is to use the helper
  added here.


privacyidea/lib/utils/__init__.py
```diff
@@ -1347,7 +1347,7 @@ def create_tag_dict(logged_in_user=None,
                 date=date,
                 client_ip=client_ip,
                 pin=pin,
-                ua_browser=request.user_agent.browser if request else "",
+                ua_browser=get_useragent_name(request),
```
jona-samuel, line 1355: Maybe we should document what is contained in the answer dict.

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -303,8 +303,10 @@ def _build_smartphone_data(token: TokenClass, challenge: str, registration_url:
                            challenge=options.get("challenge"))
     try:
         message_on_mobile = message_on_mobile.format(**tags)
-    except KeyError as e:
-        log.warning(f"Could not format the message: {e}. Using default message.")
+    except (KeyError, IndexError, ValueError) as e:
```
jona-samuel: Maby we should add AttributeError and TypeError

jona-samuel (approved):

Issue #5736, opened 2026-08-07
Challenge data is stored unencrypted in Redis while the DB path encrypts it


nilsbehlen:
### Summary

Challenge `data` is encrypted before it is written to the SQL `challenge` table, but stored as **plaintext** when the Redis challenge cache is active. With `PI_REDIS_CACHE_CHALLENGES = True` there is no SQL row at all, so Redis is the only copy — this is not a cache holding a second copy of protected data, it is the sole storage, unprotected.

Not yet released — both the Redis cache and the challenge-data encryption are on master only.

### The asymmetry

`Challenge.set_data()` (`privacyidea/models/challenge.py:120`) encrypts via `encryptPassword`, with the docstring stating *"The data is encrypted before being stored in the database since it may contain OTP values."*

`ChallengeDTO.set_data()` / `ChallengeDTO.to_payload()` (`privacyidea/lib/cache/redis.py:394`, `:410`) perform no encryption.

The plaintext is handed to Redis at `privacyidea/lib/token/auth.py:672`:

```python
db_challenge = Challenge(serial, ..., data=data if data is not None else '', ...)  # encrypts into _data
if redis_feature_enabled("challenges"):
    cache_challenge(
        ...
        data=db_challenge.data,   # <-- the DECRYPTING property getter
```

`Challenge.__init__` encrypts into `_data`, and this line reads it back out through the decrypting `data` property — an encrypt/decrypt round-trip whose only net effect is to hand plaintext to Redis.

### Reproduction

Same `create_challenge()` call, same secret value, against a real Redis 7 instance:

```
=== DB PATH ===
raw column _data : '8355e79a3dcc9acf71ff4653b1d6d7ff:dee2cee0c4c43071fe0e0a023b5799a5'
decrypted .data  : '987654'

=== REDIS PATH ===
raw redis value  : {'SER002': '{"transaction_id": "0228...", "serial": "SER002",
                    "challenge": "", "data": "987654", ...}'}
no SQL row       : True
```

### Affected data

| Source | Content of `data` | Impact |
|---|---|---|
| `lib/tokens/emailtoken.py:319`, `lib/tokens/smstoken.py:352` | the plaintext OTP, when `email.concurrent_challenges` / `sms.concurrent_challenges` is enabled | anyone who can read Redis can answer the challenge |
| `lib/tokens/pushtoken.py:913` | `display_code` for push code-to-phone | the value the user must type back — same class of secret |
[...]

PR #5792, opened 2026-08-19
Cache user and authentication lookups in Redis, and stop repeating writes on every request

nilsbehlen:
Seven commits that all remove work privacyIDEA repeats. Three add optional Redis caches, two remove writes and a table scan without needing Redis at all, and one turns the Redis CI job on for every workload. Each stands on its own.

Merged with master after the subscription-dashboard work landed; the two hand-resolved files and one behaviour change that came out of it are described at the end.

## Share user store lookups and cached authentications through Redis (44b4ad9fb)

Two new opt-in workloads for the existing Redis layer, alongside `PI_REDIS_CACHE_CHALLENGES`.

**`PI_REDIS_CACHE_USERS`** caches the three answers a resolver is asked over and over: the ID behind a login, the login behind an ID, and a user's attributes. Every one is a round trip to LDAP, SQL or HTTP. The two caches we already had cannot be shared - the `usercache` table holds only the login/ID correlation, and an LDAP resolver's cache is a dictionary inside one worker process.

It hooks in *above* the resolver, so it works for every resolver type rather than LDAP only, and *below* the custom attributes, so those are merged fresh on every read and can never be served stale. Invalidation is explicit - user updated or deleted, resolver saved or deleted, admin flush - with the TTL (`PI_REDIS_USER_CACHE_TTL`, default 300s) as the backstop rather than the mechanism. A per-resolver index set makes dropping a resolver's entries a set read instead of a keyspace scan.

**`PI_REDIS_CACHE_AUTH`** moves the `auth_cache` policy's entries out of a table that was written to in three directions per authentication: an `UPDATE` on every cache *hit* to count the use, an `INSERT` per success, and `DELETE`s for stale rows. Redis is the whole store while it is reachable, which is correct here for the same reason it is for challenges - losing an entry costs one real authentication and nothing else.

Two details worth a look in review:

- **The policy's first interval becomes the TTL.** The `4h` in `4h/5m` is exactly how long an entry can ever be used, so a cached password now physically disappears when the policy stops honouring it instead of sitting in a table until a cronjob runs. Using an entry deliberately does *not* extend its life, because the window runs from the first authentication.
- **The use counter is a Redis field of its own, raised with `HINCRBY`.** Folding it into the encrypted record would make every hit a read-modify-write, letting two concurrent authentications spend the same use and exceed `max_auths`. The table did this in one atomic `UPDATE`; this keeps that property.

This also fixes a pathology in the existing design: `add_to_cache` inserted unconditionally with no dedup, and every lookup ran Argon2 against *all* of a user's rows - so the cache got slower the more it was used. Per-entry expiry bounds that set. `pi-manage config authcache cleanup` now reports that it has nothing to do rather than a count of zero.

**Values are encrypted** before reaching Redis. The user cache holds personal data, the authentication cache an Argon2 hash of a password; neither should be readable in a Redis dump. Keys are not encrypted, because Redis has to look them up - the key space reveals who exists.

Everything degrades to the previous behaviour when Redis, or the HSM the encryption needs, is unreachable. Reading a cache must never turn a lookup or an authentication that would have succeeded into an error, and there are tests for each of those paths.

## Share the certificate health results between workers (d85b106fe)

`PI_REDIS_CACHE_HEALTH`. The dashboard's certificate panel means opening a TLS connection to every configured LDAP and Keycloak endpoint, every EntraID client-certificate credential and every admin-configured server certificate. Those results were already cached for `PI_CERT_CHECK_CACHE_SECONDS` (default 3600) - but in a dictionary belonging to one worker process, so **eight workers on three nodes probe every endpoint twenty-four times an hour** for one panel.

The per-process dictionary stays in front of the shared one, since a worker that already has the answer has no reason to ask Redis. Both use the same TTL, and `invalidate_certificate_cache()` - already wired into resolver save/delete - now drops the shared copy too, so one admin's change reaches every worker instead of only the one that served the request.

Lowest-risk of the three caches: not on the authentication path, display-only, and every failure mode (Redis unreachable, unreadable payload, failed serialisation) ends with this worker probing for itself exactly as before. Stored as plaintext, unlike the other two - it carries no credentials, but it does name internal hosts and certificate subjects, which the docs now say out loud.

## Write the client application row again, and only when it says something new (1ea027018)

**Note on the `lastseen` fix in this commit: master got there first.** This commit was written before the subscription-dashboard work landed and contained a fix for `lastseen` never being refreshed on an existing row (the update assigned `client_app.last_seen`, but the model column is `lastseen`, so it set a stray Python attribute). Master fixed exactly the same thing independently, with the identical one-word change - which is why the merge produced no conflict there - and added `test_03_update_refreshes_lastseen` for it. **No credit is claimed for that fix here; it is already on master.** What remains of this commit is the write throttle below.

With the column actually being refreshed, writing it per request is worth avoiding. `save_clientapplication` is a policy on every `/validate/check`, and recording "seen just now" cost a `SELECT` + `UPDATE` + `COMMIT` per authentication - a cluster-wide write in a replicated setup - for a timestamp neither reader needs to the second. Each worker now skips the write for a client it wrote within `PI_CLIENTAPPLICATION_WRITE_INTERVAL` (default 60s, `0` disables).

The timestamp is remembered only *after* the commit, so a failed write is retried next request instead of being suppressed for an interval, and a client never written always reaches the database.

The age is measured on the same clock the column is written from, not a monotonic one. Master's new `test_03_update_refreshes_lastseen` fakes `datetime.now()` and expects a save five simulated minutes later to land; a monotonic throttle skipped it. Measuring on the column's own clock is also what the interval actually means - how stale the recorded timestamp may become - and it makes the behaviour observable rather than hidden behind a second clock. A clock set backwards can delay a refresh by the length of the jump, which for a "last seen" column is not worth guarding against.

## Reuse the counted users between subscription checks, except when it would deny (31779c3a1)

`check_subscription` runs on every authentication carrying a known plugin's user agent, and counting users with active tokens is a `DISTINCT` across the whole `tokenowner` and `token` tables. Each worker now keeps that number for `PI_SUBSCRIPTION_COUNT_INTERVAL` (default 60s, `0` disables).

The part to review: **a kept number is only handed out while it stays within what the subscription allows.** The moment it would declare the subscription exceeded, the users are counted again - that answer refuses authentications, and tokens deleted a moment ago would otherwise keep users locked out for the rest of the interval. Because the exceeded check returns early when the count is within the limit, the scan is skipped exactly when the answer is "everything is fine", which is what virtually every request gets.

The other direction is the accepted trade: a user given a token within the interval may not be counted yet, which changes nothing for enforcement that is already probabilistic. Only the check reuses the number - the subscription overview and the statistics task still count exactly, which is why the count moved to where the allowed number is known instead of being cached inside the shared function. An expired subscription now does not count users at all.

## Document the configuration keys where the existing ones live (e00fa5851)

The reference docs covered all seven new keys, but three other places describe the Redis layer and none of them knew about the new workloads: `.env.example` (which Docker operators copy), `dev/redis.md` (which still said "not used for anything but challenges"), and `READ_BEFORE_UPDATE.md`.

The update notes matter most, because two of the changes are **on by default** and alter behaviour on upgrade with no configuration at all.

## Run the whole suite against every Redis workload in CI (97e38bb71)

`unit-tests-redis.yml` existed to prove the Redis path behaves like the database one for the whole suite, but it only ever enabled the challenge workload - the user, auth and health caches were exercised by their own tests and by nothing else. All four are now on.

Turning them on surfaced two kinds of test that cannot pass with a cache in front of what they inspect, and both keep their own backend now via the new `MyTestCase.pin_to_database()`, which names the workload it needs off and says why:

- Tests that assert the `authcache` table (empty when Redis holds the entries) and the cleanup command's output.
- Two tests that mock the resolver's `get_user_info` but read `token.user.info` *first*, which caches the real value - the mock sits below the cache and is never consulted.

Two of those took the rest of their class with them: `test_33_auth_cache` and `test_12_authcache` each set a policy and delete it at the end, so failing early leaked an `auth_cache` policy into every later test in the file. Pinning the backend stops that at the source, though those tests would be sturdier with unconditional cleanup.

## Test isolation

Three autouse fixtures in `conftest.py` clear global state that outlives a test: the LDAP resolver cache, the client-application write throttle and the kept user count.

The LDAP one **fixes an existing failure**, unrelated to everything above. `LDAPIdResolver.CACHE` is keyed on the resolver *configuration*, so two test files that configure a resolver alike share one entry and nothing ever empties it. `test_33_cache_disabled` therefore passes when its file runs alone and fails after `test_lib_user.py` - both files independently picked the search filter `(|(cn=*))` with the same "to get a unique resolver ID" comment. CI hides it because `--dist=loadfile` keeps each file on its own worker. Reproduce on master with:

    pytest tests/test_lib_usercache.py tests/test_lib_user.py tests/test_lib_resolver.py

`SubscriptionApplicationTestCase` now sets the interval to `0`: those tests assign tokens and expect the very next check to judge against them, which is the granularity the new interval deliberately relaxes. The assertions are unchanged.

## Testing

85 new tests. The cache tests skip cleanly when `TEST_REDIS_URL` is unset, so the default DB-only run is unaffected.

Verified with the **full suite against MariaDB 11** (per-worker databases, `-n 8 --dist=loadfile`, all four workloads enabled) - the closest local approximation of this job: **3212 passed, 10 failed**. All ten fail identically with Redis switched off entirely and come from the local machine, not this branch:

- `cryptography` 46.0.6 installed against the pinned 49.0.0, which reverses RFC4514 name ordering (`<Name(C=DE,…,CN=CA001)>` vs the expected `<Name(CN=CA001,…,C=DE)>`) - 7 CA/certificate failures
- `webauthn` 2.7.1 against the pinned 2.8.0, which changed an error message - 1 failure
- a `/etc/privacyidea/pi.cfg` from a local privacyIDEA install supplying a real `SECRET_KEY`, where `test_app` expects the `'t0p s3cr3t'` fallback - 2 failures

That last one is a pre-existing test-isolation wart worth knowing about independently of this PR: `test_app` reads the machine's system config, so it fails for any developer who also runs privacyIDEA locally.

## Notes for review

- **The three Redis caches are off by default** and need both `PI_REDIS_URL` and their own flag, exactly like `PI_REDIS_CACHE_CHALLENGES`.
- **The two intervals are on by default** (60s each). They change behaviour on upgrade without any configuration: `lastseen` becomes accurate to within a minute rather than to the second, and a user given a token may take up to a minute to count towards the subscription. Set either to `0` to keep the old per-request behaviour. Say so in review if you would rather these shipped opt-in - it is a one-line change to each default.
- No dependency on PR #5736 (challenge data encryption). `encryptPassword`/`decryptPassword` are already in master - that PR applies them to challenges, this one applies them to two new caches, and the code paths are disjoint.
- Deliberately **not** cached: tokens (local indexed reads, and the mutable counter/lock state is what the code works hardest to keep fresh), the audit log, and `EventCounter` (already node-partitioned to avoid replication contention).
- Internal metrics remain the obvious next candidate, but it wants write-behind and therefore an answer to who flushes Redis to the database - left out on purpose.
- `POLICY_CACHE` / `CONFIG_CACHE` and their four accessors are **removed** in their own commit (`47d4ab10c`). They had no caller anywhere since they were added in 2016, and sat in the module that now holds three working caches - `get_config_cache()` in particular reads like an accessor for an empty cache rather than an abandoned stub. The names also collide with the config caching that does exist in `SharedConfigClass`. Same archaeology as the commented-out `# @cache.memoize(10)` lines still sitting above `get_tokens`, `get_realms` and `get_resolver_object`.

## Merge with master

`privacyidea/lib/subscriptions.py` and `tests/test_lib_subscriptions.py` needed hand resolution where the dashboard work landed next to the write reductions.

In `check_subscription`, master's rewritten head stays - resolving a metered client to the application it is counted against, and ignoring a subscription with no `date_till` - and only its unconditional count of the users is dropped, because the branches below now count with the limit in hand. Master's dashboard functions keep calling `get_users_with_active_tokens` directly, which is the intended split: the number a person is *shown* stays exact, only the check reuses one.

`PluginSubscriptionStatusTestCase` mocks the user count and changes it inside a single test, so it counts on every check.


privacyidea/lib/cache/auth.py
```diff
@@ -0,0 +1,408 @@
...
+            # Not an Argon2 hash - the same case the database path treats as an
+            # old entry and discards
+            log.debug(f"Discarding a non-argon2 authentication cache entry for {username!s}@{realm!s}.")
+            _forget(client, key, [entry_id])
+            continue
+        if max_auths > 0 and record["auth_count"] >= max_auths:
+            # Used up. Drop it so the next attempt does not verify it again
+            _forget(client, key, [entry_id])
+            continue
+        _use(client, key, entry_id, record)
```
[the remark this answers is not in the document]
nilsbehlen: Accepted for now as a known gap.

The race is real: reading the count and raising it are separate round trips, so two concurrent authentications can both take the last permitted use. But the database-backed path has the identical gap between its SELECT and its UPDATE, so closing it only for Redis would make the two backends disagree about a security limit. Making it exact belongs with a change that does the same for both.

The module comment no longer claims HINCRBY makes the limit exact - it states what the increment does buy (neither authentication loses the other's) and that the limit can be exceeded once under a race, in both backends.

doc/installation/system/inifile.rst
```diff
@@ -578,16 +750,34 @@ secret file with ``PI_REDIS_URL_FILE`` (see above). Credentials embedded in the
...
+database**: restrict it to a private network, require authentication, prefer
+``rediss://``, and use at-rest encryption (encrypted volume, or a managed Redis
+with encryption) if your threat model requires it. Do not expose the Redis
+instance on a public interface.
+
+What is stored differs per workload:
+
+* Challenge data is stored as plaintext, and here Redis is **less** protected
+  than the table it replaces: the SQL ``challenge.data`` column is encrypted with
+  the server's encryption key, the Redis copy is not. The field can carry the OTP
```
jelinaunger: This does not hold anymore since PR #5738

privacyidea/api/system.py
```diff
@@ -771,14 +778,18 @@ def delete_user_cache_api():
...
-           "value": {"status": true, "deleted": 42}
+           "value": {"status": true, "deleted": 42, "flushed_resolvers": 2}
          },
          "version": "privacyIDEA unknown"
        }
     """
     row_count = delete_user_cache()
-    g.audit_object.log({"success": True, "info": f"Deleted {row_count} entries from user cache"})
-    return send_result({"status": True, "deleted": row_count})
+    flushed_resolvers = flush_user_cache()
```
jelinaunger: It is a bit confusing which function is for which cache (this is not the only place). Worth a comment or renaming them so they are self-documenting which cache they are acting on.

privacyidea/lib/cache/auth.py
```diff
@@ -0,0 +1,412 @@
...
+#
+# The counter is a field of its own rather than a member of the record so it can
+# be raised with HINCRBY, which keeps two concurrent authentications from losing
+# one another's increment the way a read-modify-write of the encrypted record
+# would. It does not make the limit exact: reading the count and raising it are
+# still two round trips, so two authentications can both see the last permitted
+# use and both take it. The database-backed cache has the same gap between its
+# SELECT and its UPDATE, so this is the behaviour it already had - a cached
+# authentication can be spent once more than ``max_auths`` says under a race.
+_USER_KEY = "pi:authcache:v1:{}:{}:{}"  # pi:authcache:v1:<username>:<realm>:<resolver>
```
jelinaunger, line 92: what is the versioning used for and why is it not used in the challenge key?

privacyidea/lib/cache/auth.py
```diff
@@ -0,0 +1,412 @@
...
+            continue
+        try:
+            record["auth_count"] = int(raw_fields.get(_COUNT_FIELD.format(entry_id), 0))
+        except (TypeError, ValueError):
+            record["auth_count"] = 0
+        entries[entry_id] = record
+    return entries
+
+
+def _forget(client, key: str, entry_ids: list) -> None:
```
jelinaunger: tye hint for the client

privacyidea/lib/cache/user.py
```diff
@@ -0,0 +1,491 @@
...
+
+    A login that resolves to nothing is not cached. The empty answer is the one
+    that changes when a user is created, and a new user that can not
+    authenticate until a cache expires is a worse failure than a repeated
+    lookup for a login that does not exist.
+
+    :param resolver: The resolver object to ask on a cache miss
+    :param resolver_name: The name of that resolver
+    :param login: The login name to look up
+    :return: whatever the resolver returns for this login
```
jelinaunger, line 272: but it is always the user id so a str, isn't it?

privacyidea/lib/authcache.py
```diff
@@ -36,12 +37,29 @@ def _hash_password(password):
...
+    :param resolver: resolver name of the user
+    :param password: the password that authenticated
+    :param max_age_seconds: how long the entry may be used, from the policy.
+        Only Redis can act on it - a database row has no lifetime of its own and
+        is removed by the cleanup job instead.
+    :return: the id of the database row, or 0 if the entry went to Redis, which
+        has no row to identify
+    """
+    auth_hash = _hash_password(password)
+    log.debug(f'Adding record to auth cache: ({username!r}, {realm!r}, {resolver!r}, {auth_hash!r})')
```
jelinaunger: even you didn't introduce this line, but auth hash is sensitive data and I think it does not have any useful information here for the log.

privacyidea/config.py
```diff
@@ -171,6 +176,12 @@ class TestingConfig(Config):
     PI_REDIS_URL = os.environ.get(ConfigKey.REDIS_URL)
     PI_REDIS_CACHE_CHALLENGES = (os.environ.get(ConfigKey.REDIS_CACHE_CHALLENGES, "false").lower()
                                  in ("true", "1", "yes", "on"))
+    PI_REDIS_CACHE_USERS = (os.environ.get(ConfigKey.REDIS_CACHE_USERS, "false").lower()
+                            in ("true", "1", "yes", "on"))
+    PI_REDIS_CACHE_AUTH = (os.environ.get(ConfigKey.REDIS_CACHE_AUTH, "false").lower()
+                           in ("true", "1", "yes", "on"))
+    PI_REDIS_CACHE_HEALTH = (os.environ.get(ConfigKey.REDIS_CACHE_HEALTH, "false").lower()
+                             in ("true", "1", "yes", "on"))
```
jelinaunger: this could be a generic is true helper function, doesn't we already have one?

tests/test_lib_cache_redis_user.py
```diff
@@ -0,0 +1,569 @@
...
+    def test_03_entry_covers_only_the_attributes_it_was_written_for(self):
+        narrow = '{"attributes": ["email", "username"], "info": {"username": "alice", "email": "a@b.c"}}'
+        # Exactly the attributes it holds, and any subset of them
+        self.assertEqual({"username": "alice", "email": "a@b.c"},
+                         _covered_user_info(narrow, ["username", "email"]))
+        self.assertEqual({"username": "alice"}, _covered_user_info(narrow, ["username"]))
+        # An attribute it was never asked for is not absent, it is unknown
+        self.assertIsNone(_covered_user_info(narrow, ["username", "mobile"]))
+        # ... and neither is "everything"
+        self.assertIsNone(_covered_user_info(narrow, None))
```
jelinaunger, line 169: Just a side note: currently the UI has a standard set of attributes it requests on the user list page. Attributes the resolver is not configured for are not contained in the return dict at all and hence never written to the cache as well. Consequently, the cache would never be used there.

privacyidea/lib/cache/user.py
```diff
@@ -0,0 +1,491 @@
...
+    changes, nothing cached under it can be trusted.
+
+    :param resolver_name: The name of the resolver
+    :return: True if the entries are gone, False if Redis could not be asked and
+             they may still be there. Callers that report what they dropped have
+             to be able to tell those apart.
+    """
+    client = _cache_client()
+    if client is None:
+        # Nothing is cached for this resolver, so there is nothing left behind
```
jelinaunger: The documentation of _cache_client() says it is also None if redis is unreachable, hence there could still be entries which are not flushed and hence this function should return false per definition

doc/installation/system/inifile.rst
nilsbehlen: Fixed by 7635cf60a (merged master's already-encrypted-challenge-data paragraph in, kept this branch's per-workload breakdown).

privacyidea/api/system.py
nilsbehlen: Fixed in 32ec44560: added a comment at each call site naming which backend (SQL usercache table vs Redis) it acts on. Didn't rename the functions themselves - delete_user_cache() predates this PR and has ~20 call sites elsewhere, out of scope here.

privacyidea/lib/cache/auth.py
nilsbehlen: Fixed in 32ec44560: added redis_lib.Redis type hints to client in _read_entries, _forget, and _use.

privacyidea/lib/cache/user.py
nilsbehlen: Fixed in 32ec44560: annotated user_id as str in cached_username, cached_user_info, invalidate_user, and _keys_of_user - matches the resolver contract, getUserId() is documented :rtype: str everywhere.

privacyidea/lib/authcache.py
nilsbehlen: Fixed in 32ec44560: dropped the hash from the log line. You're right that Argon2 itself isn't invertible, but the input is PIN+OTP - a small enough keyspace that the hash is crackable offline, so logging it defeats the point of hashing it.

privacyidea/config.py
nilsbehlen: Fixed in 32ec44560, but not by reusing is_true() directly - that creates a circular import (config.py -> lib.utils -> lib.framework -> config.py for ConfigKey), verified it breaks full app boot. Added a local _env_flag() equivalent instead.

privacyidea/lib/cache/user.py
nilsbehlen: Fixed in 32ec44560: now distinguishes caching being off (True, nothing was ever written) from Redis being unreachable (False, entries may remain), instead of returning True for both.

tests/test_lib_cache_redis_user.py
nilsbehlen: deferred to #5817

privacyidea/lib/cache/auth.py
nilsbehlen: Good catch - fixed in 8caa69e24. The challenge cache actually predates the v1 convention (it's from the already-merged #5738), and the auth/user caches already had it; aligning the challenge cache's key prefix to pi:challenge:v1:... too, rather than dropping it from the others. Updated the doc example and the tests that pin the key format.

jelinaunger (approved):

PR #5805, opened 2026-08-20
Run migration tests against real MySQL too

nilsbehlen:
## Summary
- Adds a MySQL 8.4 service container and matrix entry to `migration-tests.yml`, alongside the existing MariaDB (10.11, 11) and PostgreSQL legs.
- Includes MySQL in the backup/restore round-trip step, since `pi-manage backup` uses the same `mysqldump`/`mysql` tooling for MariaDB and real MySQL.
- Real MySQL has no `CREATE SEQUENCE` support at all (unlike MariaDB 10.3+), so this exercises the sequence-less fallback branch in `privacyidea/models/db.py` that MariaDB-only testing never hits.

### DIFF
--- a/.github/workflows/migration-tests.yml
+++ b/.github/workflows/migration-tests.yml
@@ -51,17 +51,24 @@ jobs:
           # MariaDB 10.11 is the LTS most installs run (Ubuntu 24.04 default);
           # 11.x is the rolling latest. Migrations emit version-specific DDL, so
           # both are exercised (this job is path-gated to migration/model
-          # changes, so the extra entry rarely runs). Both versions run as static
-          # service containers below — 10.11 on host port 3306, 11 on 3307 — so
-          # the image stays a literal tag; a ${{ }} expression in image: is
-          # flagged by zizmor as an unpinned reference. Each entry's db_url
-          # targets the right port.
+          # changes, so the extra entry rarely runs). All db entries run as
+          # static service containers below — 10.11 on host port 3306, 11 on
+          # 3307, MySQL on 3308 — so the image stays a literal tag; a ${{ }}
+          # expression in image: is flagged by zizmor as an unpinned reference.
+          # Each entry's db_url targets the right port.
           - label: MariaDB 10.11
             db_url: "mysql+pymysql://privacyidea:privacyidea@127.0.0.1:3306/privacyidea_migration_test"
             pip_extra: "PyMySQL"
 
           - label: MariaDB 11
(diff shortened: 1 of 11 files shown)


pablo-knight (approved):

.github/workflows/migration-tests.yml
```diff
@@ -145,18 +167,19 @@ jobs:
...
 
-      # pi-manage backup shells out to mysqldump/mysql. Install MariaDB's client
-      # (which provides the mysqldump compat name) and fail loudly if it is
+      # pi-manage backup shells out to mysqldump/mysql, which applies to both
+      # MariaDB and real MySQL. Install MariaDB's client (which provides the
+      # mysqldump compat name and talks to both) and fail loudly if it is
       # missing, so the round-trip below runs for real instead of self-skipping.
       - name: Install MariaDB client for backup round-trip
-        if: startsWith(matrix.db.label, 'MariaDB')
+        if: startsWith(matrix.db.label, 'MariaDB') || startsWith(matrix.db.label, 'MySQL')
```
jelinaunger: ubuntu-24.04 runners preinstall MySQL 8.0.46, and mariadb-client-core / mysql-client-core-8.0 conflict (both Provides:/Conflicts: virtual-mysql-client-core). So here apt removes the real MySQL client and points     
  mysqldump at mariadb-dump — on the very row added to exercise real-MySQL divergence.

<br>

It does work (libmariadb3 ships caching_sha2_password.so, so auth against 8.4 is fine), but the backup half then tests MariaDB's dumper, not Oracle's. 

<br>

Suggestion — install only for MariaDB, keep the fail-loud check for both:

<br>

- name: Install MariaDB client for backup round-trip


    if: startsWith(matrix.db.label, 'MariaDB')
    run: |
      sudo apt-get update
      sudo apt-get install -y mariadb-client

- name: Verify a dump client is present


     if: startsWith(matrix.db.label, 'MariaDB') || startsWith(matrix.db.label, 'MySQL')
     run: mysqldump --version 

<br>


Caveat: 8.0.46 client vs 8.4 server needs one green run to confirm. If it fails, either use Oracle's 8.4 client from the MySQL apt repo, or keep mariadb-client and say in the comment why — right now it reads as     
  convenience rather than a choice.

jelinaunger (approved):

.github/workflows/migration-tests.yml
nilsbehlen: good find, fixed

PR #5810, opened 2026-08-21
Fix a handful of resolver/token-list/metrics/policy edge cases

nilsbehlen:
## Summary

A batch of small, independent fixes surfaced by a `/code-review high` pass, each with a regression test that fails without its fix:

- **Resolver batch lookups**: `SQLIdResolver.get_user_info_batch` no longer swallows a DB error into a blank username (it now propagates like `LDAPIdResolver` already does, so the caller's existing per-user fallback can mark the token `**resolver error**`); `LDAPIdResolver.get_user_info_batch` now resolves every case-variant alias of a login instead of only the first one; the identical `get_usernames_batch` override body is deduplicated into a shared `UserIdResolver` helper.
- **Token list owner lookup**: a failing page-wide `TokenOwner` query now falls back to a per-token lookup instead of failing the whole page, mirroring how `_resolve_owner_logins` already recovers from a failed resolver batch call.
- **`Token.first_owner`**: the `owners` relationship now orders by id, so which owner is "first" is deterministic for a token with more than one.
- **Metrics buffering**: `_write_observations` now applies each buffered key inside its own savepoint, so one bad key (e.g. a lock-wait timeout) no longer discards every other metric sample buffered for the same request.
- **`postpolicy`**: the decorator now raises if `request=` is omitted, instead of silently defaulting to `None` and disabling every user/realm-scoped policy condition of the wrapped function — the same bug already fixed once at its two call sites.

Two related findings from the same review pass (a narrow double-race in the metrics recovery path, and a known/already-documented LDAP recursive-group-search N+1) were intentionally left out as not worth the churn right now.


privacyidea/api/lib/postpolicy.py
```diff
@@ -119,6 +119,12 @@ def __init__(self, function, request=None):
         :param request: The original request object, that needs to be passed
         :type request: Request Object
         """
+        if request is None:
+            # A missing request is not a usable default: policy functions
```
jelinaunger, line 123: Then the None default should be removed

privacyidea/lib/resolvers/SQLIdResolver.py
```diff
@@ -299,17 +299,21 @@ def get_user_info(self, user_id: int or str, attributes: list[str] = None) -> di
...
+        except ValueError as error:
+            # e.g. a non-numeric ID against an integer column -- same as get_user_info_batch
+            log.info(f"Could not look up user id {user_id!r}: {error}")
+            return userinfo
+        conditions = self._append_where_filter(conditions, self.TABLE, self.where)
+        filter_condition = and_(*conditions)
+        # A DB error here must propagate rather than being logged and swallowed: callers
+        # like getUsername() rely on it to tell "the user does not exist" (an empty
+        # result) apart from "the backend could not be reached" (an exception).
+        result = self.session.execute(select(self.TABLE).filter(filter_condition))
```
jelinaunger: the session should be rolled back on any error

jelinaunger (approved):

PR #5811, opened 2026-08-21
Add passkey policies to restrict authenticator device type and enforce userHandle

### LINKED ISSUE #5815: Passkey: allow restricting authenticator device type and enforcing userHandle binding
Add two passkey-only policies:

- `passkey_allowed_authenticator_device_types` (SCOPE.ENROLL and SCOPE.AUTH): restrict enrollment/authentication to `single_device` or `multi_device` passkeys, based on the signed `credential_device_type`.
- `passkey_enforce_user_handle` (SCOPE.AUTH, opt-in): require that the authenticator's `userHandle` matches the FIDO2 user ID recorded at enrollment, so a passkey token reassigned to a different user no longer authenticates as them without re-registering.

Implemented in PR #5811.

nilsbehlen:
Closes #5815

## Summary

Two new passkey-only policies (kept independent from the webauthn token implementation on purpose):

- `passkey_allowed_authenticator_device_types` (SCOPE.ENROLL and SCOPE.AUTH): restrict enrollment/authentication to `single_device` or `multi_device` passkeys, given as a space-separated list. This is checked against `credential_device_type`, which is derived from the signed backup-eligible flag in `authenticatorData` on every registration/authentication - not the self-reported (and unsigned) `authenticatorAttachment` string, and not AAGUID (which is only populated with enterprise attestation, off by default, so it's usually zeroed and not a realistic restriction axis). The AUTH-scope check is independent of the ENROLL-scope one and is re-verified cryptographically on every single authentication, not just checked against a value stored at enrollment.
- `passkey_enforce_user_handle` (SCOPE.AUTH, opt-in, default off): requires that the authenticator's `userHandle` matches the FIDO2 user ID recorded for the token at enrollment. privacyIDEA otherwise resolves the authenticating user purely via the credential ID's assignment in the database, so a passkey token that is unassigned and reassigned to a different user authenticates as that different user without ever re-registering the physical key. This closes that gap for deployments that want it closed, while leaving the existing (intentional) behavior as the default.

## Notable fix found along the way

Comparing `userHandle` correctly required decoding both sides to bytes rather than comparing strings: the frontend encodes `userHandle` with plain base64 (`btoa`, padded), while the stored FIDO2 user ID is base64url (unpadded) - the encoded strings differ even when the underlying bytes match.


privacyidea/api/lib/prepolicy.py
```diff
@@ -2201,6 +2203,17 @@ def fido2_auth(request, action):
                                          user_object=user_object).any())
     request.all_data[PasskeyAction.EnableTriggerByPIN] = passkey_trigger_by_pin
 
+    # Checking policy scope=SCOPE.AUTH, action=PasskeyAction.AllowedAuthenticatorDeviceTypes. Independent of the
+    # same-named ENROLL scope policy: an admin may allow enrolling both device types but only allow authentication
+    # with one of them, or vice versa.
+    request.all_data[PasskeyAction.AllowedAuthenticatorDeviceTypes] = get_policy_value_set(
+        PasskeyAction.AllowedAuthenticatorDeviceTypes, scope=SCOPE.AUTH, user=user_object)
```
jelinaunger, line 2213: on the passkey login path there is no user object hence any user conditions never match. It might make sense to match these policies after the user is resolved from the token

nilsbehlen: reordered the policy evaluation. same was true for `disabled_token_types`, so i fixed that too

jelinaunger (approved):

PR #5812, opened 2026-08-21
Drop the legacy plaintext fallback from Challenge.data

### LINKED ISSUE #5737: Clear the challenge table on upgrade instead of converting it, and drop the legacy plaintext compat code
### Summary

Challenge data encryption (v3.14, migration `a1b2c3d4e5f6`) currently *converts* existing plaintext challenge rows in place, and the model carries a permanent legacy-tolerance branch so pre-migration rows stay readable. Challenges are ephemeral — they live for one validity window, typically a couple of minutes — and privacyIDEA upgrades always involve downtime, so there is no rolling-upgrade window in which an old challenge has to survive.

Clearing the `challenge` table during the upgrade instead of converting it lets the compat code go away entirely, in the same release that introduced the need for it.

### Why now

Both the encryption and the Redis challenge cache are new in 3.14 and unreleased. If the compat code ships, it is permanent. Removing it before release costs nothing operationally: the worst case is that users who were mid-authentication at the moment of the upgrade restart their login, which is already the documented expectation for any privacyIDEA upgrade.

### Proposed change

| Where | Change |
|---|---|
| `privacyidea/migrations/versions/a1b2c3d4e5f6_encrypt_sensitive_db_fields.py:119-148` | replace the row-by-row encrypt loop with `DELETE FROM challenge`; the mirroring decrypt block at ~193 becomes a no-op. The `challenge.data` 512 -> 2000 column resize stays, since new rows need the width. The `_looks_encrypted` idempotency guard is no longer needed for this table (it stays for the SMS gateway options). |
| `privacyidea/models/challenge.py:61-71` | drop the raw fallback in the `data` property. A decrypt failure stops meaning "probably a legacy plaintext row" and becomes a real error. |
| `tests/test_lib_challenges.py:285,306` | remove `test_06_challenge_legacy_plaintext_data_readable` and `test_07_challenge_legacy_json_data_readable`. |
| `READ_BEFORE_UPDATE.md` | state that open challenges are cleared during the upgrade, so authentications in flight at that moment must be restarted. |

### Interaction with #5736

#5736 encrypts the `data` field on the Redis path. Its `_deserialize` deliberately drops entries it cannot decrypt rather than falling back to the raw value, because Redis has no legacy plaintext population to be compatible with. Today that makes the two backends inconsistent — the DB path falls back to raw, the Redis path fails closed — and the deviation has to be explained in a comment.

After this change both backends fail closed on an undecryptable value, and that comment can be reduced to a plain statement of the rule.

### Note for whoever merges the migration heads

`a1b2c3d4e5f6` is currently one of seven unmerged migration heads on master. When those heads are linearised, the challenge-clearing step must run after any other pending work against the `challenge` table, so the delete does not execute before a column change that a later head assumes.

nilsbehlen:
## Summary
- `Challenge.data` no longer falls back to returning the raw ciphertext when decryption fails; it now returns whatever `decryptPassword()` produces, so an undecryptable value surfaces as a real error instead of being treated as legacy pre-migration plaintext.
- Removes the now-unused `FAILED_TO_DECRYPT_PASSWORD` import.

The rest of #5737 was already implemented on master (the challenge table is cleared via migration `c3d4e5f6a7b8`, `READ_BEFORE_UPDATE.md` documents it, and the legacy tests were removed in `6c82a8786`). This fallback had been reintroduced by a later merge of the `5736/encrypt-challenge-data-in-redis` branch, which had forked before that removal landed.

Closes #5737

## Test plan
- [x] `tests/test_lib_challenges.py` (13 passed)
- [x] `tests/test_lib_cache_redis.py` (26 passed, 61 skipped - no Redis available locally)


privacyidea/models/challenge.py
```diff
@@ -65,14 +64,7 @@ def data(self):
...
             return raw
-        decrypted = decryptPassword(raw)
-        # decryptPassword() returns the sentinel verbatim on failure, so compare
-        # for equality: a prefix match would also discard a legitimately
-        # decrypted value that happens to start with that text.
-        if decrypted and decrypted != FAILED_TO_DECRYPT_PASSWORD:
-            return decrypted
-        # Legacy unencrypted data or decryption failure - return as-is
-        return raw
+        return decryptPassword(raw)
```
jelinaunger: Returning the FAILED_TO_DECRYPT_PASSWORD string would raise in json.loads(). Which is not too bad as it is caught there and was the same before with returning raw. However, it would be cleaner to return an empty string (same as if there is no data at all) and log decryption failed.

jelinaunger (commented):
The linked issue also states that encrypting existing challenge data in the migration can be removed, which is not contained in this PR.

nilsbehlen:
@jelinaunger Good catch, and you're right that this wasn't in the PR - fixed now.

For context on why it wasn't touched originally: the `DELETE FROM challenge` part of #5737 landed separately, in `c3d4e5f6a7b8_clear_challenge_table.py` (from the dict-only-format work), which runs immediately after `a1b2c3d4e5f6`. So the encrypt loop for `challenge.data` in `a1b2c3d4e5f6` was already dead work - every row it touched got deleted by the very next migration on any real upgrade - it just hadn't been removed yet.

I've now dropped that loop (and its downgrade counterpart) from `a1b2c3d4e5f6`, keeping only the `challenge.data` column resize (still needed for the wider encrypted JSON new challenges store) and the SMS gateway option encryption, which is unrelated.

jelinaunger (approved):

PR #5818, opened 2026-08-24
allow more ssh key types

### LINKED ISSUE #4987: SSH Token issues
- The restriction of the key type seems unnecessary, we do not use that information and the corresponding SSH-Server should handle the possible allowed key-types. We could still save the key-type as a token-info to filter for it.
- The public key is saved as an encrypted token-info in the database. Currently we can not access the decrypted value via the API. And it looks like the `SSH` Application can not either because the token-info is not decrypted here:
  https://github.com/privacyidea/privacyidea/blob/b66da1800414df3a82424aa0840199cc6af92959/privacyidea/lib/tokens/sshkeytoken.py#L164

jona-samuel:
closes #4987


privacyidea/lib/tokens/sshkeytoken.py
```diff
@@ -134,32 +178,102 @@ def update(self, param):
...
         self.add_tokeninfo("ssh_type", key_type)
         self.add_tokeninfo("ssh_comment", key_comment)
 
         # call the parents function
         TokenClass.update(self, param)
 
+        # Store an integrity checksum of the SSH key data in the encrypted
+        # OTP key field, so that a manipulation of the database entries can
+        # be detected in get_sshkey().
+        self.token.set_otpkey(compute_ssh_key_checksum(self.token.serial, key_type, key, key_comment))
```
nilsbehlen: The checksum is only kept in sync here (and in `import_token()`) — any other write path that touches the `ssh_key`/`ssh_type`/`ssh_comment` tokeninfo bypasses this recompute. In particular, `POST /token/info/<serial>/<key>` (the generic `settokeninfo` endpoint, guarded only by the ordinary `settokeninfo` policy, not `sshkey_read`) writes straight to `tokeninfo` via `lib.token.add_tokeninfo()`. An admin who legitimately fixes e.g. `ssh_comment` through that endpoint will desync the checksum, and the token then fails its own integrity check on the next `GET /token/sshkey/<serial>`, raising `TokenAdminError` and logging a false "database entries might have been manipulated" warning.

Suggestion: override `add_tokeninfo()` (and `delete_tokeninfo()`) in `SSHkeyTokenClass` to recompute the checksum whenever one of those three keys changes. That makes the token class itself the single enforcement point regardless of the caller (`update()`, the generic tokeninfo endpoint, event handlers, ...), instead of trying to patch every call site individually.

privacyidea/lib/tokens/sshkeytoken.py
```diff
@@ -23,23 +23,68 @@
...
+    part of the checksum), any manipulation of the SSH key data in the
+    database can be detected.
+
+    :param serial: The serial of the token
+    :param key_type: The SSH key type, e.g. "ssh-rsa"
+    :param key: The decrypted base64 encoded public key blob
+    :param comment: The SSH key comment
+    :return: hexlified SHA256 checksum
+    """
+    data = "\n".join([serial or "", key_type or "", key or "", comment or ""])
```
nilsbehlen: Joining the fields with a bare `"\n"` isn't collision-free: since `comment` is free text, a checksum computed over `(key_type='ssh-rsa', key='AAA', comment='X\nY')` serializes to the same string as one over `(key_type='ssh-rsa', key='AAA\nX', comment='Y')`, so the checksum doesn't uniquely bind the key/comment boundary.

Suggestion: use an unambiguous encoding instead, e.g. `json.dumps([serial or "", key_type or "", key or "", comment or ""])` — JSON escapes embedded newlines/quotes so no two distinct inputs can serialize identically. Note this changes the checksum for every already-migrated token, so it needs to ship together with the migration (or the migration needs to be updated) rather than as a follow-up.

nilsbehlen (commented):
A few findings from a review pass over the SSH-key-types changes on this branch.

privacyidea/lib/tokens/sshkeytoken.py
```diff
@@ -134,32 +178,102 @@ def update(self, param):
...
         """
-        ti = self.get_tokeninfo()
-        key_type = ti.get("ssh_type")
-        key_comment = ti.get("ssh_comment")
-        # get the ssh key directly, otherwise it will not be decrypted
-        sshkey = self.get_tokeninfo("ssh_key")
+        key_type, sshkey, key_comment = self._get_ssh_key_parts()
+        # Verify the integrity checksum
+        stored_checksum = self._get_stored_checksum()
+        if not stored_checksum:
```
nilsbehlen, line 333: `_get_stored_checksum()` returns `""` both when `key_enc`/`key_iv` are genuinely unset (never migrated) and when they're present but unreadable/corrupted (the `except Exception` branch above). This code path then reports both cases with the same "Please run the database migration..." message, which is actively misleading for the corruption case — an admin would rerun a migration that does nothing to fix corrupted ciphertext.

Suggestion: have `_get_stored_checksum()` distinguish the two (e.g. `key_enc is None` → unmigrated vs. present-but-decrypt-fails → corrupted, via a sentinel or a distinct exception) and give `get_sshkey()` a separate message for "checksum is unreadable, the token may be corrupted".

privacyidea/api/token.py
```diff
@@ -1647,6 +1647,50 @@ def get_serial_by_otp_api(otp=None):
...
+    SSH key of their own tokens.
+
+    :param serial: path component, the serial of the SSH key token.
+    :status 200: ``{"sshkey": "<type> <key> <comment>"}`` in
+        ``result.value``.
+    """
+    # check_token_action removes tokens the caller may not access and reports
+    # them in "not_authorized_serials". If our serial ended up there, deny.
+    not_authorized_serials = get_optional(request.all_data, "not_authorized_serials", [])
+    if serial in not_authorized_serials:
```
nilsbehlen: This manual re-check looks unreachable: `check_token_action()` (via the `@prepolicy` above) already raises `PolicyError` before returning when the single serial in the request is unauthorized (see `not_authorized_serials and len(all_serials) == 1` in `prepolicy.py`), and this route always has exactly one path-supplied serial. So this `if serial in not_authorized_serials: raise PolicyError(...)` can never execute.

Suggestion: drop it (and the `not_authorized_serials` lookup above it) unless there's a code path I'm missing where `check_token_action` doesn't already cover this.

privacyidea/api/token.py
```diff
@@ -86,7 +86,7 @@
                                            check_container_action, check_user_params,
                                            force_server_generate_key)
 from privacyidea.lib.challenge import (cancel_challenge, get_challenges, get_challenges_for_user,
```
nilsbehlen, line 88: Minor: these two import continuation lines lost their alignment to the opening parenthesis (looks like incidental reflow from editing a nearby import), unrelated to the SSH key feature itself. Worth a quick re-indent to keep the diff free of unrelated noise.

nilsbehlen (changes requested):

nilsbehlen (approved):

PR #5821, opened 2026-08-24
4914/conditional access auth log refinements

### LINKED ISSUE #5766: Conditional Access: Auth log refinements
* add info why an event happened (token disabled, failcounter, authz policy, ...)
* rename events:
    LOCK_USER -> LOCK_USER_TEMPORARY
    PERMANENT_LOCK_USER -> LOCK_USER_PERMANENT
* use plural for request parameter names if they support lists 
* log the endpoint (/auth, /validate/check, ...)
* CA Outcome: switch action and policy name 
     -> policy name in the top row
     -> action below

henrik-falk:


privacyidea/lib/conditional_access/engine.py
```diff
@@ -56,29 +56,32 @@ class LockoutAction(str, Enum):
...
-    LOCK_USER = "LOCK_USER"
-    PERMANENT_LOCK_USER = "PERMANENT_LOCK_USER"
+    LOCK_USER_TEMPORARY = "LOCK_USER_TEMPORARY"
+    LOCK_USER_PERMANENT = "LOCK_USER_PERMANENT"
     EMAIL_ADMIN = "EMAIL_ADMIN"
     EMAIL_USER = "EMAIL_USER"
-    BLOCK_IP = "BLOCK_IP"
-    PERMANENT_BLOCK_IP = "PERMANENT_BLOCK_IP"
+    BLOCK_IP_TEMPORARY = "BLOCK_IP_TEMPORARY"
+    BLOCK_IP_PERMANENT = "BLOCK_IP_PERMANENT"
```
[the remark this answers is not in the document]
henrik-falk: never was in master so historical outcomes dont matter.

privacyidea/migrations/versions/0147d78cbace_authentication_log.py
```diff
@@ -51,9 +52,11 @@ def upgrade():
             sa.Column('username', _unicode_case_sensitive(255), nullable=True),
             sa.Column('user_role', _unicode_case_sensitive(30), nullable=True),
             sa.Column('event_type', _unicode_case_sensitive(40), nullable=False),
+            sa.Column('reason', _unicode_case_sensitive(40), nullable=True),
             sa.Column('timestamp', sa.DateTime(), nullable=False),
             sa.Column('source_ip', _unicode_case_sensitive(50), nullable=True),
             sa.Column('client_label', _unicode_case_sensitive(1024), nullable=True),
+            sa.Column('endpoint', _unicode_case_sensitive(255), nullable=True),
```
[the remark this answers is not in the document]
henrik-falk: never was in master so historical outcomes dont matter

privacyidea/api/authentication_log.py
```diff
@@ -39,11 +41,17 @@
...
-# Each filter parameter maps 1:1 to a get_authentication_logs_paginate keyword argument; the ca_* ones filter on the
-# entry's conditional-access outcomes rather than a column of its own row, and ca_dry_run is parsed separately because
-# it is a boolean, not a list of values.
-_FILTER_PARAMS = ["resolver", "uid", "realm", "username", "user_role", "event_type", "source_ip", "serial",
-                  "transaction_id", "attempt_id", "client_label", "ca_action_type", "ca_policy_name"]
+# The list-valued filter query parameters, mapped to the get_authentication_logs_paginate keyword argument each one
+# feeds. Every one of them takes a list of values, so the query parameter is plural while the library keyword names the
+# single field it matches. The ca_* ones filter on the entry's conditional-access outcomes rather than on a column of
+# its own row; ca_dry_run is parsed separately because it is a boolean, not a list of values.
+_FILTER_PARAMS = {"resolvers": "resolver", "uids": "uid", "realms": "realm", "usernames": "username",
```
jelinaunger: why? Parameters can be plural on the lib layer as well. This would be consistent and simplifies the implementation.

privacyidea/models/authentication_log.py
```diff
@@ -76,11 +83,22 @@ class AuthenticationLog(MethodsMixin, db.Model):
...
+    reason: Mapped[str | None] = mapped_column(
+        case_sensitive_unicode(authentication_log_column_length["reason"]))
     timestamp: Mapped[datetime] = mapped_column(DateTime, default=utc_now, nullable=False)
     source_ip: Mapped[str | None] = mapped_column(
         case_sensitive_unicode(authentication_log_column_length["source_ip"]))
     client_label: Mapped[str | None] = mapped_column(
         case_sensitive_unicode(authentication_log_column_length["client_label"]))
+    # The endpoint the request authenticated against, as its request path ("/auth", "/validate/check", "/ttype/push").
+    # Nullable like the other request-scoped columns: an event recorded outside a request context (the CLI, the
+    # push_wait flow) has no endpoint to name.
```
jelinaunger: it's fine to make it not nullable, but the reasoning is wrong: an event should never be recorded outside a request context. CLI never produces auth log events and push wait runs through a request.

tests/authlog_utils.py
```diff
@@ -190,8 +193,15 @@ def assert_authentication_log_entry(entry: AuthenticationLog, user: User = None,
 
     Every other column of the authentication_log table is checked. The nullable columns default to their database default
     (None), so a column that is not passed is asserted to be empty — this enforces that a row carries *only* the data
-    it should and no leftover values. The auto-populated id and timestamp are checked for presence. The non-nullable
-    event_type is covered by the ordered list in :func:`assert_authentication_log`.
+    it should and no leftover values. ``endpoint`` and the two reason arguments are the exceptions: every row written
+    from a view carries the request path, and nearly every failed one carries a reason, neither of which says anything
+    about the outcome under test - so they are only compared when a caller passes them. Their own contracts are
```
jelinaunger: Disagree, other values tested here also occur frequently. The idea of the design making all values mandators is to enforce each information is defined explicitly and no information is silently missed in the log. It doesn't  cost us much to pass all information if we check for the auth log anyway but buys us the proof everything is logged accordingly under many different conditions and at different endpoints.

doc/conditional_access/authentication_log.rst
```diff
@@ -105,18 +106,139 @@ itself: ``USER_LOCKED``, ``IP_BLOCKED`` and ``ACCESS_DENIED``. They record why a
...
+     the enrollment was never completed.
+   ``TOKEN_TYPE_DISABLED``
+     a policy disabled this token's whole type for the request.
+   ``TOKEN_NOT_APPLICABLE``
+     the token excluded itself from this request, for example an
+     application-specific password whose service does not match.
+
+A policy refusing an otherwise valid authentication
+   ``AUTHORIZATION_POLICY``
+     an authorization policy denied the request.
```
jelinaunger: this does not give us any further insights. The reasons below are already the more specific ones (specific authz policies)

doc/conditional_access/authentication_log.rst
```diff
@@ -105,18 +106,139 @@ itself: ``USER_LOCKED``, ``IP_BLOCKED`` and ``ACCESS_DENIED``. They record why a
...
+   ``LAST_AUTH_TOO_OLD``
+     the token's last successful authentication is too long ago.
+
+The credentials
+   ``WRONG_USERSTORE_PASSWORD``
+     the user store password was wrong.
+   ``WRONG_TOKEN_PIN``
+     the token PIN was wrong, or one was given where none was expected.
+   ``WRONG_OTP``
+     the first factor was right, or not required, but the OTP was not.
```
jelinaunger: they are all already covered by the auth events (PASSWORD_FAIL, PIN_FAIL, MFA_FAIL/CHALLENGE_ANSWERED_FAIL). wrong otp might be useful

doc/conditional_access/authentication_log.rst
```diff
@@ -105,18 +106,139 @@ itself: ``USER_LOCKED``, ``IP_BLOCKED`` and ``ACCESS_DENIED``. They record why a
...
+     the response did not match the challenge.
+   ``CHALLENGE_UNKNOWN_TRANSACTION``
+     the transaction holds no challenge for this token: already consumed,
+     belonging to another token, or never issued.
+   ``CHALLENGE_EXPIRED``
+     the challenge had lapsed when the response arrived.
+   ``CHALLENGE_DECLINED_ON_DEVICE``
+     the challenge was rejected on the device.
+   ``TOKEN_NOT_FIT_FOR_CHALLENGE``
+     the response matched, but the token may no longer complete a challenge.
```
jelinaunger: here it might be interesting why it is not fit for the challenge? Does this result in the token states above?

doc/conditional_access/authentication_log.rst
```diff
@@ -105,18 +106,139 @@ itself: ``USER_LOCKED``, ``IP_BLOCKED`` and ``ACCESS_DENIED``. They record why a
...
+``/validate/triggerchallenge``
+  a challenge triggered by an administrator.
+``/validate/initialize``
+  the anonymous bootstrap of a FIDO2/passkey challenge before login.
+``/ttype/push``
+  a push challenge answered on the smartphone, which reaches the server out of
+  band.
+
+The column is empty for an event recorded outside a request, for example from
+the command line. The same value is what an *Endpoint* condition of a lockout
```
jelinaunger: authentication from the command line is not possible (see other comment), just a reminder to fix it here as well.

privacyidea/api/lib/postpolicy.py
```diff
@@ -1278,10 +1278,20 @@ def is_authorized(request, response):
...
+                # Name the policy that denied it: with several authorization policies in play, "which rule do I
+                # have to change" is the whole question the log has to answer.
+                deciding_policies = next(iter(authorized_pol.values()), None)
+                reason_detail = {"policies": deciding_policies} if deciding_policies else None
                 if context.amendable is not None:
-                    context.reclassify(AuthEventType.NOT_AUTHORIZED)
+                    # Correcting the staged event. The detail is merged, so the per-serial reasons the token layer
+                    # recorded survive alongside the policy that overrode them.
+                    context.reclassify(AuthEventType.NOT_AUTHORIZED,
+                                       reason=str(AuthEventReason.AUTHORIZATION_POLICY),
```
jelinaunger: no need for the str conversion here, it is done later on in _truncate anyway

privacyidea/api/lib/postpolicy.py
```diff
@@ -1278,10 +1278,20 @@ def is_authorized(request, response):
             # its rejection row already records why, and a NOT_AUTHORIZED row here would bury that reason and hand the
             # lockout counters an attempt the lock itself produced.
             if not context.rejected_by_conditional_access:
+                # Name the policy that denied it: with several authorization policies in play, "which rule do I
+                # have to change" is the whole question the log has to answer.
+                deciding_policies = next(iter(authorized_pol.values()), None)
+                reason_detail = {"policies": deciding_policies} if deciding_policies else None
```
jelinaunger: a helper function to create the reason details dict would be good to have a single place defining the dicts structure

privacyidea/api/lib/prepolicy.py
```diff
@@ -2869,7 +2870,11 @@ def auth_timelimit(request, action):
         result, reply_dict = check_max_auth_success(user, user_search_dict, check_validate_check=not local_admin)
 
     if not result:
-        log_authentication(AuthEventType.NOT_AUTHORIZED, request, user=user)
+        # check_max_auth_fail / _success name which limit was hit and the policy that set it. Popped from reply_dict,
+        # which is handed to the client as the error details.
+        reason, reason_detail = pop_auth_event_reason(reply_dict)
```
jelinaunger: generally, we should avoid these things. It is safe today, but a new caller might forget / not know it needs to remove this information before returning it to the client (it does not pop them already at another call site which is still safe due to a later explicit read of the information of interest).
Since both functions are only called twice I suggest changing the return type of them to clearly separate information returnable to the client and internal additional information which should not leak

jelinaunger: since this a general problem over multiple policy helper functions, it another option is to write the information to g instead of changing the signature of all these functions.

jelinaunger: Or we say for now it is fine to only log the reason (policy action) but not the explicit policy name yet. We could add this later on as well. And it is not even consistent yet, e.g. for check_last_auth_policy also the policy name is not added to the log.

privacyidea/lib/conditional_access/authentication_event_types.py
```diff
@@ -106,6 +121,150 @@ def __str__(self) -> str:
...
+    TOKEN_NOT_YET_ENROLLED = "TOKEN_NOT_YET_ENROLLED"
+    # A policy disabled this token's whole type for the request (PolicyAction.DISABLED_TOKEN_TYPES).
+    TOKEN_TYPE_DISABLED = "TOKEN_TYPE_DISABLED"
+    # The token type excluded itself from *this* request (TokenClass.use_for_authentication) - e.g. an
+    # application-specific password whose service_id does not match the one the request names.
+    TOKEN_NOT_APPLICABLE = "TOKEN_NOT_APPLICABLE"
+
+    # --- authorization, i.e. a policy refusing an otherwise valid authentication ---------------------------------
+    # An authorization policy denied the request outright (PolicyAction.AUTHORIZED = deny).
+    AUTHORIZATION_POLICY = "AUTHORIZATION_POLICY"
```
jelinaunger: ah okay, maybe AUTHORIZATION_DENIED / AUTHORIZED_DENY is more clear (no need for policy in the name as the other ones are policies as well)

privacyidea/lib/conditional_access/authentication_event_types.py
```diff
@@ -106,6 +121,150 @@ def __str__(self) -> str:
...
+
+    def __str__(self) -> str:
+        return self.value
+
+
+# Request-level precedence of the reasons, highest signal first, used when the tokens of one request failed for
+# different reasons (see reduce_request_reasons). The order follows how much the reason narrows down what the admin
+# must do: a policy decision outranks any token state, because it applies whatever the tokens look like; a permanent
+# token state (revoked) outranks a transient one (failcount, which a reset clears); and a wrong credential ranks below
+# every state, since a state is the thing that made the credential moot.
```
jelinaunger: I think this comment is misleading. If I get it correctly, the only place where concurrent reasons can occur are for no usable token (disabled, revoked, ...) so only a subset of the ranking list ever applies. Additionally, I think it is not possible to define a reasonable ranking there. All reasons are useful and important to show, reducing it to a single cause might be misleading (even the others are recorded as well, they are less prominent shown). 

I would make the reason a list. Today it is only an additional explanation and should be queryable, but it is not planned to use it as CA condition (for this we have auth events). Storing all reasons makes the precedence stuff obsolete.

tests/test_api_authentication_event_logging.py
```diff
@@ -792,7 +968,26 @@ def test_is_authorized_deny_logs_not_authorized(self):
...
+            res = self._post('/validate/check', {"user": self.username, "pass": f"{self.pin}000000"})
+            self.assertEqual(400, res.status_code, res.json)
+        finally:
+            delete_policy("authlog_deny")
+        entries = assert_authentication_log([AuthEventType.NOT_AUTHORIZED])
+        assert_authentication_log_entry(entries[AuthEventType.NOT_AUTHORIZED], user=self.user,
+                                        serials={self.serial},
+                                        reason=AuthEventReason.AUTHORIZATION_POLICY,
+                                        policies=["authlog_deny"],
+                                        reasons={self.serial: AuthEventReason.WRONG_OTP})
```
jelinaunger: not sure if I would expect this. Cause wrong OTP is not the reason this authentication failed. But I somehow get the idea that the information might be useful. It's just a bit confusing in this context to me.

privacyidea/migrations/versions/f4a1c2b6d8e3_authentication_log_reason.py
```diff
@@ -0,0 +1,110 @@
...
+            print(f"Could not add table '{TABLE}' to database.")
+            raise
+
+    try:
+        # Carry the existing classifications over: each row's single reason becomes its first (and, for historic
+        # entries, only) reason row. Done in the database rather than row by row, so a large log migrates in one
+        # statement. Rows with no reason - every success - simply produce none.
+        op.execute(sa.text(f"INSERT INTO {TABLE} (auth_log_id, reason) "
+                           f"SELECT id, reason FROM {PARENT} WHERE reason IS NOT NULL"))
+        op.drop_column(PARENT, 'reason')
```
jelinaunger: no need for this, the auth log table is not yet released, hence no existing reasons which could get lost. The whole migration script can be folded into the existing one for the authlog to have it clean at one place

privacyidea/models/authentication_log_reason.py
```diff
@@ -0,0 +1,74 @@
...
+class AuthenticationLogReason(db.Model):
+    """
+    Why an authentication-log row came out the way it did: one row per :class:`AuthEventReason` the request produced,
+    ordered by :data:`~privacyidea.lib.conditional_access.authentication_event_types.REASON_PRECEDENCE` (highest
+    signal first) as their ids ascend.
+
+    A request rarely fails for exactly one reason - a user with three tokens can have one revoked, one past its
+    failcount and one that simply got the wrong OTP - so the reasons are a **list**, not the single highest-ranked
+    one. Kept in a table of its own rather than as a column on the parent for the reason the parent's column existed
+    at all: "every NO_USABLE_TOKEN caused by the failcounter" has to be a plain indexed predicate, and neither a
```
jelinaunger: well there is not necessarily the need for an index on the reason as we do not use it on the hot path for searching the auth log. Indices are not entirely for free. Just to keep in mind and maybe double check if introducing two indices here might be an overkill for a performant search query on the auth log table.

privacyidea/lib/conditional_access/authentication_event_types.py
```diff
@@ -267,6 +265,27 @@ def order_request_reasons(reasons) -> list[AuthEventReason]:
...
+
+    :param reasons: what each token was found to be, keyed by serial
+    :param policies: the names of the policies that decided the request
+    :return: the detail dict, or ``None`` if both parts are empty
+    """
+    detail = {}
+    if reasons:
+        detail["reasons"] = {serial: str(reason) for serial, reason in reasons.items()}
+    if policies:
+        detail["policies"] = policies
```
jelinaunger, line 301: are there any policies left which are logged?  Cause I would make it consistent; either log the name always or never, but not for this policy we do, but for this one not. This could be confusing.

privacyidea/lib/token/auth.py
```diff
@@ -739,17 +740,23 @@ def check_token_list(token_object_list: list[TokenClass], passw: str, user: User
...
                         else list(token_reasons.values()))
     # Passed as recorded: order_request_reasons coerces and drops what it cannot, while converting here would raise
     # past that guard, from inside the generator, and fail the authentication over a mislabelled reason.
     ordered_reasons = order_request_reasons(deciding_reasons)
-    if ordered_reasons and reduced_event and outcome_of(reduced_event) == AuthEventOutcome.FAILURE:
-        reply_dict[AUTH_EVENT_REASON_KEY] = ordered_reasons
-        reply_dict[AUTH_EVENT_REASON_DETAIL_KEY] = {"reasons": dict(token_reasons)}
+    if token_reasons and reduced_event and outcome_of(reduced_event) == AuthEventOutcome.FAILURE:
+        # The per-serial map is recorded whenever a token was found to be something, even where none of those
+        # findings explains the winning event: "which token was in which state" is what the detail is for.
```
jelinaunger, line 763: For wrong credentials I agree, cause the credentials could have been right for a disabled token, so the reason for each token is of interest.

But if a policy let the authentication fail, the state of the tokens are not of interest in my opinion.

To generalize this: If we are already on the token level to check which token matches add the reason for all of them why they didn't match, but nowhere else.

However, I found already one exception: CHALLENGE_UNKNOWN_TRANSACTION. If the user has multiple valid tokens, this event will occur and it is right to add it to the detailed reasons in the info, but I would exclude it from the main reason list if a token exists for which the transaction ID matched. Otherwise you can have CHALLENGE_WRONG_RESPONSE and CHALLENGE_UNKNOWN_TRANSACTION next to each other which might be confusing and do not help the admin. Sure, you could also argue that the credentials might have been correct and just the transaction id was wrong and recording all reasons would simply consistent.

tests/test_api_authentication_event_logging.py
```diff
@@ -534,16 +545,18 @@ def test_the_row_lists_every_reason_while_every_token_keeps_its_own(self):
         self._assert_failed(self._authenticate(f"{self.pin}755224"))
```
jelinaunger, line 596: the same test for a challenge response token might be interesting as well (if not already there)

jelinaunger (approved):

PR #5835, opened 2026-08-27
5807/ca configurable reset on success

### LINKED ISSUE #5807: Conditional Access: Make reset on success configurable
For users a successful login can reset the event count with a successful login. This should be configurable.

henrik-falk:
A completed login floored a user policy's count at that login, unconditionally: the thresholds always measured consecutive failures since the last successful authentication. That is the right default - it keeps a legitimate user who just logged in from being re-locked by stale pre-login failures on the next typo - but it is not the only thing an admin may want a threshold to mean. Without it the threshold measures total tracked failures in the window outright.

reset_on_success is a per-policy flag defaulting to true, so existing behaviour is unchanged. It reaches only the post-response user count: a source-IP policy aggregates a signal across accounts, where one account's legitimate login must not clear it, and the pre-auth allow/deny decision is a stateless reject that lifts as the failures age out of the window. Both stay as they were.

The column goes into the existing create-table migration, as count_mode and retrigger_above_threshold did on this branch: none of the conditional-access migrations has been released yet.

The editor gets a checkbox next to the count mode, disabled for a source-IP policy so a target change explains itself rather than silently changing what the setting means.

Closes #5807


jelinaunger (commented):
Keeping all these restrictions / exceptions while making it configurable is a bit confusing to me.

For the source IP it is reasonable to me, well implemented also in the UI so it should be understandable to the users and you might be right that there is no reasonable use-case using the reset on success option for source IPs.
But there should be a Parameter Error raised in case the option is set to avoid the policy silently ignores this configuration.

But I would remove the restriction for DENY. I thinkt it can make sense to use this option there as well. Additionally, it is silently ignored and nowhere explained in the UI.
At least there should be according tests.

jelinaunger:
Also please add the new option to the templates.

jelinaunger (approved):

PR #5837, opened 2026-08-27
5819/ca unique actions per stage

### LINKED ISSUE #5819: Conditional Access: Unique actions per stage
Within a single stage most actions should only be configurable once (lock user, ...). Additionally, lock user temporarily + perma lock should be avoided.
Only email actions are allowed multiple times to enable sending different mails to different adresses

henrik-falk:
A stage could carry any number of actions of any type, including combinations that cannot do what they read as. Two LOCK_USER actions lock for whichever of the two durations is applied last; a timed and a permanent lock in one stage resolve by row order, which is not defined (LockoutPolicyStage.actions has no order_by, and the upsert refuses to downgrade a permanent restriction); an ALLOW next to a DENY is dead configuration, because the decision fails closed.

Every action is now singleton per stage except EMAIL_ADMIN and EMAIL_USER, where a second copy is the point - it notifies a second set of recipients, which is the one case a repeat does something the first copy cannot. The timed/permanent lock, timed/permanent block and ALLOW/DENY pairs are rejected outright.

The check sits inside _validate_stages, next to the duplicate-threshold rule and deliberately not beside _validate_target_actions: that one re-reads the stored stages on update, so putting it there would make a policy written before this rule impossible to rename or switch off. Only submitted stages are judged.

The rules are served per target on GET /conditionalaccess/targets rather than hard-coded in the WebUI, so the editor greys out the types a stage has already taken, flags the offending action and blocks Save from the same definition the backend enforces. A target is only told about a group it could actually form.

The engine is unchanged: its DENY-wins and no-downgrade guards stay as defence in depth for rows written before this rule.

Closes #5819


privacyidea/static_new/src/testing/mock-services/mock-conditional-access-policy-service.ts
```diff
@@ -103,6 +111,53 @@ export class MockConditionalAccessPolicyService implements ConditionalAccessPoli
...
+        return "duplicate";
+      }
+      const conflicting = groups.some(
+        (group) =>
+          group.includes(action.action_type) &&
+          earlier.some((type) => type !== action.action_type && group.includes(type))
+      );
+      return conflicting ? "exclusive" : null;
+    }
+  );
```
jelinaunger, line 130: mirror the real implementation is not the purpose of a mock function.

<br>

A mock that re-implements the code under test can't detect a change in it. conditional-access-action-item.component.ts:200,207 just delegates to the service, so the four tests at spec lines 164, 173, 180, 
  185 only exercise the mock's copy — break actionConflict in the real service and they all stay green. What they do assert is already covered against the real service in conditional-access-policy.service.spec.ts:314-352, so the copy buys no coverage and adds a second place the rule can drift.                                                                                  

<br>
                                                                                                                                                                                                            
  Suggested: make those three return canned values (() => new Set(), () => null, () => []) and let each spec drive its case with mockReturnValue. The component spec then tests the component's actual job —   
  that it forwards siblingActions/actionIndex/target, and that conflictMessage() maps "duplicate"/"exclusive"/null to the right text — and the rules stay tested once, in the service spec.

jelinaunger (approved):

PR #5838, opened 2026-08-27
5813/ca manual user lock

### LINKED ISSUE #5813: Conditional Access: Manual user lock
To the user details page add the action to manually lock the user. Optionally with a custom error message (after PR 5784 is merged).
Same could be done for an IP maybe on the IP blocklist page.

Think about if it makes sense to store the lock cause: manual or by CA

henrik-falk:
Conditional-access state was write-only from the engine: an admin could see and clear a lock, but only a policy could impose one. The user-details page now has a Lock action and the blocklist page a Block IP action, both asking whether the restriction holds until it is lifted (the default) or for a chosen duration.

Nothing in the authentication path changes: the pre-check reads the state row whoever wrote it, so a manual lock is enforced exactly like a policy lock and is cleared by the same Unlock action.

The row now records its cause, MANUAL or POLICY. That is worth storing because a manual restriction has nowhere else to record its provenance: what a policy did is a ConditionalAccessOutcome hanging off the authentication-log row of the request that triggered it, and a manual lock has no such request. The cause travels with the expiry, so it always describes the restriction now in force - a policy lock that strengthens an admin's timed one becomes a policy lock, while a manual permanent lock is never downgraded and keeps its cause. Both management pages show it as a column, and the locked-users list filters on it.

A manual write is authoritative where a policy action is not: it may replace a permanent lock with a timed one, because the engine's never-weaken rule exists to stop the firing order of two policies deciding the outcome, and an admin stating the outcome is not a race. A never-block IP is refused with an explanation rather than skipped silently, which is right for an automatic action but indistinguishable from success for a deliberate one.

user_lockout_set and blocklist_set are separate rights from the *_reset ones: clearing a restriction is recoverable, imposing one is not. An out-of-scope target is refused outright rather than quietly doing nothing.

pi-manage conditionalaccess gains lock-user and block-ip, and both list commands now print the cause. The eight conditional-access admin rights get their doc anchors, which the API docstrings already referenced.

The custom error message the issue mentions is not included: it depends on UserLockoutState.error_message from PR 5784, which is not on this branch.

Closes #5813


doc/webui/index.rst
```diff
@@ -225,6 +225,37 @@ the regular endpoints
...
+A restriction does not have to come from a policy. The *User Lock State*
+card on a user's details page offers a **Lock** action, and the *Blocklist*
+page a **Block IP** action; both ask whether the restriction lasts until an
+administrator lifts it (the default) or for a chosen duration.
+
+A manual restriction is stored in the same place a policy writes to and is
+enforced by the same pre-check, so it behaves exactly like a policy lock at
+authentication time and is cleared by the same *Unlock* / *Unblock* action.
+What is recorded alongside it is its **cause**: the *Locked Users* and
+*Blocklist* pages show a *Cause* column reading *Manual* or *Policy*, and the
```
jelinaunger: To avoid confusion with the other policies, I suggest using Conditional Access here instead

privacyidea/api/conditional_access.py
```diff
@@ -500,6 +523,66 @@ def get_user_lock():
...
+    the same pre-check: the lock is recorded with the cause ``MANUAL`` so the two can be told apart on the
+    Locked Users page, but nothing in the authentication path treats it differently. Unlike a policy lock it
+    is authoritative - it replaces a stronger lock rather than declining as a weakening.
+
+    Requires the admin policy action :ref:`policy_user_lock_set`, which is deliberately separate from
+    ``user_lock_reset``: clearing a restriction is recoverable, imposing one is not.
+
+    One user identifier is required: user or user_id
+
+    :jsonparam user: login of the user to lock
```
jelinaunger: should be username or similar to be more specific and avoid confusion with the user object

privacyidea/cli/pimanage/conditional_access.py
```diff
@@ -92,7 +94,28 @@ def list_locked_users_cmd():
...
-                   f"expires={_format_expiry(user['lock_expires_at'])}")
+                   f"expires={_format_expiry(user['lock_expires_at'])}\tcause={user['lock_cause']}")
+
+
+@conditional_access_cli.command("lock-user", help="Lock a single user by hand.")
+@click.argument("login")
+@click.option("--realm", required=True, help="The realm of the user.")
+@click.option("--resolver", help="The resolver of the user (only needed to disambiguate).")
+@click.option("--duration", type=int, help="How long the lock lasts, in seconds. Omitted, it is permanent.")
+def lock_user_cmd(login, realm, resolver, duration):
```
jelinaunger, line 105: Generally, we do not need to mirror all functions provided by the API / in the WebUI also here in the cli. But since it is already here now, we can keep it for sure.

privacyidea/lib/policy.py
```diff
@@ -2001,13 +2001,22 @@ def get_static_policy_definitions(scope=None):
...
             PolicyAction.BLOCKLIST_READ: {'type': 'bool',
                                           "desc": _("Admin is allowed to read the conditional-access blocklist."),
                                           "group": GROUP.SYSTEM},
             PolicyAction.BLOCKLIST_RESET: {'type': 'bool',
                                            "desc": _("Admin is allowed to remove entries from the "
                                                      "conditional-access blocklist."),
                                            "group": GROUP.SYSTEM},
+            PolicyAction.BLOCKLIST_SET: {'type': 'bool',
+                                         "desc": _("Admin is allowed to add an IP address to the "
+                                                   "conditional-access blocklist manually."),
```
jelinaunger: not sure if I would name it conditional-access blocklist. The moment we add the manual lock/blocks here these list are generic lock and block list.

privacyidea/api/conditional_access.py
```diff
@@ -500,6 +523,66 @@ def get_user_lock():
...
+    user = request.User
+    if not user or not user.exist():
+        user = User(uid=user_id, login=username, realm=realm, resolver=resolver)
+    if user.is_empty() or not user.exist():
+        raise ParameterError(f"The user {username or user_id}@{realm} does not exist.")
+    # A pre-flight check rather than a scoped WHERE clause: a write has exactly one target, so an
+    # out-of-scope one is refused loudly. A lock that did not happen must not look like one that did -
+    # the opposite trade-off from the DELETE, which may match several rows.
+    if not user_matches_scopes(user, get_policy_visibility_scopes(PolicyAction.USER_LOCK_SET)):
+        raise PolicyError(f"You are not allowed to lock {user.login}@{user.realm}.")
```
jelinaunger: do not include resolved user information in the error message. Either do not name the user at all (You are not allowed to lock this user) or only include the parameters passed in the request.

tests/test_api_conditional_access_state.py
```diff
@@ -287,6 +287,112 @@ def test_requires_admin(self):
...
+        # doing nothing - a lock that did not happen must not look like one that did.
+        set_policy("ca_state_scoped_set", scope=SCOPE.ADMIN, action=str(PolicyAction.USER_LOCK_SET),
+                   user="someoneelse")
+        try:
+            res = self._request("lock/user", method="POST",
+                                json_data={"user": "cornelius", "realm": self.realm1})
+            self.assertEqual(403, res.status_code, res.json)
+        finally:
+            delete_policy("ca_state_scoped_set")
+        self.assertEqual(0, UserLockState.query.count())
```
jelinaunger, line 445: add more tests here with different combinations causing unauthorized requests (restricted by realm, resolver; pass username or uid, ..) + also check the error message to ensure it is the expected reson this request failed.

privacyidea/api/conditional_access.py
jelinaunger: I would move this check before the user exist check. It just fills the gap the prepolicy check missed here and should behave consistent to it. So the user does not exist information is only provided if the admin would have the permission to act on it.

tests/test_lib_conditional_access_engine.py
```diff
@@ -712,6 +713,42 @@ def test_successful_login_resets_lock_counter(self):
         evaluate_conditional_access_policies(CAContext(self.user), AuthEventType.MFA_FAIL, now=now)
         self.assertTrue(is_user_locked(self.user))
 
+    def test_policy_lock_records_the_policy_cause(self):
+        # A policy lock names itself, so the Locked Users page can tell it from one an admin imposed.
+        now = utc_now()
+        self._make_policy(name="cause", counter_type=AuthEventType.MFA_FAIL)
+        self._seed_events(AuthEventType.MFA_FAIL, 3, timestamp=now - timedelta(seconds=100))
+        evaluate_conditional_access_policies(CAContext(self.user), AuthEventType.MFA_FAIL, now=now)
+        self.assertEqual("POLICY", self._state().lock_cause)
```
jelinaunger: you can use the enum here

tests/test_lib_conditional_access_engine.py
```diff
@@ -712,6 +713,42 @@ def test_successful_login_resets_lock_counter(self):
...
+        # The upsert refuses to downgrade a permanent lock to a timed one, so the row - cause included - is
+        # left exactly as the admin wrote it.
+        now = utc_now()
+        lock_user(self.user, now=now)
+        self._make_policy(name="cause3", counter_type=AuthEventType.MFA_FAIL)
+        self._seed_events(AuthEventType.MFA_FAIL, 3, timestamp=now - timedelta(seconds=100))
+        evaluate_conditional_access_policies(CAContext(self.user), AuthEventType.MFA_FAIL, now=now)
+        state = self._state()
+        self.assertIsNone(state.lock_expires_at)
+        self.assertEqual("MANUAL", state.lock_cause)
```
jelinaunger: do we log the reason why ca policy does not trip? We should also add a note in the auth log other info

privacyidea/static_new/src/app/components/logs/blocklist/blocklist-block-dialog/blocklist-block-dialog.component.html
```diff
@@ -0,0 +1,89 @@
...
+ License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+ SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<app-dialog-wrapper
+  (actionTriggered)="onAction($event)"
+  (wrapperClose)="close()"
+  [actions]="dialogActions()"
+  i18n-title
+  title="Block IP Address">
```
jelinaunger, line 24: the dialog is quite wide, set a max width here to make it more compact

privacyidea/static_new/src/app/components/logs/blocklist/blocklist-block-dialog/blocklist-block-dialog.component.html
```diff
@@ -0,0 +1,89 @@
...
+    CONDITIONAL_ACCESS_NEVER_BLOCK cannot be blocked.
+  </p>
+
+  <mat-form-field
+    appearance="fill"
+    subscriptSizing="dynamic"
+    class="block-dialog-ip">
+    <mat-label i18n>IP address</mat-label>
+    <input
+      matInput
```
jelinaunger, line 38: add standard input width

privacyidea/static_new/src/app/components/logs/blocklist/blocklist-block-dialog/blocklist-block-dialog.component.html
```diff
@@ -0,0 +1,89 @@
...
+      Until an administrator unblocks it
+    </mat-radio-button>
+    <mat-radio-button
+      value="timed"
+      i18n>
+      For a limited time
+    </mat-radio-button>
+  </mat-radio-group>
+
+  @if (mode() === "timed") {
```
jelinaunger: could be always visible, but disabled if permanent is selected, hence the layout does not jump

jelinaunger (approved):

PR #5839, opened 2026-08-27
5824/ca auth log client ip format

### LINKED ISSUE #5824: Conditional Access: Client & IP format in auth log
The client / IP format in the auth log should be changed to
* be able to store the source the client label / IP is derived from
* store multiple IPs

Is the override authorization client option enabled?
Is it the IP from the client or from the users device?
Store origin and proxy Ips e.g. from the X-Forwarded-For header

However, IPs should still be easily queryable. Think about a suitable db structure.

henrik-falk:
The authentication log stored one client address per row and nothing about where it came from. That address is the output of get_client_ip(), which - only when OverrideAuthorizationClient is set - walks [remote_addr] + reversed X-Forwarded-For (+ the client parameter) and picks the deepest hop the configured proxy paths permit. The peer, the rejected hops, and the fact that an override was configured at all were discarded, so the log could not answer "is this the address the connection came from, or one a proxy claimed?".

What is derived is unchanged; what is recorded is not. source_ip stays the effective address and the only one the engine counts and enforces on, so every existing index and count is untouched. Alongside it the row now carries the TCP peer (indexed, because "what came from this machine" is the second question a forensic query asks), which hop the effective address was taken from, the whole path that was considered, and whether the client label is a client_id the client chose or its User-Agent.

The chain is recorded even when no override is configured and the header is therefore ignored: recording what a request claimed is not trusting it, and everything past the peer stays client-supplied. It is capped, since X-Forwarded-For is attacker-controlled and unbounded. REMOTE_ADDR_UNMAPPED is deliberately distinct from REMOTE_ADDR - a mapping is configured but this peer may not use it - which is what makes a misconfigured proxy path visible instead of silent.

get_client_ip is now get_client_ip_info reduced to the address it selects, so the IP written to the log and the IP the engine blocks on cannot drift apart; check_proxy keeps its signature over an index-returning core.

NULL in any of the new columns means the row predates them. The WebUI honours that: the Source IP cell shows a provenance badge and expands to the peer and the claimed path, and renders no badge at all rather than guessing "direct".

Closes #5824


privacyidea/migrations/versions/0147d78cbace_authentication_log.py
```diff
@@ -68,6 +72,9 @@ def upgrade():
                         ['resolver', 'uid', 'realm', 'timestamp'])
         op.create_index('ix_authlog_ip_time', 'authentication_log',
                         ['source_ip', 'timestamp'])
+        # The TCP peer the request arrived from, the second pivot of a forensic query.
+        op.create_index('ix_authlog_peer_ip_time', 'authentication_log',
+                        ['peer_ip', 'timestamp'])
 
     except (OperationalError, ProgrammingError) as ex:
```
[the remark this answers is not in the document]
henrik-falk: migrations dont matter since nothing was on master yet

privacyidea/api/lib/utils.py
```diff
@@ -335,12 +341,31 @@ def log_authentication(event_type: AuthEventType | None, request: Request | None
...
+                            for index, hop in enumerate(client_ip_info.chain)]
+        else:
+            # before_request never got as far as publishing the derivation (an early AuthError, say). The
+            # address is still known; how it was reached is not, and is left NULL rather than guessed at.
+            source_ip = g.get("client_ip")
         if request is not None:
-            client_label = get_optional(request.all_data, "client_id") or (request.user_agent.string or None)
+            client_id = get_optional(request.all_data, "client_id")
+            if client_id:
+                client_label, client_label_source = client_id, str(ClientLabelSource.CLIENT_ID)
```
jelinaunger, line 436: no need for the str conversion here

privacyidea/api/lib/utils.py
```diff
@@ -335,12 +341,31 @@ def log_authentication(event_type: AuthEventType | None, request: Request | None
...
         log.debug("Not logging authentication event, because no event type is given.")
         return
-    client_label = None
-    source_ip = None
+    client_label = client_label_source = None
+    source_ip = peer_ip = source_ip_source = ip_chain = None
     if has_request_context():
-        source_ip = g.client_ip
+        client_ip_info = g.get("client_ip_info")
+        if client_ip_info is not None:
```
jelinaunger, line 418: A type check for ClientIpInfo might make sense here to avoid any attribute errors

privacyidea/api/authentication_log.py
```diff
@@ -42,8 +42,9 @@
 # Each filter parameter maps 1:1 to a get_authentication_logs_paginate keyword argument; the ca_* ones filter on the
 # entry's conditional-access outcomes rather than a column of its own row, and ca_dry_run is parsed separately because
 # it is a boolean, not a list of values.
-_FILTER_PARAMS = ["resolver", "uid", "realm", "username", "user_role", "event_type", "source_ip", "serial",
-                  "transaction_id", "attempt_id", "client_label", "ca_action_type", "ca_policy_name"]
+_FILTER_PARAMS = ["resolver", "uid", "realm", "username", "user_role", "event_type", "source_ip", "peer_ip",
+                  "source_ip_source", "serial", "transaction_id", "attempt_id", "client_label",
+                  "client_label_source", "ca_action_type", "ca_policy_name"]
```
jelinaunger: use plural if they can hold a list

privacyidea/lib/conditional_access/authentication_log.py
```diff
@@ -204,7 +222,14 @@ class PendingAuthEvent:
     username: str | None = None
     user_role: str | None = None
     source_ip: str | None = None
+    # How source_ip was derived: the TCP peer it arrived from, which hop it was taken from, and the whole path
+    # that was considered (see the AuthenticationLog model). None throughout for an event staged outside a
+    # request context, where there is no request to derive anything from.
+    peer_ip: str | None = None
+    source_ip_source: str | None = None
```
jelinaunger, line 237: use the enum as type here

privacyidea/lib/utils/__init__.py
```diff
@@ -652,52 +654,161 @@ def check_proxy(path_to_client, proxy_settings):
             max_idx = max(max_idx, current_max_idx)
 
     log.debug(f"Determined mapped client IP: {path_to_client[max_idx]!r}")
-    return path_to_client[max_idx]
+    return max_idx
+
+
+def check_proxy(path_to_client, proxy_settings):
```
jelinaunger: it is only used in tests, so it can be removed

privacyidea/lib/utils/__init__.py
```diff
@@ -583,7 +585,7 @@ def parse_proxy(proxy_settings):
     return proxy_set
 
 
-def check_proxy(path_to_client, proxy_settings):
+def check_proxy_index(path_to_client, proxy_settings):
```
jelinaunger: since you already touched it, please add type hints

privacyidea/static_new/src/app/components/logs/authentication-log/cells/source-ip-cell/source-ip-cell.html
```diff
@@ -0,0 +1,93 @@
...
+  @if (sourceIp()) {
+    <app-copyable [copyText]="sourceIp()!">
+      <span>{{ sourceIp() }}</span>
+      @if (showFilter()) {
+        <app-filter-value-button
+          class="margin-left-8"
+          label="Filter by this source IP"
+          i18n-label
+          [value]="sourceIp()!"
+          (filterValue)="onFilter('source_ip', $event)" />
```
jelinaunger, line 42: the filter icon does not appear. Only on the second IP

privacyidea/static_new/src/app/components/logs/authentication-log/cells/source-ip-cell/source-ip-cell.html
```diff
@@ -0,0 +1,93 @@
...
+            (filterValue)="onFilter('peer_ip', $event)" />
+        </app-copyable>
+      </div>
+    }
+    @if (hops().length) {
+      <span
+        class="info-key"
+        i18n>
+        Claimed path:
+      </span>
```
jelinaunger, line 82: there is large space below while no space above

tests/test_lib_authentication_log.py
```diff
@@ -271,6 +271,39 @@ def test_get_authentication_logs_filter_by_client_label(self):
...
+        # The peer is the second pivot: "what came from this machine", whatever it claimed to forward for.
+        log_authentication_event(event_type=AuthEventType.LOGIN_SUCCESS, resolver="res1", uid="u1", realm="r1",
+                                 source_ip="203.0.113.5", peer_ip="10.0.0.17")
+        log_authentication_event(event_type=AuthEventType.LOGIN_SUCCESS, resolver="res1", uid="u1", realm="r1",
+                                 source_ip="203.0.113.6", peer_ip="10.0.0.18")
+
+        results = get_authentication_logs(peer_ip="10.0.0.17")
+        self.assertEqual(1, len(results))
+        self.assertEqual("203.0.113.5", results[0].source_ip)
+        self.assertEqual(2, len(get_authentication_logs(peer_ip="10.0.0.*")))
```
jelinaunger: add an assertSetEqual to check exactly the expected entries are returned

jelinaunger: It also might make sense to add another entry not matching the wildcard, just to be sure

jelinaunger (approved):

PR #5840, opened 2026-08-27
5830/ca validate action value

### LINKED ISSUE #5830: Conditional Access: validate action_value when a lockout policy is saved
A lockout stage action whose `action_value` the engine cannot use is stored without complaint and then silently
never fires. `_validate_stages` deliberately treats the field as opaque —

> `action_value` may be any JSON-serializable value (its action-specific interpretation happens in the engine)
> — `privacyidea/lib/conditional_access/lockout_policy.py:532`

— so nothing checks, at save time, that the action can actually do what it says. The admin sees a saved policy with a
`LOCK_USER` action on it; no user is ever locked.

## Proposal

Validate `action_value` per `action_type` in `_validate_stages`
(`privacyidea/lib/conditional_access/lockout_policy.py:527`), which is on both the create and the update path
(`:815`, `:892`), and raise `ParameterError` like the rest of the validation there. Fail closed, as the counter-type
and condition validation already does: a policy that cannot act should not be storable.

## Also worth fixing here

- `lockout_policy.py:46` — the module docstring example writes `{"lock_duration_seconds": 600}`, which the engine
  does not read. It is the most likely source of the misconfiguration.
- The same wrong key appears in frontend fixtures:
  `conditional-access-actions-list.component.spec.ts:36`, `conditional-access-stage-item.component.spec.ts:40`,
  `conditional-access-policy.service.spec.ts:54`.
- The WebUI could mark a restricting action with no duration as invalid before the save round-trip, so the error
  arrives where the field is.

henrik-falk:
A stage action whose action_value the engine cannot use was stored without complaint and then silently never fired: _validate_stages treated the field as opaque, so a LOCK_USER with no duration - or with the lock_duration_seconds key the module docstring wrongly advertised - saved fine and locked nobody, leaving only a warning in the log.

The per-action contract is now a dispatch table validated on both the create and the update path, fail-closed like the counter-type and condition validation: LOCK_USER/BLOCK_IP need a positive duration, EMAIL_ADMIN/EMAIL_USER a non-empty subject and body, and the PERMANENT_*/ALLOW/DENY actions no value at all. The duration check is the engine's own parser (_lock_duration_seconds, now the public parse_lock_duration_seconds), so storable and usable cannot drift apart.

smtp_identifier stays optional: the SMTP server is a separate configuration object that may not exist yet, which is why the shipped MFA_BRUTEFORCE template ships it blank.

Validation is on the write path only, so a policy stored before this rule can still be enabled, disabled and renamed; re-sending its stages is the repair path.

Also fixes the docstring key that caused the misconfiguration, the same wrong key in the backend and frontend fixtures, the "email template ID" claim in the model docstring, and renders lockout_policy, conditions and lockout_policy_template in the module docs. The WebUI now reports the missing value on the action itself and blocks Save, so the error arrives where the field is instead of as a 400.

Closes #5830


privacyidea/static_new/src/app/components/conditional-access/conditional-access-edit-page/conditional-access-edit-page.component.html
```diff
@@ -317,6 +317,14 @@ <h4 i18n>Stages &amp; Actions</h4>
         Some actions are not allowed for the selected target. Change the target or remove those actions.
       </p>
     }
+    @if (!actionValuesValid()) {
+      <p
+        class="ca-stages-error"
+        i18n>
+        Some actions are missing a value they need to run: a lock or block needs a duration, an email needs a
+        subject and body. Open the action to fix it.
+      </p>
```
jelinaunger, line 341: I would expect this to be in the error summary

tests/test_lib_conditional_access_policy.py
```diff
@@ -1023,3 +1189,36 @@ def test_50_update_replaces_and_clears_the_stage_error_message(self):
...
+
+    def test_52_stage_action_docstring_does_not_claim_allow_is_a_storable_action(self):
+        # ALLOW is only ever a value of the separate AccessDecision enum (the pre-auth evaluation
+        # outcome); it is not, and never was, a member of ConditionalAccessAction, so a stage's
+        # action list can never actually carry one. The docstring must not imply otherwise.
+        import inspect
+
+        self.assertNotIn("ALLOW", ConditionalAccessAction.__members__)
+        doc = inspect.getdoc(ConditionalAccessStageAction) or ""
+        self.assertNotIn("``ALLOW``", doc)
```
jelinaunger: while these test might help to find any leftovers after the rename / action dropped. But since they are now green, all these shouldn't come back and the tests are pointless.

jelinaunger (approved):

PR #5842, opened 2026-08-27
Add API client & remembered-device management to the new WebUI

### LINKED ISSUE #5705: Unify the ecosystem-integration vocabulary (client types, policy user_agents, subscriptions) behind one backend catalog
## Summary

The vocabulary of "privacyIDEA ecosystem integrations" (Credential Provider,
Keycloak, ADFS, Shibboleth, Nextcloud, PAM, FreeRADIUS, …) is currently expressed
in **at least three independent, hardcoded places** that have already drifted out
of sync. They should be unified behind a single backend source of truth, consumed
by every place that needs it. This was scoped out of the `clients-api-keys` branch
because the subscription/licensing angle needs its own careful design.

## The three places today

1. **Client types** (API-clients feature) — hardcoded in
   `static/components/config/controllers/clientsController.js` (`$scope.clientTypes`).
   Used only as a descriptive `client_type` tag; the backend stores it free-form.
2. **Policy `user_agents`** — hardcoded in
   `static/components/config/controllers/configControllers.js`
   (`$scope.userAgentsMapping`). Used as match values for the `user_agents`
   policy attribute.
3. **Subscriptions** — `privacyidea/lib/subscriptions.py` `APPLICATIONS`
   (dict of `application → free-token limit`). Used by `CheckSubscription` for
   licensing enforcement.

All three key off the **same identifier**: the "agent" token parsed from the
`User-Agent` header by `get_plugin_info_from_useragent` (the first token before
`/`). So they are conceptually one list.

## Why this is worth doing — they have already drifted

Comparing the policy `userAgentsMapping` values against `subscriptions.APPLICATIONS`:

| integration | policy `user_agents` value | `APPLICATIONS` key | |
|---|---|---|---|
| CP | `privacyidea-cp` | `privacyidea-cp` | ✓ |
| Keycloak | `privacyIDEA-Keycloak` | `privacyidea-keycloak` | ✓ (case) |
| ADFS | `PrivacyIDEA-ADFS` | `privacyidea-adfs` | ✓ |
| Shibboleth | `privacyIDEA-Shibboleth` | `privacyidea-shibboleth` | ✓ |
| Nextcloud | `privacyidea-nextcloud` | `privacyidea-nextcloud` | ✓ |
| LDAP Proxy | `privacyIDEA-LDAP-Proxy` | `privacyidea-ldap-proxy` | ✓ |
| SimpleSAMLphp | `simpleSAMLphp` | `simplesamlphp` **and** `privacyidea-simplesamlphp` | ✓ (two keys!) |
| **PAM** | `PAM` | `privacyidea-pam` | ✗ **mismatch** |
| **Authenticator** | `privacyIDEA-App` | `privacyidea authenticator` | ✗ **mismatch** |
| FreeRADIUS | `FreeRADIUS` | — | policy-only |
| WebUI | `privacyIDEA-WebUI` | — | policy-only |

Plus `APPLICATIONS` has `demo_application`, `owncloud`, `privacyidea` (generic)
that appear in neither UI list.

**Consequence:** for PAM and the Authenticator, the policy dropdown offers one
string while subscriptions expect another — so at least one of them cannot be
matching at runtime. This is a pre-existing bug the unification must resolve, and
it needs **ground truth**: what agent token does each plugin/app actually send?

## Data-model requirement: many user-agents → one subscription

Subscriptions already need **more than one user-agent to count toward a single
subscription limit** (the `simplesamlphp` + `privacyidea-simplesamlphp` pair is an
existing example). So the model cannot be 1 integration ↔ 1 user-agent. It must be
product-centric with a **list of user-agent aliases**, e.g.:

```python
Integration(
    id="simplesamlphp",
    label="SimpleSAMLphp",
    user_agents=["simpleSAMLphp", "privacyIDEA-simpleSAMLphp"],  # all count together
    api_client=False,
)
```

## Proposed design

- A single backend catalog (e.g. `privacyidea/lib/integrations.py`), product-centric:
  `{id, label, user_agents: [...], api_client: bool}`.
- One endpoint (e.g. `GET /info/integrations`, auth-only) returning the catalog.
- Consumers filter their facet:
  - clients dropdown → entries where `api_client` → `{id→client_type, label}`;
  - policy `user_agents` picker → flatten `user_agents` into per-token options
    (keep the custom free-text escape hatch);
  - `subscriptions.py` → resolve a request's parsed UA to the product that lists
    it, and apply the product's limit, so all aliases share one bucket.
- **Keep the licensing facet (free-token limit + product→subscription grouping) in
  `subscriptions.py`, referencing the catalog** — don't smear licensing logic into
  the generic UI endpoint.

## Open questions

- **Ground truth:** the real agent token each plugin/app sends (to resolve the
  PAM / Authenticator mismatches and confirm every alias).
- **Grouping cardinality:** is it strictly N user-agents → 1 product → 1
  subscription, or can multiple *products* share one subscription (M→1)? The
  alias-list handles N→1; M→1 needs an explicit subscription-group id.
- **Where the limit lives** (catalog facet vs subscriptions-owned mapping).
- **Restricted visibility:** should the client-type / user-agent choices be
  filterable per admin via policy? (The endpoint would be the natural filter point.)
- Migration/compat for existing stored policy `user_agents` values and existing
  `Subscription` rows.

## Notes

- Licensing-critical: a wrong merge silently breaks subscription enforcement — do
  not change `APPLICATIONS` keys without verifying they match the runtime-parsed
  agent token.
- Ties into the plugin/subscription dashboard work.
- The `clients-api-keys` branch keeps its client-type list hardcoded in the
  frontend for now (with a `TODO` pointing here).

### LINKED ISSUE #5820: API keys and remembered devices UI
Add the UI

### LINKED ISSUE #5860: Pagination parameters are not validated
The `page` and `pagesize` query parameters are read with a bare `int()` and passed on unchecked:

- `privacyidea/api/token.py:463,466` (`GET /token/challenges/`)
- `privacyidea/api/token.py:701,709` (`GET /token/`)
- `privacyidea/api/container.py:207,1404`
- `privacyidea/api/clients.py:266,267` (`GET /client/<client_id>/remembered_devices`)

Three problems follow from that:

**Non-integer values give a 500.** `int("abc")` raises `ValueError`, and no error handler covers it (`api/before_after.py` registers handlers only for `AuthError`, `PolicyError`, `PrivacyIDEAError`, `ResourceNotFoundError` and `NotImplementedError`), so the request fails with an unhandled exception instead of a parameter error. `GET /token/` is reachable with role `user`, so this is not admin-only.

**Zero/negative values are dialect-dependent.** The paginating helpers compute `offset = (page - 1) * page_size` and then `.limit(page_size).offset(offset)` (`lib/token/query.py:824`, `lib/remembered_device.py:530`). SQLite clamps a negative OFFSET to 0 and answers 200; PostgreSQL rejects it:

```
SELECT 1 LIMIT 15 OFFSET -15;  ->  ERROR:  OFFSET must not be negative
SELECT 1 LIMIT -5 OFFSET 0;    ->  ERROR:  LIMIT must not be negative
```

So `GET /token/?page=0` answers 200 on SQLite and 500 on PostgreSQL. The test suite runs on SQLite by default and cannot see the difference.

**`pagesize` is uncapped.** An arbitrarily large page size defeats the reason these listings are paginated in the first place.

### Suggested fix

A shared helper that coerces both parameters, raises `ParameterError` on a non-integer, floors `page` and `pagesize` at 1 and caps `pagesize` at a supported upper bound; applied to all call sites above so the behaviour is uniform.

nilsbehlen:
## Summary
- Ports the API-clients feature (X-API-Key auth clients + rotating "remember this device" sessions, PR #5694) into the new Angular WebUI, under Policies: list/create/rotate/delete clients, and a per-client remembered-devices section with revoke by device/client/realm/user.
- Unifies the ecosystem-integration vocabulary (client types, policy `user_agents` picker, subscription dashboard/metering) behind a single backend catalog (`privacyidea/lib/integrations.py`) served via a new admin-only `GET /info/integrations`, replacing five independently-hardcoded lists across both WebUIs and the backend.
- The catalog is 3-level (agent name -> Integration -> Product): multiple integrations with distinct agent names/policy values can share one product, so subscription grouping across integrations (e.g. the two PAM binaries `pam`/`pam-passkey`, or Keycloak's `privacyidea-keycloak`/`entraid-via-keycloak`) is already handled via `Integration.product_id`.
- Merges `simplesamlphp` and `privacyidea-simplesamlphp` into a single product; they were two identically-configured products with no evidence they were ever meant to be distinct.
- Drops the Authenticator App's login-time push-token subscription nag (its own requests were already excluded from request-based metering); it keeps a real, subscribable product so the dashboard still reports its subscription state, it's just never counted or blocked.
- `client_type` on API clients is now validated against the catalog instead of accepted as free text.
- Paginates the remembered-device listing (`GET /clients/<id>/remembered_devices`) and adds an `expires_at` index plus a composite `(client_id, resolver, user_id, realm_id)` index, so a client with a very large number of remembered devices (e.g. many browsers/devices per user) doesn't load/serialise its entire device history in one request or force a full table scan on cleanup.

Closes #5820
Refs #5705 (this covers client types, the policy picker, the dashboard, and both WebUIs; per-admin restricted visibility of the catalog is deliberately left open for a follow-up)
Fixes #5860 (pagination parameters were unvalidated across the existing paginated endpoints too; fixed here rather than in a separate PR)


privacyidea/lib/integrations.py
```diff
@@ -0,0 +1,368 @@
+#  2026-08-27 Nils Behlen <nils.behlen@netknights.it>
+#             Unify the client/integration vocabulary
```
jelinaunger: adding the history here is obsolete in the age of version control.
Also the license header is missing the SPDX stuff

privacyidea/lib/integrations.py
```diff
@@ -0,0 +1,368 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+#
+# You should have received a copy of the GNU Affero General Public
+# License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+__doc__ = """Single source of truth for the privacyIDEA ecosystem's integration vocabulary.
+
+Before this module existed, the same "ecosystem integration" concept was expressed in at
+least three independent, hardcoded lists that had already drifted out of sync: the
```
jelinaunger: no need to document the old bad design

privacyidea/api/info.py
```diff
@@ -73,3 +76,24 @@ def rss():
...
+    This is administrative reference data: an integration's ``policy_value`` is what a
+    policy's ``user_agents`` condition stores, its ``id`` is what an API client's
+    ``client_type`` stores, and ``dashboard``/``section``/``ptl_slug``/``github_repo``
+    describe its row on the dashboard subscription overview. Self-service users get an
+    empty list, since none of those features are available to them.
+
+    :status 200: JSON response; the integration list is in ``result.value``.
+    """
+    if g.logged_in_user.get("role") != "admin":
+        return send_result([])
```
jelinaunger: this could be replaced by the admin_required decorator, but it would raise an error instead

privacyidea/lib/params.py
```diff
@@ -116,6 +121,57 @@ def get_optional_one_of(param, keys, default=None):
     return default
 
 
+def get_optional_int(param, key, default=None):
```
jelinaunger: type hints would be great

privacyidea/migrations/versions/6296c0889f75_remembered_devices_expires_at_index.py
```diff
@@ -0,0 +1,38 @@
+"""v3.14: Add an index on remembered_devices.expires_at
```
jelinaunger: There are three migration scripts, each adding an index to the remember device table. They could be folded into one to keep the migration history clean. They could be even added to the existing migration script adding the table, as it is not yet released.

privacyidea/api/info.py
nilsbehlen: did that too

jelinaunger (approved):

PR #5850, opened 2026-08-31
Add multi_device passkey authentication test coverage

nilsbehlen:
## Summary
- The existing `AllowedAuthenticatorDeviceTypes` authentication tests (`test_17`) only ever replayed the single_device fixture and varied the requested policy value, so there was no test evidence that a genuine backup-eligible (multi_device) assertion is correctly recognized and allowed.
- Adds a second, independently generated credential (a locally simulated ECDSA P-256 authenticator, verified against the real `py_webauthn` verification functions at fixture-generation time) to cover the multi_device allow/reject paths, plus a negative test proving the backup-eligible flag - which lives inside the signed region of `authenticatorData` - cannot be spoofed by editing the wire data post-signature.

## Test plan
- [x] `tests/test_lib_tokens_passkeytoken.py` full file run: 21/21 passed
- [x] `tests/test_api_passkey.py` full file run: 29/29 passed
- [x] `ruff check` clean on both changed files (pre-existing unrelated findings only)


tests/test_lib_tokens_passkeytoken.py
```diff
@@ -659,4 +690,50 @@ def test_18_authenticate_enforce_user_handle(self):
...
+        credential too, not just compared against whatever policy value happens to be passed in.
+        """
+        token = self._create_token_multi_device()
+
+        challenge = self._initialize_authentication_multi_device(self.authentication_challenge_multi_device_reject)
+        authentication_response = dict(self.authentication_response_multi_device_reject)
+        authentication_response["HTTP_ORIGIN"] = self.expected_origin
+        authentication_response[PasskeyAction.AllowedAuthenticatorDeviceTypes] = ["single_device"]
+        verification_result = verify_fido2_challenge(challenge["transaction_id"], token, authentication_response)
+        self.assertEqual(-1, verification_result.success)
```
jelinaunger, line 718: Also, assert the corresponding log entry here to ensure that the reason for the failed verification is the expected one.

jelinaunger (approved):

tests/test_lib_tokens_passkeytoken.py
nilsbehlen: i guess in this case its a "good" idea, since the function returns just `-1`, will do other tests too where applicable, since we are bound to the packages error messages then anyway

PR #5852, opened 2026-08-31
Improve audit writing

nilsbehlen:
Request data of arbitrary length ends up in the audit log. A value that did not fit into its column made the database reject the whole entry, so the request was not logged at all - only an error appeared in the log file. This could be triggered by sending over-long data, e.g. a long user name or user agent.

Audit values are now always shortened to what the store can hold. The former setting PI_AUDIT_SQL_TRUNCATE, which made this optional, is removed. The length of a column is read from the database and caps the configured length, so entries also fit a table whose columns have not been migrated yet.

The audit serial column is widened from 40 to 200 characters: an entry names every token that was involved in a request, and three tokens with default serials already filled the old column.

Further robustness fixes along the same path:

* /validate/check is logged before the user is resolved, so a request naming a realm or resolver that does not exist appears in the audit log as well.
* A response that matches none of several challenges now names the tokens that were challenged, restricted to the tokens of the requesting user.
* A serial or token type logged by a handler is no longer overwritten by an empty value of the response.
* get_client_ip follows the forwarding chain only as far as it consists of IP addresses instead of raising on the rest.
* truncate_comma_list terminates when the entries can not be shortened any further and cuts the list instead.
* Saving the client application no longer lets a database problem or an over-long user agent fail the request it was collected from.


doc/audit/index.rst
```diff
@@ -170,29 +170,27 @@ This will read the configuration (only the database URI) from the config file
...
+The entries to be written to the database may be longer than the column in the
+database, since they contain data from the request. Such an entry is shortened
+to the length of the column, so that it is still written: a database rejects a
+value that is too long, which would lose the whole entry. Values that hold a
+list of items, like the serials of all tokens of a request, are shortened per
+item, so that every item stays recognizable.
+
+The length of a column is read from the database, so shortening also matches a
+table whose columns differ from the ones privacyIDEA ships. If you increase a
+column length by the usual database means
```
jelinaunger: This is confusing; it first states it is read from the db hence I would expect it works out of the box, but then you say the user needs to define the length manually.

jelinaunger: okay got it now, but maybe it can be clarified a bit

READ_BEFORE_UPDATE.md
```diff
@@ -7,6 +7,27 @@
...
+  upgrade, so that the audit log can hold the serials of all tokens of a user. It is not required for privacyIDEA to
+  work: entry values are shortened to the length the audit table really has, so a database that still has the
+  40-character column keeps logging shortened serial lists instead of losing entries.
+
+* **Audit entries are always shortened to fit their column** — Values that are longer than their audit column, e.g. a
+  long user name or user agent sent by a client, are shortened so that the entry is written. Previously this required
+  `PI_AUDIT_SQL_TRUNCATE = True`; without it, the database rejected the entry and it was lost. **The setting is now
+  ignored** and shortening is always applied, which also closes the possibility to suppress one's own audit entries by
+  sending over-long request data. `PI_AUDIT_SQL_COLUMN_LENGTH` still works and is only needed if you widened columns
+  beyond what privacyIDEA ships.
```
jelinaunger, line 46: Worth noting: The migration might take a while for  mysql, mariadb and sqlite. The do an entire rebuild of the audit table as an in-place extension of the column is not possible.

privacyidea/migrations/versions/e0f1a2b3c4d5_audit_serial_column_size.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def upgrade():
+    try:
+        with op.batch_alter_table('pidea_audit', schema=None) as batch_op:
+            batch_op.alter_column('serial',
+                                  existing_type=sa.VARCHAR(length=40),
+                                  type_=sa.Unicode(length=200),
+                                  existing_nullable=True)
+    except (OperationalError, ProgrammingError) as exx:
```
jelinaunger: also catch DatabaseError here for oracle failures

jelinaunger (approved):

READ_BEFORE_UPDATE.md
nilsbehlen: ive added logging, so the person updaing knows whats happening, and also added some rough numbers for how long it might take to read_before_update

PR #5858, opened 2026-09-01
Remove removed actions from the policy templates

### LINKED ISSUE #5857: Fix inital supersuper policy template
still contains deprecated u2f policies and cannot be saved before its removed

nilsbehlen:
The `superuser` and `helpdesk` policy templates still contained four admin actions that no longer exist, so a policy created from either template could not be saved until they were removed by hand:

| Action | Status |
| --- | --- |
| `enrollU2F` | the U2F token type was removed |
| `authitems` | superseded by `fetch_authentication_items` |
| `machinetokens` | superseded by `manage_machine_tokens` |
| `mangle_machine_tokens` | removed |

The surviving equivalents (`fetch_authentication_items`, `machinelist`, `manage_machine_tokens`) were already part of both templates, so nothing is lost by dropping the dead keys. The other five templates were checked and are clean.

The templates are shipped twice, once for the old WebUI (`privacyidea/static/policy-templates`) and once for the new one (`privacyidea/static_new/public/policy-templates`). Both copies are fixed.

`PolicyTemplateTestCase` in `tests/test_lib_policy.py` guards this from now on. For every shipped template it checks that

- each action still exists in the definitions of the template scope (`validate_actions`),
- each value matches the `type` of the action definition and its list of allowed values, where the list is not built from the database,
- the template is listed in `index.json`,
- both WebUI copies are identical.

Value lists that are read from the database (resolvers, realms, SMTP and RADIUS servers) are empty on a fresh instance and are skipped.

Fixes #5857


nilsbehlen:
Superseded by #5859 — same commit, the branch was renamed to carry the issue number.

PR #5864, opened 2026-09-01
Allow push tokens to use push-capable SMS gateways

### LINKED ISSUE #5865: [Feature request] Support push delivery through explicitly push-capable SMS providers
## Problem

The push token implementation currently limits the configured delivery gateway to `FirebaseProvider` by filtering SMS gateway definitions with a hard-coded provider module.

This makes it difficult to use privacyIDEA in environments where direct Firebase Cloud Messaging access is not permitted or where push traffic must pass through an internal service. The SMS-provider abstraction already offers configurable HTTP and script providers, but push tokens cannot select them because the enrollment policy only exposes Firebase gateway definitions.

## Use case

An organization operates an internal push gateway that:

1. accepts an HTTPS POST request from privacyIDEA;
2. receives the mobile device registration token and the signed privacyIDEA push challenge;
3. applies organization-specific authentication, auditing, routing, proxy, or network controls;
4. performs the final Firebase Cloud Messaging request.

A script-based adapter is also useful for local integrations, prototypes, and deployments where the delivery mechanism is managed outside the privacyIDEA process.

## Requested behavior

Push tokens should be able to use SMS-provider implementations that explicitly declare support for structured push messages.

The selection must be opt-in rather than allowing every SMS provider automatically. Text-only providers such as SMTP, SMPP, or Sipgate should not be offered for push enrollment unless they implement the structured-message contract themselves.

For the existing generic providers:

- `HttpSMSProvider` should be able to send the device token and structured push payload as configurable JSON fields in a POST request;
- `ScriptSMSProvider` should pass the device token as an argument and a JSON-serialized push payload through standard input;
- `FirebaseProvider` should continue to work without configuration migration.

## Compatibility expectations

- Existing Firebase push configurations must remain valid.
- Existing text-message behavior of HTTP and script SMS providers must remain unchanged.
- Persistent policy and token-info names such as `firebase_config` and `firebase_token` should remain compatible unless a separate migration is introduced.
- Only providers that explicitly advertise push support should appear in the push gateway selection.
- The selected provider capability should be validated both during enrollment and before message delivery.

## Acceptance criteria

- Push-token enrollment lists all configured, explicitly push-capable SMS gateways.
- An HTTP gateway can receive a request containing a device token and the original structured push challenge.
- A script gateway receives the device token and a JSON push challenge using its documented process interface.
- Incompatible SMS providers are excluded from selection and rejected if an existing gateway definition is changed to an incompatible provider.
- Existing Firebase and text-SMS tests continue to pass.
- Tests cover capability filtering and an end-to-end push-token-to-HTTP-gateway flow.

## Related implementation

A proposed implementation is available in PR [#5864](https://github.com/privacyidea/privacyidea/pull/5864).

## AI assistance disclosure

This feature-request description was prepared with assistance from OpenAI Codex and reviewed by the submitter.

gitalexch:
## Purpose

Push tokens currently restrict their gateway configuration to `FirebaseProvider` by filtering SMS gateways with a hard-coded provider class. This prevents installations from routing privacyIDEA push challenges through an internal HTTP service or a locally managed script, even though the push token already invokes the selected gateway through the generic SMS-provider interface.

This pull request makes push delivery provider-independent while keeping compatibility with existing Firebase-based installations. It is intended for deployments where privacyIDEA sends the signed smartphone challenge to an internal gateway, and that gateway performs the final Firebase Cloud Messaging delivery.

### Explicit provider capability

`ISMSProvider` now defines:

```python
supports_push_messages = False
```

Push support is opt-in. `FirebaseProvider`, `HttpSMSProvider`, and `ScriptSMSProvider` set the capability to `True`. Other providers, such as SMTP, SMPP, or Sipgate, remain unavailable to push tokens unless they explicitly implement the structured-message contract and opt in.

This avoids presenting every SMS provider as push-compatible merely because it implements `submit_message`.

### Generic push-gateway selection

`pushtoken.py` no longer filters gateways using the hard-coded Firebase provider module. It loads each configured provider class and selects only providers that declare `supports_push_messages`.

The capability is checked in three places:

1. when producing the list of gateways available to the enrollment policy;
2. when validating the selected gateway during enrollment;
3. immediately before sending a push message, which also protects already enrolled tokens if a gateway definition is later changed to an incompatible provider.

Provider-independent names such as `push_gateway_identifier`, `push_gateway`, `device_token`, and `push_payload` are used in the delivery path.

### HTTP provider contract

`HttpSMSProvider` accepts structured push payloads through the existing additional-option template mechanism:

- `{phone}` resolves to the device registration token;
- `{message}` resolves to the original message object;
- `{otp}` remains a backwards-compatible alias;
- when a placeholder is the complete option value, the original type is preserved;
- embedded structured messages are JSON-serialized before template substitution.

For example, a gateway can be configured with:

```text
HTTP_METHOD = POST
SEND_DATA_AS_JSON = yes
device_token = {phone}
push_payload = {message}
```

The resulting request body is equivalent to:

```json
{
  "device_token": "<registration token>",
  "push_payload": {
    "nonce": "...",
    "question": "...",
    "serial": "...",
    "signature": "..."
  }
}
```

The field names remain configurable, so privacyIDEA does not impose the API schema of one particular internal gateway.

### Script provider contract

`ScriptSMSProvider` keeps the existing process interface:

- the device token is passed as the first command-line argument;
- text messages are passed unchanged through standard input;
- dictionary push payloads are serialized as JSON and passed through standard input;
- the existing wait/background and exit-code behavior is unchanged.

### Backwards compatibility

The persistent and public names `PushAction.FIREBASE_CONFIG` and `firebase_token` are intentionally retained. Renaming them would require policy and token-info migration and is outside the scope of this change.

Existing Firebase gateway behavior remains available and now uses the same capability-based path as HTTP and script gateways. Existing text-message behavior is covered by the SMS-provider regression tests.

## Tests

The focused test run completed successfully on Python 3.13:

```text
61 passed, 3 dependency deprecation warnings
```

It covers:

- capability defaults and opt-in providers;
- exclusion of a non-push SMTP gateway;
- JSON serialization for `ScriptSMSProvider`;
- structured JSON requests for `HttpSMSProvider`;
- an end-to-end `PushTokenClass -> HttpSMSProvider -> JSON body` challenge flow;
- the existing Firebase failure and polling behavior;
- existing SMS-provider regression tests.

A broader local run of the three affected test modules reached 84 passing tests and 9 failures. Those failures were existing Firebase tests where the HTTP mock was not active in this local environment and the tests attempted a real FCM request; their tracebacks ended in TLS verification rather than in the changed provider code. No network access is required by the focused tests listed above.

## AI assistance disclosure

This change was designed and implemented with assistance from OpenAI Codex. The AI was used to inspect the existing provider architecture, propose the capability-based design, prepare the implementation and tests, and draft this pull-request description. The contributor reviewed the design and resulting diff, requested clearer domain-specific naming, ran the tests described above, and remains responsible for the submitted changes.


nilsbehlen:
hi, thanks for your contribution. we will get to this shortly.

gitalexch:
> hi, thanks for your contribution. we will get to this shortly.

Have to update one of migration, the migration test failed due to hardcode, now should pass successfully.

privacyidea/lib/smsprovider/HttpSMSProvider.py
```diff
@@ -54,10 +54,31 @@
...
 class HttpSMSProvider(ISMSProvider):
 
+    supports_push_messages = True
+
+    @staticmethod
+    def _render_option_value(value, phone, message):
+        if value == "{phone}":
+            return phone
+        if value in ["{message}", "{otp}"]:
+            return message
```
nilsbehlen, line 64: Returning the raw dict here is only safe when the gateway has `SEND_DATA_AS_JSON=yes`. The default is `"no"`, and nothing in the code enforces the combination - the documentation recommends it, but a gateway that misses it fails silently and badly.

With `SEND_DATA_AS_JSON` unset, the dict reaches `data=parameter` and `requests` encodes it by iterating its *keys*:

```
BODY: push_payload=nonce&push_payload=url&device_token=t
```

Field names only - no nonce value, no signature, no payload. The gateway answers 200, `_check_success` returns `True`, `res=True`, the challenge is created, and the user is told to confirm on their phone. The same happens on GET via `params`.

Could `submit_message` reject (or JSON-serialize) a non-`str` message when `json_data` is False? A `check_configuration` that requires POST plus `SEND_DATA_AS_JSON=yes` when `ALLOW_PUSH=yes` would also catch this at configuration time rather than at the first authentication.

privacyidea/lib/smsprovider/HttpSMSProvider.py
```diff
@@ -54,10 +54,31 @@
...
+
+        serialized_message = message if isinstance(message, str) else json.dumps(message)
+        value = value.replace("{message}", serialized_message)
+        value = value.replace("{otp}", serialized_message)
```
nilsbehlen, line 68: Two changes here affect existing SMS gateways, not just push ones:

1. The early returns change the *type* of exact-match placeholders. Previously an option whose value was exactly `{otp}` went through `v.replace(...)` and then `json.loads("123456")`, arriving as the integer `123456`; it now returns the string. With `SEND_DATA_AS_JSON=yes` an existing gateway starts receiving `{"text": "123456"}` where it used to get `{"text": 123456}`, which a strictly typed remote API may reject. The same applies to `{phone}`.
2. `value.replace("{message}", ...)` now applies to every additional option of every HTTP gateway. A gateway that deliberately passes a literal `{message}` through, for the remote provider to expand, will have it consumed locally instead.

If both are intended, they need an entry in the Changelog as an upgrade note - it currently mentions only the PUSH additions.

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -1292,26 +1305,29 @@ def create_challenge(self, transactionid: str = None, options: dict = None) -> t
...
-                                                             challenge, registration_url,
-                                                             private_key_pem, options, current_presence_options)
-                    log.debug(f"Sending to firebase the smartphone_data: {smartphone_data}")
-                    res = fb_gateway.submit_message(self.get_tokeninfo("firebase_token"), smartphone_data)
+                    push_payload = _build_smartphone_data(self,
+                                                          challenge, registration_url,
+                                                          private_key_pem, options, current_presence_options)
+                    log.debug(f"Sending push payload: {push_payload}")
+                    device_token = self.get_tokeninfo("firebase_token")
+                    res = push_gateway.submit_message(device_token, push_payload)
```
nilsbehlen, line 1327: This breaks the existing failure fallback for the two newly supported gateway types.

`FirebaseProvider.submit_message` returns `False` on a non-200, which feeds the `if not res:` branch below: adjusted message, `ERROR_CHALLENGE_TEXT`, and the challenge is still created so the app can poll. `HttpSMSProvider` never returns `False` - tracing `_check_success`, every path either returns `True` or raises `SMSError` (non-2xx, `RETURN_SUCCESS` mismatch, `RETURN_FAIL` match). `ScriptSMSProvider` raises in `wait` mode as well. `SMSError` is not a `PrivacyIDEAError`, and there is no `try`/`except` here or at the `create_challenge` call site in `lib/token/auth.py`, so a gateway returning 502 turns the whole `/validate/check` into a server error with no challenge row - the fallback branch is unreachable rather than degraded, and the other tokens in the loop lose their challenges too.

`smstoken.py` wraps `submit_message` for exactly this reason. Mapping `SMSError` to `res = False` here restores the intended behavior, and is also the hook that the `result="error"` metrics case needs (see my reply on the metrics thread).

privacyidea/lib/smsprovider/ScriptSMSProvider.py
```diff
@@ -78,7 +82,7 @@ def submit_message(self, phone, message):
             # Trusted input/no user input: The scripts are created by user root and read from hard disk
             p = subprocess.Popen(proc_args, cwd=self.script_directory,   # nosec B603
                                  universal_newlines=True, stdin=subprocess.PIPE)
-            p.communicate(message)
+            p.communicate(serialized_message)
```
nilsbehlen, line 85: In `background` mode this reports success unconditionally: `p.wait()` is never called so `rcode` stays 0, and the `except` only re-raises when `background == SCRIPT_WAIT` - a missing script, a non-executable file or an `OSError` are swallowed with a log warning and the method returns `True`.

For SMS that was tolerable. For push it means `res=True`, the challenge is created, `hideResponseInput` is set, and the user waits for a notification that was never sent, with nothing surfaced anywhere. Since the delivery signal cannot exist in this configuration, I would rather `ALLOW_PUSH=yes` be rejected, or at least warned about, for a script gateway running in `background` mode.

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -1292,26 +1305,29 @@ def create_challenge(self, transactionid: str = None, options: dict = None) -> t
...
+        push_gateway_identifier = self.get_tokeninfo(PushAction.FIREBASE_CONFIG)
+        if push_gateway_identifier:
             challenge = b32encode_and_unicode(geturandom())
             if options.get("session") != ChallengeSession.ENROLLMENT:
-                if fb_identifier != POLL_ONLY:
-                    # We only push to Firebase if this token is NOT POLL_ONLY.
-                    fb_gateway = create_sms_instance(fb_identifier)
+                if push_gateway_identifier != POLL_ONLY:
+                    push_gateway = create_sms_instance(push_gateway_identifier)
+                    if not push_gateway.allows_push_messages(push_gateway.smsgateway):
```
nilsbehlen, line 1316: Worth a conscious decision: setting `ALLOW_PUSH=no` on a gateway that existing tokens reference makes every subsequent `create_challenge` raise, so one changed option locks out all users of that gateway with an error response instead of downgrading them to poll mode - which is how every other delivery problem is handled a few lines below. `test_02c` suggests this is deliberate; if so that is fine, but a log warning plus `res = False` would be more consistent with the surrounding code.

doc/configuration/sms_gateway_config.rst
```diff
@@ -110,6 +115,25 @@ parse the values as JSON and either send JSON or strings to the HTTP gateway.
...
+* ``device_token``: ``{phone}``
+* ``push_payload``: ``{message}``
+
+For PUSH messages, ``{phone}`` contains the device token rather than a phone number.
+If an option consists only of ``{message}``, the complete structured PUSH payload is
+preserved as a JSON object. When embedded in another value, the payload is serialized
+as JSON text. The existing ``{otp}`` placeholder remains an alias for ``{message}``
+for compatibility; prefer ``{message}`` in new PUSH gateway configurations.
+
+The gateway must forward the signed payload without modifying its fields, otherwise
```
nilsbehlen, line 134: Could you add a note about what a successful send means here? With an HTTP gateway a 2xx says that the gateway accepted the request, not that the notification reached the device - the Firebase leg happens after privacyIDEA is out of the loop. So the "could not be sent, please poll" fallback only fires when the gateway itself reports a failure, and an unregistered or stale device token, which Firebase reports as a non-200 today, becomes invisible. That is an acceptable consequence of proxying, but administrators should know it before they rely on the message.

Suggested addition:

```rst
.. note:: With Firebase, privacyIDEA learns from the Firebase response whether a
   notification was accepted for the device, and adjusts the message shown to the
   user when it was not. An HTTP gateway can only report that it accepted the
   request; the delivery to the push service happens afterwards and is not visible
   to privacyIDEA. Configure ``RETURN_SUCCESS`` or ``RETURN_FAIL`` so the gateway
   can report the final delivery result if you want this feedback. A Script gateway
   running in ``background`` mode reports no result at all.
```

nilsbehlen (commented):
Thanks for this - the motivation is sound and the design is the right shape. `FirebaseProvider` was already registered in `SMS_PROVIDERS` and its `submit_message` already took a dict rather than a string, so `supports_push_messages` formalizes polymorphism that existed in practice instead of adding a new abstraction. Opt-in per gateway, the class-level capability so text-only providers cannot be selected, and checking the capability at all three points are all correct choices, and keeping `firebase_config` / `firebase_token` is the right scope call.

The blocking items for me are the two silent-success paths and the exception handling; details inline. There are also two changes to the placeholder grammar that affect existing SMS gateways and are not mentioned in the Changelog.

On the delivery signal: it is fine that an HTTP gateway cannot report what Firebase will say - that is inherent to proxying and I am happy to live with it. It just needs to be explicit in the documentation, so administrators do not read the adjusted fallback message as a delivery guarantee.

I have also replied to the open Copilot threads where I agree with them.

privacyidea/lib/smsprovider/HttpSMSProvider.py
```diff
@@ -54,10 +54,31 @@
...
+        value = value.replace("{phone}", phone)
```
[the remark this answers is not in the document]
nilsbehlen: Confirmed, and it is slightly broader than a shared-gateway problem: `_mangle_phone` runs unconditionally at line 89, before the option loop, so the mangled value is what reaches every `{phone}` substitution - a gateway used *only* for push is equally affected if `REGEXP` was ever set. `REGEXP` also stays advertised in `parameters()` for push-enabled gateways, so the configuration UI actively offers it.

I would suggest keying the skip on the message being a structured payload rather than on a configuration flag, i.e. skip mangling when `isinstance(message, dict)`. That keeps `REGEXP` working for the SMS path on a gateway that serves both, and does the right thing per message.

privacyidea/lib/smsprovider/ScriptSMSProvider.py
```diff
@@ -70,6 +73,7 @@ def submit_message(self, phone, message):
 
         script_name = self.script_directory + "/" + script
         proc_args = [script_name, phone]
+        serialized_message = json.dumps(message) if isinstance(message, dict) else message
```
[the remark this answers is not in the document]
nilsbehlen: Correct - `ScriptSMSProvider.py:68` mangles `phone` before `proc_args` is built at line 75, so the device token in `argv[1]` is affected exactly as described. Same suggested fix as for the HTTP provider: skip the mangling for structured payloads rather than adding a flag.

Worth fixing together with the `background` mode behavior I commented on separately - a rewritten device token in background mode produces no error at all, since that path returns `True` unconditionally.

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -111,6 +111,20 @@
...
+    push_gateways = []
+    for gateway in get_smsgateway(identifier=identifier):
+        try:
+            package_name, class_name = gateway.providermodule.rsplit(".", 1)
+            provider_class = get_sms_provider_class(package_name, class_name)
+        except Exception as ex:
+            log.warning(f"Failed to load SMS provider {gateway.providermodule!r}: {ex!r}")
+            continue
+        if provider_class.allows_push_messages(gateway):
+            push_gateways.append(gateway)
```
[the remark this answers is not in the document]
nilsbehlen: Agreed, and the mechanism is worth spelling out: `get_sms_provider_class` delegates to `get_module_class(package, class, "submit_message")`, which only verifies that this one method exists - there is no `ISMSProvider` check - and `set_smsgateway` / `POST /smsgateway` accept an arbitrary module string. So a provider class that does not inherit `ISMSProvider` loads fine and then fails on the capability call.

The blast radius is why I would treat this as blocking: `_get_push_gateways()` is called from `PushTokenClass.get_class_info()`, so a single such gateway row breaks push policy and token configuration rendering entirely, not just that gateway's selectability.

Slight preference for moving the `allows_push_messages` call *inside* the existing `try` rather than using `getattr(..., None)`: that also skips a provider whose capability check raises, and keeps one "this gateway is unusable" log path instead of two.

privacyidea/lib/tokens/pushtoken.py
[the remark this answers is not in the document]
nilsbehlen: Agreed on the gap; `sms_send_total` is missing as well, since the push path calls `submit_message` directly rather than going through `send_sms_identifier`.

One ordering caveat on removing the Firebase-local duplicate: `FirebaseProvider.submit_message` also emits `observe(...)` plus `inc(..., result="error")` on its `except`/re-raise path for network level failures. Lifting the instrumentation to this call site only preserves that case if the call site catches exceptions - which is what my separate comment on this line asks for anyway. So the exception handling should land first and the metrics move on top of it; done in the other order, the `result="error"` case is silently lost.

nilsbehlen:
ive added two rounds of ai reviews, doubled checked by me. do you want to implement these points yourself?

gitalexch:
> ive added two rounds of ai reviews, doubled checked by me. do you want to implement these points yourself?

Thanks for feedback! My programmer skills are not enough to implement such features and stay exactly save. So, please, rewrite and implement all changes as you consider to be acceptable, I was made just a "proof of concept" and this feature is needed for me as I am using my own Firebase gateway...

nilsbehlen:
@gitalexch no worries, we will take it from here, thanks for the proof of concept.
im not sure if this will make in 3.14 as we are very busy wrapping things up, i will try to squeeze it in.

privacyidea/lib/smsprovider/HttpSMSProvider.py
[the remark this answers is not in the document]
gitalexch: Fixed in 2d49e985. Structured PUSH payloads now bypass REGEXP mangling, while ordinary SMS messages continue to use it. Regression coverage verifies that the device token is forwarded unchanged.

privacyidea/lib/smsprovider/ScriptSMSProvider.py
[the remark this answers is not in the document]
gitalexch: Fixed in 2d49e985. Script gateways now skip REGEXP mangling for structured PUSH payloads, preserving the device token; SMS behavior is unchanged. Script launch failures also raise SMSError instead of being reported as success.

privacyidea/lib/tokens/pushtoken.py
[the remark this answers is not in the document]
gitalexch: Fixed in 2d49e985 by moving allows_push_messages() into the existing try block. Providers without the capability method, or whose capability check raises, are now skipped and logged instead of breaking class-info/policy listing. A regression test covers this case.

privacyidea/lib/tokens/pushtoken.py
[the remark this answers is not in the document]
gitalexch: Fixed in 2d49e985. PUSH and SMS delivery metrics now wrap the provider-independent submission path, including the exception/error case, and the Firebase-local instrumentation was removed to avoid duplicates. Tests cover success, false returns, and exceptions.

privacyidea/lib/smsprovider/HttpSMSProvider.py
gitalexch: Fixed in e37ac18d. PUSH-enabled HTTP gateway configuration now requires HTTP_METHOD=POST and SEND_DATA_AS_JSON=yes, and submit_message enforces the same invariant at runtime for structured payloads. Tests cover configuration and runtime rejection.

privacyidea/lib/smsprovider/HttpSMSProvider.py
gitalexch: Fixed in 2d49e985. Raw-value preservation is now limited to structured dict PUSH payloads, so legacy SMS {phone}/{otp} rendering and JSON type conversion remain unchanged. Literal {message} substitution is likewise part of the structured PUSH path, and the compatibility behavior is documented in the Changelog.

privacyidea/lib/tokens/pushtoken.py
gitalexch: Fixed in 2d49e985. Provider submission is now wrapped at the push call site; exceptions are logged and mapped to res=False, preserving challenge creation when polling is allowed. The wrapper also records the provider-independent error metrics.

privacyidea/lib/smsprovider/ScriptSMSProvider.py
gitalexch: Addressed in 2d49e985. Failures to start the script now raise SMSError in every mode instead of returning success. Background mode still cannot report final delivery, so that limitation is now stated explicitly in the gateway documentation.

privacyidea/lib/tokens/pushtoken.py
gitalexch: This remains an intentional configuration error: a token explicitly references a gateway that has been made non-PUSH-capable, so silently treating that mismatch as a transient delivery failure would hide an administrator error. Runtime delivery failures still degrade to polling through res=False. The existing test documents this distinction.

doc/configuration/sms_gateway_config.rst
gitalexch: Added in 2d49e985. The documentation now distinguishes Firebase device acceptance from HTTP gateway request acceptance, recommends RETURN_SUCCESS/RETURN_FAIL for final-result feedback, and notes that Script background mode reports no result.

nilsbehlen:
we will merge this soon, but we need to finish our own preparations first. this is not forgotten

nilsbehlen (approved):

PR #5866, opened 2026-09-02
CA - authentication widget

### LINKED ISSUE #5803: Conditional Access: Event count summary widget
- [ ] widget with summary of event counts
- [ ] histogram in widget similar to the histogram in the authentication log table actions

jelinaunger:


henrik-falk (approved):
maybe we want sliders here too and until first entry (all data) option too

PR #5868, opened 2026-09-02
Feature/landing page setting

### LINKED ISSUE #5783: webui setting: landing page
since we have settings now, we can let people select their landing page.
while it should be dashboard by fault (and dashboard should be always enabled, no more policy), we can offer a list of possibilities, like:
- Dashboard
- Token
- User
- Audit
- Authlog

I dont think policies or CA policies makes sense here

henrik-falk:


nilsbehlen:
Review notes on the branch changes (excluding the master merges).

### 1. `styles/table.scss:11` — transparent `$table-bg` breaks sticky headers

Material emits `background: inherit` on `thead` / `.mat-mdc-header-row` / `.mat-mdc-header-cell` so a pinned header picks up the table element's opaque background. `table { background-color: $table-bg }` (line 275) was the only thing feeding that chain, so with `$table-bg: transparent` the header is transparent and rows scroll visibly through the header text. Only the even columns stay opaque, since `th:nth-child(even)` still paints `--mat-sys-surface-variant` (line 310), so the header reads half see-through, half striped. Affects every sticky-header table — Audit (`sticky: true` in `audit.component.html`), containers, token lists.

Suggestion: keep the table transparent for the depth gradient, but give the header row its own explicit opaque background instead of letting it inherit one.

### 2. `ui-settings.component.ts:107` — a failed reset aborts the rest of the reset

`resetLandingPage()` is the only leg of the `forkJoin` in `resetSettings()` that can error; `appearanceService.resetToDefaults()`, `themeService.setTheme()` and `setShowLoadingUrls()` all pipe `catchError(() => of(null))`. If `DELETE /user/settings/starting_page` fails (500, or 401 on an expired JWT), the `forkJoin` errors and `switchLocale("en")` never runs — so "Reset to defaults" leaves the language untouched after theme, appearance and loading-urls were already reset.

Suggestion: `catchError(() => of(null))` inside `resetLandingPage()`, matching its siblings — that keeps the swallow-and-notify convention in the service, where `fail()` has already shown the toast.

### 3. `ui-preferences.service.ts:139` — post-login navigation waits on an unbounded GET

Navigation after a successful login now depends on `resolveLandingPath$` → `landingPage$()` → `GET /user/settings`. `landingPage$()` handles the error case but not a request that never settles: if the connection drops right after `/auth` returned, the user is authenticated with a stored JWT and a running session timer but stays on the login form with no feedback.

Suggestion: bound it in `landingPage$()` — a `timeout({ first: 5000 })` ahead of the existing `catchError` falls back to the policy-driven default. That one method is the choke point for all three callers (login component, `loginGuard`, admin route).

### 4. `admin.routes.ts:82-87` — the empty-path landing redirect is unreachable

`app.routes.ts:60` full-matches `/` to `login` before the `LayoutComponent` route is reached, so `/` always goes to `/login` and `loginGuard` makes the landing decision. This route never runs, and `self-service.routes.ts` has no equivalent — but the `path.slice(1)` relative-redirect handling makes it look load-bearing.

Suggestion: drop the block along with the five imports that exist only for it (`inject`, `resolveLandingPath$`, `AuthService`, `UiPreferencesService`, `map`); no spec covers it. Alternatively make it live by removing the `/` → `login` redirect in `app.routes.ts`, so an authenticated admin lands on their landing page without the login bounce — but that is a routing behaviour change, probably not for this PR.

### 5. `ui-preferences.service.ts:153` — a failed save leaves the dropdown showing an unsaved value

`setLandingPage()` returns `void` and fire-and-forgets, unlike `setShowLoadingUrls()` / `setTheme()`, which return the settled write. When the POST fails, `fail()` shows a toast but the cache signal is not updated — so `landingPage()` still returns the old value while `mat-select` keeps the user's pick. Because the `[value]` binding did not change, Angular never re-writes it, and the dropdown shows e.g. "Audit" while the next login still lands on Token, until a reload.

Suggestion: return the observable from `setLandingPage()` for symmetry, and hold the selection in a writable signal in the component (`linkedSignal(() => landingPage())`, bound `[(value)]`) that is set optimistically and reset from `landingPage()` on error, so the control snaps back to what is actually stored.

nilsbehlen (approved):

PR #5870, opened 2026-09-03
Fixing error raises

### LINKED ISSUE #5829: Post-policies of `/validate/check` are skipped whenever the endpoint raises
`/validate/check` (and its `/validate/radiuscheck` alias) carries 14 `@postpolicy`                            
  decorators. None of them run when the view function raises an exception: the                                  
  error is handled by the error handlers in `privacyidea/api/before_after.py`,                                  
  which build the response with `send_error()` and never pass it through the                                    
  post-policy chain.                                                                                            
                                                                                                                
  Most of the 14 only act on a successfully built authentication response and are                               
  correctly no-ops on an error path. Two of them are not:                                                       
                                                                                                                
  * `hide_specific_error_message` — the specific reason for the failure leaks                                   
    despite the policy being enabled.                                                                           
  * `construct_radius_response` — `/validate/radiuscheck` returns a JSON body and a                             
    non-400 status code instead of the documented empty `204`/`400`.

These are not exotic paths: an unknown serial raises `ResourceNotFoundError`, a                               
  serial/user mismatch and the FIDO2 argument checks raise `ParameterError`, and a                              
  revoked token raises `TokenAdminError`.

jona-samuel:
closes #5829


privacyidea/api/before_after.py
```diff
@@ -527,13 +528,121 @@ def after_request(response):
...
     response = hide_version(request, response)
 
+    # No caching! Applied last, to the final response object, so a shaped
+    # replacement response still carries the no-cache guarantee.
+    response.headers['Cache-Control'] = 'no-cache'
+
+    return response
+
+
+def shape_validate_error_response(request, response):
```
jelinaunger, line 549: this should be the single source of truth. Actually, three places of more or less (they already drifted) the same implementation exist: postpolicy decorators, in auth errorhandler, and here. Remove the other two as they are now covered both by after request. But use the policy decorator function here (move them out of the policydecorator files and into a suitable helper file).

jelinaunger (approved):

PR #5878, opened 2026-09-04
Let a token class own the token info it maintains

nilsbehlen:
A token info entry can hold what a token authenticates with — the public key of a passkey, the server a RADIUS token forwards to, the answers of a questionnaire token — next to the free-form metadata an administrator keeps there. The generic token info endpoints wrote both alike, so there was no way to tell "a note on this token" from "the value that decides how this token authenticates".

## What changes

**Every token class declares the entries it maintains itself.** Either as exact keys or as a whole namespace such as `radius.`. A class only declares what it introduces, the effective set is the union over the class hierarchy, so `TotpTokenClass` inherits HOTP's `2step_*` and the base class' `tokenkind` without repeating them. A questionnaire token declares its whole token info, because its questions *are* the keys and are chosen during the enrollment.

**The write path is split.** `TokenClass.add_tokeninfo()` / `delete_tokeninfo()` are the request-facing entry points and refuse a declared key with a `PolicyError`. `write_tokeninfo()` / `remove_tokeninfo()` are the writers for server side code, and `add_tokeninfo_dict()` / `set_tokeninfo()` use them, so the token classes, enrollment and token import keep writing those entries. Enforcing in the token class rather than in the endpoint means the token info endpoints, the token event handler, `update()`, import and the CLI are all covered by one check.

**The entries that do change over the lifetime of a token keep a way to be set.** `POST /token/set` now also accepts them as `<tokentype>.<key>`, e.g. `remote.user`, restricted to the namespace of the token's own type and audited per field like its other fields. `count_auth_max`, `hashlib`, `validity_period_*` and friends already had that endpoint. An offline refill token can be deleted, which revokes the refill, but not set to a given value.

**Enrollment parameters belong to the enrollment.** `init_token()` updates an existing token when it is called with a known serial, so `update()` runs both for an enrollment and against a token that is already in use. The `<tokentype>.*` parameters are now only taken while the token is being enrolled, which covers a fresh token, a two-step enrollment and a token in `clientwait` or `verify`.

**The WebUI follows the server.** `GET /token/` reports `readonly_info_keys` and `undeletable_info_keys` per token, computed server side so the frontend does not duplicate the rule and dynamic keys work by themselves. The token detail info card shows a restricted entry with its value but without an edit field, and only offers to delete the ones that can be deleted. It also no longer resends the whole token info on save.

## Fixed along the way

- An omitted value type cleared the type of an existing entry, which replaced an encrypted value with its plaintext. It now leaves the type unchanged, and `add_tokeninfo()` at lib level no longer drops the `value_type` it accepts. This was previously worked around in the SSH key token class only.
- The reserved `.type` suffix is refused as a key of its own. A stored `<key>.type` row shadowed the entry that is synthesized from the value type of `<key>`, and made the value be read with the wrong type.
- An unusable sign count or public key fails a passkey authentication with a log line instead of raising out of the request.

## Notes for the reviewer

- The provenance of a key decides whether it is declared, not how bad it would be to change it. `device_type` and `backed_up` are re-derived from the signed `authenticatorData` on every authentication, but they are still declared, because a policy condition on `tokeninfo.device_type` reads the stored value.
- `test_lib_tokenclass.py::test_11_every_tokeninfo_key_in_the_code_is_declared` parses the package and asserts that every key passed to `get_tokeninfo` / `add_tokeninfo` / `delete_tokeninfo` is covered by some declaration, so a new key cannot be added unnoticed. That check is what found `refilltoken` and `offline_counter`, which the offline machine application writes across token types.
- The three CLI actions that write token info report and skip a restricted entry and carry on with the remaining tokens, because a run selects tokens across types and a key can be owned by one type and free-form on another.

## Behaviour changes to be aware of

- `POST` / `DELETE /token/info/<serial>/<key>` returns 403 for an entry a token class maintains. `settokeninfo` keeps working for free-form entries.
- `ssh_key` and `ssh_comment` are no longer editable through the token info endpoint. They are covered by the SSH key integrity checksum, and re-enrolling the key is the supported way to change it.
- A re-`init` on an already enrolled token no longer rewrites its `<tokentype>.*` entries. `POST /token/set` does.
- An out-of-tree token class that calls `self.add_tokeninfo()` for a key a parent class declares has to use `write_tokeninfo()`. `add_tokeninfo` is not documented as plugin API.

Documented under `settokeninfo` and `set` in `doc/policies/admin.rst`, with a `Changelog` entry.


privacyidea/api/token.py
```diff
@@ -1325,6 +1343,17 @@ def set_api(serial=None):
                                        f"validity_period_start={validity_period_start!r}, "})
         res += set_validity_period_start(serial, user, validity_period_start)
 
+    # Token info entries a token class maintains but that change over the lifetime of a token, e.g.
+    # "remote.user" or "phone". A parameter counts as one when some token class declares it settable, so the
+    # fixed fields above and any framework parameter are unaffected.
+    settable_info = {key: value for key, value in request.all_data.items()
+                     if key not in HANDLED_SET_FIELDS and is_settable_tokeninfo_key(key)}
```
jelinaunger: is_settable_tokeninfo_key() is called once per request parameter here. It calls get_token_class_dict() on every invocation, which is uncached: it walks the ~30 modules from get_token_list(), calls dir() on
  each, and runs inspect.isclass / issubclass over every name in every module namespace. So a /token/set with six parameters rebuilds the whole token class registry six times, and set_token_type_info() does 
  it again per key. Not a hot path, but the declarations are class attributes that can't change at runtime — resolving the class list once per call, or memoizing the union of settable keys, would avoid it.  
  _collect_tokeninfo_declaration() already caches on the same grounds.

privacyidea/api/token.py
```diff
@@ -1325,6 +1343,17 @@ def set_api(serial=None):
...
+    if settable_info:
+        for key, value in sorted(settable_info.items()):
+            # The value can be a secret, e.g. the shared secret of a RADIUS server, so only the key is audited
+            g.audit_object.add_to_log({'action_detail': f"{key!s} set, "})
+        res += set_token_type_info(serial, settable_info, user=user)
```
jelinaunger: The audit records writes that didn't happen. The action_detail line is added for every requested key, before set_token_type_info() runs — but that function filters per token (own_info, continue), so a key 
  that's settable on some other type is dropped silently. POST /token/set serial=<an HOTP token> phone=+49123 logs phone set in the audit while nothing is written and res doesn't move. This is the only field
  in set_api where audited and written can diverge; the other setters write to every matched token, so their audit lines are always accurate.

privacyidea/api/token.py
nilsbehlen: Fixed in 5c1e6b746. `get_settable_tokeninfo_keys(keys)` resolves the class list once for the whole set of keys, and `set_token_type_info()` uses it for its validation loop as well, so a six-parameter `/token/set` walks the token modules once instead of twelve times (0.86 ms -> 0.19 ms measured). `is_settable_tokeninfo_key()` stays as the single-key wrapper.

Deliberately not memoized process-wide: `get_token_list()` folds in `PI_TOKEN_MODULES`, so the set of classes is a property of the application, not of the process. Resolving once per call keeps that intact.

privacyidea/api/token.py
nilsbehlen: Confirmed and fixed in 5c1e6b746. `set_token_type_info()` now returns the keys it wrote, listed once per token they reached, so its length is still the count the endpoint reports and the audit can follow the write instead of preceding it. The event handler's truthiness check on the return value is unaffected by the type change.

The cross-type case in `test_02e_token_type_entries_are_written_through_token_set` now asserts the audit as well. Against the old ordering it fails with `'radius.user' unexpectedly found in 'radius.user set, '` on the remote token, which is exactly the divergence you described.

jelinaunger (approved):

PR #5886, opened 2026-09-07
4914/ca auth log reason detail dialog

### LINKED ISSUE #5887: Conditional Access: auth log add reason detail dialog

henrik-falk:
The per-serial reason detail is a nested map - which token was found to be what, and which policies decided - and the Info column could only fold it into one line of compact JSON, which is unreadable at that width and the widest thing in the table. It moves to the Reasons column, where it belongs: an icon button next to the reasons opens a dialog listing the findings as the serial/reason table they are, with the deciding policies under them.

The Reasons column becomes a cell component of its own (cells/reason-cell), joining the other columns that render a list rather than a scalar; the reason lines and their styles move over unchanged. The button appears only for an entry that actually recorded a detail, so it never opens an empty dialog.

The Info cell skips the reason_detail key and exports hasInfoContent, which the table now asks before sizing the column: a row whose only key is the skipped one would otherwise make the Info column claim its 300px for a cell that renders nothing - and that is the common case, since most failed requests record nothing else there.

An entry can carry a detail and no reason of its own: the token that decided the outcome records none when the event type already names the cause (a wrong PIN is logged as PIN_FAIL), and the findings of the tokens the request never got to check are deliberately not promoted to the row. Such a cell would be an empty column beside a lone button, so it shows a muted "see details" instead.

### DIFF
--- a/privacyidea/static_new/src/app/components/logs/authentication-log/reason-detail-dialog/reason-detail-dialog.html
+++ b/privacyidea/static_new/src/app/components/logs/authentication-log/reason-detail-dialog/reason-detail-dialog.html
@@ -1,0 +1,78 @@
+<!--
+ (c) NetKnights GmbH 2026,  https://netknights.it
+
+ This code is free software; you can redistribute it and/or
+ modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
+ as published by the Free Software Foundation; either
+ version 3 of the License, or any later version.
+
+ This code is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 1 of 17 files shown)


privacyidea/static_new/src/app/components/logs/authentication-log/reason-detail-dialog/reason-detail-dialog.html
```diff
@@ -0,0 +1,78 @@
...
+          }
+        </tbody>
+      </table>
+    </section>
+  }
+  @if (data.policies.length) {
+    <section class="reason-detail-section">
+      <h4
+        class="reason-detail-heading"
+        i18n>Deciding policies</h4>
```
jelinaunger: I would move this up to be shown before the tokens as the policy is the crucial part for the outcome / event, the token reasons are just side notes

jelinaunger (approved):

PR #5893, opened 2026-09-07
Back up and restore PostgreSQL databases

### LINKED ISSUE #5891: pi-manage backup does not support PostgreSQL
`pi-manage backup create` and `pi-manage backup restore` only handle SQLite and MySQL/MariaDB. With a PostgreSQL `SQLALCHEMY_DATABASE_URI`, both commands print `unsupported SQL syntax: postgresql+psycopg2` and exit 2 — so a PostgreSQL installation has no working backup or restore, and finds out at recovery time. The documentation for backup/restore does not mention which engines are supported.

`backup restore` is worse than a clean failure: it extracts the archive over `/` (restoring `pi.cfg` and, if included, the encryption key) and only then bails out on the database, so it half-succeeds.

### What it should do

Dump with `pg_dump` and restore with `psql`, so PostgreSQL works the same way MySQL/MariaDB does today:

- dump as plain SQL with `--clean --if-exists`, so the dump can be replayed into a database that still holds the old schema (a plain `pg_dump` only applies to an empty database)
- restore with `--single-transaction` and `ON_ERROR_STOP`, so a failing restore does not leave a half-restored schema
- pass the password via the environment, not on the command line
- report a missing `pg_dump`/`psql` with the package to install, instead of a traceback
- the target database and its role still have to exist — the restore does not create them

### Also worth fixing along the way

- The engine is matched against a hardcoded list of driver strings, so `mysql+mysqldb` and other legitimate drivers of a supported engine fall into the "unsupported" branch.
- The URI is parsed with `urlparse`, which does not percent-decode — a password containing e.g. `%40` in the URI is passed to the client raw and authentication fails.
- A URI without a host (local socket) crashes instead of connecting over the socket.
- Nothing stops a dump from one engine being fed to another: the restore picks the dump out of the archive by filename and then uses whatever URI it finds, so a MySQL dump can be piped into `psql`.

### Scope

- PostgreSQL support in `pi-manage backup create` / `restore`
- test coverage, including the backup round-trip test on PostgreSQL in the migration-tests workflow (needs a `postgresql-client` matching the server version — `pg_dump` refuses to dump a newer server)
- document the supported engines, the required client packages, and that restore overwrites the target database

nilsbehlen:
Fixes #5891.

`pi-manage backup` handled SQLite and MySQL/MariaDB only. With a PostgreSQL `SQLALCHEMY_DATABASE_URI` both commands printed `unsupported SQL syntax: postgresql+psycopg2` and exited 2 — and `backup restore` reached that point only after extracting the archive over `/`, leaving the configuration replaced and the database untouched.

### What this does

The engine is derived from the SQLAlchemy backend name instead of matching the URI scheme against a list of driver strings, so every driver of a supported engine is recognised (`mysql+mysqldb`, `postgresql+psycopg`, …). Each engine has its own dump/restore helper and both commands classify the URI and dispatch. PostgreSQL is dumped with `pg_dump` and restored with `psql`.

Notable flags and the reason for each:

| Flag | Why |
| --- | --- |
| `--clean --if-exists` | without it a plain dump only applies to an *empty* database |
| `--no-owner --no-privileges --no-comments` | lets a role that owns neither the schema nor its objects replay the dump; a comment on the public schema is dumped whenever it differs from the server default, and applying it as a non-owner aborts the whole restore |
| `--single-transaction`, `ON_ERROR_STOP` | all-or-nothing restore instead of a half-restored schema |
| `--no-password` (both commands) | a scheduled backup must fail rather than wait on a password prompt |
| `PGPASSWORD` in the child environment | keeps the password off the command line and out of a file in the config directory, which the archive would contain |

Host and port are omitted when the query part of the URI carries them, because SQLAlchemy lets those override the URI — that is where privacyIDEA itself connects.

The dump's suffix inside the archive now records the engine that wrote it (`.pgsql` is new, `.sql` and `.sqlite` unchanged), and a restore refuses a dump written by another engine instead of feeding it to a client that cannot read it. A missing client command is reported with the package to install instead of a traceback.

### Fixed on the way past

- Parsing with `make_url()` decodes the URI, so a percent-encoded password (`p%40ss`) no longer reaches the client raw and fails authentication.
- A URI without a host connects over the local socket instead of crashing on `shlex.quote(None)`.
- MySQL option-file values are quoted and their backslashes doubled. An unquoted value ends at a `#` and the parser expands backslash escapes inside the value, so a password containing either made **every** MySQL backup fail with an access-denied error (verified against a live MariaDB).
- The MySQL restore hands the dump to the client as an open binary file instead of reading it into memory and decoding it with the locale encoding — a large dump or one written in another encoding did not survive that.

### Tests, CI, docs

The round-trip test covers PostgreSQL too, gated on the dump client of the respective engine, and restores a second time into the populated database. The migration workflow runs it on the PostgreSQL leg with a client from the PostgreSQL project's repository, since `pg_dump` refuses to dump a server newer than itself and the client Ubuntu ships is older than the `postgres:17` service. `compose-dev.yml` gained the maintained 10.x MariaDB LTS next to 11. The pi-manage chapter now documents the supported engines, the client packages per engine, and what a restore does and does not create — including the MySQL/MariaDB caveat that the dump carries its original database name.

### Verification

Round trips driven through the real CLI against live servers: PostgreSQL 17.9, MariaDB 11 and MariaDB 10.11 (each with its own client). The non-owner restore case was reproduced both ways — psql exit 3 without `--no-comments`, exit 0 with it. `tests/cli/test_cli_pimanage.py`: 50 passed. ruff and zizmor clean.

### DIFF
--- a/privacyidea/cli/pimanage/backup.py
+++ b/privacyidea/cli/pimanage/backup.py
@@ -21,21 +21,65 @@ import configparser
 import os
 import pathlib
 import re
-import shlex
 import shutil
 import subprocess
 import sys
 import tarfile
 from datetime import datetime
-from urllib.parse import urlparse
+from typing import IO, NoReturn
 
 import click
 from dateutil.tz import tzlocal
 from flask import current_app
 from flask.cli import AppGroup
 from flask.config import Config
+from sqlalchemy.engine.url import URL, make_url
+from sqlalchemy.exc import ArgumentError
 
-MYSQL_DIALECTS = ["mysql", "pymysql", "mysql+pymysql", "mariadb+pymysql"]
+SQLITE = "sqlite"
+MYSQL = "mysql"
+POSTGRESQL = "postgresql"
+
+# Backup implementation per SQLAlchemy backend name (the part of the URI before
+# the "+"), so every driver of a supported engine is recognised.
+BACKEND_FAMILIES = {
+    "sqlite": SQLITE,
+    "mysql": MYSQL,
+    "mariadb": MYSQL,
+    "postgresql": POSTGRESQL,
+    "postgres": POSTGRESQL,
+}
+
+FAMILY_NAMES = {SQLITE: "SQLite", MYSQL: "MySQL/MariaDB", POSTGRESQL: "PostgreSQL"}
+
+# The client packages providing the dump/restore commands, used in error messages.
+CLIENT_PACKAGES = {MYSQL: "mariadb-client (or mysql-client)", POSTGRESQL: "postgresql-client"}
+
+# Suffix of the database dump inside the backup archive. The suffix records which
+# engine wrote the dump, so a restore can refuse to feed it to a different one.
+# ".sql" and ".sqlite" are the names earlier versions wrote and stay unchanged.
+DUMP_SUFFIXES = {SQLITE: ".sqlite", MYSQL: ".sql", POSTGRESQL: ".pgsql"}
+DUMP_FAMILIES = {suffix: family for family, suffix in DUMP_SUFFIXES.items()}
+DUMP_FILE_PATTERN = re.compile(r"dbdump-\d{8}-\d{4}(?P<suffix>\.sqlite|\.pgsql|\.sql)$")
+
+# libpq environment variables for the connection parameters that can appear in
+# the query part of a PostgreSQL URI. Parameters outside this mapping are
+# reported as ignored instead of being dropped silently.
+POSTGRESQL_ENV_PARAMS = {
+    "host": "PGHOST",
+    "port": "PGPORT",
+    "sslmode": "PGSSLMODE",
+    "sslrootcert": "PGSSLROOTCERT",
+    "sslcert": "PGSSLCERT",
+    "sslkey": "PGSSLKEY",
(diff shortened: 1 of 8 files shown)


privacyidea/cli/pimanage/backup.py
```diff
@@ -260,38 +292,286 @@ def backup_restore(backup_file, keep_db_uri):
...
+def _postgresql_env(parameters: dict) -> dict[str, str]:
+    """Environment for pg_dump and psql.
+
+    The password is passed in the environment rather than on the command line,
+    where it would be visible in the process list, and rather than in a file in
+    the config directory, which would end up inside the backup archive. The
+    remaining connection parameters have no command line option and reach the
+    client as the libpq variable of the same meaning.
+    """
+    env = dict(os.environ)
```
henrik-falk, line 515: Maybe clear the known PG* libpq variables (PGHOST, PGPORT, PGUSER, PGDATABASE, plus everything in POSTGRESQL_ENV_PARAMS) before copying os.environ, so an ambient PGHOST/PGPASSWORD etc. can't silently redirect a dump when the URI omits that parameter.

henrik-falk (approved):

PR #5901, opened 2026-09-08
standardise dialogue widths

### LINKED ISSUE #5841: Long single-line strings cause dialogues to become very wide
Some dialogues become very wide because they contain long strings of text on a single line.
This is not an issue of the dialogue wrapper. These dialogues have to constrain their own width.

frankmer:
To prevent the dialogue from becoming a single long line of text, I added standardised dialogue widths.
S: Material dialog minimum width.
M: The width of two L inputs, which is also a good width for reading.
L: M width plus 300 px extra space for other content, such as a QR code.


henrik-falk (approved):
done

PR #5902, opened 2026-09-09
CA - fix counting for multi-master

jelinaunger:
Order the attempt reduction by time, not by row id                                                     
                                                                                                                
  ## Why                                                                                                        
                                                                                                                
  The attempt reduction that feeds the conditional-access counters read the                                     
  `authentication_log.id` as a total time order, in three places:                                               
  `_count_matching_attempts` used it for the `since_last_success` reset point,                                  
  for each attempt's representative row, and for the floor that drops attempts                                  
  older than the last success. The model stated that outright — "ordering an                                    
  attempt's rows by `id` reconstructs the full chain".                                                          
                                                                                                                
  That id is a plain autoincrement, so the order it gives is the insertion order                                
  of a **single writer** only. Under multi-master replication — Galera with                                     
  `wsrep_auto_increment_control`, which privacyIDEA documents as an HA option —                                 
  each node writes from its own offset and applies remote write sets                                            
  asynchronously, so a row committed later on one node can carry a lower id than                                
  one committed earlier on another. Two consequences, in both directions:

- a failure that followed a success drops out of the count, and the lockout                                   
    fires an attempt late — the wrong direction for a security control;                                         
- a success that followed a failure fails to reset the counter, locking out a                                 
    user who did authenticate.                                                                                  
                                                                                                                
`#5790` hit the neighbouring symptom in the audit table (ids no longer 1 apart)                               
  and its `_get_id_stride()` documents the assumption this PR removes: "one node                                
  is the steady writer for a given offset".                                                                     
                                                                                                                

## What changed                                                                                               
                                                                                                                
  **`(timestamp, id)` replaces `id` as the ordering key** (`engine._row_order`).                                
  The timestamp is written by the application, so it holds across writer nodes;                                 
  the id stays the tiebreaker and is exact for the rows of one flush — same                                     
  transaction, same node, ascending ids. This also makes the `PER_ATTEMPT`                                      
  reduction consistent with `_count_events`, which already floored the                                          
  `PER_REQUEST` counter on `max(timestamp)`. 

**`authentication_log.timestamp` carries microsecond precision** for that key                                 
  to mean anything. A plain `DateTime` renders `DATETIME` with `fsp=0` on                                       
  MySQL/MariaDB and `DATE` on Oracle, both of which collapse a whole second into                                
  one instant; PostgreSQL and SQLite already keep microseconds. Changed in the                                  
  table's own migration, which is unreleased, so no follow-up revision.                                         
  `DATETIME(6)` occupies the 8 bytes the index key budget there already charged,                                
  so no index changes.                                                                                          
                                                                                                                
  **A side effect worth naming:** this fixes an under-count that needed no                                      
  cluster at all. `_count_events` floors on `timestamp > last_success`, which on                                
  MySQL with second precision dropped *every* failure landing in the same second                                
  as the success — single node, no replication involved.                                                        
                                                                                                                
  **The statistics reduction follows** (second commit). It picked its                                           
  representative with two `MAX` aggregates over the id and a join back on                                       
  `coalesce(success_id, latest_id)`, so it would have disagreed with the engine                                 
  about which row classifies an attempt. A single `ROW_NUMBER` window ranks an                                  
  attempt's rows instead — `LOGIN_SUCCESS` first, then newest by                                                
  `(timestamp, id)` — so rank 1 *is* the representative and the rule reads the                                  
  way the engine states it. The outer join is a plain id equality now, and the                                  
  partition is what the aggregate grouped by, so rows without an `attempt_id`                                   
  still count individually.  

This is privacyIDEA's first window function, because no aggregate ranks on a                                  
  column *pair*. Every backend the suite runs against has it: MySQL 8.0+,                                       
  MariaDB 10.2+, PostgreSQL, SQLite 3.25+, Oracle — against a CI matrix of                                      
  MariaDB 10.11 and 11, MySQL 8.4, PostgreSQL 17, SQLite 3.45.                                                  
                                                                                                                
  ## What is left                                                                                               
                                                                                                                
  The residual is the clock offset between privacyIDEA instances: events falling                                
  within it are ordered by id, i.e. arbitrarily. That window is a few                                           
  milliseconds under NTP, and two authentication events of one subject that                                     
  close together are no more distinguishable than the ordering makes them. It is                                
  written down on the model rather than left implied. Exact-microsecond ties are                                
  not the concern — the only structured source of those is several rows in one                                  
  flush, where the id tiebreaker is correct by construction.   

 ## Tests                                                                                                      
                                                                                                                
  Two regression tests, each seeding rows whose ids contradict their timestamps —                               
  the inversion a multi-master cluster produces:                                                                
                                                                                                                
  - `test_count_attempts_representative_is_the_newest_timestamp_not_the_highest_id`                             
    and `..._since_last_success_floors_on_the_timestamp_not_the_id` for the engine;                             
  - `test_representative_is_the_newest_timestamp_not_the_highest_id` for the                                    
    statistics, asserting both the chart total and that `count_user_attempts`                                   
    agrees, so the two reductions cannot drift apart again.                                                     
                                                                                                                
  All three were verified to fail with the timestamp removed from the ordering                                  
  key, and pass with it. `test_lib_conditional_access_engine.py`,                                               
  `test_lib_authentication_log.py`, `test_api_authentication_log.py` (395) and                                  
  `test_api_conditional_access.py` (114) pass.


privacyidea/lib/conditional_access/engine.py
```diff
@@ -607,10 +631,10 @@ def _count_matching_attempts(rows: Sequence[AuthenticationLog], tracked_types: s
     All rows sharing an ``attempt_id`` are one authentication attempt. Its representative — the event that classifies
     the whole attempt — is the :attr:`AuthEventType.LOGIN_SUCCESS` row if the attempt ever logged in (a completed
     success is terminal: a later stray answer replayed on the same, now-answered challenge maps to the same
-    ``attempt_id`` but must not undo the success), otherwise the **latest** row by ``id``. Row ``id`` is the insertion
-    order, which orders the multichallenge steps correctly independent of event type — e.g. a wrong answer *then* a
-    continue reads as in-progress (latest = the continue), while a continue *then* a wrong answer reads as failed
-    (latest = the fail), which an event-type ranking could not distinguish.
+    ``attempt_id`` but must not undo the success), otherwise the **latest** row by :func:`_row_order`. That is the
+    insertion order, which orders the multichallenge steps correctly independent of event type — e.g. a wrong answer
```
henrik-falk: That is chronological order. Timestamp first, id only as the tiebreaker

henrik-falk (approved):

PR #5907, opened 2026-09-09
Keep the triggering client in the push text when the phone polls

### LINKED ISSUE #5787: Make the triggering client available to the push notification text (also when polling)

frankmer:
Renders the notification while the request of the authenticating client is still at
hand, stores it in the challenge data and reads it back when the smartphone polls.


privacyidea/lib/tokens/pushtoken.py
```diff
@@ -107,6 +107,11 @@
 AVAILABLE_PRESENCE_OPTIONS_NUMERIC = [f'{x:02}' for x in range(100)]
 ALLOWED_NUMBER_OF_OPTIONS = list(range(2, 11))
 DEFAULT_NUMBER_OF_PRESENCE_OPTIONS = 3
+# Caps so that neither a client header nor a long policy text can overflow the 2000
+# characters the encrypted challenge data has to fit into.
+MAX_CLIENT_TAG_LENGTH = 256
+MAX_STORED_QUESTION_LENGTH = 512
+MAX_STORED_TITLE_LENGTH = 128
```
jelinaunger, line 114: The caps count plaintext characters, while Challenge.set_data (privacyidea/models/challenge.py:142) stores encryptPassword(json.dumps(data)) into Unicode(2000):                     
                                                                                                                                                                                                                  
  - json.dumps defaults to ensure_ascii=True → each non-ASCII char becomes a 6-char \uXXXX escape                                                                                                                 
  - encryptPassword hexlifies → roughly doubles again (32 + 1 + 2*padded_len)                                                                                                                                     
                                                                                                                                                                                                                  
  Effective budget is therefore ~980 plaintext chars, and 512+128 only stays under it for pure ASCII — which is exactly what the new tests use, and SQLite ignores column lengths, so the tests can't catch it.   
  Measured for a presence challenge with a 512-char é question and 128-char ü title: 4006 plaintext chars → ~8065 chars of ciphertext against the 2000-char column.         

<br>

  Reachable from a request header: werkzeug decodes headers as latin-1, so a 256-char high-Latin-1 User-Agent plus a push_text_on_mobile policy containing {ua_string} gets you there. On MySQL in strict mode    
  /validate/check raises a DataError; non-strict truncates silently, decryptPassword then fails, get_data() returns {}, and the challenge loses mode/correct_answer/smartphone_confirmed — the precise failure the
  new comment and test 16b claim to prevent.                                                                    


<br>

  Fix direction: enforce the budget on the encoded size (cap len(json.dumps(...)) or the resulting ciphertext), or use ensure_ascii=False with a byte budget, instead of character counts.

privacyidea/lib/tokens/pushtoken.py
```diff
@@ -301,6 +295,8 @@ def _build_smartphone_data(token: TokenClass, challenge: str, registration_url:
                            recipient={"givenname": user.info.get("givenname") if user else "",
                                       "surname": user.info.get("surname") if user else ""},
                            challenge=options.get("challenge"))
+    for tag in ("ua_string", "ua_browser", "action"):
+        tags[tag] = tags[tag][:MAX_CLIENT_TAG_LENGTH]
```
jelinaunger, line 299: add a debug log message if a tag was truncated. This also holds for the other strings gonna be truncated now.

henrik-falk (approved):

PR #5910, opened 2026-09-09
CA - Bind a principal's own log entries to their account, not their login name

jelinaunger:
### The gap                                                                                                   
                                                                                                                
  `get_policy_visibility_scopes` built a user-role caller's boundary from                                       
  `realms=[realm], usernames=[username]`, and the admin's own-entries alternative                               
  in `get_authentication_log_visibility_scopes` did the same — although every                                   
  authentication-log row also records the `(resolver, uid, realm)` identity that                                
  the engine counts its subjects by (`ix_authlog_user_event_time`).                                             
                                                                                                                
  A login is not an identity, which cuts both ways:                                                             
                                                                                                                
  * **False negatives, in normal operation.** `user.login` is stored as typed                                   
    unless the resolver has multiple login names, and the boundary matches                                      
    case-sensitively, so a user who logs in as `Alice` one day and `alice` the                                  
    next sees only the casing-matching half of their own history. A login rename                                
    loses them everything recorded before it.                                                                   
  * **A false positive.** Delete an account and recreate the same login name in                                 
    that realm, and the new account reads the previous one's authentication                                     
    history — source IPs, serials, timestamps, failure reasons.                                                 
                                                                                                                
  The remembered-device feature in this same codebase binds to                                                  
  `(resolver, user_id, realm_id)` and documents why: so a device "can never be                                  
  recognised for a *different* account that later reuses a freed login name".                                   
  The log's own boundary should hold to that.

### The change                                                                                                
                                                                                                                
  * `AuthenticationLogVisibilityScope` gains a **`uids`** dimension, enforced in                                
    all three consumers — the auth-log query (`visibility_condition`), the lock                                 
    query (`state._visibility_condition`) and `user_matches_scopes`. A dimension                                
    honoured in only some of them would silently widen whichever boundary carried                               
    it.                                                                                                         
  * New **`own_entries_scope(login, realm)`** (`lib/policies/helper.py`) resolves                               
    a principal to `(resolver, uid, realm)` and returns the scope for exactly                                   
    that account. It reuses the user object `before_request` already resolved for                               
    a user-role caller (comparing login and realm first, so a subject taken from                                
    the request parameters cannot slip in) and otherwise does one lookup.                                       
  * Both own-entries paths build their scope from it: the user branch of                                        
    `get_policy_visibility_scopes`, and the admin's alternative in                                              
    `get_authentication_log_visibility_scopes` — which every log-reading endpoint                               
    goes through (listing, attempt summary, CA outcome statistics).                                             
                                                                                                                
  No schema change: the columns the boundary now matches on are already recorded. 

### Behaviour decisions worth a reviewer's attention                                                          
                                                                                                                
  1. **Fail closed on an unresolvable identity.** A user then sees nothing rather than every entry in their realm    
  that carries their login name — which, once that name has been reassigned, is another account's history. A scoped admin keeps their policy                               
     scope without the own-entries alternative. The alternative — falling back to                               
     the login name — is the bug being fixed, so it is not offered.                                             
  2. **Entries that resolved to no account are nobody's own.** A row with                                       
     `resolver`/`uid` NULL that names a login (a `USER_UNKNOWN`, say) is now                                    
     visible to administrators only. A user therefore no longer sees                                            
     "someone tried my login name and it never resolved" — under a reused name                                  
     such a row cannot be attributed to either account.                                                         
  3. **A local database admin keeps login-name matching.** With neither realm nor                               
     resolver there is no identity to bind their rows to, so they stay matched by                               
     username plus `user_role=admin-internal`; the role is what keeps a same-named                              
     realm user's entries out. This branch is documented as inherent rather than                                
     pending.


privacyidea/lib/policies/helper.py
```diff
@@ -168,15 +168,55 @@ def get_admin_audit_params() -> dict:
...
+    if not (login and realm):
+        return None
+    user = getattr(request, "User", None)
+    # before_request resolves a user-role caller's own account already (resolve_logged_in_user), so reusing it saves
+    # a second resolver lookup; login and realm are compared because request.User otherwise carries a subject taken
+    # from the request parameters.
+    if not (user and user.login == login and user.realm == realm.lower()):
+        try:
+            user = User(login=login, realm=realm)
+        except Exception as ex:
```
henrik-falk, line 199: Consider narrowing this to `except (UserError, ResolverError)

henrik-falk (approved):

privacyidea/lib/policies/helper.py
jelinaunger: There is a wide range of possible errors which can occur in the resolvers, so I keep it this way

PR #5911, opened 2026-09-09
4914/ca dry run enforced since

henrik-falk:
Reset conditional-access counters when a policy leaves dry-run

A policy that had dry_run=True and accumulated failures during a trial window was, on the very next request after dry_run was turned off, judged against the full accumulated history and could immediately lock/block - the exact scenario dry-run exists to protect against.

Add ConditionalAccessPolicy.enforced_since, set when dry_run flips True -> False, and floor the count window there in engine._effective_window_seconds so only failures from that point on are counted.


tests/test_lib_conditional_access_engine.py
```diff
@@ -999,6 +1000,30 @@ def test_dry_run_outcome_has_no_expiry_when_the_duration_is_misconfigured(self):
...
+        # Failures accumulated while the policy was dry-run must not count once it starts enforcing - flipping
+        # dry_run off is the transition dry-run exists to make safe.
+        policy, _stages = self._make_policy(name="was_dry", counter_type=AuthEventType.MFA_FAIL, dry_run=True)
+        self._seed_events(AuthEventType.MFA_FAIL, 3)
+        evaluate_conditional_access_policies(CAContext(self.user), AuthEventType.MFA_FAIL)
+        self.assertFalse(is_user_locked(self.user))  # still dry-run: no lock
+
+        policy.dry_run = False
+        policy.enforced_since = utc_now()
+        db.session.commit()
```
jelinaunger: use the lib function here, setting it directly on the orm object does not cover that enforce_since is automatically set on dry_run toggle

privacyidea/lib/conditional_access/policy.py
```diff
@@ -1139,6 +1140,11 @@ def update_conditional_access_policy(
     Existing locks/blocks written before the change are timed and expire on their
     own, so no stale state is left enforced.
 
+    Turning ``dry_run`` off (``True`` -> ``False``) sets ``enforced_since`` to now, so the count functions floor
```
jelinaunger, line 1166: actually this happens on every toggle, but I think it should only be as stated here.

privacyidea/lib/conditional_access/engine.py
```diff
@@ -819,6 +819,28 @@ def _count_scoping(policy: ConditionalAccessPolicy) -> "tuple[list | None, Calla
...
+    exclude the very failures the transition is meant to start counting. Once the configured window has elapsed
+    since ``enforced_since`` this returns the configured width unchanged. ``enforced_since`` in the future (a naive
+    clock skew) is clamped to an empty window rather than a negative one.
+
+    :param policy: the policy whose window is computed
+    :param window_end: the instant the window ends (already normalized to naive UTC by the caller)
+    :return: the window width in seconds to hand to the counters
+    """
+    window_seconds = policy.time_window_seconds
+    if policy.enforced_since is not None:
```
jelinaunger, line 979: might make sense to only do this if dry run is disabled

privacyidea/migrations/versions/f4a7c8e2b910_ca_policy_enforced_since.py
```diff
@@ -0,0 +1,67 @@
...
+def downgrade():
+    try:
+        op.drop_column('conditional_access_policies', 'enforced_since')
+    except (OperationalError, ProgrammingError) as exx:
+        msg = str(exx.orig).lower()
+        if "no such column" in msg or "does not exist" in msg or "check that it exists" in msg:
+            print("Column 'enforced_since' already removed.")
+        else:
+            print("Could not remove column 'enforced_since' from table 'conditional_access_policies'.")
+            raise
```
jelinaunger: this can be folded into the existing migration script

jelinaunger (commented):
Documentation is missing. While is might make sense, it might not be clear that it behaves like this. It might also make sense to explicitly not have the reset. It could also be optional with a dialog appearing on the dry-run disable action.

privacyidea/migrations/versions/173d32328846_conditional_access_policies.py
```diff
@@ -52,6 +60,43 @@ def _create_table(table_name, *columns):
...
+        connection = op.get_bind()
+        connection.execute(sa.text(
+            "UPDATE conditional_access_policies SET enforced_since = :enforced_since "
+            "WHERE dry_run = :dry_run"
+        ).bindparams(sa.bindparam("enforced_since", utc_now(), type_=sa.DateTime()),
+                     sa.bindparam("dry_run", False, type_=sa.Boolean())))
+    except Exception as exx:
+        print(f"Could not backfill 'enforced_since' for existing conditional-access policies: {exx}")
+        raise
+
```
jelinaunger: this is unnecessary as ca is not released yet so no existing columns to be updated

jelinaunger (approved):

PR #5919, opened 2026-09-09
4914/ca realm deletion warning

henrik-falk:
Warn when deleting a realm still referenced by a conditional-access policy

A USER_REALM condition's value is never re-validated after write time, so a deleted realm silently changes what the condition matches: IN fails safe (matches fewer requests), but NOT_IN fails open - no user can ever have a deleted realm again, so the exclusion widens to match every request, silently expanding a DENY/LOCK_USER policy to the whole installation.


privacyidea/lib/realm.py
```diff
@@ -173,6 +173,43 @@ def get_default_realm():
...
+                  ConditionalAccessPolicyCondition.value).join(
+        ConditionalAccessPolicyCondition,
+        ConditionalAccessPolicyCondition.policy_id == ConditionalAccessPolicy.id).where(
+        ConditionalAccessPolicyCondition.condition_type == ConditionType.USER_REALM)
+    for policy_name, operator, value in db.session.execute(stmt).all():
+        if value and realm_name in value:
+            if operator == ConditionOperator.NOT_IN:
+                log.warning(f"Conditional-access policy '{policy_name}' excludes realm '{realm_name}' via "
+                            f"NOT_IN. Deleting this realm will widen the policy to match every request, "
+                            f"since no user can have a deleted realm.")
```
jelinaunger: Actually, there is no widening. The deleted realm stays in the ca policy definition, hence the behaviour for the other realms stays the same. It just does not make sense anymore, since no user of this realm exist.

<br>

Still a warning should be shown due to the following reasons:
1. CA policy becomes invalid / un-saveable
2. Re-creating a realm with the same name silently rebinds the condition
3. users migrated to another realm lose their exemption

<br>

Additionally, I would show the warning in the UI / return from the API, similar as for the internal user attributes. Just with the difference we do not delete any ca policy, but show a warning like "This realm is still defined in the following CA policies. Are you sure you want delete this realm?"

privacyidea/api/realm.py
```diff
@@ -411,6 +412,34 @@ def get_default_realm_api():
     return send_result(res)
 
 
+@realm_blueprint.route('/<realm>/delete-warnings', methods=['GET'])
+@log_with(log)
+@prepolicy(check_base_action, request, PolicyAction.RESOLVERDELETE)
+def get_realm_delete_warnings_api(realm=None):
```
jelinaunger: Not sure the /realm/<realm>/delete-warnings endpoint is really required, for a few reasons:                                                                                                                                           
                                                                                                                                                                                                                                        
  - The DELETE path already covers this on its own, and the PR keeps the full 908/909 error flow at realm-table.component.ts:658-672. Drop getRealmDeleteWarnings and everything still works — you just get two dialogs instead of one  
    when a realm has both custom attributes and a CA reference.                                                                                                                                                                         
  - The warning text now lives twice: Python f-strings in lib/realm.py and $localize strings in realm-table.component.ts:577-585.                                                                                                       
  - The comment at realm-table.component.ts:655 correctly notes the pre-check can go stale, so the error path can never be retired — both paths stay forever.                                                                           
  - The usual argument for a pre-check ("an error can only return prose, the endpoint returns localizable lists") doesn't really apply: send_error(..., details=...) (api/lib/utils.py:149) already emits a top-level "detail" key and  
    AuthError uses it end to end (lib/error.py:114 → before_after.py:663). Only UserError is unwired, because privacyidea_error (before_after.py:740) drops details.                                                                    
                                                                                                                                                                                                                                        
  An alternative that gets the same single up-front dialog: have delete_realm collect all confirmable blockers before raising, replace 908 with one generic code instead of adding 909, and put the per-blocker lists in detail. Worth  
  doing now if at all — 908 is on master but only in v3.14dev*, so renumbering is still free. Either way, delete_custom_attributes and confirm_ca_policies should stay separate flags (one deletes data, the other only proceeds despite
  a dangling reference), and the prose message needs to stay because pimanage uses it as its confirm prompt (cli/pimanage/pi_config.py:194).                                                                                            

<br>
                                                                                                                                                                                                                                        
  However, if you decide to keep the endpoint, could you at least:                                                                                                                                                                      
                                                                                                                                                                                                                                        
  - give the warning messages a single source so the two copies can't drift,                                                                                                                                                            
  - add API-level tests for the new route (the diff only touches tests/test_lib_realm.py),

jelinaunger (approved):

PR #5937, opened 2026-09-10
Record the answer to a push challenge in the audit log

### LINKED ISSUE #5754: Push decline reason is not written to the audit log
The audit log should indicate whether a push response was rejected and, if so, the reason.
However, there is no indication of whether it was rejected, nor is the reason provided.

frankmer:
Now the admin can trace what actually happened. Before, there was just a successful POST /ttype/<ttype> request without any information.


nilsbehlen (approved):

nilsbehlen:
i have added a split for the event for conditional access, so that cancel and unknown_trigger are separate events since unknown_trigger is a higher value signal that deserves separate processing

PR #5945, opened 2026-09-11
4914/ca review fix pass

henrik-falk:


nilsbehlen:
The `count_before` fix has no test for `count_mode=PER_ATTEMPT`, and it looks like it does not actually hold in that mode.

`_exclude_own_rows` layers the id exclusion onto the `row_filter`, but `_count_attempts` applies `row_filter` to the *reduced representative*, not to the rows feeding the reduction — as its own docstring warns, narrowing the reduction's input misclassifies rather than narrows. This request's row is the newest, so it *is* the representative of its attempt, and excluding it drops the whole attempt from `count_before` even when an earlier request of that same attempt already contributed it. So `count_before` is one too low and a fire-once action fires again on every further request of the attempt — the exact refire the second commit says it fixed by switching from `attempt_id` to `own_row_ids`.

Reproduced against this branch with a `PER_ATTEMPT` twin of `test_fire_once_does_not_refire_on_a_later_request_of_the_same_multi_request_attempt`:

```python
self._make_policy(name="lock3", counter_type=AuthEventType.MFA_FAIL, count_mode=CountMode.PER_ATTEMPT)
self._seed_attempts(AuthEventType.MFA_FAIL, 2)

own = tuple(self._seed_attempt("att-shared", [AuthEventType.MFA_FAIL]))
first = evaluate_conditional_access_policies(CAContext(self.user, own_row_ids=own), AuthEventType.MFA_FAIL)

own2 = tuple(self._seed_attempt("att-shared", [AuthEventType.MFA_FAIL]))
second = evaluate_conditional_access_policies(CAContext(self.user, own_row_ids=own2), AuthEventType.MFA_FAIL)

own3 = tuple(self._seed_attempt("att-shared", [AuthEventType.MFA_FAIL]))
third = evaluate_conditional_access_policies(CAContext(self.user, own_row_ids=own3), AuthEventType.MFA_FAIL)
```

Outcomes are `1 / 1 / 1`; expected `1 / 0 / 0`. The count stays at 3 for all three requests (one attempt, counted once), but `count_before` is 2 every time.

The shipped `PER_ATTEMPT` templates escape this because their only action is `DENY`, which `_build_stages` defaults to `retrigger_above_threshold`. An admin-configured fire-once `LOCK_USER` or `EMAIL_ADMIN` on a `PER_ATTEMPT` policy does not — it re-locks and re-mails on every retry within one challenge-response login.

Fix would be to exclude the rows *before* the reduction for the attempt counters (pass the ids into `_count_attempts` and skip them in `_count_matching_attempts`'s first pass) rather than filtering the representative afterwards, plus the missing `PER_ATTEMPT` test.

nilsbehlen:
Rest of the review of this branch against `4914/conditional_access`. The seven claimed fixes hold up otherwise: the migration chain is linear and single-headed again (`b8c9d0e1f2a3 → e0f1a2b3c4d5 → 0147d78cbace → 173d32328846 → c1a9f7e2b840 → b2f5c9e1a7d4 → d3e8b1c47f92`) with `e0f1a2b3c4d5` byte-identical to master; F13 converged on the base's own `rejected_by_conditional_access` rather than the branch's `gate_rejected`; and all the affected suites pass locally (engine 234, policy/request-context/migration 143, authentication_log 120, api_conditional_access 112, api_lib_policy 39, remembered_device 44). The branch is 2 commits behind the base (`1ea1cabcc` + its merge); `git merge-tree` says that merge is conflict-free.

Ordered by how much they matter.

---

### The merge dropped both commits' `tests/test_api_conditional_access.py` delta

`git diff origin/4914/conditional_access HEAD -- tests/test_api_conditional_access.py` is empty, although 5d91aa91e added 51 lines there and 58120557f reworked them (+40/−15). The merge resolved that file to the base version, which #5941 had rewritten wholesale. The F14 code change in `_refuse` went with it.

Losing that part is *correct*: `_refuse` no longer exists, because #5941 made a tripping request keep its own outcome, so there is no response left to strip offline OTPs or the remember-device cookie from. F14 is superseded, not fixed here.

What survives is the header-stripping guard added for it in `privacyidea/lib/remembered_device.py:424`, whose own docstring says it is "not currently reachable". Either drop that guard along with the rest of F14, or keep it and say in the PR description that F14 was superseded by #5941 — right now the commit message claims a fix that is not in the tree. Worth re-checking the merge resolution generally; nothing guarantees that test file was the only casualty.

### The column retrofit in `0147d78cbace` should be deleted, not extended

None of the conditional-access migrations have ever shipped, so this branch only has to be the tip of the chain — it does not have to repair a database stamped from an earlier form of the file. That premise is the whole of the F19 column retrofit (`privacyidea/migrations/versions/0147d78cbace_authentication_log.py:110-119`, plus the paragraph at :93-99), so it is dead weight and should go, together with `tests/test_migration_0147d78cbace.py`, which reproduces exactly that starting state.

(For what it is worth, it also would not have done the job: `id`, `event_type`, `timestamp`, `auth_log_id` and `reason` are all `nullable=False`, contrary to the inline claim, and the in-place amendments include a rename (`event_id` → `id`, 94c0eb348) and collation/length changes to existing columns (b0e4e031a, 12b17a2f0) that adding missing columns cannot repair. Moot now, but the paragraph asserting "only new columns and indexes were folded in later" is not accurate.)

Two things to keep:

* **F18 stays.** It is not self-correction — `e0f1a2b3c4d5` is on master, so re-parenting it under the CA chain would strand a database already stamped there. Restoring it byte-identical and chaining `0147d78cbace` after it is exactly what makes this branch the tip.
* **The pre-existing index retrofit stays** (`:125-131`), but on the autocommit rationale its docstring already gives: `create_table` and each `create_index` are separate autocommitting statements, so a run that creates the table and then fails leaves a table with no indexes and a re-run must still add them. That is partial-failure recovery, not drift repair — and it cannot apply to columns, since `create_table` is a single statement. The trailing sentence "The same amended-in-place history applies here" should go with the rest, since it justifies the index retrofit on the drift story instead.

If the migration test is worth keeping at all, it wants re-framing to that partial-failure premise (create the table alone, run `upgrade()`, assert the indexes appear) rather than "stamped before the later columns existed".

### `_exclude_own_rows` also filters the reset-on-success floor lookup

`_exclude_own_rows` puts the id exclusion into `sql_filters`, and `_count_events` spreads those into the `LOGIN_SUCCESS` `max(timestamp)` subquery that computes the `reset_on_success` floor, not just into the count. So `count_before` means "as if this request had never existed" rather than "before this request's rows joined it": if one of the request's own rows is the success that floors the window, excluding it moves the floor backwards, `count_before` is computed over a *wider* range than `count` and can exceed it, making `count_before < threshold <= count` unsatisfiable.

Latent today — no current flow stages a `LOGIN_SUCCESS` followed by tracked failure rows in one request — but the fix is small: pass the exclusion through `_count_events` as its own argument so it reaches the counting statement only.

### Four new ruff violations in `engine.py`

`engine.py` is ruff-clean on `4914/conditional_access` and has four errors here: E501 at 949 and 1798 (over the 120-char limit), E731 at 970 and 972 (assigned lambdas). The two `combined_row` lambdas would read better as a nested `def` anyway.

### `CAContext` docstring loses the blank line that ends the field list

`privacyidea/lib/conditional_access/context.py:95` — the new `:ivar own_row_ids:` block consumed the blank line that separated the field list from the "Note what is deliberately *absent*" paragraph, so docutils hits a field list ending without a blank line and folds the Note into the field body. One blank line after `not computed.`.

### `doc/conditional_access/policies.rst` will contradict itself after merging the base

`1ea1cabcc` (on the base, not yet merged here) adds a paragraph roughly twenty lines below this branch's edit, saying a fire-once action fires "only on the request where the count *equals* the threshold" — the exact claim the edit at :149 replaces. Different hunks, so the merge lands both silently. Its conclusion still holds (a subject already past the threshold stays unrestricted, hence re-trigger on the top stage); only the premise needs rewording to the new "carried the count to at or above the threshold" phrasing.

### Note: `count_before` costs a second count query

Every post-auth evaluation whose count already owns a stage now runs the counting query twice, and under `PER_ATTEMPT` that means materialising the subject's whole in-window row set as ORM objects a second time. The early return when no stage is triggered keeps it off the path for unaffected subjects, so this is a note rather than a defect — but if it ever shows up, `_count_attempts` could return both counts from one fetch, the exclusion being a pure Python predicate anyway.

nilsbehlen (approved):

PR #5956, opened 2026-09-14
Stop the test suite from exhausting the PostgreSQL connection limit

nilsbehlen:
Running the full suite against PostgreSQL sporadically fails with

```
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) connection to server
at "127.0.0.1", port 5432 failed: FATAL:  sorry, too many clients already
```

usually in `tests/test_lib_audit.py`, which is not where the connections went.

## Why only PostgreSQL, and why only Python 3.10

`postgres:17` defaults to `max_connections=100`, minus three superuser-reserved,
so 97 are usable. MariaDB and MySQL default to 151 and SQLite has no server, so
PostgreSQL is simply the only backend whose ceiling sits inside the suite's
working set.

The connections are released when the cyclic garbage collector reclaims the
objects that own the pools, so how many are open at any moment depends on GC
timing. Measured on the same commit, same `-n 4`, same database, Python 3.10
peaked at 78 concurrent connections where 3.13 peaked at 54, and under 3.10 all
four workers sat at 27-36 at once rather than one. 78 of 97 is a thin margin.

## What is fixed

**`tests/test_migrations.py` stranded connections permanently.** The `flask_app`
fixture builds an application per test and Flask-SQLAlchemy gives each one its
own engine, held in a `WeakKeyDictionary` keyed on the application. Dropping the
application does not close the pool. Measured: **+2 per test, 22 left open at the
end of the file**, never released, on 3.10 and 3.13 alike. Disposing the engines
in the fixture teardown takes that to **0**, and the file also runs faster
(37.3s -> 26.7s) because it no longer builds up connection pressure.

**Finalizers that can never run.** `sqlaudit`, `SQLIdResolver` and `sqlstats`
defer their session cleanup to a request finalizer. Outside a request - pi-manage,
a cron job, a background thread - nothing tears the context down, so the callback
never runs and only keeps the object, and its open connection, alive for the life
of the process. `lib/metrics.py` already guards this with `is_request_context()`;
these three now do too. In one file 63 of 462 audit objects were built outside a
request; after the change the number of finalizers registered matches the number
of audit objects built inside a request exactly.

**Engine ownership.** With the default "null" registry each object gets a private
engine, but only the audit module disposed it - the resolver and the monitoring
module closed the session and left the engine's connection to the user store open
until garbage collection. With the "shared" registry the audit module had the
opposite problem: it disposed an engine other requests still expect to find
connections in, defeating the pooling that registry exists to provide. Ownership
now decides.

**The ceiling.** One connection per request is what `PI_ENGINE_REGISTRY_CLASS =
"null"` means, and that is the shipped default the tests exercise; the suite is
right to run that way. Raising the test server's limit to 300 is what actually
removes the failure mode. A service container in a workflow takes no command, so
CI writes the setting and restarts the server.

## Testing

Full suite on the combination that fails, Python 3.10 + PostgreSQL, `-n 4
--dist=loadfile`, plus the directly affected suites run consecutively
(`test_lib_pooling`, `test_lib_lifecycle`, `test_lib_audit`,
`test_lib_monitoringstats`, `test_lib_resolver`, `test_lib_metrics`,
`test_lib_user`, `test_lib_usercache`, `tests/cli/`, `test_lib_periodictask`,
`test_lib_tasks`, `test_api_monitoring`, `test_api_audit`) - all green.

Note on the numbers: the deterministic wins are the migration fixture (22
stranded -> 0) and the library fixes. The whole-suite peak under `-n 4` is
dominated by the per-request engine churn that the "null" registry implies by
design, and moves much less - which is why the raised ceiling is part of this
rather than an afterthought.


nilsbehlen:
not going to raise coverage on those files, it stays as it was before

henrik-falk (approved):
The is_request_context() guard added around register_finalizer in SQLIdResolver.py, sqlaudit.py and sqlstats.py doesn't just skip a finalizer that would never run and it also disables the non-request callers that do invoke call_finalizers() (tests/base.py:461, the pitokenjanitor teardowns), so those objects engines keep their pooled connections until GC, which works against the PR's own goal; the fix is an app.teardown_appcontext hook, as init_ca_session already uses.

frankmer (approved):

PR #5960, opened 2026-09-15
Count users with and without tokens in the token usage widget

frankmer:
- The token usage widget now also shows how many users own a token and how many do not, for one realm at a time. The realm can be chosen in the widget and is stored in its settings.
- New endpoint `GET /token/ownercount` (admin only, `tokenlist` action) counts the distinct token owners, overall and per resolver.
- New `has_tokens` filter on `GET /user/`. The widget's two user counts link to the user list with this filter set.
- SCIM resolver: listed users now include a `userid`, so they can be matched against token owners.


privacyidea/api/token.py
```diff
@@ -775,6 +776,52 @@ def list_api():
...
+        out of every realm's count, as it is by the ``has_tokens``
+        filter of :http:get:`/user/`. Several realms may be given as a
+        comma-separated list. Without this parameter every realm the
+        caller may see is counted.
+    :status 200: ``result.value`` is ``{"count": <number>,
+        "by_resolver": {<resolver name>: <number>}}``.
+    """
+    realm = get_optional(request.all_data, "realm")
+    realms = [r.strip() for r in realm.split(",") if r.strip()] if realm else None
+    allowed_realms = getattr(request, "pi_allowed_realms", None)
```
nilsbehlen: **`tokenlist` realm scope is applied to the owner realm**

`request.pi_allowed_realms` is filled by `check_admin_tokenlist` from the *token* realms of the policy, and every other consumer enforces it against `TokenRealm` (`_create_token_query`, `lib/token/query.py:130`). Here it is handed to `get_token_owners_per_resolver()`, which joins `TokenOwner.realm_id`, so the two scopes disagree in both directions:

- a token with tokenrealm A owned by a user in realm B is listable by an admin scoped to A, but contributes nothing to that admin's count;
- a token owned by a user of realm A whose tokenrealm is only B is *not* listable by that admin, yet its owner **is** counted — an admin holding `tokenlist` on A and no `userlist` learns about token ownership outside the realms they may read.

The owner-realm semantics of the `realm` *query parameter* are deliberate and documented — it is reusing the token-realm *authorization* list for them that leaks. Either intersect with `TokenRealm` as everywhere else, or gate the endpoint on a right that is explicitly owner-realm-based.

privacyidea/static/src/app/components/dashboard/widgets/tokens-widget/tokens-widget.component.ts
```diff
@@ -108,8 +184,46 @@ export class TokensWidgetComponent extends DashboardWidget implements OnInit {
...
   });
 
+  readonly userCounts = computed<UserTokenCounts>(() => {
+    const owners = this.dataRef()?.value()?.owners?.result?.value;
+    const userList = this.usersRef()?.value();
+    const skippedResolvers = userList?.detail?.skipped_resolvers ?? [];
+    const users = userList?.result?.value;
+    const withTokens = owners ? this.reachableOwners(owners, new Set(skippedResolvers)) : null;
+    // A resolver shared between realms reports its users once per realm; the owner count does not.
+    const total = users && new Set(users.map((user) => `${user.resolver}\u0000${user.username}`)).size;
```
nilsbehlen: **The two counts de-duplicate users by different rules**

Realm X has resolverA and resolverB, both serving a login `alice`, and both alices own a token:

- `get_user_list()` keys `users_dict` on `(username, realm)` (`lib/user.py:1046`) → `GET /user/?realm=X` returns **one** alice, so `total = 1`.
- `get_token_owners_per_resolver()` counts distinct `(resolver, user_id)` pairs (`lib/token/query.py:1010`) → `GET /token/ownercount?realm=X` returns **2**, so `withTokens = 2`.

The widget then renders "Users with tokens: 2" for a realm whose user list holds 1 user, and `withoutTokens = Math.max(0, 1 - 2) = 0` hides the contradiction instead of surfacing it. Clicking through to `/users?realm=X&has_tokens=True` shows 1 row beside the count of 2.

The same split also misfiles individual users: `owner_keys[user_tuple]` (`lib/user.py:1048`) is written only for the **first** resolver that reports a given username, so a token-less alice in resolverA masks the token-owning alice in resolverB under `has_tokens`.

Both sides need one identity. Since the user list is what the click-through lands on, counting owners per `(username, realm)` is the consistent choice; if `(resolver, user_id)` is the intended identity instead, then `withoutTokens` must not be derived by subtracting from a differently-keyed total.

privacyidea/lib/user.py
```diff
@@ -997,6 +1017,8 @@ def get_user_list(param: dict | None = None, user: User | None = None,
                 user_list = resolver.getUserList(search_dict, list(requested_user_store_attributes))
                 succeeded_resolvers.add(resolver_name)
                 for user_info in user_list:
+                    # Read before the attribute stripping below, which may drop it from the record.
+                    user_id = str(user_info.get("userid") or "")
```
nilsbehlen: **A `userid` of `0` collapses to the empty string**

`str(user_info.get("userid") or "")` is a falsy test, not a `None` test. A resolver with an integer userid column and a row with id `0` yields `""` while `TokenOwner.user_id` holds `"0"`, so the triple never matches: the user is reported as owning no token under `has_tokens=True` and is wrongly listed under `has_tokens=False`.

And because `get_token_owner_keys()` uses `user_id or ""` as well (`lib/token/query.py:1042`), that same `""` **does** match any owner row with a NULL or empty `user_id` in the same realm+resolver — flipping the answer the other way for a different user.

```python
user_id = "" if (uid := user_info.get("userid")) is None else str(uid)
```

and the same treatment on the query side.

privacyidea/lib/token/query.py
```diff
@@ -990,6 +990,58 @@ def get_num_tokens_in_realm(realm: str, active: bool = True) -> int:
...
+    :return: A dict mapping each resolver name to its number of token owners
+    """
+    owners = select(TokenOwner.resolver, TokenOwner.user_id).select_from(TokenOwner).distinct()
+    if realms is not None:
+        owners = owners.join(Realm, Realm.id == TokenOwner.realm_id).where(
+            func.lower(Realm.name).in_([r.lower() for r in realms]))
+    per_resolver = owners.subquery()
+    rows = db.session.execute(
+        select(per_resolver.c.resolver, func.count()).select_from(per_resolver).group_by(per_resolver.c.resolver))
+    return {resolver or "": count for resolver, count in rows}
```
nilsbehlen: **NULL and `''` resolver names collapse to one dict key**

On MySQL/MariaDB and PostgreSQL `TokenOwner.resolver` can hold both `NULL` and `''` (only Oracle folds them together). `{resolver or "": count for resolver, count in rows}` then produces two entries keyed `""`, the later one wins, and the earlier count disappears.

`api/token.py:821` returns `sum(per_resolver.values())` as `count`, so the invariant promised in the docstring ("those numbers add up to `count`") holds only because the lost rows are missing from both numbers.

Coalesce in SQL so the grouping happens in the database:

```python
resolver_col = func.coalesce(TokenOwner.resolver, "")
```

privacyidea/lib/user.py
```diff
@@ -882,6 +893,13 @@ def get_user_list(param: dict | None = None, user: User | None = None,
         search_dict['username'] = param['user']
     log.debug('Changed search key to username: %s.', search_dict['username'])
 
+    # None leaves the user list as it is, True/False keeps only the users that do respectively
+    # do not own a token. Resolved after the resolvers answered, see below. An empty value does
+    # not filter, as with realm and resolver above.
+    has_tokens = get_optional(param, "has_tokens")
+    has_tokens = is_true(has_tokens) if has_tokens not in (None, "") else None
```
nilsbehlen: **An unrecognised `has_tokens` value silently inverts the filter**

`is_true()` (`lib/utils/__init__.py:925`) accepts only `1`, `"1"`, `True`, `"True"`, `"true"`, `"TRUE"`. Everything else is `False`, so `GET /user/?realm=corp&has_tokens=yes` — a natural spelling, and one `is_true` does not know — returns exactly the users that own **no** token, with a 200 and nothing to indicate the parameter was misread. Same for `has_tokens=Tru`.

The docstring's "``True`` keeps only … ``False`` only …" already describes a closed set; parse the three states explicitly and raise `ParameterError` on anything else.

privacyidea/static/src/app/services/user/user.service.ts
```diff
@@ -551,6 +572,18 @@ export class UserService extends FilterableTableService implements UserServiceIn
...
 
+  /**
+   * Users of one realm, or of every realm when none is given, reduced to login name and resolver —
+   * the pair that identifies a person across realms. Unlike {@link usersResource} this is not bound
+   * to the globally selected realm. Unreachable resolvers are named in the response detail.
+   */
+  fetchUsernames(realm?: string): Observable<PiResponse<UserData[], UserListResponseDetail | undefined>> {
+    return this.http.get<PiResponse<UserData[], UserListResponseDetail | undefined>>(this.baseUrl, {
+      headers: this.authService.getHeaders(),
+      params: { attributes: "username,resolver", ...(realm ? { realm } : {}) }
```
nilsbehlen: **`include_custom_attributes=false` is missing here**

`api/user.py:207` defaults the flag to `True`:

```python
include_custom_attributes = (is_true(request.all_data.get("include_custom_attributes", True))
                             and is_attribute_at_all())
```

As soon as one custom user attribute exists anywhere in the DB, `is_attribute_at_all()` is true, so `get_user_list()` runs `get_attributes()` once per user (`lib/user.py:1031`) and appends/strips `userid` besides. A 20 000-user realm pays 20 000 extra SELECTs on every dashboard load and on every refresh.

`usersResource` in this same file (line 488) already sends `include_custom_attributes: false` — `fetchUsernames` should too. One line.

privacyidea/static/src/app/components/dashboard/widgets/tokens-widget/tokens-widget.component.ts
```diff
@@ -120,57 +234,135 @@ export class TokensWidgetComponent extends DashboardWidget implements OnInit {
...
         })
       )
     );
+    if (!this.canListUsers()) {
+      this.usersRef.set(null);
+      return;
+    }
+    this.usersRef.set(
+      // A failure here only costs the "without tokens" row, not the whole widget.
+      this.store.load(this.usersStoreKey(realm), () =>
```
nilsbehlen, line 330: **The whole user list is downloaded only to take its length**

`load()` fetches `GET /user/?realm=X&attributes=username,resolver` and `userCounts()` reduces the response to a single number. On a 100 000-user LDAP realm that walks the whole directory and ships every login to the browser on each dashboard visit, and again on each `DashboardDataStore.refreshAll()`.

This PR already establishes the right shape for the other half of the pair — `/token/ownercount` returns a number, not a list. A user-count endpoint (or a `count` / `pagesize=0` mode on `GET /user/`) is the fix at the same depth, and it is also the cheapest route out of the de-duplication mismatch flagged on `userCounts()`, because one endpoint can then define the identity for both numbers.

privacyidea/lib/token/query.py
```diff
@@ -990,6 +990,58 @@ def get_num_tokens_in_realm(realm: str, active: bool = True) -> int:
...
+    assignment without a realm - a token given to a user that has none - yields the empty realm and
+    therefore matches no user of any realm, just as the owner count of a realm leaves it out.
+
+    :param resolvers: Only return the triples of these resolvers. None returns the triples of every
+        resolver, an empty list returns nothing.
+    :return: A set of (lowercase realm name, resolver name, user id) triples
+    """
+    owners = select(Realm.name, TokenOwner.resolver, TokenOwner.user_id).select_from(TokenOwner).outerjoin(
+        Realm, Realm.id == TokenOwner.realm_id).distinct()
+    if resolvers is not None:
```
nilsbehlen, line 1016: **Every owner triple is materialised for what is often a single-user membership test**

`lib/user.py:1073` calls this with `resolvers=sorted(succeeded_resolvers)` and nothing else, so `GET /user/?realm=corp&username=alice&has_tokens=True` — one user in the result — still builds a set of every `(realm, resolver, user_id)` triple of every resolver that answered. With 200 000 assigned tokens that is a 200 000-element set of 3-tuples per request.

The realms in `realm_filters` and the user ids already collected in `owner_keys` are both available at the call site; passing them down (or pushing the membership test into SQL) keeps the query proportional to the page being listed.

privacyidea/static/src/app/components/dashboard/widgets/tokens-widget/tokens-widget.component.ts
```diff
@@ -120,57 +234,135 @@ export class TokensWidgetComponent extends DashboardWidget implements OnInit {
...
+    }
+    const id = this.instance()?.id;
+    if (id) {
+      this.layoutService.updateWidgetSettings(id, { [REALM_SETTING]: realm });
+    }
+    // The load is given the new realm: the instance input only carries it after the next
+    // change detection run, so reading it back here would still yield the previous one.
+    // Dropped, or the entries of every realm ever picked would be refetched on each refresh.
+    if (previous) {
+      this.store.invalidate(this.storeKey(previous));
```
nilsbehlen: **The "stored realm no longer exists" fallback abandons its store entries**

With `settings.realm = "oldrealm"`, `realmResource.hasValue()` is still false during construction, so `realm()` returns `"oldrealm"` (line 119) and the effect calls `load("oldrealm")` — creating `dashboard:tokens:oldrealm` and `dashboard:token-users:oldrealm`. When the realms arrive and the name is gone, `realm()` falls back to the default and loads that, but only `selectRealm()` invalidates what it leaves behind.

The two orphaned entries stay in the store, and `DashboardDataStore.refreshAll()` (`user-utils-panel.component.ts:270`) refetches everything it holds — so every global dashboard refresh re-issues the token counts plus a full user listing for a realm that no longer exists, indefinitely. `authentication-activity-widget.component.ts:406` and `conditional-access-widget.component.ts:586` both carry comments spelling out that this cleanup is required.

Invalidate the previous realm's entries in the fallback path too, not only in `selectRealm()`.

privacyidea/static/src/app/services/user/user.service.ts
```diff
@@ -286,8 +301,9 @@ export class UserService extends FilterableTableService implements UserServiceIn
         // the allowed-filter list and the emitted API param key stays canonical. Only the key is
         // normalized. The value is passed through case-preserving, so case matching is the resolver's
         // decision rather than the frontend's.
-        .map((token) => [token.key.toLowerCase(), token.value] as const);
-      return buildFilterParams(entries, this.allFilterKeys());
+        .map((token) => [token.key.toLowerCase(), token.value] as const)
+        .map(([key, value]) => [key, booleanKeys.has(key) ? toBooleanParam(value ?? "") : value] as const);
```
nilsbehlen: **An unparsable `has_tokens` is dropped instead of sent**

Typing `has_tokens: y` (or `on`, or `yes`) into the user filter: `toBooleanParam("y")` returns `undefined` (`utils/filter.utils.ts:76`), `toWildcardParam` then sees an empty trimmed value and emits no parameter at all. `GET /user/` runs unfiltered, the full realm is listed, and the filter chip still reads `has_tokens: y`.

`TokenService.toParamValue` (`token.service.ts:151`) does the opposite for the same class of key: `toBooleanParam(value) ?? value`, forwarding the raw value and letting the backend decide. Worth picking one behaviour across the two services — forwarding at least lets the server reject it, once the backend stops reading unknown values as `False`.

privacyidea/static/src/app/services/dashboard/dashboard-layout.service.ts
```diff
@@ -162,6 +164,13 @@ export class DashboardLayoutService implements DashboardLayoutServiceInterface {
     this.persistIfLive();
   }
 
+  public updateWidgetSettings(id: string, settings: WidgetSettings): void {
```
nilsbehlen, line 167: **`updateWidgetSettings()` overwrites rather than merges**

The method sits on the public `DashboardLayoutServiceInterface` and is named "update", but `{ ...widget, settings: { ...settings } }` discards `widget.settings` entirely. Today only `REALM_SETTING` exists, so nothing is lost; the moment a second setting is added (a sort order, a collapsed flag), `tokens-widget.component.ts:277` calling `updateWidgetSettings(id, { realm })` will erase it on every realm switch.

Either merge into the existing settings, or rename to `setWidgetSettings` so callers can see that full replacement is the contract.

privacyidea/static/src/app/components/dashboard/widgets/tokens-widget/tokens-widget.component.ts
nilsbehlen: **Minor, same line:** the `(resolver, username)` `Set` can never be smaller than the array it wraps.

`load()` always calls `fetchUsernames(realm)` with exactly one realm, and `get_user_list()` already de-duplicates on `(username, realm)` across every resolver of that realm — so no two entries in the response can share a username, let alone a `(resolver, username)` pair. `new Set(...).size` always equals `users.length`, and the comment "A resolver shared between realms reports its users once per realm" documents a case this call cannot produce.

It matters mainly because it reads like a guard against the counting mismatch flagged in the other comment on this line, and it isn't one. Either drop it for `users.length`, or make the key match what the backend actually de-duplicates on.

privacyidea/static/src/app/services/user/user.service.ts
```diff
@@ -34,12 +34,19 @@ import {
...
   FilterableTableServiceInterface
 } from "@services/table-utils/filterable-table-service";
-import { buildFilterParams, filterParamsEqual } from "@utils/filter.utils";
+import { buildFilterParams, filterParamsEqual, toBooleanParam } from "@utils/filter.utils";
 import { Observable, of } from "rxjs";
 import { catchError, map } from "rxjs/operators";
 
 const apiFilterKeys = ["description", "email", "givenname", "mobile", "phone", "resolver", "surname", "username"];
 
+// Not a user store attribute but privacyIDEA's own record, hence hidden behind the advanced toggle
```
nilsbehlen: **The comment describes an advanced toggle the user table does not have**

`user-table.component.ts:148` builds `filterKeywords = [...apiFilterKeys, ...advancedApiFilterKeys]` and `user-table.component.html:61` feeds that one flat list to `appFilterAutocomplete` — `has_tokens` is autocompleted exactly like `username` or `email`, not hidden behind anything.

The "passed through unwrapped" half is accurate; it's the placement rationale that will mislead whoever changes this next.

The accompanying spec has the same blind spot: *"offers every advanced keyword alongside the plain column filters"* asserts against `MockUserService.advancedApiFilterKeys`, which is empty, so it passes even if `has_tokens` is removed from the list entirely. Asserting on the literal key would let it fail.

nilsbehlen (commented):
Review of the branch as of f9268589a — 13 findings, ordered by severity below. Each one is an inline comment on the line it applies to.

**Would block a merge**
1. `api/token.py:816` — the `tokenlist` realm scope is a *token*-realm restriction everywhere else, but is applied here to the *owner* realm: the count diverges from what the same admin may list, in both directions.
2. `tokens-widget.component.ts:194` — "with tokens" and "without tokens" come from two endpoints that de-duplicate users by different keys; the widget can show more owners than the realm has users and clamps the difference to 0.
3. `lib/user.py:1021` — `str(userid or "")` turns a user id of `0` into `""`, which never matches its owner row and wrongly matches NULL/empty ones.
4. `lib/token/query.py:1018` — `resolver or ""` maps both NULL and `''` to one key, so one of the two counts is dropped, including from the reported total.
5. `lib/user.py:900` — `has_tokens=yes` is read as `False`, i.e. it returns the exact inverse of what was asked, with a 200.
6. `user.service.ts:583` — `fetchUsernames()` omits `include_custom_attributes=false`, costing one extra SELECT per user in the realm on every dashboard load.

**Worth fixing, not merge-blockers** — none of these produce a wrong answer to a well-formed request today.

7. `tokens-widget.component.ts:363` — the entire user list is fetched just to take its length; a count endpoint is the fix at the same depth (and the cheapest route out of 2).
8. `lib/token/query.py:1039` — every owner triple is materialised even when one user is being listed.
9. `tokens-widget.component.ts:280` — the realm-no-longer-exists fallback abandons its store entries, which every global refresh then refetches forever.
10. `user.service.ts:305` — an unparsable `has_tokens` is dropped client-side; `TokenService` forwards the raw value for the same key class.
11. `dashboard-layout.service.ts:167` — `updateWidgetSettings()` replaces the settings object instead of merging into it.
12. `tokens-widget.component.ts:194` — the `(resolver, username)` `Set` can never shrink the array it wraps.
13. `user.service.ts:43` — the comment claims an advanced toggle that does not exist, and the spec that covers it asserts against an empty mock list.

nilsbehlen (changes requested):

privacyidea/api/token.py
frankmer: Resolved by removing /token/ownercount and get_token_owners_per_resolver(). The widget now takes both user numbers from the new GET /user/count, guarded by userlist.

privacyidea/static/src/app/components/dashboard/widgets/tokens-widget/tokens-widget.component.ts
frankmer: Both numbers now come from one get_user_list() pass via GET /user/count, so they always add up and there is no need to clamp.

privacyidea/lib/user.py
frankmer: Fixed on the listing side via the new _owner_key(), where only None counts as missing, so 0 becomes "0". The query side keeps user_id or "" because a user without an id now gets no owner key at all and can no longer match a NULL or empty owner row.

frankmer:
/token/ownercount is removed. The widget now gets both user numbers from the new GET /user/count, which counts one get_user_list() pass, so count and with_tokens share the user list's identity and always add up.
Separately, tokenowner.token_id had no index, now added as index=True plus migration b7c1e4d2a9f3. I benchmarked the count query on the schema the migrations actually build, before and after, on SQLite, MariaDB 11, MySQL 8.4, PostgreSQL 17 and Oracle XE 21. Only SQLite gains, where it goes from unusable to instant. MySQL and MariaDB already index the foreign key, so the migration adds nothing there, and PostgreSQL and Oracle measured the same time with and without it. It therefore costs the production databases nothing and only brings a SQLite development database in line.

PR #5966, opened 2026-09-16
Allow self-service users to list service IDs for applspec enrollment

### LINKED ISSUE #5611: Self-service users cannot enroll Application Specific Password tokens — service_id dropdown requires admin-only /serviceid/ endpoint
### Description

The "Application Specific Password" (`applspec`) token enrollment form is shared between the Admin UI and the self-service User UI (`token.enroll.applspec.html`). Its `service_id` field is a **required** `<select>` dropdown with no free-text fallback:

```html
<select class="form-control"
        required
        id="sevice_id"
        ng-model="form.service_id"
        ng-options="type as desc for (type, desc) in formInit.service_ids">
</select>
```

The dropdown options are populated via `$scope.loadAvailableServiceIDs()` (`tokenControllers.js`), which calls `ConfigFactory.getServiceid()` → `GET /serviceid/`.

That endpoint's blueprint is grouped under `@admin_required` in `before_after.py`:

```python
@tokengroup_blueprint.before_request
@serviceid_blueprint.before_request
@admin_required
def before_admin_request():
    before_request()
```

This is a hard role check on the JWT, independent of any policy. No `admin`-scope policy (e.g. `serviceid_list`) can grant a `role=user` session access to it, since the restriction is enforced in code before policy evaluation ever runs.

### Impact

A realm user opening the enrollment wizard for an `applspec` token gets `403 Authentication failure. You do not have the necessary role (['admin']) to access this resource!` on the dropdown's data call. `formInit.service_ids` stays empty, and since the field is `required`, **the form can never be submitted** — self-service enrollment of this token type is effectively broken for every non-admin user, regardless of the `enrollAPPLSPEC` user-scope policy action being granted.

### Steps to reproduce

1. Grant a realm user `enrollAPPLSPEC: true` via a `scope: user` policy.
2. Log in to the self-service WebUI as that user.
3. Open the token enrollment wizard, select "Application Specific Password".
4. Observe: Service ID dropdown is empty / browser console shows 403 on `GET /serviceid/`; form cannot be submitted.

### Suggested fix

One of:
- Allow `GET /serviceid/` (list only, not add/delete) for `role=user` sessions, scoped by an existing or new user-facing policy action, mirroring how other read-only listings are exposed to self-service users.
- Add a free-text fallback for `service_id` in `token.enroll.applspec.html` when the dropdown fails to load / is empty, so self-service enrollment doesn't hard-depend on the admin API.

### Environment

- eduMFA 2.9.3
- Confirmed the same `@admin_required` grouping exists in current `privacyidea/privacyidea` main branch — likely inherited upstream, not eduMFA-specific.

Also reported as https://github.com/eduMFA/eduMFA/issues/1196 , report generated using Sonnet 5

samuraiii:
Closes #5611

## AI involvement

This PR was written by **Claude Code (Anthropic)**, working autonomously end-to-end at the request of a user administering a privacyIDEA-derived deployment (eduMFA) who ran into this bug in practice. Claude did the source investigation, designed and implemented the fix, wrote the new test, ran the full local test suite plus Ruff/Bandit/codespell, and iterated on CI failures via a throwaway draft PR against the requester's own fork before opening this one. The user reviewed the diff and security implications and authorized submission. I (the user) take responsibility for the change and am available to address review feedback.

## Problem

`GET /serviceid/` was gated by the blueprint-wide `@admin_required` check in `before_after.py`. The `applspec` (Application Specific Password) token enrollment form — in both the legacy AngularJS WebUI and the new WebUI — has a required `service_id` dropdown with no free-text fallback, populated from this endpoint. A self-service user granted `enrollAPPLSPEC` therefore got a 403 on the dropdown's data call, `service_id` stayed empty, and the enrollment form could never be submitted, regardless of the granted enrollment policy.

## Fix

* Moved `serviceid_blueprint` from the admin-only `before_admin_request` group to the mixed admin/user `before_user_request` group in `privacyidea/api/before_after.py`.
* `GET /serviceid/` now relies purely on the existing `check_base_action(SERVICEID_LIST)` prepolicy, which is scope-aware. `SERVICEID_LIST` is now also registered under `SCOPE.USER` (`privacyidea/lib/policy.py`), so admins can explicitly grant or restrict it for self-service users the same way they already can for admins.
* `POST`/`DELETE` (add/delete a service ID) keep a hard admin-only gate via `@admin_required`, now applied directly to those two view functions in `privacyidea/api/serviceid.py` instead of inheriting it from the blueprint.
* Added `tests/test_api_serviceids.py::test_02_selfservice_user_can_list_but_not_modify`, which checks a user-role token gets 200 on `GET /serviceid/` and 401 on both `POST` and `DELETE`.
* Documented the new `serviceid_list` user policy action in `doc/policies/user.rst` and updated the module docstring / endpoint docstring in `privacyidea/api/serviceid.py`.
* Changelog entry added.

## Security note

On an install with **zero** user-scope policies defined, `GET /serviceid/` becomes readable by any authenticated realm user by default (privacyIDEA's existing "no policies of this scope ⇒ default-allow" model for user-scope actions). This is consistent with how every other user-scope action already behaves (e.g. `assign`, `resync`) and the data involved is low-sensitivity (service ID names/descriptions meant for the enrollment dropdown), but it is a genuine widening of read access relative to the previous hard-401 behavior, so flagging it explicitly for review. `POST`/`DELETE` are unaffected — `@admin_required` is a hard role check independent of policy state.

## Testing

* `tests/test_api_serviceids.py` (existing + new test): pass
* `tests/test_lib_serviceid.py`, `tests/test_api_machines_serviceid.py`, `tests/test_api_audit.py`: pass (15 tests, adjacent blueprints/layers)
* `tests/test_api_system.py`, `tests/test_api_info.py`, `tests/test_api_token.py`: pass (172 tests, other blueprints sharing the edited `before_request` group)
* Full CI (Ruff, Bandit, codespell, CodeQL, MariaDB/PostgreSQL/Redis unit tests on Python 3.10 and 3.14) run green via a throwaway draft PR against my own fork before opening this one.
* Not verified: the actual browser enrollment flow in either WebUI (no running server/frontend build in this environment). The backend fix addresses the root cause both frontends were blocked on.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nilsbehlen:
Hi and thank you for opening this pullrequest.
Do you want to implement feedback yourself, or should we just add to your PR?

samuraiii:
I thing you can add.

nilsbehlen:
hi again, i will update this next week, so it is not forgotten.

nilsbehlen:
Notes for when I rework this, mostly about the WebUI half, which got stricter on the other side of the fence in the meantime.

#5968 gates every WebUI request on the right the endpoint actually enforces. For service IDs that landed in `service-id.service.ts` as

```ts
// serviceid_list exists only in the admin scope, so a self-service user can never list the service IDs.
readonly canListServiceIds = computed<boolean>(
  () => !this.authService.isSelfServiceUser() && this.authService.actionAllowed("serviceid_list")
);
```

plus an enrollment block whose message reads "Application specific password tokens cannot be enrolled in self-service: the service IDs are only available to administrators."

Both are accurate against master today — this PR is what makes them wrong, so the edit belongs here rather than in #5968. The part to watch is that nothing fails loudly if it is forgotten: if only the API change lands, `GET /serviceid/` starts answering users while the WebUI still hard-blocks them, the feature ships as a no-op, and `tests/test_api_serviceids.py` passes either way.

Three things move with the API change:

- drop `!isSelfServiceUser()` from `canListServiceIds` — `actionAllowed("serviceid_list")` alone is then correct for both roles
- drop the comment above it, which is the justification for that conjunct
- collapse `serviceIdsUnavailableReason` to its `serviceid_list` branch and remove the `serviceId.unavailableSelfService` string

`serviceIdResource`'s route predicate needs nothing: `enrollAPPLSPEC` is already in the user scope via `ui_enroll: ["admin", "user"]`.

The default install works out. `ui_get_rights` returns every action of a scope when that scope has no active policies, so registering `SERVICEID_LIST` under `SCOPE.USER` grants it to every self-service user on an install with no user-scope policies, and the block lifts without anyone configuring anything.

On the API change itself: moving the blueprint to the mixed group and putting `@admin_required` on the POST and DELETE routes directly keeps every write route gated, which checks out.

PR #5968, opened 2026-09-16
Skip WebUI requests the admin has no rights for or whose result they cannot use

### LINKED ISSUE #5940: Do not send requests the admin has no right for
An `httpResource` decides whether to send its request almost entirely from the current route: "am I on
a page that might display this data?". Whether the logged-in admin is actually allowed to read that data is only
checked in some of the services. For an admin whose rights are cut down by an admin policy (helpdesk-style roles), the
WebUI therefore sends requests that the server correctly answers with 403.

A resource whose endpoint is policy-gated must check the corresponding right before it
fires — and where the data is genuinely needed for something the admin *is* allowed to do, the UI should say which right is missing instead of silently showing nothing.

The pattern already exists in the code base, it is just not applied consistently.

### Example that triggered this (token enrollment page, privacyIDEA servers)

`PrivacyideaServerService.remoteServerResource`
(`services/privacyidea-server/privacyidea-server.service.ts:63-72`) fires whenever
`ContentService.onTokenEnrollmentLikely()` is true and the caller is not a self-service user. It never checks
`privacyideaserver_read`, and the failure is piped into `handleResourceError` at line 91, so a restricted admin gets the
error snackbar quoted above on:

* Tokens → Enroll Token (`/tokens/enrollment`)
* Token details (included for rollover)
* Token wizard
* Container templates (list, create, details)

The admin cannot use the data either way: the privacyIDEA-server list is only needed for the `remote` token type, and
`remote` is not offered in the type dropdown unless the admin has `enrollREMOTE` (the list comes from
`GET /auth/rights`). The navigation entry and the External Services → privacyIDEA Servers page are *already* gated on
`privacyideaserver_read` (`navigation.component.html:877`, `privacyidea-servers.component.ts:85`) — only the background
request is not. The service is provided at the layout route (`app.routes.ts:90`) and injected by the user-utils panel,
so the resource is live for the whole session and re-evaluates on every navigation.

Worth fixing together with it: an admin who *does* have `enrollREMOTE` but not `privacyideaserver_read` would, after a
pure "skip the request" change, get an empty server dropdown with no explanation — the same trap that was fixed for
email actions in cd84b72bb ("Keep email actions offered without smtpserver_read and hint at the missing right"). Note
the asymmetry: RADIUS and certificate enrollment read their names from `/system/names/*`, gated by the *enroll* right,
whereas the remote token needs the separate config-read right `privacyideaserver_read` — so `enrollREMOTE` alone is not
enough to enroll a remote token in the WebUI.

frankmer:
Closes #5940


privacyidea/static/src/app/components/token/token-enrollment/enroll-asp/enroll-applspec.component.ts
```diff
@@ -81,9 +67,16 @@ export class EnrollApplspecComponent extends EnrollTokenBase<ApplspecEnrollmentD
...
     required(f);
-    disabled(f, () => this.disabled() || this.generateOnServer() || this.authService.checkForceServerGenerateOTPKey("applspec"));
+    disabled(
+      f,
+      () => this.disabled() || this.generateOnServer() || this.authService.checkForceServerGenerateOTPKey("applspec")
+    );
   });
 
+  // Mirrors the fetch condition of serviceIdResource: without it the field becomes a free-text input.
+  readonly serviceIdsListable = computed<boolean>(
```
nilsbehlen, line 77: **2. Self-service users are told they need admin-only rights**

**Breaks:** applspec is user-enrollable (`applicationspecificpasswordtoken.py:81-83`: `'user': ["enroll"]`, `'ui_enroll': ["admin", "user"]`), but `serviceid_list` exists only in `SCOPE.ADMIN`, so `ui_get_rights(SCOPE.USER)` can never contain it and `serviceIdsListable()` is *always* false for role `user`. Every self-service user enrolling an ASP token therefore gets the red hint "listing the service IDs needs the serviceid_list and enrollAPPLSPEC rights" - naming one right no user can ever hold, and one (`enrollAPPLSPEC`) they must already have to be on that form at all.

**Fix:** Include the `isSelfServiceUser()` conjunct that `serviceIdResource` starts with, and give the user role its own wording - for a user nothing is missing, the list simply is not available to them.

privacyidea/static/src/app/components/token/token-enrollment/enroll-remote/enroll-remote.component.ts
```diff
@@ -89,6 +77,14 @@ export class EnrollRemoteComponent extends EnrollTokenBase<RemoteEnrollmentData>
   });
 
   remoteServerOptions = this.privacyideaServerService.remoteServerOptions;
+  // Mirrors the fetch condition of remoteServerResource: without it the identifier becomes a free-text input.
+  readonly remoteServersListable = computed<boolean>(
+    () => this.authService.actionAllowed("privacyideaserver_read") && this.authService.actionAllowed("enrollREMOTE")
+  );
+
+  onRemoteServerInput(identifier: string): void {
```
nilsbehlen, line 85: **5. An empty remote server in the fallback input aborts enrollment silently**

**Breaks:** Leave the free-text fallback empty and click Enroll: `buildEnrollmentArgs` hits `if (!this.remoteServer()) return null;` (line 101) without `markAsTouched()`, and `<mat-error>Remote Server is required` never renders because this control has no `ngControl`, so `MatFormField._control.errorState` stays false. Nothing happens and nothing is shown.

**Fix:** Do what `enroll-applspec` does in this same PR - hold the identifier in a `remoteServerId = signal<string>("")`, wrap it in `form(...)` with `required(f)`, bind `[formField]`, and build `remote.server_id` from it.

nilsbehlen (commented):
Reviewed this branch's own two commits (`0bf135372` "Skip WebUI requests the admin has no rights for..." and `b2ea1fab0` "fix tests"); the two master merges (`b81a6ff41`, `d551641c8`) are out of scope.

I checked all 18 policy action names used here against `privacyidea/lib/policies/actions.py` and the `@prepolicy` decorators - every one matches the right the endpoint actually enforces, including the `tokenlist` -> `manage_machine_tokens` correction for `GET /machine/token` and the admin-only `configread` on `GET /system/` (`check_admin_base_action`). `ui_get_rights` returns all scope actions when no policies exist, so default installations are not gated off, and all five render sites of `app-enroll-token-type-switch` fall under `onTokenEnrollmentLikely()`, so none of the three new `*Listable` predicates currently yields an empty select without a fallback.

12 findings, ranked by severity. **1-8 are the functional ones, 9-12 are quality items.** Findings 2, 5, 6 and 8-12 are inline below. Findings 1, 3, 4 and 7 point at files this PR does not touch, so GitHub cannot anchor them (not even file-level) - they are written out here in full, in their ranked positions.

---

### 1. Email token type config is still a dead end - `privacyidea/static/src/app/components/configuration/token-type-config/token-types/email-config/email-config.component.html:36` *(functional)*

**Breaks:** `smtpServerResource` (`smtp.service.ts:69-81`) already returns `undefined` without `smtpserver_read`. An admin with `systemwrite` but no `smtpserver_read` opens Configuration -> Token Types -> Email: `smtpServers()` is `[]`, the `<mat-select>` renders with zero options, and the hint still says "Select a predefined SMTP server configuration". There is no input fallback, so `email.mailserver` can never be set by that admin - the exact dead end the SMS half of this PR fixes, on the same page.

**Fix:** Give it the same `@if (smtpServersListable()) { <mat-select> } @else { <input matInput> }` treatment `sms-config.component.html` just got.

### 3. `GET /token/` is still sent without `tokenlist` from three resources - `privacyidea/static/src/app/services/token/token.service.ts:694` *(functional)*

**Breaks:** `GET /token/` is guarded by `check_admin_tokenlist(PolicyAction.TOKENLIST)` (`api/token.py:639-640`). An admin with `userlist` but not `tokenlist` opens Users -> \<user\> details: `userTokenResource` (gated only on `onUserDetails()`) sends `/token/?user=...`, gets a 403 and an error toast. `tokenDetailResource` (line 654) and `tokenSerialResource` (line 620) have the same hole. `tokenResource` at line 733 already carries `if (this.authService.role() === "admin" && !this.authService.actionAllowed("tokenlist")) return undefined;`, so the inconsistency sits inside one file.

**Fix:** Add the same `tokenlist` conjunct to those three predicates.

### 4. The machine card on token details now renders permanently empty - `privacyidea/static/src/app/components/token/token-details/token-details.component.html:197` *(functional)*

**Breaks:** `machine.service.ts:247` now returns `undefined` without `manage_machine_tokens` (admin scope only). An admin with `tokenlist` but not `manage_machine_tokens` opens an sshkey or hotp token: `@if (tokenType() === "sshkey" || tokenType() === "hotp")` still renders `<app-token-details-machine>`, `tokenApplications()` stays `undefined`, and the card shows an empty Machines list with no explanation. Before this change the 403 at least produced an error notification.

**Fix:** Gate the card on `manage_machine_tokens` too, the way the navigation entry already is (`navigation.component.html:331`: `actionsAllowed(["machinelist", "manage_machine_tokens"])`).

### 7. RADIUS token type config has the same dead end - `privacyidea/static/src/app/components/configuration/token-type-config/token-types/radius-config/radius-config.component.html:35` *(functional)*

**Breaks:** The parent binds `[radiusServers]="systemService.radiusServers()"` (`token-type-config.component.html:66`), and that resource returns `undefined` without `enrollRADIUS` (`system.service.ts:176-180`; `/system/names/radius` is gated on `enrollRADIUS` in `api/system.py:492`). An admin who may configure token types but may not enroll RADIUS tokens sees an empty select plus the "Select a predefined" hint and cannot set `radius.identifier` at all.

**Fix:** Same free-text fallback as the SMS gateway field.

privacyidea/static/src/app/components/token/token-enrollment/enroll-asp/enroll-applspec.component.ts
nilsbehlen, line 77: **6. The fetch predicate is re-derived per component instead of read from the service**

**Breaks:** The comment says this mirrors `serviceIdResource`, but that predicate is `!isSelfServiceUser() && (onExternalServiceIds() || (onTokenEnrollmentLikely() && enrollAPPLSPEC)) && serviceid_list` - three conjuncts the computed drops. `enroll-remote.component.ts:81` and `sms-config.component.ts:49` do the same. It is harmless today only because every render site of the enroll switch happens to fall under `onTokenEnrollmentLikely()`; add one render site, or one route to a resource, and you get an empty select with no fallback and no hint.

**Fix:** Expose a single `canList` signal on each service and have both the `httpResource` predicate and the template read it - `authentication-log.service.ts:218` (`canRead`) already works that way.

privacyidea/static/src/app/components/token/token-enrollment/enroll-asp/enroll-applspec.component.html
```diff
@@ -42,15 +42,28 @@
...
+            <mat-option [value]="option">{{ option }}</mat-option>
+          }
+        </mat-select>
+      } @else {
+        <input
+          matInput
+          [formField]="serviceIdForm" />
+      }
+      @if (!serviceIdsListable()) {
+        <mat-hint
```
nilsbehlen, line 61: **8. The instruction is a `mat-hint`, so Material hides it exactly when it is needed**

**Breaks:** In the fallback branch the input is bound to `serviceIdForm`, which carries `required(f)`. As soon as `markAsTouched()` fires on an empty field, `MatFormField._getDisplayedMessages()` returns `'error'` and renders only `<mat-error>Service ID is required` - the "Type the name" guidance is removed from the DOM at the moment the user needs it. `enroll-remote.component.html:47` has the same problem.

**Fix:** Put the "type it by hand" wording somewhere that survives an error state: the label/placeholder, or the error text itself.

privacyidea/static/src/app/components/token/token-enrollment/enroll-remote/enroll-remote.component.html
```diff
@@ -27,15 +27,30 @@
...
+          }
+        </mat-select>
+      } @else {
+        <input
+          matInput
+          required
+          [value]="remoteServer()?.id ?? ''"
+          (input)="onRemoteServerInput($event.target.value)" />
+      }
+      @if (!remoteServersListable()) {
```
nilsbehlen, line 47: **9. Redundant second `@if` on a condition that already has an `@else`**

**Breaks:** `@if (remoteServersListable()) {...} @else {...}` is immediately followed by `@if (!remoteServersListable())`, so one decision is evaluated and tracked twice and the two blocks can drift apart when edited. `enroll-applspec.component.html:60` duplicates it the same way; `sms-config.component.html` gets it right with a single `@if/@else` pair covering both the control and the hint.

**Fix:** Move the hint into the existing `@else`.

privacyidea/static/src/app/components/token/token-enrollment/enroll-remote/enroll-remote.component.ts
```diff
@@ -89,6 +77,14 @@ export class EnrollRemoteComponent extends EnrollTokenBase<RemoteEnrollmentData>
...
+    this.remoteServer.set(identifier ? ({ id: identifier } as RemoteServer) : null);
```
nilsbehlen, line 86: **10. `{ id: identifier } as RemoteServer` fabricates a half-built object**

**Breaks:** The cast leaves `identifier`, `name`, `url` and `tls` undefined while the type declares them present, which switches off the only check that would catch a consumer reading a field this path never sets. The select branch of the same template renders `{{ option.name }}`, so if `remoteServersListable()` ever flips to true after a value was typed, the label is `undefined` and the object-identity `[value]` comparison matches no listed server either.

**Fix:** Keep a plain `remoteServerId` string signal and map it to `remote.server_id` at build time rather than inventing a `RemoteServer`.

privacyidea/static/src/app/services/resolver/resolver.service.spec.ts
```diff
@@ -253,6 +253,14 @@ describe("ResolverService", () => {
     expect(resolverService.editableResolvers()).toEqual(["ldap1", "ldap2", "ldap3"]);
   });
 
+  it("should not request a single resolver without resolverread", () => {
```
nilsbehlen, line 256: **11. This test cannot detect a wrong policy action name**

**Breaks:** `(authService.actionAllowed as jest.Mock).mockReturnValue(false)` disallows *every* action, so the assertion still passes if `selectedResolverResource` were gated on, say, `policyread` instead of `resolverread`. That is precisely the class of mistake this PR had to avoid in ~20 places - and hit once already (`tokenlist` -> `manage_machine_tokens`).

**Fix:** Drive `authData.rights` like the other new tests in this PR, or pin the name the way `event.service.spec.ts:327` does with `action !== "eventhandling_read"`.

privacyidea/static/src/app/services/privacyidea-server/privacyidea-server.service.spec.ts
```diff
@@ -50,6 +51,9 @@ describe("PrivacyideaServerService", () => {
     httpMock = TestBed.inject(HttpTestingController);
     notificationService = TestBed.inject(NotificationService);
     contentService = TestBed.inject(ContentService) as unknown as MockContentService;
+    authService = TestBed.inject(AuthService) as unknown as MockAuthService;
+    // The servers are only requested by an admin who may read them; on enrollment only if remote tokens may be enrolled.
```
nilsbehlen, line 55: **12. Comment lines over the 120-character limit**

**Breaks:** This comment is 121 characters and `service-id/service-id.service.spec.ts:54` is 122; the project line length is 120. Prettier does not reflow comments, so `npm run format:write` will not clean them up.

**Fix:** Wrap both onto a second line.

privacyidea/static/src/app/components/token/token-enrollment/enroll-asp/enroll-applspec.component.ts
frankmer: Disabled the select when the list cannot be loaded and added a tooltip with the reason on it and on the Enroll button.

frankmer: Added a canList signal to every service, for example canListRemoteServers, and used it in both the httpResource and the templates.

privacyidea/static/src/app/components/token/token-enrollment/enroll-asp/enroll-applspec.component.html
frankmer: Moved the reason into a tooltip on the disabled select and on the Enroll button.

privacyidea/static/src/app/components/token/token-enrollment/enroll-remote/enroll-remote.component.ts
frankmer: Replaced the RemoteServer object with a plain remoteServerId string and set the option value to the database id that remote.server_id expects.

nilsbehlen (approved):

PR #5976, opened 2026-09-17
refuse to start with an unusable PI_HASH_ALGO_LIST or PI_HASH_ALGO_PARAMS

### LINKED ISSUE #5926: An unusable PI_HASH_ALGO_LIST is only discovered at the first login
An unusable value in ``PI_HASH_ALGO_LIST`` is accepted at startup and only fails when
something first verifies a hash - which in practice means the first login attempt.

`CryptContext(["nosuchscheme"])` raises `KeyError: "no crypt handler found for algorithm:
'nosuchscheme'"`. That is not a `ValueError`, so none of the `except ValueError` handlers
around hash verification catch it, and a single typo turns into an unhandled traceback:

- `crypto.py:287` - `verify_with_pepper()`, i.e. **administrator login** - unguarded
- `lib/challengeresponsedecorators.py:111` - unguarded
- `models/token.py:302` - guarded with `except ValueError`, which a `KeyError` passes
  straight through
- the authentication cache, since #5914

So an installation can start, look healthy, and then answer every login with a 500. The
administrator who made the typo has no way back in through the UI either, since their own
login is one of the paths that breaks.

`PI_HASH_ALGO_PARAMS` has the same shape of problem: an unknown parameter name reaches
passlib only when the first hash is created.

## Proposed change

Build the `CryptContext` once, when the application starts, and fail there with a message
naming the configuration entry and the unknown scheme.

Note that `_check_config()` (`app.py:208`), which already refuses to start without
`PI_ENCFILE` and `PI_PEPPER`, is **not** the right place on its own: it is called only
from `create_docker_app()` (`app.py:544`) and never from `create_app()`, so a validation
placed there would not cover the hand-written `pi.cfg` where such a typo is most likely.

A side benefit: `pass_hash()` and `verify_pass_hash()` construct a `CryptContext` on every
call. At about 0.03 ms that is not a performance problem, but a context built once is the
natural place for the validation.

## Verification

- An unknown scheme in `PI_HASH_ALGO_LIST` refuses to start, with the entry and the
  scheme in the message.
- A valid list still starts, for both `create_app()` and `create_docker_app()`.
- An unknown key in `PI_HASH_ALGO_PARAMS` is reported the same way.

frankmer:
Closes #5926


nilsbehlen (commented):
Findings are on the individual lines. One more that cannot be anchored inline, because the file is not part of this diff:

**`doc/installation/system/inifile.rst` (around line 105) -- `PI_HASH_ALGO_LIST` / `PI_HASH_ALGO_PARAMS` are not documented as a hard startup failure.**

An admin following the current advice ("take care that the previously used one is still included", "further information on possible parameters in the PassLib documentation") has no indication that a typo in either key now takes down the whole server -- and every `pi-manage` command with it -- instead of failing at login as before. A short note next to those two paragraphs, showing the error text, belongs with this change.

privacyidea/app.py
```diff
@@ -272,6 +273,25 @@ def _check_config(app: Flask):
         app.config[ConfigKey.NO_RESPONSE_SIGN] = True
 
 
+def _check_hash_config(app: Flask):
```
nilsbehlen: **Consider building the context where it is used instead of re-implementing it here.**

This function hardcodes the two shapes `crypto.py` happens to use today (a param-less context for `verify_pass_hash`, `DEFAULT_HASH_ALGO_PARAMS` + config params for `pass_hash`), and `app.py` now has to import `CryptContext`, `DEFAULT_HASH_ALGO_LIST` and `DEFAULT_HASH_ALGO_PARAMS` for it. Any later change in `crypto.py` -- adding `default=`/`deprecated=`, changing the merge order, reading a third config key -- leaves this check validating something the runtime no longer builds, and nothing fails.

A `build_pass_context()` in `crypto.py`, cached in `get_app_local_store()` and primed once by both factories, would validate at startup by construction, could not drift, and would drop the ~30 us `CryptContext` rebuild that `pass_hash`/`verify_pass_hash` currently pay on every PIN check and every `/auth`. It would also be the single place to fix the missing-backend and parameter problems noted below.

privacyidea/app.py
```diff
@@ -272,6 +273,25 @@ def _check_config(app: Flask):
...
+    # passlib only reports an unusable scheme or parameter when the CryptContext is built, which
```
nilsbehlen: **This premise is not true, and it is why the check below is incomplete.**

Measured against passlib: building the `CryptContext` reports neither a missing backend (that is raised at `.hash()`/`.verify()`) nor an out-of-range parameter value (that is warned about and clamped, never raised). Construction only catches an unknown scheme name and an unknown parameter name.

"i.e. the first login" is also narrower than reality -- `pass_hash` is reached by setting a token PIN, by password reset and by `pi-manage admin add`, so the first failure need not be a login.

Suggestion: describe what is actually checked, so the next reader does not trust a guarantee the code does not give.

privacyidea/app.py
```diff
@@ -272,6 +273,25 @@ def _check_config(app: Flask):
...
+    # otherwise happens at the first hash verification, i.e. the first login.
+    hash_algo_list = app.config.get(ConfigKey.HASH_ALGO_LIST, DEFAULT_HASH_ALGO_LIST)
+    try:
+        pass_ctx = CryptContext(hash_algo_list)
```
nilsbehlen: **The check misses the main class of unusable `PI_HASH_ALGO_LIST`: passlib resolves hash backends lazily.**

`CryptContext(...)` construction succeeds for a scheme that has no working backend, so the first login still fails. Verified on this branch:

- `PI_HASH_ALGO_LIST = ["bcrypt", "pbkdf2_sha512"]` -- `_check_hash_config` passes, `create_app()` returns, and the first `pass_hash("secret")` raises `ValueError` from the bcrypt backend.
- `PI_HASH_ALGO_LIST = ["argon2", ...]` on a host without `argon2_cffi` -- the context constructs fine and `.hash()` raises `passlib.exc.MissingBackendError`, which is a `RuntimeError` subclass and therefore not even caught by the `(KeyError, ValueError, TypeError)` tuple below.

A missing backend is the likeliest cause of the reported symptom, so as it stands the PR leaves it in place.

Suggestion: actually exercise the contexts -- `pass_ctx.hash(<dummy>)` on the params context, plus `get_crypt_handler(scheme).has_backend()` for every scheme in the list, since `verify_pass_hash` uses all of them and not just the default -- and add `MissingBackendError` to the caught set.

privacyidea/app.py
```diff
@@ -272,6 +273,25 @@ def _check_config(app: Flask):
...
+    except (KeyError, ValueError, TypeError) as e:
+        raise RuntimeError(f"'{ConfigKey.HASH_ALGO_LIST}' is not usable: {e.args[0] if e.args else e}") from e
+    if not pass_ctx.schemes():
+        raise RuntimeError(f"'{ConfigKey.HASH_ALGO_LIST}' must contain at least one hash algorithm")
+
+    hash_algo_params = dict(DEFAULT_HASH_ALGO_PARAMS)
+    hash_algo_params.update(app.config.get(ConfigKey.HASH_ALGO_PARAMS, {}))
```
nilsbehlen: **This line is outside the `try`, so a non-mapping `PI_HASH_ALGO_PARAMS` escapes as a bare `ValueError` instead of this function's diagnostic.**

Docker deployment with `PRIVACYIDEA_PI_HASH_ALGO_PARAMS=argon2__rounds=5` (not valid JSON): `from_prefixed_env` keeps it as a `str`, and `dict.update("argon2__rounds=5")` raises `ValueError: dictionary update sequence element #0 has length 1; 2 is required` out of `create_docker_app`, with no mention of `PI_HASH_ALGO_PARAMS` anywhere in the traceback -- the opposite of what this function exists to do. Same for a list or an int.

Suggestion: move the `update()` inside the `try` below; it raises the same exception types the handler already catches.

privacyidea/app.py
```diff
@@ -272,6 +273,25 @@ def _check_config(app: Flask):
...
+    try:
+        CryptContext(hash_algo_list, **hash_algo_params)
```
nilsbehlen: **A parameter for a scheme that is not in the list is silently dropped, and this check calls the config usable.**

`CryptContext(["argon2", "pbkdf2_sha512"], argon2id__rounds=5)` constructs without error and hashes with the default rounds. `PI_HASH_ALGO_PARAMS = {"argon2id__rounds": 5}` is a very plausible typo -- the stored hash reads `$argon2id$` -- and the result is that the server starts, this check says nothing, and argon2 keeps `rounds=9`. Same for `{"bcrypt__rounds": 5}` when bcrypt is not listed.

That is a silently ineffective security setting inside the exact scope this function claims to cover.

Suggestion: after building the context, assert that every `<scheme>__` prefix in `hash_algo_params` names a scheme in `pass_ctx.schemes()`, and raise otherwise.

nilsbehlen: **Out-of-range parameters are clamped by passlib with only a warning, so a silently weakened configuration passes this check.**

`CryptContext(["pbkdf2_sha512"], pbkdf2_sha512__rounds=0)` constructs -- three `PasslibHashWarning`s, no exception -- and then hashes as `$pbkdf2-sha512$1$...`, i.e. one round. With `PI_HASH_ALGO_PARAMS = {"pbkdf2_sha512__rounds": 0}`, or any value below a scheme's minimum, the server starts, this check reports the config usable, and every admin password and token PIN is hashed with effectively no work factor.

Suggestion: run both `CryptContext` calls inside `warnings.catch_warnings()` with `simplefilter("error", passlib.exc.PasslibWarning)`, so a clamped value is reported like a rejected one.

privacyidea/app.py
```diff
@@ -546,6 +566,7 @@ def create_app(config_name="development",
             DEFAULT_LOGGING_CONFIG["handlers"]["file"]["filename"] = app.config.get(ConfigKey.LOGFILE)
         _setup_logging(app, DEFAULT_LOGGING_CONFIG)
 
+    _check_hash_config(app)
```
nilsbehlen: **While you are here: `create_app()` still never calls `_check_config()`.**

`_check_config` is called only from `create_docker_app`. `create_app` is the factory behind the WSGI entry point, `create_silent_app`, and therefore every `pi-manage` / `pi-token-janitor` / cron invocation -- and it validates neither `PI_ENCFILE` nor `PI_PEPPER`. An install whose `PI_ENCFILE` path is wrong starts cleanly and fails on the first token operation, which is the same "only discovered at first use" bug class this PR is about.

Adding `_check_config(app)` one line above the new call would close it.

tests/test_app.py
```diff
@@ -177,6 +193,84 @@ class Config(TestingConfig):
...
+        self.pi_logger = logging.getLogger("privacyidea")
+        self.pi_state = (self.pi_logger.handlers[:], self.pi_logger.level, self.pi_logger.propagate)
+
+    def tearDown(self):
+        self.logger.handlers = self.orig_handlers
+        self.logger.level = self.level
+        self.pi_logger.handlers, self.pi_logger.level, self.pi_logger.propagate = self.pi_state
+
+    @contextmanager
+    def isolated_config_file(self):
```
nilsbehlen: **Copy-pasted verbatim from `AppTestCase`, including a comment that describes the other class's assertions.**

"comparing the app config against config.py only works against an empty config file" -- `HashConfigTestCase` compares nothing against `config.py`, it checks the hash configuration. And this copy is dead code, see the comment on `_create_app`.

Suggestion: hoist one module-level contextmanager and use it from both classes.

tests/test_app.py
```diff
@@ -177,6 +193,84 @@ class Config(TestingConfig):
...
+        # A pi.cfg installed on the machine running the tests overwrites the values from
+        # config.py, so comparing the app config against config.py only works against an
+        # empty config file.
+        with tempfile.NamedTemporaryFile(suffix=".cfg") as config_file:
+            with mock.patch.dict(os.environ, {ENV_KEY: config_file.name}):
+                yield
+
+    @staticmethod
+    def _create_app(**hash_config):
```
nilsbehlen: **`isolated_config_file` is never called, so these tests run against the machine's `/etc/privacyidea/pi.cfg` and its database.**

`_create_app` passes no `config_file` and sets no `PRIVACYIDEA_CONFIGFILE`, so `create_app()` falls back to `DefaultConfigValues.CFG_PATH`. Verified locally: the test app comes up with `SQLALCHEMY_DATABASE_URI = "sqlite:////.../data.sqlite"` and `PI_HEALTH_CERT_PROBES` from the installed `pi.cfg`, and `test_05` runs the full factory, so `_setup_node_configuration()` writes a `NodeName` row into that database.

Worse for what these tests are for: a `pi.cfg` that sets `PI_HASH_ALGO_LIST` or `PI_HASH_ALGO_PARAMS` is loaded *after* the injected `Config` class and silently overrides the value under test, so on such a machine tests 01-05 assert against the machine's config instead of the injected one.

Suggestion: use the `isolated_config_file` contextmanager that is already sitting unused in this class, or pass an explicit `config_file` (`tests/testdata/test_pi.cfg`) the way `MyTestCase` does. `_create_docker_app` right below already guards against exactly this by patching `DefaultConfigValues.CFG_PATH`.

nilsbehlen: **Each test builds a complete Flask app for a check that only needs config validation.**

`test_05` and `test_07` run the factory to completion, so every run registers an `atexit` closure that holds a whole Flask app alive until interpreter exit and then calls `db.engine.dispose()` in an app context. `_setup_logging` also mutates the module-level `DEFAULT_LOGGING_CONFIG` / `DOCKER_LOGGING_CONFIG` dicts (`filename` and `level`) for the rest of the pytest process; `tearDown` restores the logger objects but not those dicts.

tests/test_app.py
```diff
@@ -177,6 +193,84 @@ class Config(TestingConfig):
...
+        # The Docker factory also reads the pi.cfg of the machine the test runs on.
+        with mock.patch.object(DefaultConfigValues, "CFG_PATH", os.path.join(dirname, "no-such-pi.cfg")), \
+                mock.patch.dict(os.environ, env):
+            return create_docker_app()
+
+    def test_01_unknown_scheme_refuses_to_start(self):
+        with self.assertRaisesRegex(RuntimeError, "PI_HASH_ALGO_LIST.*nosuchscheme"):
+            self._create_app(PI_HASH_ALGO_LIST=["argon2", "nosuchscheme"])
+
+    def test_02_empty_list_refuses_to_start(self):
```
nilsbehlen: **This regex matches both `RuntimeError` branches, so the test does not pin the empty-list behaviour it is named for.**

`"PI_HASH_ALGO_LIST"` is satisfied by the `... is not usable` message as well as by `... must contain at least one hash algorithm`. If the `if not pass_ctx.schemes()` branch were removed and some other error happened to fire, this test would still pass.

Suggestion: assert the distinguishing text (`at least one hash algorithm`), the way tests 01/03 assert on the offending token.

nilsbehlen (changes requested):

nilsbehlen (approved):

PR #5985, opened 2026-09-20
Consider the login in User.exist() to detect users deleted from the store

### LINKED ISSUE #4788: User.exist() not really checks if the user exist

anushamukka9:
User.exist() only checked uid and realm_id, so a user deleted from the user store still counted as existing as long as those identifiers were known. That breaks the orphaned-token and orphaned-settings detection, which relies on exist() to spot principals that are gone. The login is resolved from the user store when the User object is built, so requiring it makes exist() reflect whether the user is actually still there. Added a regression test for a uid that no longer resolves to a login.

Fixes #4788


nilsbehlen:
hello. thanks for trying, but that problems scope is alot larger, and we want to tackle it properly, which we cant do right now since we are about to finish the next release. Also improving this might result in behavior changes, so its scheduled for 3.15 currently. Thanks again!

